PR #31, opened 2015-11-24
Client cleanup

azawlocki:


poc/golemPy/golem/clientconfigdescriptor.py
```diff
@@ -52,6 +52,33 @@ def __init__(self):
         self.app_version = ""
         self.eth_account = ""
 
+    def init_from_app_config(self, app_config):
+        """
```
chfast: Multiline doc string should be written as:

```
"""Multiline
text
""""
```

Looks bad, but you don't want to have `\n` at the beginning.

poc/golemPy/golem/network/transport/message.py
```diff
@@ -1827,6 +1827,13 @@ def __init__(self, num=0, sig="", timestamp=None, dict_repr=None):
     def dict_repr(self):
         return {MessageNewNodes.NUM_STR: self.num}
 
+def init_manager_messages():
+    """ Add manager messages to registered messages list"""
```
chfast: Also no space at the beginning.

PR #53, opened 2015-12-01
Change the way KeysAuth and derivatives decide on keys on disk location

chfast:


chfast:
Side question: can we remove RSA-based manager?

badb:
We may remove RSA or leave if it for test purpose. It was added to test if auth methods are general.

chfast:
I will leave it then. Any comments about the code changes?

golem/core/keysauth.py
```diff
@@ -131,6 +131,23 @@ def generate_new(self, difficulty):
         return
 
     @staticmethod
+    def __get_key_loc(file_name_pattern, uuid):
```
badb: Maybe we should set root_path as a variable in instance KeysAuth? It won't be static method, but use will be able to set path in config file.

PR #68, opened 2015-12-03
Client id to key

badb:


badb:
Important changes:
- app_cfg is a little bit nicer
- client_id => node_name
- resource server is indexing resource peers by key_id instead of client_id
- p2pserver is indexing all peers by key_id instead of client_id
- tasks are indexing known client by key_id instead of client_id

examples/gnr/task/blenderrendertask.py
```diff
@@ -132,10 +132,10 @@ def __init__(self,
         for frame in frames:
             self.frames_given[frame] = {}
 
-    def query_extra_data(self, perf_index, num_cores=0, client_id=None):
+    def query_extra_data(self, perf_index, num_cores=0, node_id=None, node_name=None):
 
-        if not self._accept_client(client_id):
-            logger.warning(" Client {} banned from this task ".format(client_id))
+        if not self._accept_client(node_id):
+            logger.warning(" Client {} banned from this task ".format(node_name))
```
chfast: - If `node_name` is None the warning message will be incorrect. Probably the better default value for `node_name` would be `''` or `'<unnamed>'`.
- Is the leading space needed?

examples/gnr/ui/TaskDetailsDialog.ui
```diff
@@ -175,7 +175,7 @@
       <item row="0" column="0">
        <widget class="QLabel" name="label_6">
         <property name="text">
-         <string>Node ID</string>
+         <string>Node name</string>
```
chfast: "Node Name" of "Node name"?

golem/network/p2p/peersession.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, conn):
 
         # Information about peer
         self.degree = 0
-        self.node_id = None
+        self.node_name = None
```
chfast: Shouldn't that be an empty string by default?

golem/network/transport/message.py
```diff
@@ -929,7 +935,7 @@ class MessageReportComputedTask(Message):
 
     SUB_TASK_ID_STR = u"SUB_TASK_ID"
     RESULT_TYPE_STR = u"RESULT_TYPE"
-    NODE_ID_STR = u"NODE_ID"
+    NODE_NAME_STR = u"NODE_NAME"
```
chfast, line 938: Why that repeats many times?

chfast:
It would be nice to separate big mechanical changes (like renames) from functional changes in the future.

examples/gnr/task/blenderrendertask.py
badb: `node_name` should never be None in no-test scenario. But yeah, for unpredicted or test scenarios, we should initialize it with blank string.

golem/network/transport/message.py
badb, line 938: Because it's very old and ugly way of making messages (each message has its own constants). Maybe we should change someday... (but it's not crucial change).

PR #75, opened 2015-12-04
Docker image build script and GNR compute node

chfast:


golem/environments/checkcmd.py
```diff
@@ -12,5 +14,6 @@ def check_cmd(cmd, no_output=True):
         else:
             rc = subprocess.call([pref_cmd, cmd])
             return rc == 0
-    except Exception, err:
+    except Exception:
+        logger.warn("Command {} not found".format(cmd))
```
badb: This warning may be misleading. There are problems with too broad exception clause in many places in the [code](https://github.com/imapp-pl/golem/issues/34). We should change this to [OSError](https://docs.python.org/2/library/subprocess.html#exceptions) and add some support for ValueError and CalledProcessError (with different warnings).

azawlocki: I'd just change 'Exception' to 'subprocess.CalledProcessError' and let other exceptions pass through.

badb: But this function is used mainly to check whether command exists in a system. If it doesn't OSError is thrown. CalledProcessError is called only if this program exists, but return error.

azawlocki: In this case the program is 'which'/'where' and its returns error if the command we test for does not exist. Our code assumes that 'which'/'where' is available, if it's not then an exception should be thrown

chfast:
I have rewritten check_cmd() to something that not require another external program like `which`.

PR #76, opened 2015-12-04
Dummy task

azawlocki:
1) A task class implementing a minimal (sort of) subset of golem.task.taskbase.Task interface.
2) A test script that starts a headless Golem client and submits a single instance of a dummy task.


chfast:
Looks good to me. You can merge at will.

PR #81, opened 2015-12-07
multiple addresses to connect with from node

badb:


azawlocki:
At some point it would be good to separate the Blender-specific part of node.py (~3 lines of code) from the more generic stuff that can be reused with other environments.

azawlocki:
Sorry, didn't want to close this yet!

examples/gnr/node.py
```diff
@@ -4,27 +4,66 @@
...
+
+def __parse_ipv4(addr_arg):
+    host, port = addr_arg.split(":")
+    port = int(port)
+    return TCPAddress(host, port)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='GNR Compute Node')
+    parser.add_argument('--connect', type=str)
```
chfast, line 53: 1. Here we can add param `nargs='*'`. That will parse `--connect addr1 addr2` into `['addr1', 'addr2']`. No need to manually parse coma-separated lists. More info: https://docs.python.org/3/library/argparse.html#nargs
2. I would consider name other than `--connect`, preferably `--peer`, or `--peers`.

examples/gnr/node.py
```diff
@@ -4,27 +4,66 @@
...
+    client = create_client()
+    blender_env = BlenderEnvironment()
+    blender_env.accept_tasks = True
+    client.environments_manager.add_environment(blender_env)
+    client.start_network()
+
+    if args.connect:
+        addresses = parse_connect(args.connect)
+        for addr in addresses:
+            client.connect(addr)
```
chfast, line 65: I would like to avoid to call `connect` directly. I want to just add peers and let the Client implementation what to do with that information. Ideally:

```
client.add_peers([])
client.start()
```

I can be sure `add_peers()` is non-blocking. Not sure about `connect()`.

However, current version can be accepted for now.

azawlocki, line 65: In `P2PService` there's a number of lists and dicts such as `free_peers`. The function `client.add_peers()` could add peer addresses to `free_peers`. It seems that `P2PService.sync_network()`, which is called periodically, tries to connect to nodes from `free_peers` if more peers are needed.

chfast, line 65: What about that suggestion?

PR #85, opened 2015-12-09
Tests for Ethereum SubAccounts contract

chfast:
The contract solidity code is so far here: https://chriseth.github.io/browser-solidity/?gist=ac140b941584ca36be92
I copy the EVM code and ABI spec to python manually. 
Please review the contract as well.

Tests are quite basic, not every edge case is tested. But I want to do it in iterations as it is quite boring job to write tests. Besides I expect the contract to change also.

Later on I will create a cost model for the contract (in form of Python function) and will compare that with the reality in the tests.

Travis CI fails to build it because of some dependencies problems. You may take a look at that. Anyway we've just reached the limit of 100 fee builds on Travis anyway. CircleCI is free when used with one container (it looks to be enough for now), so I will disintegrate unless objections.


chfast:
I've added the contract source code to the repo to allow commenting.

golem/contracts/SubAccounting.sol
```diff
@@ -0,0 +1,75 @@
...
+    function withdraw(uint value) external {
+        uint balance = _balance[msg.sender];
+        if (balance >= value) {
+            // FIXME: send() can fail if sending to another contract.
+            msg.sender.send(value);
+            _balance[msg.sender] -= value;
+        }
+    }
+
+    function transferExt(uint[] payments) external {
```
azawlocki: I don't like the name `transferExt`. 'Ext' is for 'external'?

golem/contracts/SubAccounting.sol
```diff
@@ -0,0 +1,75 @@
...
+        return _balance[msg.sender];
+    }
+
+    function deposit() external {
+        // TODO: Use it as an anonymous method?
+        _balance[msg.sender] += msg.value;
+    }
+
+    function withdrawAll() external {
+        // TODO: Should we explicitly forbid full payout?
```
azawlocki, line 21: Or we can store real_balance + 1 in _balance, so that it is never 0 even after withdrawAll(), and have another function (closeAccount()?) for explicitly shutting down an account. Not sure if it's worth extra complication, maybe in future.

golem/contracts/SubAccounting.sol
```diff
@@ -0,0 +1,75 @@
+contract SubAccounting {
+    mapping(address => uint) _balance;
+
+    event Transfer(address indexed from, address indexed to, uint256 value);
+
+    function balanceOf(address addr) constant external returns (uint) {
+        return _balance[addr];
+    }
+
+    // This function is redundant, but might make some clients simplier.
```
azawlocki: simpler

golem/contracts/SubAccounting.sol
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        // Send left value to sender account (conditional to safe storage
+        // modification costs).
+        if (value > 0)
+            _balance[msg.sender] += value;
+    }
+
+    // This function allows batch payments using sent value and sender balance
+    function transfer(bytes32[] payments) external {
```
azawlocki, line 56: was `uint[] payments` in `transferExt`

golem/contracts/SubAccounting.sol
```diff
@@ -0,0 +1,75 @@
...
+            bytes32 payment = payments[i];
+            address addr = address(payment);
+			uint v = uint(payment) / 2**160;
+            if (v > value)
+                break;
+            _balance[addr] += v;
+            value -= v;
+            Transfer(msg.sender, addr, v);
+        }
+
```
azawlocki, line 48: The same for loop as in `transferExt` (almost). Can we extract it to a separate function, or have one existing function call another?

golem/contracts/SubAccounting.sol
chfast: Yes. It's an `transfer()` alternative that transfers only external value. I'm not sure it is worth it, but is should be good at least for comparison. Currently it is not even tested.

Is `transferExternalValue` a better name?

golem/contracts/SubAccounting.sol
chfast, line 21: I have similar ideas about that.

In the end DApp can/should handle that. We can keep it simple here. DApp/Golem can add another level of abstraction, like:
- create account: `deposit(1)`
- withdraw all: `withdraw(balance() - 1)`
- close account: `withdraw(balance())`

I plan to send the contract for community review too.

golem/contracts/SubAccounting.sol
chfast, line 56: `bytes32` and `uint` are binary identical to contracts. However, it turned out it is easier to pack `uint96` and `address` into `bytes32` than into `uint` in Python. In other words `bytes32` is better for Python and humans as it informs the data is not a simple number.

In general I proposed to extend the ABI with structs like `(uint96, address)`, but that will take at least some time to be accepted. https://github.com/ethereum/EIPs/issues/21

golem/contracts/SubAccounting.sol
azawlocki, line 21: ok

golem/contracts/SubAccounting.sol
chfast, line 48: We can do both. But we will loose gas on that. It is not a big deal for now, but prevents reliable `transfer` with `transferExt` comparison.

golem/contracts/SubAccounting.sol
azawlocki: I think so. You decide.

azawlocki:
After the dependency on 'rlp' is fixed you can go ahead and merge!

chfast:
Explicit dependencies of `rlp` and `pycrypto`/`pycryptodome` removed.

I also added tests for `transferExternalValue()`. It is less than 0.5% better than `transfer()` in worst case (2 recipients).

Would you like to check that part?

golem/contracts/SubAccounting.sol
badb, line 21: What are the pros/cons of making method anonymous?

chfast, line 21: Anonymous method is called when contract cannot match any other method. In case a transaction with value and without data is sent, we can still add the value to the sender's sub-account.

badb, line 21: So it's just safer for user who may send empty message with value by mistake? Are there any cost differences? If not, it may be a good idea.

chfast, line 21: No additional costs. It's just solidity feature. As you said, we can catch some user's mistakes with that and prevent the user from loosing money.

I will track contract's TODO list in #88

PR #122, opened 2016-01-15
Changes for enabling network tests with IMUNES

azawlocki:
You can ignore the script `tests/imunes/imunes.py` for now. It works fine but needs some explanation, I'll add a wiki page. I'm also not sure it `tools/` is a good place for docker files, mabe just `docker/`?


chfast:
## [Current coverage](https://codecov.io/github/imapp-pl/golem?ref=5c558e5c9951bc78685fc849c9c003b953af2635) is `36.93%`

> Merging **#122** into **develop** will decrease coverage by **-0.59%** as of [`5c558e5`](https://codecov.io/github/imapp-pl/golem/commit/5c558e5c9951bc78685fc849c9c003b953af2635)

``` diff
@@            develop    #122   diff @@
=======================================
  Files           254     256     +2
  Stmts         20862   21001   +139
  Branches       2329    2339    +10
  Methods           0       0       
=======================================
- Hit            7829    7756    -73
- Partial         253     256     +3
- Missed        12780   12989   +209
```

> Review entire [Coverage Diff](https://codecov.io/github/imapp-pl/golem/compare/eef24ec8e263db650f056aa31fa7e3fd3d8c5744...5c558e5c9951bc78685fc849c9c003b953af2635) as of [`5c558e5`](https://codecov.io/github/imapp-pl/golem/commit/5c558e5c9951bc78685fc849c9c003b953af2635)

---

### [Uncovered Suggestions](https://codecov.io/github/imapp-pl/golem/features/suggestions?ref=5c558e5c9951bc78685fc849c9c003b953af2635)
1. `+0.12%` via [...siptrustsimulator.py#261...285](https://codecov.io/github/imapp-pl/golem/golem/ranking/simulator/gossiptrustsimulator.py?ref=5c558e5c9951bc78685fc849c9c003b953af2635#261) 
2. `+0.11%` via [.../task/pbrtgnrtask.py#72...95](https://codecov.io/github/imapp-pl/golem/gnr/task/pbrtgnrtask.py?ref=5c558e5c9951bc78685fc849c9c003b953af2635#72) 
3. `+0.10%` via [...siptrustsimulator.py#64...84](https://codecov.io/github/imapp-pl/golem/golem/ranking/simulator/gossiptrustsimulator.py?ref=5c558e5c9951bc78685fc849c9c003b953af2635#64) 
4. _[See 7 more...](https://codecov.io/github/imapp-pl/golem/features/suggestions?ref=5c558e5c9951bc78685fc849c9c003b953af2635)_

> Powered by [Codecov](https://codecov.io). Updated on successful CI builds.

gnr/node.py
```diff
@@ -1,7 +1,13 @@
 """GNR Compute Node"""
 
 import os
-import pickle
+
+try:
+    import cPickle as pickle
```
chfast: Why `cPickle` is better?

badb: "First, cPickle can be up to 1000 times faster than pickle because the former is implemented in C" - source: https://docs.python.org/2/library/pickle.html#module-cPickle

gnr/node.py
```diff
@@ -37,9 +43,10 @@ class Node(object):
     def __init__(self, **config_overrides):
         self.client = create_client(**config_overrides)
 
-    def initialize(self):
+    def initialize(self, enable_blender):
         self.client.start_network()
-        self.load_environments(self.default_environments)
+        if enable_blender:
+            self.load_environments(self.default_environments)
```
chfast: That looks weird. Can we make it more generic by providing environments to load, or adjust "default environments" before starting a node?

golem/network/transport/session.py
```diff
@@ -87,7 +87,7 @@ def interpret(self, msg):
         """
         self.last_message_time = time.time()
 
-        # print "Receiving from {}:{}: {}".format(self.address, self.port, msg)
+        print "Receiving from {}:{}: {}".format(self.address, self.port, msg)
```
chfast: We must use logs here.

tests/gnr/test_gnr_node.py
```diff
@@ -22,7 +23,7 @@ def setUp(self):
         AppConfig.CONFIG_LOADED = False
 
     @patch('gnr.node.reactor')
-    def test_help(self, mock_reactor):
+    def test_hep(self, mock_reactor):
```
chfast: Why?

scripts/pickle2json.py
```diff
@@ -0,0 +1,19 @@
+if __name__ == "__main__":
```
chfast: This is not needed I think.

golem/network/transport/session.py
badb: I will leave it commented as it was. It generates a lot of information which is useful for some very specific network debugging scenarios. Putting this in logs just doesn't make sens most of the time.

tests/gnr/test_gnr_node.py
azawlocki: No idea.

gnr/node.py
chfast: But it was not needed in the first place. Does it improve performance?

badb: The question was if why is `cPickle` better, not if it's really necessary ;) It probably doesn't make a huge difference for current small objects. Maybe if there are many resource files. Python3 uses cPickle as default pickle implementation and we're using cPickle in different places in the code... Just guessing, I don't now what was the real reason.

azawlocki: I've simplified this to `import cPickle as pickle` (without `try ... except`). That's how we import `cPickle` in other places.

chfast:
OK for me.

PR #137, opened 2016-01-26
add one payment for the whole task

badb:


golem/network/transport/message.py
```diff
@@ -1554,6 +1554,29 @@ def dict_repr(self):
         return {MessageNatPunchFailure.NAT_PUNCH_FAILURE_STR: True}
 
 
+class MessageRewardPaid(Message):
+    Type = TASK_MSG_BASE + 25
```
azawlocki, line 1558: Can you please add a simple comment: when is this msg sent, who is the sender and who is the receiver?

golem/task/taskserver.py
```diff
@@ -255,6 +252,9 @@ def subtask_accepted(self, task_id, reward):
             self.decrease_trust_payment(task_id)
         self.task_keeper.remove_waiting_for_verification(task_id)
 
+    def subtask_accepted(self, subtask_id, reward):
+        logger.debug("Subask {} result accepted".format(subtask_id))
```
azawlocki, line 256: A typo: "Subask"

golem/task/tasksession.py
```diff
@@ -262,10 +263,14 @@ def send_result_rejected(self, subtask_id):
...
-    def send_reward_for_task(self, subtask_id, reward):
+    # TODO: by default this may be
+    def send_reward_for_task(self, task_id, reward):
+        self.send(MessageRewardPaid(task_id, reward))
+
+    # TODO: use this method
+    def send_message_subtask_accepted(self, subtask_id, reward):
         """ Inform that results pass verification and confirm reward
-        :param str subtask_id:
+        :param str task_id:
```
azawlocki, line 273: But the param's called `subtask_id`.

golem/transactions/ethereum/ethereumpaymentskeeper.py
```diff
@@ -43,3 +43,8 @@ def __init__(self, key_id, port, addr, node_name, node_info, eth_account):
         AccountInfo.__init__(self, key_id, port, addr, node_name, node_info)
         self.eth_account = eth_account
 
+    def __eq__(self, other):
+        ethereum_eq = self.eth_account == other.eth_account
```
azawlocki, line 47: I'd make this comparison case insensitive.

badb, line 47: Are ethereum addresses case insensitive?

tests/golem/transactions/test_paymentskeeper.py
```diff
@@ -180,3 +184,17 @@ def test_database(self):
...
         self.assertTrue(xyzCalled)
+
+
+class TestAccountInfo(unittest.TestCase):
+    def test_comparison(self):
+        k = EllipticalKeysAuth()
+        a = AccountInfo(k.get_key_id(), 5111, "10.0.0.1", "test-test-test", Node())
+        b = AccountInfo(k.get_key_id(), 5111, "10.0.0.1", "test-test-test", Node())
+        self.assertEqual(a, b)
+        n = Node(prv_addr="10.10.10.10", prv_port=1031, pub_addr="10.10.10.10", pub_port=1032)
```
azawlocki, line 195: We should also check that `AccountInfos` with same key and other fields different are equal.

golem/task/tasksession.py
badb, line 273: This method need general changes, the whole description should be change and it's not used right now. I'll fix description (and method) when payment and verification protocol will be updated.

golem/network/transport/message.py
badb, line 1558: Done in develop. It's not a final version, only a mockup with less misleading name that previous MessageSubtaskResultAccepted.

golem/transactions/ethereum/ethereumpaymentskeeper.py
chfast, line 47: Depends what you mean. Ethereum account is a 160-bit integer, so no. Python implementation usually uses 20-byte array (still no), but it is common to present is as hex-encoded string, usually lowercase. If this input comes from a user we might consider that.

I would add some tests covering such cases.

badb, line 47: It comes from user input and then it is read from config file right now, but as I understand in the future it will be generated from lightwallet. Maybe we should jest add class EthereumAddress that has some parsing / checking /comparison method?

azawlocki, line 47: Looking at the tests in `TestEthereumPaymentsKeeper` below I assumed the addresses are given as strings containing 20-digit hex numbers. Separate class for addresses is a good idea.

chfast, line 47: I would keep it as pyethereum does for now, what is a list of 20 bytes. Conversion from string is done by `str.decode('hex')`.

PR #138, opened 2016-01-27
ethereum addresses class + small fixes

badb:


golem/transactions/ethereum/ethereumpaymentskeeper.py
```diff
@@ -41,10 +52,34 @@ class EthAccountInfo(AccountInfo):
...
         account_eq = AccountInfo.__eq__(self, other)
         return ethereum_eq and account_eq
 
+
+class EthereumAddress(object):
+    """ Keeps information about ethereum addresses in normalized format
+    """
+    @classmethod
+    def parse(cls, address):
+        if len(address) in range(40, 50):
```
chfast: Where the 50 comes from?

golem/transactions/ethereum/ethereumpaymentskeeper.py
```diff
@@ -41,10 +52,34 @@ class EthAccountInfo(AccountInfo):
...
+            address = address.lower()
```
chfast, line 65: Why to lower only in this case?

golem/transactions/ethereum/ethereumpaymentskeeper.py
```diff
@@ -41,10 +52,34 @@ class EthAccountInfo(AccountInfo):
...
         ethereum_eq = self.eth_account == other.eth_account
         account_eq = AccountInfo.__eq__(self, other)
         return ethereum_eq and account_eq
 
+
+class EthereumAddress(object):
+    """ Keeps information about ethereum addresses in normalized format
+    """
+    @classmethod
+    def parse(cls, address):
```
chfast, line 63: This function does not even return a EthereumAddress.

golem/transactions/ethereum/ethereumpaymentskeeper.py
```diff
@@ -41,10 +52,34 @@ class EthAccountInfo(AccountInfo):
...
+        if len(address) in range(40, 50):
+            address = address.lower()
+        return normalize_address(address)
+
+    def __init__(self, address):
+        self.address = None
+        try:
+            self.address = self.parse(address)
+        except Exception as err:
+            logger.warning("Can't set Ethereum address, {} is not a proper value: {}".format(address, err))
```
chfast, line 73: Exception replaced with a log? WTF?

golem/transactions/ethereum/ethereumpaymentskeeper.py
```diff
@@ -1,5 +1,16 @@
+import logging
+from rlp.utils import encode_hex
+
+try:
+    from pyethereum.utils import normalize_address
```
chfast: Is there any problem with pyethereum package?

golem/transactions/ethereum/ethereumpaymentskeeper.py
```diff
@@ -41,10 +52,34 @@ class EthAccountInfo(AccountInfo):
     """ Information about node's payment account and Ethereum account. """
     def __init__(self, key_id, port, addr, node_name, node_info, eth_account):
         AccountInfo.__init__(self, key_id, port, addr, node_name, node_info)
-        self.eth_account = eth_account
+        self.eth_account = EthereumAddress(eth_account)
```
chfast, line 51: Maybe we should use simply `normalize_address()`?

golem/transactions/ethereum/ethereumutils.py
```diff
@@ -53,3 +62,22 @@ def ceil32(x):
 
 def to_string_for_regexp(value):
     return str(value)
+
+
+def sha3(seed):
+    return sha3_256(to_string(seed))
+
+
+def normalize_address(x, allow_blank=False):
```
chfast: Is that a copy from pyethereum?

golem/transactions/ethereum/ethereumpaymentskeeper.py
chfast: The repo is named `pyethereum`, but Python package `ethereum`. I think it should be `from ethereum.utils`.

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -21,4 +26,13 @@ def global_pay_for_task(self, task_id, payments):
         :param dict payments: all payments group by ethereum address
         """
         eth_connector = EthereumConnector(ETH_CONN_ADDR)
-        eth_connector.pay_for_task(self.eth_account, task_id, payments)
+        if self.eth_account.get_str_addr():
+            eth_connector.pay_for_task(self.eth_account.get_str_addr(), task_id, payments)
+        else:
+            logger.warning("[!IMPORTANT!] Can't pay for task {}, no ethereum address set".format(task_id))
```
chfast: Yes, this is very important... log.

golem/transactions/ethereum/ethereumpaymentskeeper.py
badb: Yeah, you're right it should be 51. normalize_address function accepts strings with length 40, 42, 48 and 50.

golem/transactions/ethereum/ethereumpaymentskeeper.py
badb, line 65: Because only those cases (ok, 51...) arguments are supposed to be hex-encoded strings. Other lengths are accepted when they are 20 or 24-bytes arrays.

golem/transactions/ethereum/ethereumtransactionsystem.py
badb: Oh, come on. Pay for task it's a mockup right now anyway and it's functionality it's commented out. When we add real integration with Ethereum we should think how to solve this situation properly... We will see "very important" log and remember about it, but it won't mess up test nodes...

golem/transactions/ethereum/ethereumpaymentskeeper.py
badb, line 73: See the comment at the bottom. It doesn't make sense to throw back address exception right now.

golem/transactions/ethereum/ethereumpaymentskeeper.py
badb, line 63: Why should it?

golem/transactions/ethereum/ethereumpaymentskeeper.py
badb: I still cannot install it properly on Windows. Can I?

golem/transactions/ethereum/ethereumutils.py
badb: It is. All methods in this file are. I know it's terribly ugly, just temporary solution until I find a way to use pyethereum..

golem/transactions/ethereum/ethereumpaymentskeeper.py
badb, line 51: `normalize_address` returns binary address format. We need a method that return hexencoded string in some places, so I think it's useful to have this class which gives easy access to both representations.

chfast, line 51: I don't agree. We should choose single internal representation. Convert only in place where different representation is expected.

golem/transactions/ethereum/ethereumpaymentskeeper.py
chfast, line 63: It looks like "named constructor". If it is not designed to be used in other places make it private.

golem/transactions/ethereum/ethereumtransactionsystem.py
chfast: Ok. I'm usually add `FIXME` comment in such places to track "small tasks" like that.

badb:
I'm merging this branch, even though there are still some doubts, because there are some useful features. We may come back to ethereum address class design during ethereum integration.

PR #142, opened 2016-01-27
fix issue

badb:
https://github.com/imapp-pl/golem/issues/140

### DIFF
--- a/gnr/customizers/configurationdialogcustomizer.py
+++ b/gnr/customizers/configurationdialogcustomizer.py
@@ -30,15 +30,15 @@ class ConfigurationDialogCustomizer(Customizer):
     @staticmethod
     def du(path):
         try:
-            return subprocess.check_output(['du', '-sh', path]).split()[0].replace(",", ".")
+            size = int(subprocess.check_output(['du', '-sb', path]).split()[0])
         except (OSError, subprocess.CalledProcessError):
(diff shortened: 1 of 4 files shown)


badb:
This value is only used as a string, so it's also possible to make a fix in tests and not in function code. But we may need to want to use number in the future so changing a function may be a better solution.

gnr/customizers/configurationdialogcustomizer.py
```diff
@@ -30,7 +30,7 @@ def load_data(self):
     @staticmethod
     def du(path):
         try:
-            return subprocess.check_output(['du', '-sh', path]).split()[0]
+            return subprocess.check_output(['du', '-sh', path]).split()[0].replace(",", ".")
```
chfast: Maybe it's better to skip `-h` option.

chfast: Is it faster than pure Python? E.g. http://stackoverflow.com/questions/1392413/calculating-a-directory-size-using-python

chfast:
Ok. However, consider adding test for `get_dir_size()`: https://codecov.io/github/imapp-pl/golem/compare/017d052...47be47e

azawlocki:
Fine for me but let's try to meet this 80% codecov target.

PR #144, opened 2016-01-29
Ethereum PoC

chfast:
Payments with Ethereum Proof of Concept. The `eth.py` script shows all needed uses cases.


golem/ethereum/client.py
```diff
@@ -0,0 +1,123 @@
...
+    __client_rpc_port = None
+
+    @staticmethod
+    def __start_client_subprocess():
+        if not Client.__client_subprocess:
+            assert not Client.__client_rpc_port
+            rpcport = find_free_net_port(9001)
+            basedir = path.dirname(__file__)
+            # Data dir must be set the class user to allow multiple nodes running
+            datadir = path.join(appdirs.user_data_dir('golem'), 'ethereum9')
```
badb: Is `ethereum9` a typo or network number?

chfast: No. We are using Ethereum private network with id 9. I added the id to the dir name to make Ethereum network swap easier.

golem/ethereum/client.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+log = logging.getLogger('golem.eth.rpc')
+
+
+def find_free_net_port(start_port):
+    open_ports = set(c.laddr[1] for c in psutil.net_connections())
+    while start_port in open_ports:
+        start_port += 1
+    return start_port
```
chfast: @badb Here is a helper function that finds free port number on host. That may be useful for others. Should I move it?

badb: Yes, good idea.

chfast:
All comments addressed.

This PR contains also code for Ethereum contracts. It would be better to submit them separately but I hope we can live with that.

Issue #146, opened 2016-02-03
Set proper public_address config parameter


badb:
In default branch right now there is a bug, which can be easily fixed by adding public_address field to config descriptor and appconfig, initialized with empty value. But we have to determine what should be the proper init value and is it really necessary.

PR #151, opened 2016-02-08
Fix for #146

azawlocki:
The option `--public-address` is now only valid for `tests/imunes/node.py`, not for `gnr/node.py`. Specifying `--public-address` results in bypassing `Node's` `collect_network_info`  (stun is not used) and setting `Node.pub_addr` and `Node.priv_addr` to the specified value.

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -41,10 +41,11 @@ def create_client(**config_overrides):
     config_desc.init_from_app_config(app_config)
 
     for key, val in config_overrides.iteritems():
-        # For safety we only allow here overriding properties that are
-        # fields of ClientConfigDescriptor. Hence the getattr() here:
-        getattr(config_desc, key)
-        setattr(config_desc, key, val)
+        if hasattr(config_desc, key):
+            setattr(config_desc, key, val)
+        else:
+            raise AttributeError(
+                "Can't override nonexistent config attribute '{}'".format(key))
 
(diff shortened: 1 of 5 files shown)


golem/client.py
```diff
@@ -41,6 +41,9 @@ def create_client(**config_overrides):
     config_desc.init_from_app_config(app_config)
 
     for key, val in config_overrides.iteritems():
+        # For safety we only allow here overriding properties that are
+        # fields of ClientConfigDescriptor. Hence the getattr() here:
+        getattr(config_desc, key)
```
chfast: There is also `hasattr()` builtin function. It might be more explicit.

tests/imunes/test_imunes_node.py
```diff
@@ -4,6 +4,7 @@
 from golem.core.simpleenv import SimpleEnv
 from gnr.renderingenvironment import BlenderEnvironment
 from golem.tools.testwithappconfig import TestWithAppConfig
+# from golem.client import create_client
```
chfast: Remove.

golem/client.py
azawlocki: You're right, never write in comments what you can write in code. Fixed in 2eff0831aad71e868cf53f0a6654bba19bb77f80

tests/imunes/test_imunes_node.py
azawlocki: Removed.

PR #167, opened 2016-02-18
Ethereum contract: Lottery

chfast:


golem/ethereum/contracts/Lottery.sol
```diff
@@ -8,13 +8,6 @@ contract Lottery {
...
     }
 
-    function payout() {
-        if (msg.sender == owner) {
-            if (msg.sender.send(ownerDeposit))
-                ownerDeposit = 0;
-        }
-    }
-
     // in real-life this struct can fill 2 storage words
```
badb: Is it still true?

chfast: Yes. In stage 1 we have: value (32b), payer (20b) and maturity (should fit 8b). In stage 2 random value replaces maturity and payer.

badb:
Looks great, but we still should consider possibility of sending `check` for someone else and receiving a percantage of the prize.

chfast:
"Someone else" would need to know the proof. So only the lottery participants can do it. Let's consider that later on.

PR #175, opened 2016-02-23
Quickfixes

badb:


badb:
Fixes: https://github.com/imapp-pl/golem/issues/164

PR #211, opened 2016-03-16
IPFS resource exchange

mfranciszkiewicz:
Computed files are zipped, encrypted with AES-256 and published via IPFS.

Current issues:
- task result _hashes_ need to be persisted on disk and loaded on demand
- it is not possible to switch between the vanilla exchange method and IPFS
- there is no control over the IPFS daemon


golem/resource/ipfs/resourcesmanager.py
```diff
@@ -0,0 +1,206 @@
...
+        common_prefix = self.task_common_prefixes.get(task_id, '')
+        return path.replace(common_prefix, '', 1)
+
+    # def copy_resources(self, new_resource_dir):
+    #     copy_file_tree(self.resource_dir, new_resource_dir)
+    #     file_names = next(os.walk(self.resource_dir))[2]
+    #     for f in file_names:
+    #         os.remove(os.path.join(self.resource_dir, f))
+    #
+    # def change_resource_dir(self, resource_dir):
```
badb: Why is it commented out?

PR #216, opened 2016-03-18
Issue74

badb:


badb:
Works really well and looks nice! There are some tests needed and then we can merge it.

PR #224, opened 2016-03-22
Rename TCPAddress to SocketAddress

badb:


chfast:
If only renames go ahead.

PR #229, opened 2016-03-25
Refactoring for transaction system

badb:


golem/task/taskkeeper.py
```diff
@@ -7,43 +7,112 @@
...
         self.removed_task_timeout = remove_task_timeout
         self.environments_manager = environments_manager
 
     def is_supported(self, th_dict_repr):
+        """ Checks if task described with given task header dict representation may be computed
+        by this node. This node must support proper environment, be allowed to make computation cheaper than with
+        max price declared in task and have proper application version.
+        :param dict th_dict_repr: task header dictionary representation
+        :return bool: True if this node may compute a task
+        """
```
chfast, line 111: In the future we should keep track of the reason the task was rejected.

golem/task/taskkeeper.py
```diff
@@ -7,43 +7,112 @@
...
-            return None
-
     def check_environment(self, th_dict_repr):
+        """ Checks if this node supports environment necessary to compute task described with task header.
+        :param dict th_dict_repr: task header dictionary representation
+        :return bool: True if this node support environment for this task, False otherwise
+        """
         env = th_dict_repr.get("environment")
         if not env:
             return False
```
chfast: Doesn't it mean that the task does not require any environment?

golem/task/taskkeeper.py
```diff
@@ -52,9 +121,18 @@ def check_environment(self, th_dict_repr):
         return self.environments_manager.accept_tasks(env)
```
chfast, line 124: It would be better to allow Environment Manager to decide what it means "task is supported". It can be:
- is environment available?
- is environment enabled?
  etc.

golem/task/taskmanager.py
```diff
@@ -56,6 +57,8 @@ def __init__(self, node_name, node, listen_address="", listen_port=0, key_id="",
 
         self.use_distributed_resources = use_distributed_resources
 
+        self.comp_task_keeper = CompTaskKeeper()
+
```
chfast: Why it was added?

golem/transactions/incomeskeeper.py
```diff
@@ -79,6 +79,12 @@ def get_state(self, task_id, from_node):
         except ReceivedPayment.DoesNotExist:
             return None
 
+    @staticmethod
+    def get_waiting():
```
chfast: I call then "awaiting" payments. I want also avoid "pending" because in Ethereum it means transactions that has been passed to the network and wait for being included in a block.

tests/golem/transactions/ethereum/test_ethereumincomeskeeper.py
```diff
@@ -0,0 +1,22 @@
...
+from golem.tools.testwithdatabase import TestWithDatabase
+from golem.transactions.ethereum.ethereumincomeskeeper import EthereumIncomesKeeper
+
+
+class TestEthereumIncomesKeeper(TestWithDatabase):
+    def test_get_income(self):
+        ik = EthereumIncomesKeeper()
+        e = EllipticalKeysAuth()
+        node = e.get_key_id()
+        self.assertIsNone(ik.get_income(node, 0))
```
chfast, line 13: I would advice using plain `assert`. PyTest parses the failures very well. Like 

```
assert ik.get_income(node, 0) is None
```

chfast:
Looks ok. Only comments for considerations.

golem/task/taskkeeper.py
badb, line 111: https://github.com/imapp-pl/golem/issues/231

golem/task/taskmanager.py
badb: TaskKeeper was previously responsible for two things - keeping information of all tasks (task headers) in the network and keeping information about tasks computed on this machine. I split it into two classes: TaskHeaderKeeper and CompTaskKeeper. The information that TaskHeaderKeeper holds is temporary and may be dropped when Golem is closed. Information that CompTaskKeeper holds is permanent and should be saved in the database. Therefore I thought that CTK better suits TaskManager which holds other permanent information about tasks than TaskServer which probably shouldn't.

tests/golem/transactions/ethereum/test_ethereumincomeskeeper.py
badb, line 13: Why? Simple asserts are considered more pythonic and readlible, but unittest methods produces nicer error message and looks nicer to me.

golem/task/taskkeeper.py
badb, line 124: So your suggesting that we should rename methods (supported -> available) and (accept_tasks -> enabled) and then add method "supported" which checks both?

chfast, line 124: Something like that. And only `supported()` should be used by other modules. But that can be done in later changes.

badb, line 124: https://github.com/imapp-pl/golem/issues/235

PR #252, opened 2016-04-06
Luxtask

badb:


gnr/gnrapplicationlogic.py
```diff
@@ -313,18 +314,20 @@ def get_environments(self):
...
 
-    def _test_task_computation_finished(self, success, est_mem=0, error=""):
-        if success:
-            self.progress_dialog_customizer.show_message("Test task computation success!")
+    def _test_task_computation_success(self, results, est_mem):
+        self.progress_dialog_customizer.show_message("Test task computation success!")
+        if self.customizer.new_task_dialog_customizer:
+            self.customizer.new_task_dialog_customizer.test_task_computation_finished(True, est_mem)
+
+    def _test_task_computation_error(self, error=""):
```
chfast: Make `error` mandatory.

chfast:
What is LocalComputer? It looks very suspicious to me. Another convenient class that takes a lot of params and configures other objects?

And why it is not submitted as separated PR?

tests/gnr/test_gnrapplicationlogic.py
```diff
@@ -67,17 +75,17 @@ def test_run_test_task(self):
         task_type.task_builder_type.return_value = ttb
         logic.task_types["TESTTASK"] = task_type
         logic.run_test_task(ts)
-        time.sleep(0.5)
+        time.sleep(2)
```
chfast: You can wait in a loop for the expected result. See `wait_for` helper for Ethereum node.

badb: Oh, sorry, that was just an experiment and I forgot to change it back!

badb:
LocalComputer is suppose to run some docker containers when task that was defined on this machine requires it (to run benchmarks, tests, verify or join computation, whatever). Most of the code comes from our old TaskTester. It prepares directories, resources and callbacks  and starts DockerThreads. Do you have a better idea how to do it?

It's here because it was needed to run merge for luxtask (and wasn't really needed early) and right now is not used by other classes.

chfast:
I lack full picture. Is it possible to have single execution stack. I believe it should be possible to run benchmarks, test tasks and run others' task using the same infrastructure. We should not need TaskTester, TaskRunner, LocalComputer, etc.

Similar problem is with tasks  themselves, why the need to have BlenderTask, LuxTask, ... **and** BlenderTaskBuilder, LuxTaskBuilder, ...

chfast:
In general it looks good. You can merge it if you need it now. I can check details later if required.

PR #254, opened 2016-04-06
Speed up build

chfast:


chfast:
@badb @mfranciszkiewicz anyone?

badb:
How can we check if it's really working? What should happen if we run again CI tests on this branch, docker container won't be build again?

chfast:
They will be built faster.

PR #259, opened 2016-04-08
Continuation on Ethereum payments

chfast:


badb:
TestEthereumTransactionSystem.test_init fails - can't remove test dir - chaindata is used. Same problem with EthereuClientTest.test_client.

chfast:
Thanks. I will take care of that.

chfast:
Blocked by https://github.com/imapp-pl/golem/pull/279.

chfast:
@badb Can you test again on non-virtualized Windows?

golem/transactions/ethereum/paymentprocessor.py
```diff
@@ -58,28 +59,45 @@ def __init__(self, client, privkey):
         # TODO: Maybe it should not be the part of this class
         # TODO: Allow seting timeout
         # TODO: Defer a call only if payments waiting
-        scheduler = LoopingCall(lambda: self.sendout())
+        scheduler = LoopingCall(lambda: self.run())
```
badb: LoopingCall(self.run) is not working?

badb:
Seems to be working :+1:

golem/transactions/ethereum/paymentprocessor.py
chfast: Probably is. I always forget that in Python function is a value.

PR #261, opened 2016-04-08
IPFS optimizations

mfranciszkiewicz:


golem/core/fileshelper.py
```diff
@@ -56,6 +58,77 @@ def get_dir_size(dir_, report_error=lambda _: ()):
     return size
 
 
+def common_dir(arr, plus_sep=None, ign_case=None):
```
chfast, line 61: I would call it `commonprefix` as in Python 3.

golem/resource/ipfs/client.py
```diff
@@ -39,80 +38,100 @@ def read(self, count):
...
         if not os.path.exists(tmp_dir):
             os.makedirs(tmp_dir)
 
-        for member in tar_file:
-            tar_file.extract(member, tmp_dir)
+        with cls.lock:
+            if os.path.exists(dst_path):
+                os.remove(dst_path)
+
+        tar_file.extractall(tmp_dir)
```
chfast, line 108: Why not extracting directly to `dst_path`?

mfranciszkiewicz, line 108: Just to be extra safe and avoid collisions while extracting (tmp_dir is a uuid).

PR #268, opened 2016-04-12
Removed some more uses of changing renderer in blender task; Removed min subtask time

magdasta:


magdasta:
finished removing changing blender renderer, removed min subtask time

PR #275, opened 2016-04-14
Gui

badb:


chfast:
screenshot from 2016-04-20 18-04-48
1. The visual style is quite weird.
2. Left side buttons are bigger than the container. The container should be configured the way it has the size of contents (no sizing by hand).
3. "Advanced options" make the window suddenly much bigger, and when hidden the windows stays big (see picture). Can you consider placing "Show advanced options" button on the top of right pane (will be just button there by default)?

chfast:
Moreover, should you remove `admmain.py` and others to clean up the gnr a bit?

badb:
1. What do you suggest to make it less "weird"?
2. Fixed, check please. 
3. It shouldn't make the window much bigger. Can you send me some additional screenshots, how it looks at the beginning and what happens after showing advance option and then after hiding them?
4. Yeah, of cource, also `gnradmmain.py` and some other classes. But I was thinking about doing this as a separated task / PR, because it's a lot of work. Is it ok?

badb:
sreen different

I checked the version with task specific options on the right and advance on the bottom but it makes the window too high and is much less logical for me. 

Anyway, it's pretty strange why you have your form labels aligned to the right. In my app it's aligned to the left and it's set that way in the form layout.

PR #285, opened 2016-04-19
remove temporary files left by tests

badb:
For: https://github.com/imapp-pl/golem/issues/239

### DIFF
--- a/golem/tools/testwithappconfig.py
+++ b/golem/tools/testwithappconfig.py
@@ -1,6 +1,6 @@
 from mock import Mock
 from os import path
-from golem.core.keysauth import KeysAuth
+from golem.core.keysauth import KeysAuth, EllipticalKeysAuth
 from golem.tools.testdirfixture import TestDirFixture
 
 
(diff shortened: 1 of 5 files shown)


golem/tools/testwithappconfig.py
```diff
@@ -13,5 +13,7 @@ def setUp(self):
     def tearDown(self):
         if hasattr(KeysAuth, '_keys_dir'):
             del KeysAuth._keys_dir
+        if hasattr(EllipticalKeysAuth, '_keys_dir'):
+            del EllipticalKeysAuth._keys_dir
 
-        super(TestWithKeysAuth, self).tearDown()
+        TestDirFixture.tearDown(self)
```
chfast, line 19: What's wrong with `super()`. I usually change it in favor of `super()`.

badb, line 19: It was nothing wrong in this case, PeerSession was just a complicated issue and I thought that maybe there are some problems with mulitple inheritance somewhere (multiple inheritance is why I prefer this notation). Forgot to change it back.

PR #287, opened 2016-04-19
Resource manager - multiple path separators support (windows), tweaked retries

mfranciszkiewicz:


tests/golem/resource/ipfs/test_ipfs_resourcemanager.py
```diff
@@ -8,6 +9,14 @@
 from golem.tools.testdirfixture import TestDirFixture
 
 
+def list_endswith(longer, shorter):
```
badb: What was this function supposed to do?

golem/resource/ipfs/client.py
```diff
@@ -6,13 +6,14 @@
...
 from functools import wraps
 from threading import Lock
-from twisted.internet import threads
 from types import FunctionType
 
 import ipfsApi
 import requests
 from ipfsApi.commands import ArgCommand
 from ipfsApi.http import HTTPClient, pass_defaults
+from twisted.internet import reactor
```
badb: This reactor is imported too early and you cannot start a gui application because of that (try admmain.py or main.py).

PR #288, opened 2016-04-20
Case insensitive file formats

magdasta:


gnr/task/scripts/docker_luxtask.py
```diff
@@ -34,7 +34,7 @@ def find_flm(directory):
     try:
         for root, dirs, files in os.walk(directory):
             for names in files:
-                if names[-4:] == ".flm":
+                if names.upper().endswith() == ".FLM":
```
badb: `endswith` takes at least one argument

PR #289, opened 2016-04-20
Remove dead code touching QApplication

chfast:


chfast:
This simple change cleans up the code around QApplication. Do you want it @badb?

PR #311, opened 2016-04-26
Prepare for making Ethereum Payments public.

chfast:
Public repo: https://github.com/imapp-pl/ethereum-payments.


chfast:
@badb, can we include that in tests? It just moves some files (as can be seen in https://github.com/imapp-pl/ethereum-payments).

PR #312, opened 2016-04-26
Active peers table in Main window -> Status

mfranciszkiewicz:


chfast:
The code looks very nice.

golem/client.py
```diff
@@ -437,6 +437,9 @@ def get_status(self):
             msg += "Budget: {}\n".format(self.transaction_system.budget)
         return msg
 
+    def get_peers(self):
+        return self.p2pservice.peers.values()[:]
```
badb: What is `[:]` for?

mfranciszkiewicz: This shallow copy is unnecessary. Thanks for pointing that out.

PR #319, opened 2016-05-02
Lux blend benchmarks

magdasta:


gnr/benchmarks/benchmarkrunner.py
```diff
@@ -0,0 +1,61 @@
...
+                               True, BenchmarkRunner.RUNNER_WARNING,
+                               BenchmarkRunner.RUNNER_SUCCESS)
+        # probably this could be done differently
+        self.benchmark = benchmark
+        
+    def _get_task_thread(self, ctd):
+        if ctd.docker_images:
+            return LocalComputer._get_task_thread(self, ctd)
+        else:
+            #TODO not allow running tasks outside docker?
```
badb: Yeah, we should probably run only docker tasks as benchmarks.

tests/gnr/benchmarks/test_benchmark.py
```diff
@@ -0,0 +1,58 @@
...
+from gnr.task.blenderrendertask import BlenderRendererOptions
+
+from gnr.renderingdirmanager import get_benchmarks_path
+
+class TestBlenderBenchmark(unittest.TestCase):
+    def setUp(self):
+        self.benchmark = Benchmark()
+    
+    def verify_log_helper(self, file_content):
+        with tempfile.NamedTemporaryFile() as file_:
```
badb: Can you use `TempDirFixture` for this test and create temp file `self.tempdir`?

badb:
Benchmarks in the preferences tab are correctly run and their new values are saved properly in the config file. But if user click "change button" which is supposed to save a new config, the benchmarks value is reset to zero and those zeros are save in the config file.

badb:
It seems like there are some problems with test_paymentprocessor.py. I'm guessing that this branch is just too old. @magdasta Can you rebase develop or merge it into this branch and then check if payment test is still an issue?

PR #322, opened 2016-05-04
Timeout

badb:


chfast:
I don't think it is a good change. It's too chaotic.
1. Name are wrong. Timeout is a time duration, like 30 min. Deadline is absolute point in time.
2. A static description of a task should have timeout set (if user wants it).
3. Timeout changes into deadline when the user starts a task. Deadline is what the network knows.

badb:
Hm, you're probably right with the names. I'll change it later. 

What do you mean by static description? Task definition (like in gt file)?
1) User defines timeouts for his task and subtasks. 
2) When user start a task, a task timeout is changed to deadline. Subtask timeout is not changed. 
3) When a computing user gets a subtask to compute than subtask timeout is changed to deadline. 
Do we agree on this?

chfast:
Except 3). Computing node gets a subtask with a deadline. That's is actually independent on task having timeout/deadline at all. 

I would suggest implementing good 3) first. And leaving tasks without any time constraints.

badb:
1. Names are fixed, points in times are now deadlines. 
2. All deadlines are in UTC. 
3. Computing node get a subtask with a deadline. 
4. Task state "failure" is changed into "timeout", so user would now what really happened.

Issue #212, opened 2016-03-17
Valgrind returns errors for c++ taskcollector


magdasta:
All of the errors say "Conditional jump or move depends on uninitialised value(s)". Seems like a problem with FreeImage library or a valgrind bug (? https://bugs.launchpad.net/ubuntu/+source/valgrind/+bug/1094586). We should investigate if this is dangerous.

[log.txt](https://github.com/imapp-pl/golem/files/177944/log.txt)

badb:
@banasiakadam60 Is this one finished?

banasiakadam60:
Yes, it can be closed

PR #326, opened 2016-05-05
Valgrind returns errors for c++ taskcollector #212

banasiakadam60:
Remove passing temporary values by reference

### DIFF
--- a/gnr/taskcollector/TaskCollector.cpp
+++ b/gnr/taskcollector/TaskCollector.cpp
@@ -88,7 +88,7 @@ void FreeImageErrorHandler(FREE_IMAGE_FORMAT fif, const char *message) {
 	if (fif != FIF_UNKNOWN) {
 		std::cerr << FreeImage_GetFormatFromFIF(fif) << " Format\n";
 	}
-	std::cerr << message << " ***\n";
+	std::cerr << message << " ***" << std::endl;
 }
 
 // ----------------------------------------------------------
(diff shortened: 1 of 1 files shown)


gnr/taskcollector/TaskCollector.cpp
```diff
@@ -88,7 +88,7 @@ void FreeImageErrorHandler(FREE_IMAGE_FORMAT fif, const char *message) {
 	if (fif != FIF_UNKNOWN) {
 		std::cerr << FreeImage_GetFormatFromFIF(fif) << " Format\n";
 	}
-	std::cerr << message << " ***\n";
+	std::cerr << message << " ***" << std::endl;
```
chfast, line 91: Actually `\n` is better because `std::endl` additionally and unnecessary flushes the stream.

gnr/taskcollector/TaskCollector.cpp
```diff
@@ -109,20 +109,20 @@ class TaskCollector {
 	}
 	~TaskCollector() = default;
 
-	bool addImgFile(const std::string& pathName, int flag = 0)  {
+	bool addImgFile(const std::string pathName, int flag = 0)  {
```
chfast, line 112: Does it actually fix the issue reported by valgrind?

In any case `const` is misleading when passing argument by value.

gnr/taskcollector/TaskCollector.cpp
```diff
@@ -307,35 +307,35 @@ int main(int argc, char *argv[]) {
...
-        
+
+    //to be sure the ordering is proper
+    std::sort(argv + 3, argv + argc);
+
 	for (int i = 3; i < argc; ++i) {
 		if (std::string(argv[i]).find("Alpha") == std::string::npos) {
 			if (!taskCollector->addImgFile(argv[i])) {
-				std::cerr << "Can't add file: " << argv[i] << "\n";
+				std::cerr << "Can't add file: " << argv[i] << std::endl;
```
chfast, line 320: Here `'\n'` can be used instead of `"\n"`.

chfast:
You probably should use https://github.com/imapp-pl/golem clone to make pr (just use your own branch) instead of using a fork in https://github.com/banasiakadam60/golem.

E.g. CircleCI does not build PR from forks by default.

gnr/taskcollector/TaskCollector.cpp
banasiakadam60, line 112: As you can see, this function is called from main. Parameters are taken from argv. This mean, char\* is casted to the temporary std::string and the reference to this temporary object is passed to addImgFile. So we need to call copy ctor.

> In any case const is misleading when passing argument by value.

Why do you think so? It just protect from accidentally change of the value of param. If you want only read it, IMO, it should be marked as const.

gnr/taskcollector/TaskCollector.cpp
banasiakadam60, line 91: Yes, you're right.

gnr/taskcollector/TaskCollector.cpp
banasiakadam60, line 320: Right.

banasiakadam60:
@chfast Ok, commit now available here: https://github.com/imapp-pl/golem/pull/345
This one can be abandoned,

PR #329, opened 2016-05-05
Docker security enhancements + container and VM resource management

mfranciszkiewicz:


golem/docker/machine/machine_manager.py
```diff
@@ -0,0 +1,386 @@
...
+            status = self.docker_machine_command('status')
+            status = status.strip().replace("\n", "")
+            return status == 'Running'
+        except Exception as e:
+            logger.error("DockerMachine: failed to check docker-machine status: {}"
+                         .format(e.message))
+        return False
+
+    def __docker_machine_images(self):
+        command = self.docker_machine_command('list')
```
badb: `docker_machine_command` call the command so here you receive a result of a command and run the result.

requirements.txt
```diff
@@ -1,6 +1,7 @@
 appdirs>=1.4
 click>=4.0
-docker-py>=1.6.0
+docker-py==1.7.2
```
chfast, line 3: How about `>=`?

mfranciszkiewicz, line 3: Newest version breaks client API version compatibility on CircleCI

requirements.txt
```diff
@@ -20,4 +21,4 @@ Twisted>=15.4
 ethereum-rpc-client
 devp2p>=0.5.1
 cffi>=1.1.2
-https://github.com/ethereum/pyethereum/tarball/develop#egg=ethereum
+ethereum>=1.2.0
```
badb: Can we change this to ethereum>=1.1.0 
Higher versions still aren't working on windows.

chfast: You can. In case some tests fail I will take care of that later.

PR #332, opened 2016-05-10
Multiple seeds

mfranciszkiewicz:


golem/model.py
```diff
@@ -167,3 +170,16 @@ class NeighbourLocRank(BaseModel):
     class Meta:
         database = db
         primary_key = CompositeKey('node_id', 'about_node_id')
+
+
+class KnownHosts(BaseModel):
+    ip_address = CharField()
```
chfast, line 176: I think we should keep the address as a single string (e.g. 172.0.0.1:8888). DB does not need to know the pieces of the address.

golem/network/p2p/p2pservice.py
```diff
@@ -77,11 +79,24 @@ def new_connection(self, session):
...
+
         try:
-            socket_address = SocketAddress(self.config_desc.seed_host, self.config_desc.seed_port)
-            self.connect(socket_address)
-        except AddressValueError, err:
-            logger.error("Invalid seed address: " + err.message)
+            with db.transaction():
+                self.__remove_redundant_hosts_from_db()
+                hosts = KnownHosts.select()
+                self.__sync_seeds(hosts)
```
chfast: Does it have to be a single transaction?

golem/network/p2p/p2pservice.py
```diff
@@ -92,6 +107,28 @@ def connect(self, socket_address):
                                       P2PService.__connection_failure)
         self.network.connect(connect_info)
 
+    def add_known_peer(self, node, ip_address, port):
+        is_seed = node.is_super_node() if node else False
+
+        try:
+            with db.transaction():
+                KnownHosts.delete().where(
```
chfast, line 136: Here you want "insert or replace" SQL query. Maybe peewee supports it.
More: http://stackoverflow.com/questions/3634984/insert-if-not-exists-else-update.

golem/network/p2p/p2pservice.py
```diff
@@ -196,11 +239,11 @@ def remove_peer(self, peer_session):
             pc.status = PenConnStatus.Failure
             self._remove_pending_sockets(pc)
 
-        for p in self.peers.keys():
-            if self.peers[p] == peer_session:
-                del self.peers[p]
-                self.peer_order.remove(p)
-                self.suggested_address.pop(p, None)
+        for key_id in self.peers.keys():
```
chfast: Usually I'm against mixing function and non-functional changes in a single PR.

golem/network/p2p/p2pservice.py
```diff
@@ -788,11 +829,27 @@ def __sync_peer_keeper(self):
         if peers_to_find:
             self.send_find_nodes(peers_to_find)
 
+    def __sync_seeds(self, hosts=None):
+        if hosts:
+            self.seeds = filter(lambda x: x.is_seed, hosts)
+        else:
+            with db.transaction():
```
chfast: You don't have to surround every query with `db.transaction()`. It is useful only if you want multiple queries be _all or nothing_.

golem/network/p2p/p2pservice.py
```diff
@@ -92,6 +107,28 @@ def connect(self, socket_address):
                                       P2PService.__connection_failure)
         self.network.connect(connect_info)
 
+    def add_known_peer(self, node, ip_address, port):
```
chfast, line 131: Any changes we can add unit test for that function?

golem/network/p2p/p2pservice.py
mfranciszkiewicz, line 136: Yes, peewee supports upsert queries. The thing is, it matches all data, not only the required ip address and port.

golem/network/p2p/p2pservice.py
mfranciszkiewicz: It's also useful when the transaction isolation level is not known (the underlying db engine will probably change), and we want to read only committed data.

PR #366, opened 2016-05-17
Changes to docker image caching #2

mfranciszkiewicz:


badb:
It looks like it's a little bit quicker, but there may be problem if we change Dockerfiles. But we will need a better solution for docker management, anyway, so I guess it's sufficient for now.

PR #396, opened 2016-05-30
Trading system prototype #5

banasiakadam60:
Proposal for a new trading system

### DIFF
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -3,6 +3,7 @@ import logging
 import os
 import struct
 import time
+from threading import Timer
 
 from golem.network.transport.message import MessageHello, MessageRandVal, MessageWantToComputeTask, \
     MessageTaskToCompute, MessageCannotAssignTask, MessageGetResource, MessageResource, MessageReportComputedTask, \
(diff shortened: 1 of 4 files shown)


golem/task/tasksession.py
```diff
@@ -322,20 +326,15 @@ def send_nat_punch(self, asking_node, dest_node, ask_conn_id):
...
-            self.send(MessageTaskToCompute(ctd))
+        if trust < self.task_server.config_desc.computing_trust:
+            warn = "Client not trusted: {} need at least {}".format(trust, self.task_server.config_desc.computing_trust)
+            logger.error(warn)
+            self.send(MessageCannotAssignTask(msg.task_id, warn))
         else:
-            self.send(MessageCannotAssignTask(msg.task_id, "No more subtasks in {}".format(msg.task_id)))
+            self.waiting_clients.append(msg)    # add node to waiting list
+            if not self.is_launched:            # if timer is not launched
+                Timer(120 - self.task_server.config_desc.trading_ratio, self.__select_offer, ()).start()  # launch it
```
badb, line 336: Maybe it would be better to user deferLater or other twisted method? I'm affraid of some synchronization problem and parallel access (and you don't use any locks here).

golem/task/tasksession.py
```diff
@@ -47,6 +48,9 @@ def __init__(self, conn):
 
         self.result_owner = None  # information about user that should be rewarded (or punished) for the result
 
+        self.waiting_clients = []   # list of clients waiting for a task
```
badb, line 51: Session is only open between two specific clients. There won't be any other clients in this class.

Issue #333, opened 2016-05-10
Change testing window layout


badb:
OK button is misleading. We should remove the possibility to close testing window while test isn't finished or add posibility to stop it. 
We should also remove possibility to close the window by clicking "x" in right corner.

banasiakadam60:
BTW. Maybe I will also disable 'ok' buttons on tests in 'Provider' tab?

badb:
Good idea!

banasiakadam60:
Can it be closed?

PR #406, opened 2016-06-07
Change testing window layout #333

banasiakadam60:
Disable 'ok' button when test is in progress and remove 'x' button

### DIFF
--- a/gnr/gnrapplicationlogic.py
+++ b/gnr/gnrapplicationlogic.py
@@ -295,6 +295,8 @@ class GNRApplicationLogic(QtCore.QObject):
 
             self.progress_dialog = TestingTaskProgressDialog(self.customizer.gui.window)
             self.progress_dialog_customizer = TestingTaskProgressDialogCustomizer(self.progress_dialog, self)
+            self.progress_dialog_customizer.gui.window.setWindowFlags(QtCore.Qt.Dialog | QtCore.Qt.FramelessWindowHint)
+            self.progress_dialog_customizer.button_enable(False)    # disable 'ok' button
             self.progress_dialog.show()
 
             self.tt.run()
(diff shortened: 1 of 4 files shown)


gnr/gnrapplicationlogic.py
```diff
@@ -295,6 +295,8 @@ def run_test_task(self, task_state):
 
             self.progress_dialog = TestingTaskProgressDialog(self.customizer.gui.window)
             self.progress_dialog_customizer = TestingTaskProgressDialogCustomizer(self.progress_dialog, self)
+            self.progress_dialog_customizer.gui.window.setWindowFlags(QtCore.Qt.Dialog | QtCore.Qt.FramelessWindowHint)
```
badb: The problem with frameless windows is that it's can be quite invisible. Is it possible to add frame, but with closing button not enabled?
testing

banasiakadam60: I know it looks awful, but you cannot just disable this button. It depends on window manager, not qt. Only thing, which can be done is to overwite event closeEvent, which is raised when user click 'x'. If you do this, button is still enable, but nothing happens when 'x' is clicked, so it may look like the suspension of the application. And of course, inheritance of QDialog is needed to do this, and at this moment, we have an object of this type.
[Here is a topic about this](http://stackoverflow.com/questions/18256459/qdialog-prevent-closing-in-python-and-pyqt)

badb: Hm. So let's keep the "x" but let's make testing button not enabled earlier (before test is started). 

Later we can show a warning after x is pressed and kill the test (but there's no easy method to kill docker computation before timeout right now).

badb:
It would be good to block "test task" button just after the test is started (so if someone close testing window, he won't be able to run test again). Same should be done with the recount button. 
Or maybe even whole "new task" form should be disable while testing took place.

banasiakadam60:
@badb OK, working on it. But it will force user to restart the application. Is it correct?

badb:
Only if the test hang out completely. But then it may be impossible to run another test (because test folder will be blocked), so user will have to restart Golem to add a new task anyway.

banasiakadam60:
OK, thanks for the explanation :)

badb:
Are you able to add some tests quicklyl and resolve the conflicts? It would be nice to merge this branch soon.

banasiakadam60:
Can I merge this branch? After doing this, I'll send the tests.

badb:
Just merge develop into this branch - make just step 1 form resolving conflicts instruction and then push the results here.

banasiakadam60:
I'm not able to debug this bug. Each time I got Segfault. I'll revert commit with tests.

PR #409, opened 2016-06-13
Restart

badb:


mfranciszkiewicz:
There are issues with restarting tasks / sub-tasks:
- for a partially computed task:
  - restarting a task does not reset the preview box
  - a certain amount of time has to pass before sending new sub-tasks (timeout-related issue?)
  - restarting a computed sub-task changes the corresponding preview part to a red stripe (internally marked as failed?), computing nodes are unable to retrieve new sub-tasks
- for a fully computed task:
  - restarting a sub-task changes the progress value, but the preview remains intact (not sure if this is intentional)
  - task requests are rejected by the requestor (`No more subtasks in ...` on the provider's side)

Issue #290, opened 2016-04-20
Scale image preview


badb:
Right now there is a fixed preview size set and cannot be changed. It look bad for smaller images or images with completely different ratio.

PR #421, opened 2016-06-24
Scale image preview #290

banasiakadam60:


gnr/ui/appmainwindow.py
```diff
@@ -20,15 +20,22 @@ def __init__(self):
...
+        """
         if tab_name.lower() == 'new_task':
             self.ui.testTaskButton.setEnabled(enable)
-            self.ui.finishButton.setEnabled(enable)
             self.ui.showAdvanceNewTaskButton.setEnabled(enable)
             self.ui.addResourceButton.setEnabled(enable)
             self.ui.saveButton.setEnabled(enable)
             self.ui.loadButton.setEnabled(enable)
             self.ui.taskTypeComboBox.setEnabled(enable)
+            if success:
```
badb: It's not necessary. If the task is finished properly "start task" button will made enabled by other function from customizer. 
It's also a good practice to keep separate issues in separate commits.

badb:
Seems to behave nicely! But maybe we should remove the frame from the preview?

gnr/ui/appmainwindow.py
banasiakadam60: Sorry, I pushed it here by mistake.

banasiakadam60:
OK, removed.

banasiakadam60:
You meant something like that?

gnr/ui/appmainwindow.py
```diff
@@ -15,15 +15,21 @@ def __init__(self):
...
 
         self.ui.setupUi(self.window)
+        self.ui.previewLabel.setFrameStyle(QFrame.NoFrame)
         self.ui.previewLabel.setPixmap(QPixmap(path.join(get_golem_path(), "gnr", "ui", "nopreview.png")))
 
     def show(self):
         self.window.show()
 
-    def setEnabled(self, tab_name, enable):
+    def setEnabled(self, tab_name, enable, success=True):
```
badb: "success" parameter is no longer needed.

badb:
Yes, looks better in my opinion.

chfast:
The code looks good to me too. If everybody happy let's merge that and move on.

PR #429, opened 2016-06-27
Prepare task for serialization

badb:
Remove lambdas and functions from task variables to make possible to serialize a task instance. 
Serialized tasks may be later saved in the database and resumed after restart.

### DIFF
--- a/golem/task/taskbase.py
+++ b/golem/task/taskbase.py
@@ -1,8 +1,11 @@
 import abc
+import logging
 import time
 from copy import deepcopy
 
 from golem.core.variables import APP_VERSION
+
+logger = logging.getLogger(__name__)
 
 
 class TaskHeader(object):
@@ -106,6 +109,7 @@ class Task(object):
             if self.listeners[i] is listener:
                 del self.listeners[i]
                 return
+        logger.warning("Trying to unregister listener that wasn't registered.")
 
     @abc.abstractmethod
     def initialize(self, dir_manager):
--- a/golem/task/taskbase.py
+++ b/golem/task/taskbase.py
@@ -91,9 +91,9 @@ class Task(object):
         self.listeners = []
 
     def __getstate__(self):
-        state = deepcopy(vars(self))
-        del state['listeners']
-        return state
+        state_attr = vars(self).keys()
+        state_attr.remove('listeners')
+        return {attr: deepcopy(getattr(self, attr)) for attr in state_attr}
 
     def __setstate__(self, dict_):
         self.__dict__ = dict_
@@ -180,7 +180,6 @@ class Task(object):
         :return bool: True if task passed verification, False otherwise
         """
         return  # Implement in derived class
-
 
 
     @abc.abstractmethod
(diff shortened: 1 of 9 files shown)


golem/task/taskbase.py
```diff
@@ -1,8 +1,12 @@
-import time
 import abc
+import logging
+import time
+from copy import deepcopy
 
 from golem.core.variables import APP_VERSION
 
+logger = logging.getLogger(__name__)
```
chfast: Please give it a explicit name. E.g. "golem.task".

golem/task/taskbase.py
```diff
@@ -56,7 +60,15 @@ def __init__(self):
         self.docker_images = None
 
 
-class Task:
+class TaskEventListener(object):
+    def __init__(self):
```
chfast, line 64: Is the empty constructor needed? Is it abstract class?

golem/task/taskbase.py
```diff
@@ -76,7 +88,28 @@ def __init__(self, header, src_code):
...
+        self.__dict__ = dict_
+        self.listeners = []
+
+    def register_listener(self, listener):
+        assert isinstance(listener, TaskEventListener)
+        self.listeners.append(listener)
+
+    def unregister_listener(self, listener):
+        assert isinstance(listener, TaskEventListener)
+        for i in range(len(self.listeners)):
```
chfast: Sorry, but no!

Aggressive version would be `self.listeners.remove(listener)`.
Defensive version would be `if listener in self.listeners: self.listeners.remove(listener)`.

golem/task/taskmanager.py
```diff
@@ -27,11 +28,21 @@ def subtask_status_updated(self, subtask_id):
         pass
 
 
+class TMTaskEventListener(TaskEventListener):
```
chfast: This seems to be overdesigned. Unless there are issues with multiple inheritance in this case, can you make the TaskManager the TaskEventListener? It is enough to rename `notice_task_update()` to `notify_update_task()`.

golem/task/taskmanager.py
```diff
@@ -60,6 +71,7 @@ def __init__(self, node_name, node, listen_address="", listen_port=0, key_id="",
         self.listeners = []
         self.activeStatus = [TaskStatus.computing, TaskStatus.starting, TaskStatus.waiting]
         self.use_distributed_resources = use_distributed_resources
+        self.task_events_listener = TMTaskEventListener(self)
```
chfast: Not needed if TaskManager would implement TaskEventListener interface. See previous comment.

golem/task/taskbase.py
```diff
@@ -76,7 +88,26 @@ def __init__(self, header, src_code):
         self.header = header
         self.undeletable = []
 
-        self.notify_update_task = lambda task_id: None
+        self.listeners = []
+
+    def __getstate__(self):
+        state = deepcopy(vars(self))
+        del state['listeners']
```
mfranciszkiewicz: Could we avoid (deep)copying listeners?

PR #437, opened 2016-06-29
change check_subtask_wrapper into react_to_key

badb:
Change decorator that checks existance of object in dictionary to the one that catches the key error.


chfast:
Can we have single `react_to_key_error`? Also, better name might be `handle_key_error`.

Can be merged.

Issue #428, opened 2016-06-27
Change log management


badb:
Log files are growing too quickly. Possible solutions:
- turn off debug message
- save errors and warnings into separate files
- don't put computation logs in the golem log
- change golem log file more often 
- ... ?

chfast:
Some suggestions:
1. I'm not sure it is standard or not, but I've seen and like the config in form of `":warning;golem.*:info;golem.pay:debug"`.
2. Our logging.ini config file sucks. I tried to use it many times but still cannot figure out how it works if at all. Can we replace that with something usable?
3. Let's have "info" level by default.

PR #441, opened 2016-06-30
Logging

badb:
https://github.com/imapp-pl/golem/wiki/Logging

Resolves: #428


gnr/main.py
```diff
@@ -12,7 +13,10 @@
 @click.command()
 @click.option('--datadir', '-d', type=click.Path())
 def main(datadir):
-    config_logging()
+    if datadir:
```
chfast: Can you move it to golemapp.start()?

gnr/logging.ini
```diff
@@ -1,28 +1,39 @@
...
+level: WARNING
+handlers: consoleHandler
+
+[logger_golem]
+level: INFO
+qualname: golem
+handlers: fileHandlerWithTimeRot
+
+[logger_pay]
+level: DEBUG
```
chfast: golem.pay does not have to be DEBUG. Is it here as an example?

tests/gnr/task/test_gnrtask.py
```diff
@@ -29,6 +29,7 @@ def test_gnr_task(self):
             self.assertEqual(task.get_results(subtask_id), False)
         self.assertTrue(any(["not my subtask" in log for log in l.output]))
 
+
```
chfast, line 22: Ooops.

chfast:
Level of golem.pay is to high by default. Can be merged after fixed.

gnr/logging.ini
badb: Should I comment it out or remove?

chfast: Change to INFO.

chfast:
After merging in `handle_key_error` changes I have problem with `test_gnrtask.py`. I'm not sure which version is final as the same changes are applied many times in different branches. Anyway, functions like `GNRTask.get_stdout()` do not throw KeyErrors but have `handle_key_error` modifier.

Please remove my last two commits if needed.

badb:
Fixed. Those functions should not have `handle_key_error` modifier (subtask_id is checked on the higher level).

Issue #398, opened 2016-05-30
Better names for frames


badb:
When user define output file name to "outputfile.exr" then Blender frames will have names like `ouptufile0001.exr`. Therefore we may not support frames scenes that have more then 9999 frames. There should be possibility to define the number of leading zeros.

PR #442, opened 2016-06-30
Better names for frames

banasiakadam60:
Change padding in filenames


chfast:
You can link PR with issues by:
Closes #398.

gnr/customizers/renderingmainwindowcustomizer.py
```diff
@@ -155,8 +155,8 @@ def __set_preview(self, t):
     def __get_frame_name(self, definition, num):
         output_name, ext = os.path.splitext(definition.output_file)
         frame_num = definition.renderer_options.frames[num]
-        output_name += str(frame_num).zfill(4)
-        return output_name + ext
+        return "{}_{}_{}{}".format(output_name, num,
+                                   str(frame_num).zfill(len(str(len(definition.renderer_options.frames)))), ext)
```
badb: Uhm, it's not a very good approach. You're taking into account only the number of frames which may have nothing in common with frames numbers. For example we may have 3 frames with numbers [11232, 13134, 23131]. 
So we should take into account the maximum numer frame, not number of frames. 
And the best approach is to allow user to define this format (number of leading zeros).

gnr/task/scripts/docker_blendertask.py
```diff
@@ -45,13 +45,13 @@ def run_blender_task(outfilebasename, scene_file, script_src, start_task,
...
         print(cmd, file=sys.stderr)
         exit_code = exec_cmd(cmd)
         if exit_code is not 0:
             sys.exit(exit_code)
 
 
-run_blender_task(params.outfilebasename, params.scene_file, params.script_src,
-                 params.start_task, params.frames, params.output_format)
+run_blender_task(params.outfilebasename, params.scene_file, params.script_src, params.start_task,
+                 params.frames, params.output_format, params.total_tasks if params.total_tasks else 4)
```
badb: Also name total tasks may be very misleading here. It may have nothing in common with number of frames.

badb:
It's very unclear what new options does. Some explanations and better default value is a must. 
Additionally, it's a very specific options strongly connected with "frame render tasks". So it should be in task specific options (and only for Blender). Also leading zero should not be part of task_definition but element in options (renderer_options) in definition.

banasiakadam60:
OK, but what should be the default value? At this moment it set to 0 and defined as `max(definition.leading_zeros, max(frame_numbers))` in case of default (0) or incorrect (too low) value.

badb:
I'm not sure (3 or 4)? My problem with default 0 is that I'm not sure if I should write "000" or "3". 
Potential problem with `max(definition.leading_zeros, max(frame_numbers))` is that you're calling it two times: in customizer (ok) and in blender task script. Blender task script works only on small subset of frames, so it this two methods may generate completelly different results.
It may not be a problem - there's a chance that blender finds max frame number in the full scene, so the final file will have proper name, but please check it.

banasiakadam60:
I've checked it yesterday and it looks fine. But ofc I'll check it one more time :)
Maybe I'll add some additional checkbox to determine whether the user wants to define padding, or wants it to be generated automaticly (basing on the frames numbers)?

badb:
As you wish. I don't like the idea of additional checkbox, but I'm not sure what's better.

badb:
Ok, checkbox looks nice. One additional suggestion: maybe you can change name 'leading zeros' to 'pad to length' or something like that? 
Problems:
- Subtasks results are named properly, but final pictures have old names. For example part of the subtask result has name frame_5_00005.exr, but final picture has names frame005.exr. 
- Subtasks numbers in final frame numbers are unnecessary (and sometimes plain wrong), so output file name for that task should be frame_00005.exr and not frame_5_00005.exr. This subtask number is only userd to distinct different part of frames received from computing nodes.

badb (changes requested):
Now it's another way around. Whole frame has proper name, but part of the frames are named as a whole frame and therefore override one another producing final empty frame. To reproduce: choose more subtasks than frames, eg. frames 3-4, number of subtasks 10.

gnr/customizers/renderingmainwindowcustomizer.py
```diff
@@ -157,7 +157,7 @@ def __get_frame_name(self, definition, num):
         output_name, ext = os.path.splitext(definition.output_file)
         frame_num = definition.renderer_options.frames[num]
         padding = max(len(str(max(definition.renderer_options.frames))), definition.renderer_options.pad_to_length)
-        return "{}_{}{}".format(output_name, str(frame_num).zfill(padding), ext)
+        return "{}_{}_{}{}".format(output_name, num, str(frame_num).zfill(padding), ext)
```
badb: Final frame have name just like "abc_00007.png". Frame is saved correctly, but in the GUI it's displayed like "abc_0_00007.png" which doesn't exist.

gnr/ui/BlenderWidget.ui
```diff
@@ -145,27 +145,70 @@
...
+       </item>
+      </layout>
+     </item>
+     <item row="6" column="0">
+      <widget class="QCheckBox" name="paddingCheckBox">
+       <property name="enabled">
+        <bool>false</bool>
+       </property>
+       <property name="toolTip">
+        <string>Set number of leading zeros in frames name</string>
```
badb: I still have a feeling that this description is misleading. More precisely we're padding the frame number in frame name with leading zeros to the given width.

badb:
I have a few more ideas how to improve it (maybe we can replace padding GUI elements with Blender-like "#" notation in output names. I'll try them out soon.

badb (dismissed):

badb:
Alternative approach. User define output file name for frames with "#", ex. "output_image_#####_ver_1.PNG". It's compatibile with Blender approach, so it may be more natural for the users. If you don't like it and prefare previous version with GUI, feel free to reverse the last commit.

mfranciszkiewicz (approved):

Issue #438, opened 2016-06-30
Allow setting Golem Client name


chfast:
Setting the name could be helpful to identify tester in the network.

badb:
Note: we should remember to change `node_name` variable in `gnrapplicationlogic` (if #444 is merged).

PR #444, opened 2016-06-30
WebSocket RPC + GUI in a separate process

mfranciszkiewicz:
Expected future improvements on RPC:
- reducing the size and complexity of exchanged objects
- choosing of a more lightweight and platform-independent serialization method (e.g. common to web apps)
- keyring / key auth implementation (basic support included)

GUI changes proposed:
- preview updates should be performed in a separate process

Also, a better process monitor might be required.


chfast:
Only if you know how to do it, you could squeeze some of the WIP commits to make the history look better.

badb:
We must change `start task` method just after this update, start task just after add (and remove start task button and context menu method) or made button and method not enabled just after click.

gnr/customizers/configurationdialogcustomizer.py
```diff
@@ -4,6 +4,7 @@
 
 from PyQt4 import QtCore
 from PyQt4.QtGui import QMessageBox, QPalette
+from twisted.internet.defer import inlineCallbacks
```
badb: This isn't used.

gnr/gnrapplicationlogic.py
```diff
@@ -231,9 +293,10 @@ def _get_builder(self, task_state):
         if hasattr(task_state.definition, "renderer"):
             task_state.definition.task_type = task_state.definition.renderer
 
-        return self.task_types[task_state.definition.task_type].task_builder_type(self.client.get_node_name(),
-                                                                                  task_state.definition,
-                                                                                  self.client.datadir)
+        builder = self.task_types[task_state.definition.task_type].task_builder_type(self.node_name,
```
badb, line 296: We would like to have "change_node_name" function in the future available and `node_name` is set only at the beginning. It's not really important right now, just a look-out for the future.

golem/client.py
```diff
@@ -162,21 +178,27 @@ def start_network(self):
         self.task_server.task_manager.register_listener(ClientTaskManagerEventListener(self))
         self.p2pservice.connect_to_network()
 
+        self.do_work_task.start(0.1, False)
```
badb: Can you move it to `start` function?

tests/golem/core/test_processmonitor.py
```diff
@@ -0,0 +1,100 @@
...
+
+        if all_stopped:
+            break
+        else:
+            time.sleep(0.5)
+
+
+class TestProcessMonitor(unittest.TestCase):
+
+    def test_monitor(self):
```
badb, line 50: On windows I get an error: 

> PicklingError: Can't pickle <function run_exit at 0x02CBDFB0>: it's not found as test_processmonitor.run_exit

gnr/gnrapplicationlogic.py
mfranciszkiewicz, line 296: Will `change_node_name` be a logic's method or a part of `change_config`? If so, there will be no problem keeping it up to date. Inserting a yield / deferred call here complicates testing.

badb, line 296: Part of `change_config` probably, but I'm not sure yet. Anyway, it's all right, I've just added a node to #438 to remember about this.

PR #445, opened 2016-07-01
Netcode: TCP server and dummy task test updates

mfranciszkiewicz:
Main change:
- private IP address connection attempts are limited to reachable private networks (time required to establish a connection is reduced)

Misc. changes in dummy task tests:
- redundant code removed
- return code check fixed (may return 0)

### DIFF
--- a/golem/core/hostaddress.py
+++ b/golem/core/hostaddress.py
@@ -1,10 +1,9 @@
-import socket
-import os
 import logging
 import netifaces
+import os
+import socket
 
 import ipaddress
-import struct
 import stun
 
 from variables import DEFAULT_CONNECT_TO, DEFAULT_CONNECT_TO_PORT
@@ -70,23 +69,6 @@ ip4Addresses = ip_addresses
 get_host_addresses = ip_addresses
 
 
-def __parse_ipv4_networks(networks):
-    result = []
-    for net_class in networks:
-        ip_net = ipaddress.ip_network(net_class)
-        result.append((int(ip_net.network_address), int(ip_net.netmask)))
-    return result
-
-
-__ipv4_private_networks = [
-    u'10.0.0.0/8',
-    u'127.0.0.0/8',
-    u'172.16.0.0/12',
-    u'192.168.0.0/16'
-]
-__parsed_ipv4_private_networks = __parse_ipv4_networks(__ipv4_private_networks)
-
-
 def ip_address_private(address):
     if address.find(':') != -1:
         try:
@@ -96,16 +78,11 @@ def ip_address_private(address):
                          .format(address, exc.message))
             return False
     try:
-        a = struct.unpack('!I', socket.inet_aton(address))[0]
+        return ipaddress.IPv4Address(unicode(address)).is_private
     except Exception as exc:
         logger.error("Cannot parse IPv4 address {}: {}"
                      .format(address, exc.message))
         return False
-
-    for n_addr, n_mask in __parsed_ipv4_private_networks:
-        if (a & n_mask) == n_addr:
-            return True
-    return False
 
 
 def ip_network_contains(network, mask, address):
--- a/golem/core/hostaddress.py
+++ b/golem/core/hostaddress.py
@@ -2,6 +2,9 @@ import socket
 import os
 import logging
 import netifaces
+
+import ipaddress
+import struct
 import stun
 
 from variables import DEFAULT_CONNECT_TO, DEFAULT_CONNECT_TO_PORT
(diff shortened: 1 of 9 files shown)


golem/network/transport/network.py
```diff
@@ -51,6 +51,11 @@ def set_session_factory(self, session_factory):
     # Protocol function
     def connectionMade(self):
         """Called when new connection is successfully opened"""
+        try:
+            self.transport.setTcpKeepAlive(1)
+        except AttributeError:
```
chfast, line 60: Please comment when this exception can happen and what it means.

golem/network/transport/tcpserver.py
```diff
@@ -23,6 +24,10 @@ def __init__(self, config_desc, network):
         """
         Server.__init__(self, config_desc, network)
         self.cur_port = 0  # current listening port
+        self.use_ipv6 = config_desc.use_ipv6 if config_desc else False
+        self.ipv4_networks = ip_networks()
+        self.ipv6_networks = ip_networks(use_ipv6=True) \
+            if self.use_ipv6 else []
```
chfast: Can it be

```
if self.use_ipv6:
    self.ipv6_networks = ...
```

golem/network/transport/tcpserver.py
```diff
@@ -143,6 +151,30 @@ def _add_pending_listening(self, req_type, port, args):
         pl.args["listen_id"] = pl.id
         self.pending_listenings.append(pl)
 
+    def _is_address_accessible(self, socket_addr):
+        if not socket_addr:
```
chfast, line 162: Comment about this case.

golem/network/transport/tcpserver.py
```diff
@@ -143,6 +151,30 @@ def _add_pending_listening(self, req_type, port, args):
...
+                return self._is_address_in_network(addr, self.ipv6_networks)
+            return self._is_address_in_network(addr, self.ipv4_networks)
+        return True
+
+    @staticmethod
+    def _is_address_in_network(addr, networks):
+        for net, mask in networks:
+            if ip_network_contains(net, mask, addr):
+                return True
+        return False
```
chfast: What do you thing about Python magic like:

```
any(ip_network_contains(net, mask, addr) for net, mask in networks)
```

tests/golem/core/test_hostaddress.py
```diff
@@ -10,3 +10,51 @@ def testGetHostAddress(self):
     def testGetHostAddress2(self):
         self.assertEqual(get_host_address('10.30.100.100'), '10.30.10.216')
         self.assertEqual(get_host_address('10.30.10.217'), '10.30.10.216')
+
+    @unittest.skip("Find network testing framework")
```
chfast: Why skip?

tests/golem/task/dummy/runner.py
```diff
@@ -165,7 +165,7 @@ def run_simulation(num_computing_nodes=2, num_subtasks=3, timeout=120,
 
     def monitor_subprocess(proc):
         global task_finished
-        while not proc.returncode:
+        while proc.returncode is None:
```
chfast, line 168: That would saved me some time yesterday :)

golem/core/hostaddress.py
```diff
@@ -37,10 +40,79 @@ def ip_addresses(use_ipv6=False):
...
+__ipv4_private_networks = [
+    u'10.0.0.0/8',
+    u'127.0.0.0/8',
+    u'172.16.0.0/12',
+    u'192.168.0.0/16'
+]
+__parsed_ipv4_private_networks = __parse_ipv4_networks(__ipv4_private_networks)
+
+
+def ip_address_private(address):
```
chfast, line 72: I don't have to check details here to be sure I've done it right, but...
Why haven't you tried using `ipaddress` package (backport from Python 3), e.g. https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address.is_private.

golem/network/transport/network.py
mfranciszkiewicz, line 60: TCP keepalive is a native connectivity check built in a networking stack. If the underlying transport is not TCP-based, `setTcpKeepAlive` method will not be available.

chfast, line 60: Usually, by "comment" I mean add comment in the code for future generations.

chfast:
In general looks good. The biggest concern is `ipaddress` package not used where it could be. Try to address the issues and I hope we could merge it ASAP.

golem/network/transport/network.py
mfranciszkiewicz, line 60: I was going to, and also wanted to respond.

golem/core/hostaddress.py
mfranciszkiewicz, line 72: See my note below

golem/core/hostaddress.py
```diff
@@ -37,10 +40,79 @@ def ip_addresses(use_ipv6=False):
...
+            return False
+    try:
+        a = struct.unpack('!I', socket.inet_aton(address))[0]
+    except Exception as exc:
+        logger.error("Cannot parse IPv4 address {}: {}"
+                     .format(address, exc.message))
+        return False
+
+    for n_addr, n_mask in __parsed_ipv4_private_networks:
+        if (a & n_mask) == n_addr:
```
mfranciszkiewicz: This bit makes the check much faster for multiple addresses

chfast: Have you measured the difference? Is it worth it? We are doing a prototype here. I know it is very slow already, but also less code to maintain is better.

golem/network/transport/tcpserver.py
```diff
@@ -26,8 +26,10 @@ def __init__(self, config_desc, network):
         self.cur_port = 0  # current listening port
         self.use_ipv6 = config_desc.use_ipv6 if config_desc else False
         self.ipv4_networks = ip_networks()
-        self.ipv6_networks = ip_networks(use_ipv6=True) \
-            if self.use_ipv6 else []
+        if self.use_ipv6:
+            self.ipv6_networks = ip_networks(use_ipv6=True)
+        else:
+            self.ipv6_networks = []
```
chfast, line 32: I think it will never be used when `use_ipv6 == False`. So you don't have to even set it.

chfast, line 32: Or alternatively, you can have just `self.networks` containing v4 or v6 networks depending on `self.use_ipv6`.

chfast:
Address as many issues as you thing is reasonable and merge.

golem/core/hostaddress.py
mfranciszkiewicz: I've done synthetic tests. What I haven't done is app profiling focused on processing delay, which will surely show that the impact is small, as there are other areas in need of optimization. I'll remove that code to keep it simple.

PR #448, opened 2016-07-04
Running lux and blender benchmarks on docker config change

magdasta:


mfranciszkiewicz:
The _Estimated Lux / Blender performance_ values are not updated in the UI. Besides that, it works as it should.

magdasta:
Are you sure they are not? Some time is needed to recount the performance, but I believe after a while they **are** updated

mfranciszkiewicz:
(Tested on Windows) VM reconfiguration takes place in a separate `Thread`, the `done_callback` is also ran in that thread. Unlike `QThread`, `Thread` knows nothing about Qt signals. It either has to be a `QThread` or the main thread.

badb:
Seems to work now, but I'm not sure if it's the best method to do this or the best function name. Feel free to change it to something nicer.

PR #451, opened 2016-07-07
Re-enable CPU core assignment for Docker containers

mfranciszkiewicz:
Available CPU cores are read from Golem's process CPU affinity settings.


golem/docker/config_manager.py
```diff
@@ -35,8 +48,9 @@ def build_config(self, config_desc):
             max_resource_size = config_desc.max_resource_size
 
             with self._try():
```
badb, line 50: Btw. is it possible to use `except Exception` in `_try` method instead of just `except`?

PR #459, opened 2016-07-11
using proper benchmark result

magdasta:


golem/task/taskserver.py
```diff
@@ -23,6 +23,12 @@ def __init__(self, node, config_desc, keys_auth, client, use_ipv6=False):
         self.keys_auth = keys_auth
         self.config_desc = config_desc
 
+        self.performance = {"DEFAULT" : self.config_desc.estimated_performance,
```
badb: Hm, I don't think it's a good idea to put this list here.  I suggest adding get_performance method to enviroment and getting this value from there.

golem/task/taskserver.py
```diff
@@ -72,14 +78,16 @@ def request_task(self):
         theader = self.task_keeper.get_task()
         if theader is not None:
             trust = self.client.get_requesting_trust(theader.task_owner_key_id)
+            task_type = theader.environment
+            performance = self.performance[task_type]
```
badb: `TaskServer` may ask `self.task_keeper` to get performance from environment using `environments_manager`. See comment above.

PR #461, opened 2016-07-11
Fix button styles #460

banasiakadam60:


gnr/customizers/renderercustomizer.py
```diff
@@ -75,6 +75,8 @@ def _setup_connections(self):
             self.gui.ui.outputFileLineEdit.textChanged,
             self.gui.ui.outputFormatsComboBox.currentIndexChanged,
             self.gui.ui.outputFileLineEdit.textChanged,
+            self.gui.ui.framesLineEdit.textChanged,
```
badb: `renderercustomizer` is for all renderers and frame's options are only used for framerenders (ie. Blender but not Lux). This should be changed in `framerendercustomizer`.

gnr/customizers/renderercustomizer.py
```diff
@@ -129,12 +129,17 @@ def _res_y_changed(self):
         self.logic.change_verification_option(size_y_max=self.gui.ui.outputResYSpinBox.value())
         self.logic.task_settings_changed()
 
+    def _frames_changed(self):
```
badb: Can you move it to FrameRendererCustomizer?

badb:
I think it can be merged after moving this one function.

PR #467, opened 2016-07-12
golemapp.py: recognize --multiprocessing-fork option

mfranciszkiewicz:
`--multiprocessing-fork` and it's argument are required by `multiprocessing` on Windows. This mainly concerns the packaged version.

### DIFF
--- a/golemapp.py
+++ b/golemapp.py
@@ -18,7 +18,9 @@ from gnr.node import GNRNode  # TODO: This also configures the logging.
 @click.option('--task', '-t', multiple=True, type=click.Path(exists=True),
               callback=GNRNode.parse_task_file,
               help="Request task from file")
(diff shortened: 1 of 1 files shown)


golemapp.py
```diff
@@ -18,7 +18,9 @@
 @click.option('--task', '-t', multiple=True, type=click.Path(exists=True),
               callback=GNRNode.parse_task_file,
               help="Request task from file")
-def start(gui, payments, datadir, node_address, peer, task):
+@click.option('--multiprocessing-fork', nargs=1, default=None)
```
chfast, line 21: I think it should be `@click.option('--multiprocessing-fork', is_flag=True)` if there is no value passed with.

mfranciszkiewicz, line 21: In this case, there is a value

PR #468, opened 2016-07-12
subtask timeout / deadline

badb:
Rename subtask_ttl to subtask_timeout and add deadline to CompTaskDef.
It does the same thing as `timeout` branch, but change only the subtask deadline (not task deadline). 

That's the change that requires protocol changes, because it's incompatibile with current  protocol version.


gnr/task/renderingtask.py
```diff
@@ -240,6 +239,7 @@ def _new_compute_task_def(self, hash, extra_data, working_directory, perf_index)
         ctd.performance = perf_index
         ctd.working_directory = working_directory
         ctd.docker_images = self.header.docker_images
+        ctd.deadine = timeout_to_deadline(self.header.subtask_timeout)
```
mfranciszkiewicz: Misspelled `deadline`

mfranciszkiewicz:
1. I've encountered some `TypeError`s:
   
   ```
   File "/opt/golem/golem/task/taskserver.py", line 71, in sync_network
     self.__remove_old_tasks()
   File "/opt/golem/golem/task/taskserver.py", line 757, in __remove_old_tasks
     nodes_with_timeouts = self.task_manager.remove_old_tasks()
   File "/opt/golem/golem/task/taskmanager.py", line 313, in remove_old_tasks
     if get_current_time() > s.deadline:
   exceptions.TypeError: can't compare datetime.datetime to str
   ```
   
   ```
   unsupported operand type(s) for -: 'str' and 'datetime.datetime'
   [...]
   /opt/golem/golem/client.py:262:task_resource_collected
   /opt/golem/golem/task/taskcomputer.py:112:task_resource_collected
   /opt/golem/golem/core/common.py:59:deadline_to_timeout
   ```
2. Could you change protocol versions in `peersession.py` and `tasksession.py`?

badb:
1. They were caused by typo. But that's a good example - we should add some validation function for important subtask parameters before we start computation (or send them to the network). I'll make a sepearate issue for that. 
2. Done!

PR #482, opened 2016-07-13
logging in multiprocess app

badb:
It seems to work, but is ugly.  If someone has a better idea, feel free to propose a change.


mfranciszkiewicz:
I suggest PR #488 as a more lightweight solution

badb:
#488 seems to solve the problem and is less problematic. But I'm leaving the branch open, maybe we will need to use this method later.

PR #484, opened 2016-07-13
Metadata

badb:


golem/client.py
```diff
@@ -351,6 +353,17 @@ def get_requesting_trust(self, node_id):
             return self.ranking.get_requesting_trust(node_id)
         return None
 
+    def get_description(self):
+        try:
+            account = Account.get_or_create(node_id=self.get_client_id())
```
mfranciszkiewicz: `get_or_create` returns a tuple. Maybe it should be
`account, _ = Account.get_or_create(node_id=self.get_client_id())`
instead?

badb: Documentation said that it should, but it doesn't and account, _ doesn't work on my machine. Maybe it's a version issue.

PR #486, opened 2016-07-14
Collector optimization

magdasta:
Closes https://github.com/imapp-pl/golem/issues/397.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -10,7 +10,7 @@ dependencies:
   pre:
     - sudo apt-get install -y python-qt4 pyqt4-dev-tools
     - if [ ! -e /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/PyQt4 ]; then ln -s /usr/lib/python2.7/dist-packages/PyQt4/ /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/; ln -s /usr/lib/python2.7/dist-packages/sip.so /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/; ln -s /usr/lib/python2.7/dist-packages/sip.x86_64-linux-gnu.so /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/; fi
-    - if [ ! -e /usr/local/bin/ipfs ]; then wget http://52.40.149.24:9999/go-ipfs_v0.4.1_linux-amd64.tar.gz; tar xvfz go-ipfs_v0.4.1_linux-amd64.tar.gz; sudo mv go-ipfs/ipfs /usr/local/bin/ipfs; /usr/local/bin/ipfs init; fi
+    - if [ ! -e /usr/local/bin/ipfs ]; then wget http://dist.ipfs.io/go-ipfs/v0.4.1/go-ipfs_v0.4.1_linux-amd64.tar.gz; tar xvfz go-ipfs_v0.4.1_linux-amd64.tar.gz; sudo mv go-ipfs/ipfs /usr/local/bin/ipfs; /usr/local/bin/ipfs init; fi
     - /usr/local/bin/ipfs config --json Bootstrap "[]"
     - /usr/local/bin/ipfs config --json SupernodeRouting.Servers "[]"
     - /usr/local/bin/ipfs config --json Addresses.Swarm '["/ip6/::/tcp/4001", "/ip6/::/udp/4002/utp", "/ip4/0.0.0.0/udp/4002/utp"]'
--- a/circle.yml
+++ b/circle.yml
@@ -10,7 +10,7 @@ dependencies:
   pre:
     - sudo apt-get install -y python-qt4 pyqt4-dev-tools
     - if [ ! -e /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/PyQt4 ]; then ln -s /usr/lib/python2.7/dist-packages/PyQt4/ /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/; ln -s /usr/lib/python2.7/dist-packages/sip.so /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/; ln -s /usr/lib/python2.7/dist-packages/sip.x86_64-linux-gnu.so /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/; fi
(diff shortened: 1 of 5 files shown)


gnr/taskcollector/TaskCollector.cpp
```diff
@@ -153,17 +147,21 @@ class AddTaskCollector : public TaskCollector {
...
+		bitmap_ptr firstChunk = GenericLoader(*it);
+		
+		const auto width = FreeImage_GetWidth(firstChunk.get());
+		const auto height = FreeImage_GetHeight(firstChunk.get());
+		const auto type = FreeImage_GetImageType(firstChunk.get());
+		
+		bitmap_ptr finalImage(FreeImage_Copy(firstChunk.get(), 0, height, width, 0));
 
-		auto RGBChunkWorker = [=, &finalImage](const bitmap_ptr& el)
+		auto RGBChunkWorker = [=, &finalImage](const std::string el)
```
mfranciszkiewicz: It should be ok to pass by reference.

circle.yml
```diff
@@ -10,7 +10,7 @@ dependencies:
   pre:
     - sudo apt-get install -y python-qt4 pyqt4-dev-tools
     - if [ ! -e /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/PyQt4 ]; then ln -s /usr/lib/python2.7/dist-packages/PyQt4/ /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/; ln -s /usr/lib/python2.7/dist-packages/sip.so /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/; ln -s /usr/lib/python2.7/dist-packages/sip.x86_64-linux-gnu.so /home/ubuntu/virtualenvs/venv-system/lib/python2.7/site-packages/; fi
-    - if [ ! -e /usr/local/bin/ipfs ]; then wget http://dist.ipfs.io/go-ipfs/v0.4.1/go-ipfs_v0.4.1_linux-amd64.tar.gz; tar xvfz go-ipfs_v0.4.1_linux-amd64.tar.gz; sudo mv go-ipfs/ipfs /usr/local/bin/ipfs; /usr/local/bin/ipfs init; fi
+    - if [ ! -e /usr/local/bin/ipfs ]; then wget http://52.40.149.24:9999/go-ipfs_v0.4.1_linux-amd64.tar.gz; tar xvfz go-ipfs_v0.4.1_linux-amd64.tar.gz; sudo mv go-ipfs/ipfs /usr/local/bin/ipfs; /usr/local/bin/ipfs init; fi
```
chfast, line 13: How is that related?

magdasta, line 13: It isn't. It was a workaround for tests on Circle not running.

chfast, line 13: Please removed and if needed submit as a separated change.

gnr/taskcollector/TaskCollector.cpp
```diff
@@ -109,23 +109,17 @@ class TaskCollector {
...
 
-	bool addImgFile(std::string pathName, int flag = 0)  {
+	bool addImgFile(std::string pathName)  {
 		if (pathName.empty())
 			return false;
-		auto img = GenericLoader(pathName, flag);
-		if (!img)
-			return false;
-		chunks.emplace_back(std::move(img));
+		chunks.emplace_back(pathName);
```
chfast: Small optimization possibility here: you can move the `pathName` to the `chunks` vector instead of copying it.

```
chunks.emplace_back(std::move(pathName))
```

Or alternatively pass the `pathName` to the function by reference.

```
bool addImgFile(const std::string& pathName)
```

gnr/taskcollector/TaskCollector.cpp
```diff
@@ -223,24 +223,31 @@ class PasteTaskCollector : public TaskCollector {
...
+		
 		const auto it = chunks.begin();
-		const auto width = FreeImage_GetWidth(it->get());
+		bitmap_ptr firstChunk = GenericLoader(*it);
+		const auto width = FreeImage_GetWidth(firstChunk.get());
 		auto height = 0;
 		
-		for (auto i = chunks.begin(); i != chunks.end(); ++i)
-			height += FreeImage_GetHeight(i->get());
+		for (auto i = chunks.begin(); i != chunks.end(); ++i) {
```
chfast: Foreach `for` construction can be used here. It would look like:

```
for (auto& chunk : chunks) {
    bitmap_ptr img= GenericLoader(chunk);
    ...
}
```

gnr/taskcollector/TaskCollector.cpp
```diff
@@ -223,24 +223,31 @@ class PasteTaskCollector : public TaskCollector {
...
+			bitmap_ptr chunk = GenericLoader(*i);
+			height += FreeImage_GetHeight(chunk.get());
```
chfast: This is a bit problematic as we need to open a chunk just to get its height and it is closed just after.

chfast:
There are small issues that should be addressed. Can be merged afterwards.

gnr/taskcollector/TaskCollector.cpp
```diff
@@ -150,37 +158,36 @@ class AddTaskCollector : public TaskCollector {
...
 		
-		const auto width = FreeImage_GetWidth(firstChunk.get());
-		const auto height = FreeImage_GetHeight(firstChunk.get());
 		const auto type = FreeImage_GetImageType(firstChunk.get());
 		
 		bitmap_ptr finalImage(FreeImage_Copy(firstChunk.get(), 0, height, width, 0));
+		unsigned int res_x = width;
 
-		auto RGBChunkWorker = [=, &finalImage](const std::string& el)
+		auto RGBChunkWorker = [=, &finalImage, &res_x](const std::string& el)
```
chfast: There is easier way to pass `this->width` to lambda function. You must add `this` in the capture expression:

```
auto RGBChunkWorker = [this, =, &finalImage](...
```

and IIRC you must use `this` explicitly inside the lambda as `this->width`.

gnr/taskcollector/TaskCollector.cpp
```diff
@@ -132,6 +134,12 @@ class TaskCollector {
 		auto img = finalize();
 		return GenericWriter(img, outputPath, EXR_FLOAT);
 	};
+	void set_width(int w) {
```
chfast: It's better to keep the same type. `width` is `unsigned int`, `w` is `int`.

chfast:
I'm not sure passing width and height to task collector explicitly is so good idea. @badb ?

chfast:
Now looks good to me.

PR #490, opened 2016-07-14
Name

badb:
Allow user to change node_name
Using node_name in DirManager removed.


mfranciszkiewicz:
One small issue: empty node names are accepted

PR #495, opened 2016-07-14
monitor, v1

badb:
monitor implementation


golem/monitor/serialization/defaultserializer.py
```diff
@@ -0,0 +1,27 @@
+import json
+
+
+class DefaultJSONSerializer(object):
```
chfast: Is it oneliner packed into a dadicated class?

PR #497, opened 2016-07-15
Create log file directories

mfranciszkiewicz:


gnr/gnrstartapp.py
```diff
@@ -82,10 +82,11 @@ def start(self, client, logic_service_info):
 def start_gui_process(queue, datadir, rendering=True, gui_app=None, reactor=None):
 
     if datadir:
-        config_logging(path.join(datadir, GUI_LOG_NAME))
+        log_name = path.join(datadir, GUI_LOG_NAME)
     else:
-        config_logging(GUI_LOG_NAME)
+        log_name = GUI_LOG_NAME
 
+    config_logging(log_name, datadir=datadir)
```
chfast, line 89: As the `log_name` is a full path it does not make sense to pass also the datadir. You can extract the directory name easily by `os.path.dirname()`.

golem/core/common.py
```diff
@@ -57,8 +59,16 @@ def func_wrapper(*args, **kwargs):
         return func_wrapper
 
 
-def config_logging(logname=LOG_NAME):
+def config_logging(logname=LOG_NAME, datadir=None):
     """Config logger"""
+    if datadir:
+        try:
+            os.makedirs(datadir)
+        except OSError as exc:
```
chfast, line 67: I think it is simpler to write it as

```
if not path.exists(datadir):
    os.makedirs(datadir)
```

gnr/gnrstartapp.py
mfranciszkiewicz, line 89: `log_name` is not always a full path. Note the `if ... else` above.

golem/core/common.py
mfranciszkiewicz, line 67: Yes it is, but it doesn't take multiple processes into account. It might (and does) happen at the same time, leading to an error.

gnr/gnrstartapp.py
chfast, line 89: I precisely mean `log_file` is either `datadir + LOG_NAME` or `LOG_NAME`. So datadir is either included in `log_name` or is `None`.

mfranciszkiewicz, line 89: Well, my choosing of `datadir` as a keyword is probably misleading. It could really be any directory, e.g. `/var/log/golem`.

golem/core/common.py
chfast, line 67: Understood.

Issue #201, opened 2016-03-15
Show payments and computation time for task


badb:
They should be clearly visible from the requestor window for active tasks.

PR #498, opened 2016-07-15
Show payments and computation time for task #201 Names

banasiakadam60:
[WIP] Show payments


badb:
1) I got an error after starting: QFormLayoutPrivate::setItem: Cell (4, 1) already occupied, 
2) 
glitch
3) Default task name is not shown in new task form (but later added properly to the tab). 
4) I would change column order, but I'm not sure which one may be the best. Maybe we should ask Julian about it later. But it's not very important now.

gnr/ui/tasktableelem.py
```diff
@@ -31,6 +37,12 @@ def __build_row(self):
         self.status_item = QTableWidgetItem()
         self.status_item.setText(self.status)
 
+        self.timer_item = QTableWidgetItem()
+        self.timer_item.setText('00:00:00')
+
+        self.cost_item = QTableWidgetItem()
+        self.cost_item.setText('0,000000')
```
badb: We should probably use `0.000000` notation.

banasiakadam60: OK

golem/appconfig.py
```diff
@@ -152,6 +152,7 @@ def load_config(cls, datadir, cfg_file_name=CONFIG_FILENAME):
                                  eth_account=ETH_ACCOUNT_NAME,
                                  min_price=MIN_PRICE,
                                  max_price=MAX_PRICE,
+                                 task_name="",
```
badb: Hm, it's not necessary to put it config. Let's move it to some renderdefaults or environment options.

banasiakadam60: Ok, done

badb:
Can you try to merge develop into this branch and then check if it passes the tests?
1) There are two price fields, remove the one at the top. 
2) Luxrender tasks are still called Blender_date
3) Cloning the task doesn't change the name
4) I think maybe the right column order should be: name, state, progress bar, value, time, id (but it's up for discussion).

PR #505, opened 2016-07-19
Payments: use (and show) deposit balance

chfast:
The deposit balance (ethers got from computations) is used as the first choise for outgoing payments.

### DIFF
--- a/gnr/gnrapplicationlogic.py
+++ b/gnr/gnrapplicationlogic.py
@@ -198,12 +198,11 @@ class GNRApplicationLogic(QtCore.QObject):
         self.client.get_balance().addCallback(self._update_payments_view)
 
     def _update_payments_view(self, result_tuple):
-        b, ab = result_tuple
+        b, ab, deposit = result_tuple
         if not (b and ab):
             return
 
         rb = b - ab
-        deposit = 0  # TODO: Get current deposit value.
         total = deposit + b
         ether = 1.0 / 10 ** 18
         fmt = "{:.6f} ETH"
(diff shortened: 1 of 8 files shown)


gnr/gnrapplicationlogic.py
```diff
@@ -198,12 +198,11 @@ def update_payments_view(self):
         self.client.get_balance().addCallback(self._update_payments_view)
 
     def _update_payments_view(self, result_tuple):
-        b, ab = result_tuple
+        b, ab, deposit = result_tuple
```
badb, line 201: Throws exception when transaction system is turn off

golem/ethereum/paymentprocessor.py
```diff
@@ -109,6 +110,21 @@ def balance(self, refresh=False):
...
+            addr = keys.privtoaddr(self.__privkey)
+            r = self.__client.call(_from=addr.encode('hex'),
+                                   to=self.BANK_ADDR.encode('hex'),
+                                   data=data.encode('hex'),
+                                   block='pending')
+            if r is None or r == '0x':
+                self.__deposit = 0
+            else:
+                self.__deposit = int(r, 16)
+            log.info("Deposit: {}".format(self.__deposit / float(10**18)))
```
badb, line 125: Can you use  ETH from golem.core.common here?

gnr/gnrapplicationlogic.py
chfast, line 201: Fixed.

golem/ethereum/paymentprocessor.py
chfast, line 125: Not the best idea. We need a function or a type to represent ETH amount. First try should be pyethereum library.

badb, line 125: https://github.com/ethereum/pyethereum/blob/develop/ethereum/utils.py#L408 denoms?

chfast, line 125: Yep. I'm working on it. But will submit as a separated PR.

PR #506, opened 2016-07-19
keys shouldn't depend on client's name

### LINKED ISSUE #502: Ethereum private key depends on client's name
Looks like Ethereum account's private key depends on client's name so if the name is changes the Ethereum address changes too. It is reversible if the name is changed back.

I guess the Ethereum private key should be derived from client's private key.

badb:
Closes #502


chfast:
Works and looks!

PR #507, opened 2016-07-20
running test task for first and last frame

magdasta:


badb:
Nice. Can you add a simple test?

magdasta:
@badb Done

PR #512, opened 2016-07-20
subtask value in eth

### LINKED ISSUE #511: Price in subtask window should be show in ethers
Subtasks detail price is shown in wei.

badb:
Resolves #511


chfast:
Conflicts with #513.

Issue #202, opened 2016-03-15
Allow to add names for tasks


badb:
User should be able to find and recognise his tasks easily. Right now each task has it's own task id and when we click on its row we can see the details. User should be able to add specific name that is visible only in local golem instance.

PR #514, opened 2016-07-20
#202 #201

banasiakadam60:
Allow user to define task name + show computation cost and time


banasiakadam60:
@badb Can it be merged? I've moved all commits related with #201 and #202 from 'names' branch to this one to clean up this mess I made. I've joined this two issues becouse both need a change in AppMainWindow.ui, so I could pushed it as one PR, or two related PRs. IMO one PR is better option.

badb:
It have to wait with being merge - there's not time to add it to relase 0.3 version (because there is no time to test it and still a lot of debugging with other problems. 
One new PR is ok imho. Can you add some tests to pass codecdov patch? 
And one small issue: default lux render task is loaded properly after clicking "reset to defaults", but not after changing task type combo box.

banasiakadam60:
OK. Oh, thanks, I have somehow missed this issue :)

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -59,3 +60,15 @@ def get_incoming_payments(self):
...
+        """
+        Get cost of subtasks
+        :param subtasks: List of subtasks ID
+        :return: Cost of subtasks
+        """
+        cost = 0.0
+        if self.payments_keeper:
+            for subtask in subtasks:
+                cost += self.payments_keeper.get_payment(subtask)
+        return cost / denoms.ether
```
badb: One small thing: in other places where ether division is implemented there is also imported float division from python 3. Can you add from **future** import division here?

banasiakadam60: OK, done :)

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -59,3 +59,15 @@ def get_incoming_payments(self):
...
+    def get_payment_for_subtasks(self, subtasks):
+        """
+        Get cost of subtasks
+        :param subtasks: List of subtasks ID
+        :return: Cost of subtasks
+        """
+        cost = 0
+        if self.payments_keeper:
+            for subtask in subtasks:
+                cost += self.payments_keeper.get_payment(subtask)
```
mfranciszkiewicz: Wouldn't it be better to select multiple `Payment.cost`s in a single query? If you're concerned with query size, you can:
- split the query into batches
- later on, when a task <- subtask mapping is stored in the db, query with `task_id` (hopefully soon)

Plus, how about storing a (lazily initialized) sum and updating it on each payment?

mfranciszkiewicz: On second thought: I think we should postpone the optimizations and merge this PR.

PR #539, opened 2016-07-28
Scaling preview

magdasta:


badb:
Great! Work very well. 
But can you make the subtaks frame thinner? It's very thick in that version.  That thinckness makes problem from #541 very visible (and possibly, however I'm not 100% sure make it even worse).

magdasta:
Thanks. Will take a look at the frame thing

badb:
Still problems with 1920x1080, 30 subtasks. Preview strips are connected in the wrong order and I'm not able to mark the last subtask (number 30).

Issue #549, opened 2016-08-05
Better error messages


banasiakadam60:
We should show more precise error messages when testing tasks eg. not enough RAM

PR #550, opened 2016-08-05
Better error messages #549

banasiakadam60:


magdasta:
Are we sure errors 246 and 247 are result of too low RAM limit?

banasiakadam60:
I found it here: https://github.com/imapp-pl/golem/wiki/Common-problems

gnr/gnrapplicationlogic.py
```diff
@@ -518,13 +518,21 @@ def test_task_computation_success(self, results, est_mem):
...
         self.customizer.gui.setEnabled('new_task', True)  # enable everything on 'new task' tab
         if self.customizer.new_task_dialog_customizer:
             self.customizer.new_task_dialog_customizer.test_task_computation_finished(False, 0)
 
+    @staticmethod
+    def __parse_error_message(error_msg):
+        if not error_msg:
+            return u"Unknown error"
+        if any(code in error_msg for code in ['246', '247', '500']):
+            return u"RAM limit is too low. Consider increasing max memory usage"
```
badb: Maybe better solution is to make it: error_message + "There is a chance that you RAM limit..."

magdasta:
Then looks fine for me.

banasiakadam60:
@badb OK, I'll change it

Issue #554, opened 2016-08-05
Update last used directory


banasiakadam60:
Every time I open a main scene file, open file dialog starts in Golem's program directory. It should start in the directory which it was in last time, so I do not have to switch to the same directory every time I load a scene file.

badb:
https://github.com/imapp-pl/golem-docs/issues/5

PR #555, opened 2016-08-05
Update last used directory #554

banasiakadam60:


badb:
It's nice that default directory is user directory. But that's the only one improvement I'm afraid. If you reload the task or change task type, custmizer is generated from the beginning and saved paths are reset.

banasiakadam60:
OK, ofc I can change it and keep paths in some external class (I suppose it will be best solution), but I'm not sure... I thought that if user change task type, let's say from Blender to Lux, we shouldn't show him dir with blender tasks. But, as I said, if you want, I can change it.
btw. What do you mean by 'reload task'? Loading saved task?

badb:
Yeah, reload = load again from .gt file

Hm. I've checked QFileDialog docs and theres saveState function.

> Saves the state of the dialog's layout, history and current directory.
> 
> Typically this is used in conjunction with QSettings to remember the size for a future session. A version number is stored as part of the data.

Maybe it can be easily used to save those directiores and then to be restored not only between tasks but also sessions. Just an idea, not sure if it's really going to work.

badb:
Right now it always open the home directory, now matter what you set or opened.

banasiakadam60:
Yyy... I've just checked this and it's work fine? At least on my KUbunu. I even had to use QSettings in tests, because it was using last used directory when testing. Did you check it on Ubuntu, or Windows?

badb:
Windows.  Always open home directory, doesn't matter if I set scene in different directory by hand or load from the .gt.

banasiakadam60:
OK, I'll check it on Windows. I've tested this on KUbuntu and works fine, on pyqt wiki it was written it works similar on every supported platform, but it looks like it doesn't.

mfranciszkiewicz:
Confirmed - always opens user's home folder on Windows.

badb:
I think I find the reason - settings weren't saved on my computer because there was a problem with access right to the default file. I've changed the settings file, please check that it's working properly on Linux.

mfranciszkiewicz (approved):

PR #559, opened 2016-08-08
Test task progress bar #558

banasiakadam60:
Simple progress bar

### DIFF
--- a/gnr/gnrapplicationlogic.py
+++ b/gnr/gnrapplicationlogic.py
@@ -489,6 +489,7 @@ class GNRApplicationLogic(QtCore.QObject):
         self.br.run()
 
     def _benchmark_computation_success(self, performance, label):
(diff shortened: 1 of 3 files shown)


badb:
The most irritating and misleading progress bar ever, but probably better then previous empty test  window :) 
Maybe we'll add some serious estimation later, right now it should be sufficient.

badb:
Are you going to change something more here?

banasiakadam60:
Haha, thanks :) As I said last time, for me we shouldn't use this progress bar as an estimator (IMO it can be misleading) and this pr can be closed. But I can try to find some way to show proper progress.

badb:
We may show proper progress when we find out how to measure subtasks progress.

banasiakadam60:
OK, so maybe for now let's merge this PR, and in the mean time I'll try to check the possibilities to show progress

PR #561, opened 2016-08-09
Various memory leak fixes

mfranciszkiewicz:


gnr/gnrstartapp.py
```diff
@@ -184,5 +184,3 @@ def start_app(datadir=None, rendering=False,
         start_client_process(queue, start_ranking, datadir, transaction_system)
     except Exception as exc:
         print "Exception in Client process: {}".format(exc)
-
-    process_monitor.exit()
```
badb, line 188: Why do you remove this line?

mfranciszkiewicz, line 188: The reason was that some new errors related to killing the geth subprocess popped up.

When you interrupt the app with Ctrl+C, all subprocesses receive the shutdown signal. If you close the GUI, the process monitor detects a dead subprocess and exits the application. The app shuts down either way (or at least the 2 main processes). We currently have problems with terminating geth, but that's an issue to resolve in a separate PR.

golem/network/p2p/p2pservice.py
```diff
@@ -265,19 +265,18 @@ def remove_peer(self, peer_session):
...
         """
-        pc = self.pending_connections.get(peer_session.conn_id)
-        if pc:
-            pc.status = PenConnStatus.Failure
+        self.remove_pending_conn(peer_session.conn_id)
 
         for p in self.peers.keys():
-            if self.peers[p] == peer_session:
-                self.peer_order.remove(p)
+            if self.peers.get(p, None) == peer_session:
```
badb: This None isn't needed.

PR #563, opened 2016-08-10
Run tasks in Docker containers without root privileges

mfranciszkiewicz:
Unlike `su` and running Docker containers as a specified user, both the _effective_ `UID` and `GID` are identical to Golem's user `UID` and `GID` (Linux).

Other changes:
- updated runner scripts
- bumped image versions

### DIFF
--- a/gnr/task/images.ini
+++ b/gnr/task/images.ini
@@ -1,3 +1,3 @@
-golem/base images/Dockerfile.base 1.1
-golem/blender images/Dockerfile.blender 1.1
-golem/luxrender images/Dockerfile.luxrender 1.1
+golem/base images/Dockerfile.base 1.2
+golem/blender images/Dockerfile.blender 1.2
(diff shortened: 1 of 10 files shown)


badb:
I get on windows:

> no such file or directory
> Error response from daemon: Container command not found or does not exist. 

It may be a problem with working directory: https://github.com/docker/compose/issues/2950 
Do you have some ideas or should I investigate further on my machine?

mfranciszkiewicz:
Thanks, I'll sort it out.

paulperegud:
Gosu README.md was updated only 3 days ago and it mentions using chroot --userspec as an option:

``` bash
$ docker run -it --rm debian:jessie chroot --userspec=nobody / grep ^[GU]id: /proc/self/status
Uid:    65534   65534   65534   65534
Gid:    65534   65534   65534   65534
```

paulperegud:
chroot requires capabilities we drop while starting container. So, instead I've used su-exec

PR #565, opened 2016-08-10
Sorting payments by value instead of lexicographically

### LINKED ISSUE #457: Sorting by value in payment table
Values in payment table are sorted in lexicographical order.

magdasta:
resolves #457

### DIFF
--- a/gnr/customizers/paymentsdialogcustomizer.py
+++ b/gnr/customizers/paymentsdialogcustomizer.py
@@ -60,8 +60,6 @@ class SmartTableItem(QTableWidgetItem):
             if len(t2) == 0:
                 t2 = "0.0"
             return float(t1) < float(t2)
-        if t1.isdigit() and t2.isdigit():
-            return int(t1) < int(t2)
         return t1 < t2
             
             
--- a/gnr/customizers/paymentsdialogcustomizer.py
+++ b/gnr/customizers/paymentsdialogcustomizer.py
@@ -36,31 +36,22 @@ class PaymentsDialogCustomizer(Customizer):
 
 class SmartTableItem(QTableWidgetItem):
     def __lt__(self, other):
-        t1 = self.text()
-        t2 = other.text()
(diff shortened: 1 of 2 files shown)


gnr/customizers/paymentsdialogcustomizer.py
```diff
@@ -34,18 +34,49 @@ def _add_income(self, income_info):
...
+            return float(t1) < float(t2)
+        if t1.endswith("%"):
+            t1 = t1[:-1]
+            t2 = t2[:-1]
+            if len(t1) == 0:
+                t1 = "0.0"
+            if len(t2) == 0:
+                t2 = "0.0"
+            return float(t1) < float(t2)
+        if t1.isdigit() and t2.isdigit():
```
badb: One small problem - subtask_id is digit, but it's a little bit too long to realise that some values are shorter than others.  We may 
a) change the way subtask id is generated
b) add some predefine types to this new class
c)  leave it, it's not very important.

gnr/customizers/paymentsdialogcustomizer.py
```diff
@@ -34,18 +34,47 @@ def _add_income(self, income_info):
...
+    def __lt__(self, other):
+        t1 = self.text()
+        t2 = other.text()
+        if t1 is None:
+            return True
+        if t2 is None:
+            return False
+        t1 = str(t1)
+        t2 = str(t2)
+        if t1.endswith("ETH") or t2.endswith("ETH"):
```
mfranciszkiewicz: Maybe something more like this?

```
        for postfix in ['ETH', '%']:
            if t1.endswith(postfix) or t2.endswith(postfix):
                l = len(postfix)
                t1 = self.to_float(t1[:-l])
                t2 = self.to_float(t2[:-l])
                return t1 < t2
        return t1 < t2

    @staticmethod
    def to_float(t):
        try:
            return float(t)
        except Exception:
            return 0.0
```

PR #566, opened 2016-08-12
limit challenge len significantly

badb:


mfranciszkiewicz:
I would use a stronger source of randomness here, e.g. `random.SystemRandom`. Either way, we should merge this.

PR #578, opened 2016-08-16
small diagnostic changes

badb:
- Remove rand_val from peer session - it may create potential security issue in the future and is not needed
- Remove doubled "nice" argument from vmsnaphsot


golem/network/p2p/p2pservice.py
```diff
@@ -184,6 +184,7 @@ def get_diagnostics(self, output_format):
         for peer in self.peers.values():
             peer = PeerSessionInfo(peer).__dict__
             del peer['node_info']
+            del peer['rand_val']
```
mfranciszkiewicz: Or maybe just remove `node_info` and `rand_val` keys in `PeerSessionInfo`?

PR #582, opened 2016-08-18
Better peer management

mfranciszkiewicz:
- `remove_on_disconnect` replaced with a removed == stored session check 
- problematic `redundant_peers` method removed
- peer management methods using a `Lock`

### DIFF
--- a/golem/network/p2p/p2pservice.py
+++ b/golem/network/p2p/p2pservice.py
@@ -4,6 +4,7 @@ import time
 from collections import deque
 
 from ipaddress import AddressValueError
+from threading import Lock
 
 from golem.core.simplechallenge import create_challenge, accept_challenge, solve_challenge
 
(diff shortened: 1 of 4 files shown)


golem/network/p2p/p2pservice.py
```diff
@@ -268,27 +272,28 @@ def remove_peer(self, peer_session):
...
-        self.suggested_conn_reverse.pop(p, None)
+        peer_id = peer_session.key_id
+        stored_session = self.peers.get(peer_id)
 
-        if p in self.free_peers:
-            self.free_peers.remove(p)
-        if p in self.peer_order:
-            self.peer_order.remove(p)
-
-        self.__send_degree()
```
badb, line 283: Why do you remove this?

mfranciszkiewicz: I moved it to `remove_peer_by_id` which will be called if `stored_session == peer_session`

badb:
I got several errors on Windows in `test_peersession` `test_react_to_hello`.
Line 80-81 `key_id` is removed from `peers`.
Line 88 Actual call is 'Too many peers' not 'Duplicate peers'.

mfranciszkiewicz:
`opt_peer_num` was missing in the mocked config object. Surprisingly it didn't matter much on Linux.

PR #593, opened 2016-08-25
Warn user if blender task is missing additional files

magdasta:


gnr/task/blenderrendertask.py
```diff
@@ -399,6 +399,28 @@ def _change_scope(self, subtask_id, start_box, tr_file):
         extra_data['output_format'] = self.output_format
         return extra_data, (0, 0)
 
+    def after_test(self, results, tmp_dir):
+        ret = []
+        if results and results.get("data"):
+            for filename in results["data"]:
+                if filename.lower().endswith(".log"):
+                    fd = open(filename, "r")
```
badb: Consider using "with" with files and "not in".  It's just a suggestion, it's not important.

badb:
It will save our testers (and us - answering the questions about magenta) a lot of time :+1:

PR #602, opened 2016-08-31
Save stats

### LINKED ISSUE #601: All time stats
We keep stats about number of computed subtasks in current session, but it may be cool to keep also "all time" stats in database.

badb:
Resolves #601

### DIFF
--- a/gnr/gnrapplicationlogic.py
+++ b/gnr/gnrapplicationlogic.py
@@ -256,9 +256,10 @@ class GNRApplicationLogic(QtCore.QObject):
 
         self.customizer.gui.ui.knownTasks.setText(str(known_tasks))
         self.customizer.gui.ui.supportedTasks.setText(str(supported))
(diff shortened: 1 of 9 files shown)


mfranciszkiewicz:
Global stat update is prone to race conditions. Tested with:

```
    def test_for_race_conditions(self):

        n_threads = 10
        n_updates = 100
        n_expected = n_threads * n_updates

        sk = IntStatsKeeper(CompStats)

        def increase_stat():
            n = 0
            while n < n_updates:
                sk.increase_stat("computed_tasks")
                n += 1

        threads = [Thread(target=increase_stat) for _ in xrange(0, n_threads)]

        for t in threads:
            t.start()
        for t in threads:
            t.join()

        assert sk.session_stats.computed_tasks == n_expected
        assert sk.global_stats.computed_tasks == n_expected
```

PR #613, opened 2016-09-06
change restart mechanism

badb:
- Restart PreviewUpdaters in Blender
- Don't change preview paths to None (clean images instead). 
- Don't clean list of subtasks when restaring the task. Just restart them.

### DIFF
--- a/tests/gnr/task/test_blenderrendertask.py
+++ b/tests/gnr/task/test_blenderrendertask.py
@@ -93,6 +93,8 @@ class TestBlenderFrameTask(TempDirFixture):
         print str_
         assert path.isfile(str_)
 
+        assert len(self.bt.preview_file_path) == len(self.bt.frames)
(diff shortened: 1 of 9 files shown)


mfranciszkiewicz:
Blender task preview starts behaving incorrectly if you:
1. Restart a partially computed task
2. Restart that task after completion

During computation, half of the preview is marked in red. On 100%, the preview is incomplete. 
Although I don't think anyone would want to that, it may be a sign of some deeper issue.

tests/gnr/task/test_blenderrendertask.py
```diff
@@ -300,7 +306,13 @@ def test_update_frame_preview(self):
         self.assertTrue(img.size == (300, 200))
         img = Image.open(file4)
         self.assertTrue(img.size == (300, 200))
-        
+
+        print bt.use_frames
```
mfranciszkiewicz: Let's just remove this line

Issue #571, opened 2016-08-15
Node name not visible in the network monitor on the first application run


magdasta:
Nothing serious, but testers tend to get confused because of it.

badb:
We may solve it by additing config update monitor method.

PR #616, opened 2016-09-07
broadcast new node name

paulperegud:
On config change send hello to peers if node name was changed.
This will allow other users see node name for new nodes.


paulperegud:
fixes issue #571 (together with PR #618 )

PR #618, opened 2016-09-08
add on_config_update method to monitor

paulperegud:
Update node name in monitor when node name is changed in client


paulperegud:
fixes issue #571 (together with PR #616)

badb:
That shouldn't be the final solution of the problem, we will need `updateConfig` monitor model and method in the future.  
But it's good enough for now and works, so let's merge it.

PR #621, opened 2016-09-08
CLI

### LINKED ISSUE #548: Fully functional CLI
Would be useful for systems not supporting graphics / accessed remotely. Testers point that it's very important.

mfranciszkiewicz:
1. Command hierarchy is built from decorated classes and methods in `golem.interface.client.*`
2. Command structure is translated to `argparse` parsers

Moreover:
- WebSocket RPC implies connection overhead
- CLI does not require importing `golem.client.*` classes; startup time is improved at the expense of maintaining separate interfaces
- GUI and CLI logic isn't shared; requires refactoring on GUI side
- GUI and CLI shouldn't be used simultaneously; GUI logic assumes sole control of the client (expect crashes)
- regarding commands:
  - `tasks load [file]` automatically tests and starts the task
  - currently there are no commands giving insight into node reputation

Progress:
- [x] Command implementations
- [x] Interactive mode
- [x] Unified command results / feedback
- [x] Tests
- Output formatting
  - [x] Console
  - [x] JSON
- [x] Packaging

Resolves #548


golem/core/fileshelper.py
```diff
@@ -143,3 +146,21 @@ def inner_dir_path(path, directory):
...
+    :return str: directory size in human readeable format (eg. 1 Mb) or "-1" if an error occurs.
+    """
+    try:
+        size = int(subprocess.check_output(['du', '-sb', path]).split()[0])
+    except (OSError, subprocess.CalledProcessError):
+        try:
+            size = int(get_dir_size(path))
+        except OSError as err:
+            import logging
+            logging.getLogger(__file__).info("Can't open dir {}: {}".format(path, str(err)))
```
badb: Previously ig logged to "gnr.gui' not a file. It's probably not a problem to log here, but do we have final policy about naming loggers in `golem`?

golem/interface/client/environments.py
```diff
@@ -0,0 +1,51 @@
...
+
+    sort = Argument(
+        '--sort',
+        choices=table_headers,
+        optional=True,
+        default=None,
+        help="Sort environments"
+    )
+
+    @command(argument=sort, help="Show environments")
```
paulperegud, line 19: We might start calling envs "applications" in UI. In line with {application, task, subtask} naming.

golemcli.py
```diff
@@ -0,0 +1,55 @@
...
+        interactive=('-i', '--interactive'),
+        address=('-a', '--address'),
+        port=('-p', '--port'),
+    )
+
+    args = sys.argv[1:]
+
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(*arguments['interactive'], dest="interactive", action="store_true", default=False)
```
paulperegud: This switch really needs some explanation in golemcli --help

paulperegud: It's also hard to understand because there is no practical difference between "golemcli -i" and "golemcli" if no extra arguments are provided

tests/golem/test_client.py
```diff
@@ -192,6 +196,47 @@ def test_quit(self):
         c.quit()
 
 
+class TestClientRPCMethods(TestWithDatabase):
+
+    @patch('golem.network.p2p.node.Node.collect_network_info')
+    def test_get_node(self, _):
```
badb, line 202: Can you add c.quit() in these functions? There are problems on windows otherwise.

badb (approved):

tests/golem/test_client.py
mfranciszkiewicz: Done

golem/core/fileshelper.py
mfranciszkiewicz: The `__file__` is there by mistake (it was supposed to be `__name__`). I've renamed the logger to `golem.core`.

golemcli.py
mfranciszkiewicz: OK, the interactive mode will have to be set explicitly

PR #623, opened 2016-09-13
less annoying progress bar, second attempt

badb:


paulperegud (commented):
tested after rebasing onto develop, works properly

Issue #423, opened 2016-06-24
Add ability to cancel tests and benchmarks


badb:
Basically we need a method to stop the docker jobs on demand.

PR #629, opened 2016-09-15
add abort test functionality

paulperegud:
split test in UI into preparation and execution phases
add ability to abort test by killing VM in execution phase
client: limit number of concurrent tests to one
enable and disable 'ok' and 'abort' buttons in appropriate moments
disable closing of dialog window via close window button


paulperegud:
fixes #423

mfranciszkiewicz:
I am unable to cancel the tests simply because the window pops up after they're done. Guess I personally don't need an abort button. But maybe it should be opened earlier on?

PR #630, opened 2016-09-15
saving benchmarks results even if save button in options wasnt clicked

### LINKED ISSUE #603: Estimated performance display
Estimated performance in preference tab is displayed with two many decimal places.

magdasta:
Resolves #603


mfranciszkiewicz:
All other changes also get updated. I think it would be better to save benchmark results separately.

Issue #619, opened 2016-09-08
Add table headers in task table


badb:
The columns right now are sometimes to narrow to see full task name / task id. There should be headers that allow user to change width of the column.

PR #632, opened 2016-09-15
Set headers visible and changed their order

magdasta:
#619

### DIFF
--- a/gnr/ui/AppMainWindow.ui
+++ b/gnr/ui/AppMainWindow.ui
@@ -38,7 +38,16 @@
       <property name="sizeConstraint">
        <enum>QLayout::SetNoConstraint</enum>
       </property>
-      <property name="margin">
+      <property name="leftMargin">
+       <number>5</number>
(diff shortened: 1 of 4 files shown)


mfranciszkiewicz:
The `progress` column takes more than half of the view after maximizing the window. Perhaps some reordering / width scaling would help?

paulperegud:
@mfranciszkiewicz @magdasta  This PR is hanging here for a long time. Have you decided on how to proceed?

mfranciszkiewicz:
@paulperegud Not really. Feel free to make changes.

mfranciszkiewicz:
I took the liberty of changing the resize behavior. Users can now resize columns manually.

paulperegud (approved):
LGTM

badb:
One small potential problem: if user resize one column manually, than those changes are lost aftert window is resized. Anyway, it's probably not very important and this is much better than current "solution" .

PR #636, opened 2016-09-27
Ranking thread-safety

### LINKED ISSUE #624: Make reputation system thread safe

paulperegud:
Ranking class is thread-safe as it is. I've added a test and demoted some methods to non-public.

resolves #624

### DIFF
--- a/tests/golem/ranking/test_ranking.py
+++ b/tests/golem/ranking/test_ranking.py
@@ -31,7 +31,8 @@ class TestRankingDatabase(TestWithDatabase):
         thread2.join()
         thread3.join()
         thread4.join()
-        self.assertEqual(0.0, expected)
+        result = r.get_computing_trust("ABC")
+        self.assertEqual(result, expected)
 
     def test_local_rank(self):
         self.assertIsNone(RankingDatabase.get_local_rank("ABC"))
--- a/tests/golem/ranking/test_ranking.py
+++ b/tests/golem/ranking/test_ranking.py
@@ -1,39 +1,15 @@
-import mock
 from threading import Thread
+from unittest import TestCase
+
+from mock import patch, MagicMock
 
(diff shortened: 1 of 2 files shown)


tests/golem/ranking/test_ranking.py
```diff
@@ -1,8 +1,38 @@
...
+        thread2.start()
+        thread3 = Thread(target=run)
+        thread3.start()
+        thread4 = Thread(target=run)
+        thread4.start()
+        thread1.join()
+        thread2.join()
+        thread3.join()
+        thread4.join()
+        self.assertEqual(0.0, expected)
```
mfranciszkiewicz: `expected` is assigned a value prior to the actual test. The outcome doesn't change, though, and the assertion still holds.

paulperegud: Fixed. Was trying not to rely on value 0.0 and omitted last step.

tests/golem/ranking/test_ranking.py
```diff
@@ -1,8 +1,39 @@
...
+from threading import Thread
+
 from golem.tools.testwithdatabase import TestWithDatabase
-from golem.ranking.ranking import RankingDatabase
+from golem.ranking.ranking import Ranking, RankingDatabase, RankingStats
+from golem.client import Client
 
 
 class TestRankingDatabase(TestWithDatabase):
+    def test_increase_trust_thread_safety(self):
```
badb: This test doesn't work on windows:
1) `OperationalError: database is locked` occurs
2) `AssertionError: 0.01098901098901099 != 0.0`

badb (changes requested):

badb:
I've added some locks so test passes on my machine, but feel free to revert and do something else.

PR #660, opened 2016-10-20
Fix: Show a progress dialog for benchmarks

mfranciszkiewicz:
Main thread is no longer blocked by `join`ing the benchmark's thread. The progress dialog is shown prior to (not after) the test.


banasiakadam60:
LGTM

PR #661, opened 2016-10-21
adding extension to out file name

magdasta:


mfranciszkiewicz (changes requested):
The extension is not updated in the text edit on Ubuntu 14.04. Task def contains a correct value.

badb:
If you change the output format then new output format is added to the task name instead of being changed. For example you define output file as: `/home/user/outputfile.PNG`. Then you change 'output format' to 'JPEG'. Output file is not changed at the beginning, but if you click "test task" or click `output file` line edit it is changed to `/home/user/outputfile.PNG.JPEG`.

mfranciszkiewicz (approved):

Issue #664, opened 2016-11-14
Remove pickle


banasiakadam60:
Remove picke and replace it with JSON

paulperegud:
CBOR for network, JSON for local storage.

banasiakadam60:
#681

PR #681, opened 2016-11-21
Pickle #664

banasiakadam60:


golem/monitor/serialization/defaultserializer.py
```diff
@@ -1,4 +1,4 @@
-import json
+import jsonpickle as json
```
badb: It's not the same library: restult `TypeError: encode() got an unexpected keyword argument 'indent'
`

badb:
`ERROR - golem.task.taskserver - Wrong task header received Invalid signature`

PR #696, opened 2016-12-02
Refactor gnr 

### LINKED ISSUE #307: Transfer GNR to GolemApp
New plan:
634e52a4-ac3b-11e6-970a-4935a7e44585

badb:
Resolves #307 

There are still many things that should be changed, but maybe it's a good idea to merge it in current form and fix next thing later. 

Main changes:
- different directory structure, some files moved
- ui files renamed to views
- customizers directories renamed to controllers
- some parts of renderingapplication logic moved to applicationlogic (finally we want to remove this class completely)
- appsmanager (temporary solution) that loads config, so golem and gui classes won't have to import things from specific apps
- rendering environment splited into blender environment and lux environment
- scene file editor splitted into blender and lux editor
- find_task_script removed 
- main.py removed (it can be replaced by golemapp.py)
- generateui now scans not only gui directory but also apps directories for .ui files


apps/blender/benchmark/benchmark.py
```diff
@@ -13,17 +13,17 @@ def __init__(self):
         
         self.normalization_constant = 9360
         
-        self.blender_task_path = os.path.join(get_benchmarks_path(), "blender", "blender_task")
+        self.blender_task_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "test_task")
         
         self.task_definition.output_file = os.path.join(tempfile.gettempdir(), "blender_benchmark.png")
         self.task_definition.tasktype = "Blender"
-        self.task_definition.renderer = "Blender"
+        self.task_definition.task_type = "Blender"
```
mfranciszkiewicz, line 23: Do we need both `tasktype` and `task_type`?

apps/rendering/task/renderingdirmanager.py
```diff
@@ -1,34 +1,24 @@
...
 
 
-def get_task_scripts_path():
-    return path.join(get_golem_path(), "gnr", "task", "scripts")
-
-
-def find_task_script(script_name):
-    scripts_path = get_task_scripts_path()
+def find_task_script(task__file__, script_name):
+    scripts_path = path.abspath(path.join(path.dirname(path.abspath(task__file__)), "..", "resources", "scripts"))
```
mfranciszkiewicz: Perhaps we could use package directories (e.g. `rendering.__path__`) to avoid `..`

gui/view/AppMainWindow.ui
```diff
@@ -2906,7 +2906,7 @@
   </customwidget>
  </customwidgets>
  <resources>
-  <include location="gnr/ui/icons.qrc"/>
+  <include location="gui.ui/icons.qrc"/>
```
mfranciszkiewicz: `gui/view/icons.qrc`

mfranciszkiewicz (commented):
There is room for directory structure improvements in subsequent PRs.

paulperegud:
Provider node does not seem to process tasks on Linux. To replicate start two nodes and prevent them from communicating with nodes that use different code using this small nftables config script: https://gist.github.com/paulperegud/de1d572218e4a7f84e194b501bcfc4dd

EDIT: this seems to be an issue caused by misconfiguration on my side.

PR #699, opened 2016-12-05
removed invalid_logname test

bloose007:
removed invalid_logname test because it was not invalid on my system  and created the directories.

### DIFF
--- a/tests/golem/core/test_common.py
+++ b/tests/golem/core/test_common.py
@@ -40,17 +40,6 @@ class TestConfigLogging(TempDirFixture):
     @patch('logging.config.fileConfig')
     def test(self, file_config):
 
-        invalid_logname = os.path.join(self.path, "/<\0?:/*!\">", "dir", "log.txt")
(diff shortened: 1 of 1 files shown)


badb:
This path was not invalid because it contains strange symbols, but because `join` changed it to a proper (but strange) path inside root directory. Therefore, this test passes for normal users, but failed for root. 
We can assume that our tests should be run only for non-root users, but then modify this test to try to create a normal log files inside root directory or accept and remove this part of the test. 
I vote for accepting this PR.

paulperegud:
Issue of symbols allowed in filenames is really complicated and IMHO this test should not be there. Right now it was testing permissions, which is wrong.

mfranciszkiewicz:
The issue is not complicated at all - a filename cannot contain a null character (*NIX systems) and a specific set of characters on Windows (':' among others). Apparently, `os.path.join` rewrote the `\0` character and the test passed.

PR #703, opened 2016-12-07
changed test_dir_size test to check if user is root

bloose007:
changed test_dir_size test to check if user is root before testing permissions.

### DIFF
--- a/tests/golem/core/test_fileshelper.py
+++ b/tests/golem/core/test_fileshelper.py
@@ -4,6 +4,7 @@
 import os
 import re
 import shutil
+import getpass
 
 from golem.core.common import get_golem_path, is_windows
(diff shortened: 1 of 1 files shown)


banasiakadam60:
LGTM

tests/golem/core/test_fileshelper.py
```diff
@@ -43,10 +44,11 @@ def test_dir_size(self):
             new_size = get_dir_size(self.testdir)
             self.assertGreaterEqual(new_size, size)
 
-            errors = []
-            get_dir_size(self.testdir, report_error=errors.append)
-            self.assertEqual(len(errors), 1)
-            self.assertIs(type(errors[0]), OSError)
+            if getpass.getuser() != 'root':
```
mfranciszkiewicz, line 47: `if os.geteuid() != 0:`
might be less straightforward but IMHO is better.

Issue #667, opened 2016-11-15
Pass in the except block


banasiakadam60:
http://docs.openstack.org/developer/bandit/plugins/try_except_pass.html
We've got some
```
try:
    (...)
except:
    pass
```
blocks in code. We should consider at least log warning about this exceptions. List of try-except-pass blocks:
```
./gnr/application.py:19
./golem/docker/config_manager.py:70
./golem/docker/job.py:196
./golem/http/stream.py:308
./scripts/imunes/imunes_helper.py:166
./scripts/imunes/imunes_helper.py:736
./scripts/packaging/package_creator.py:244
./scripts/packaging/package_creator.py:393
./scripts/packaging/package_creator.py:625
./scripts/packaging/package_creator.py:761
```

PR #710, opened 2016-12-13
Add error messages to the empty try-except blocks

banasiakadam60:
#667 

### DIFF
--- a/golem/http/stream.py
+++ b/golem/http/stream.py
@@ -305,8 +305,10 @@ class ChunkStream:
             # read remaining data
             try:
                 self.__read(drain=True)
-            except:
-                pass
+            except StopIteration as ste:
+                logger.debug("Cannot read remaining data: {}".format(ste.message))
+            except socket.error:
+                logger.debug("Socket error")
             logger.debug("Disconnecting socket: closing socket")
             # dispose of the socket
             self.sock.close()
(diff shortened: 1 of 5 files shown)


golem/http/stream.py
```diff
@@ -305,8 +305,10 @@ def __disconnect(self):
             # read remaining data
             try:
                 self.__read(drain=True)
-            except:
-                pass
+            except StopIteration as ste:
```
mfranciszkiewicz: `StopIteration` will always be raised while working on valid streams; I think we don't need a debug log here.

gui/application.py
```diff
@@ -16,8 +16,8 @@ def __init__(self, app_logic, mainWindowClass):
         try:
             # Linux check might suffice if X11 was the only option available
             QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_X11InitThreads)
-        except Exception:
-            pass
+        except Exception as ex:
+            logger.warning("Error occurred when setting up Qt: {}".format(ex.message))
```
mfranciszkiewicz: This block will always fail on Windows.

badb:
Can you change printing / logging to print exception rather than exception message? Not all exceptions have properly set "message" variable.

banasiakadam60:
Done ;)

PR #714, opened 2016-12-15
Hotfix: enable local mining

chfast:
This change enables mining in local Ethereum node. This make the private Ethereum network more robust and not depending too much on single central node on pebble server.

### DIFF
--- a/golem/ethereum/client.py
+++ b/golem/ethereum/client.py
@@ -25,7 +25,7 @@ class Client(object):
             assert Client.node.datadir == datadir, \
                 "Ethereum node's datadir cannot be changed"
         if not Client.node.is_running():
-            Client.node.start(rpc=True)
+            Client.node.start(rpc=True, mining=True)
         self.web3 = Web3(KeepAliveRPCProvider(host='localhost', port=Client.node.rpcport))
 
     @staticmethod
--- a/golem/ethereum/node.py
+++ b/golem/ethereum/node.py
@@ -1,7 +1,6 @@
 from __future__ import division
 
 import atexit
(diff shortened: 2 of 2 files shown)


banasiakadam60:
If we have `mining` enable by default, maybe we should remove this line: https://github.com/golemfactory/golem/blob/develop/golem/ethereum/node.py#L108 ? It is duplicated by https://github.com/golemfactory/golem/blob/develop/golem/ethereum/node.py#L131 with probably another address.

golem/ethereum/client.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, datadir, nodes=None):
             assert Client.node.datadir == datadir, \
                 "Ethereum node's datadir cannot be changed"
         if not Client.node.is_running():
-            Client.node.start(rpc=True)
+            Client.node.start(rpc=True, mining=True)
```
banasiakadam60, line 28: If we have mining enable by default, maybe we should remove this line: https://github.com/golemfactory/golem/blob/develop/golem/ethereum/node.py#L108 ? It is duplicated by https://github.com/golemfactory/golem/blob/develop/golem/ethereum/node.py#L131 with probably another address.

banasiakadam60 (changes requested):

golem/ethereum/client.py
chfast: Good point.

chfast: Done.

banasiakadam60 (approved):

mfranciszkiewicz (approved):

PR #737, opened 2016-12-30
Initial monitor improvement

jiivan:
* Activated sending incomes and expenditures to monitor
 * A lot of initial refactoring


golem/client.py
```diff
@@ -101,7 +101,7 @@ def __init__(self, datadir=None, transaction_system=False, connect_to_known_host
                          key=self.keys_auth.get_key_id(),
                          prv_addr=self.config_desc.node_address)
 
-        logger.info('Client "{}", datadir: {}'.format(self.config_desc.node_name, datadir))
+        log.info('Client "%s", datadir: %s', self.config_desc.node_name, datadir)
```
badb, line 105: What is the reason to replace format with %?

golem/client.py
```diff
@@ -700,24 +699,24 @@ def __do_work(self):
 
             try:
                 self.p2pservice.sync_network()
-            except Exception as exc:
-                logger.error("p2pservice.sync_network failed: {}".format(exc))
+            except:
```
badb, line 705: Maybe it's a better idea to keep Exception here (until we switch to Python3 at least).

tests/golem/monitor/test_seralizer.py
```diff
@@ -0,0 +1,16 @@
+from unittest import TestCase
```
badb: Typo in a test name

jiivan: T/Y. Fixed that.

golem/client.py
jiivan: log.exception is a convenience method that appends current traceback to LogError, so there is no need to manually reinvent this if this is what you meant.

golem/client.py
jiivan: Actually the main point of this change is to avoid string formatting unless it's necessary. After this change, formatting will occur only for LOG_INFO and below.

gui/startapp.py
```diff
@@ -197,3 +198,9 @@ def start_app(start_ranking=True, datadir=None,
         print(u"Exception in Client process: {}".format(exc))
 
     process_monitor.exit()
+
+    def sigint_handler(signum, frame):
```
badb: After discussion with mfranciszkiewicz: this function should be called before `start_client_process`. 
Also this file is only used to start gui, and we'll probably need it also for console interface. Can you try to move it to `golemapp.py`?

jiivan: Sure. Done it.

PR #742, opened 2017-01-03
Fix versions of dependencies

chfast:
- Fix web3.py version to 3.4.4, it contains some dependency fixes.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,9 +18,9 @@ qt4reactor==1.6
 requests==2.8.1
 six==1.10.0
 Twisted>=16.5
-web3==3.4.4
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -18,7 +18,7 @@ qt4reactor==1.6
 requests==2.8.1
 six==1.10.0
 Twisted>=16.5
-web3
+web3==3.4.4
```
banasiakadam60: Maybe it would be better to set it to '>='?

chfast:
@mfranciszkiewicz @badb this ok?

mfranciszkiewicz (approved):

PR #761, opened 2017-01-13
More tests for golem core

banasiakadam60:


golem/core/keysauth.py
```diff
@@ -258,16 +266,19 @@ def generate_new(self, difficulty):
         """
         min_hash = self._count_min_hash(difficulty)
         priv_key = RSA.generate(2048)
-        pub_key = priv_key.publickey()
+        pub_key = str(priv_key.publickey().n)
```
jiivan, line 269: Isn't this var (pub_key) just overwritten in a while loop and then right after it?

tests/golem/core/test_common.py
```diff
@@ -36,28 +37,36 @@ def test_call_with_dec(self):
...
+        logname_u = unicode(self.logname)
+        dirname_u = unicode(self.dirname)
 
         config_logging(logname_u)
 
         assert os.path.exists(dirname_u)
         file_config.assert_called_with(ANY,
                                        defaults={'logname': logname_u.encode('unicode-escape')},
                                        disable_existing_loggers=False)
+        file_config.called = False
```
jiivan: Is this necessary? If it is, then maybe .reset_mock()?

tests/golem/core/test_hostaddress.py
```diff
@@ -28,34 +29,109 @@ def mock_ifaddresses(*args):
...
+def is_ip_address(address, ip_v4, ip_v6):
+    """
+    Check if @address is correct IP address
+    :param address: Address to be checked
+    :param ip_v4: Check if is IPv4 address
+    :param ip_v6: Check if is IPv6 address
+    :return: True if is correct, false otherwise
+    """
+    import socket
+
```
jiivan: Altough it's just a helper test function may I suggest the following:

1. Use just one flag ip_v6, because it's either ipv4 or ipv6 and there can't be both. ipv4 can be default.
1. Use this flag to determine first argument of socket.inet_pton
1. remove c&p code

jiivan (changes requested):

tests/golem/core/test_hostaddress.py
banasiakadam60: Done

tests/golem/core/test_common.py
banasiakadam60: Done

golem/core/keysauth.py
banasiakadam60: It have to be overwritten

PR #763, opened 2017-01-14
Add CircleCI and Codecov badges to README

MitchellCash:


badb:
Too many brackets (see preview). Also we can appveyor flag: https://ci.appveyor.com/api/projects/status/40ffpq86bl00um02?svg=true

MitchellCash:
@badb yikes, don't know how I missed both of those errors! All fixed now.

PR #765, opened 2017-01-16
Test of apps.core.benchmark

jiivan:


banasiakadam60 (approved):

apps/core/benchmark/benchmarkrunner.py
```diff
@@ -41,16 +40,16 @@ def run(self):
...
+            if task_thread.error_msg:
+                logger_msg += " " + task_thread.error_msg
+            logger.warning(logger_msg)
+            self.error_callback(task_thread.error_msg)
+            return
+
+        res, _ = task_thread.result
+        if res and ("data" in res):
+            if self.benchmark.verify_result(res["data"]):
+                self.success_callback(self.benchmark.normalization_constant / (self.end_time - self.start_time))
```
badb: This test fails on windows generally, not only on appveyor.

jiivan: Thank you. Fixed that

PR #766, opened 2017-01-16
Repeat rmtree in tests if it fails

badb:
Quite ugly solution, but should reduce appveyor test-failures.


chfast:
rmtree has incorrect implementation on Windows. Might be fixed in some Python 3+.

PR #771, opened 2017-01-18
Dockerhub

### LINKED ISSUE #744: Publish golem docker images to DockerHub
I think this can save us a lot of time for CI and alpha testers if we publish docker images and they are just downloaded instead of being built.

badb:
Resolves #744


setup.py
```diff
@@ -65,7 +71,30 @@ def try_building_docker_images():
             finally:
                 os.chdir(cwd)
 
-try_building_docker_images()
+
+def try_pulling_docker_images():
+    if not try_docker():
+        return
```
chfast: I think we should print error and stop the script.

chfast (changes requested):
When there is problem with docker in setup.py we should stop the script with error. The same for missing Qt tools.

chfast (approved):

mfranciszkiewicz (approved):

PR #777, opened 2017-01-19
Qt4 -> Qt5

gborowik:
OSX QT5 xhyve


chfast (commented):
Can you push Qt only changes first?

apps/core/gui/controller/addresourcesdialogcustomizer.py
```diff
@@ -1,5 +1,5 @@
-from PyQt4 import QtCore
-from PyQt4.QtGui import QHeaderView
+from PyQt5 import QtCore
+from PyQt5.QtWidgets import QHeaderView
```
badb, line 2: There are some delays while checking /  unchecking files (but it's probably not very improtant).

gui/view/subtasktableentry.py
```diff
@@ -36,7 +36,7 @@ def __build_row(self):
 
         self.progressBarInBoxLayoutWidget = QWidget()
         boxLayout = QVBoxLayout()
-        boxLayout.setMargin(3)
+        #boxLayout.setMargin(3)
```
badb, line 39: Remove?

badb (approved):

PR #778, opened 2017-01-20
Better tests for golem/docker

banasiakadam60:


golem/docker/job.py
```diff
@@ -58,6 +58,9 @@ def __init__(self, image, script_src, parameters,
         :param str work_dir: directory for temporary work files
         :param str output_dir: directory for output files
         """
+        from golem.docker.image import DockerImage
+        if not isinstance(image, DockerImage):
+            raise TypeError('Incorrect image type: {}. Should be: DockerImage'.format(type(image)))
```
jiivan, line 63: Is it really necessary?

gborowik (approved):

jiivan (approved):

tests/golem/docker/machine/test_docker_machine_manager.py
```diff
@@ -140,35 +140,34 @@ def test_build_config(self):
         config = MockConfig(0, 768, 512)
 
         dmm = MockDockerMachineManager()
-        assert dmm.virtual_box_config == dmm.defaults
+        self.assertEqual(dmm.virtual_box_config, dmm.defaults)
```
chfast: I don't know what you have decided, by py.test picks up `assert` very nicely -- you will get much better error report than from `assertEqual`.

golem/docker/job.py
banasiakadam60: Later in `_prepare` it is used: `image=self.image.name`. So, IMO, it will be better if it fail at the beginning, rather than later, forcing people to debbuging, and searching what's wrong.

tests/golem/docker/machine/test_docker_machine_manager.py
banasiakadam60:
```
>       assert 1 == 2
E       AssertionError: assert 1 == 2
```
```
>       self.assertEqual(1, 2)
E       AssertionError: 1 != 2
```
Both are similar for me.  In some cases using `self.assertXXX` is more readable, eg. `assert foo()` you don't know if it check if function returns True, or just check if it's returns anything. If you use `self.assertTrue(foo())`, then it's clear. And, IMO, it will be better to always do assertions in one way. But, of course, if you want, I can revert it.

jiivan: unittest's assert methods cover vast amount of checks. And it's probably better to be consistent than using assert statements for boolean checks and unittest methods for more complicated cases.

chfast: You can use `asssert` for every check. py.test will analyse whole expression tree attached and will detect comparison operators, will show values of function call arguments, and more. Try it.

tests/golem/docker/test_docker_blender_task.py
```diff
@@ -182,10 +183,62 @@ def test_blender_test(self):
...
+        self.assertIsNotNone(task.src_code)
+        self.assertIsInstance(task.header, TaskHeader)
+        self.assertEqual(task.header.task_id, '7220aa01-ad45-4fb4-b199-ba72b37a1f0c')
+        self.assertEqual(task.header.task_owner_key_id, '')
+        self.assertEqual(task.header.task_owner_address, '')
+        self.assertEqual(task.header.task_owner_port, 0)
+        self.assertIsInstance(task.header.task_owner, Node)
+        self.assertEqual(task.header.subtask_timeout, 1200)
+        self.assertEqual(task.header.node_name, 'some_node')
+        self.assertEqual(task.header.resource_size, 8493997)
```
badb: AssertionError: 8494053L != 8493997

tests/golem/docker/test_blender_job.py
```diff
@@ -42,8 +42,7 @@ def test_blender_job(self):
 
         params = {
             "outfilebasename": "out",
-            "scene_file": DockerJob.RESOURCES_DIR + "/" +
-                          path.basename(scene_files[0]),
+            "scene_file": path.join(DockerJob.RESOURCES_DIR, path.basename(scene_files[0])),
```
badb: with self._create_test_job(script=task_script_src, params=params) as job:
            job.start()
            exit_code = job.wait()
>           self.assertEqual(exit_code, 0)
E           AssertionError: 1 != 0

tests/golem/docker/test_docker_blender_task.py
banasiakadam60: That's strange. It is calculated as:
```
        resource_size = 0
        task_resources = set(filter(os.path.isfile, task_resources))
        for resource in task_resources:
            resource_size += os.stat(resource).st_size
```
Does it failed on Windows or on Linux?

tests/golem/docker/test_blender_job.py
banasiakadam60: Fixed

tests/golem/docker/test_docker_blender_task.py
banasiakadam60: Fixed

tests/golem/docker/test_docker_blender_task.py
```diff
@@ -182,10 +183,62 @@ def test_blender_test(self):
...
+        self.assertEqual(task.header.estimated_memory, 0)
+        self.assertEqual(task.header.min_version, '0.3')
+        self.assertEqual(task.header.docker_images[0].repository, 'golem/blender')
+        self.assertEqual(task.header.docker_images[0].tag, 'latest')
+        self.assertEqual(task.header.max_price, 10.2)
+        self.assertIsNone(task.header.signature)
+        self.assertEqual(task.undeletable, [])
+        self.assertEqual(task.listeners, [])
+        self.assertEqual(len(task.task_resources), 2)
+        self.assertTrue(task.task_resources[0].endswith('docker_blendertask.py'))
```
badb: E       AssertionError: False is not true

Resources are still defined as set: 
  "resources":{
    "py/set":[

chfast: Can you replace `assertTrue` with `assert` and run again with `py.test`?

badb:
```
>       assert task.task_resources[0].endswith('docker_blendertask.py')
E       AssertionError: assert <built-in method endswith of unicode object at 0x052A0458>('docker_blendertask.py')
E        +  where <built-in method endswith of unicode object at 0x052A0458> = 'C:\Sources\golem/apps/blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'.endswith
```

chfast: That's exactly what I'm talking about @jiivan @banasiakadam60.
Compare 
```
E AssertionError: False is not true
```
with
```
>       assert task.task_resources[0].endswith('docker_blendertask.py')
E       AssertionError: assert <built-in method endswith of unicode object at 0x052A0458>('docker_blendertask.py')
E        +  where <built-in method endswith of unicode object at 0x052A0458> = 'C:\Sources\golem/apps/blender/benchmark/test_task/scene-Helicopter-27-cycles.blend'.endswith
```

banasiakadam60: OK, done
@badb could you check if this error still occurs? Because I cannot reproduce it

chfast (approved):

PR #780, opened 2017-01-23
Improved ResourceManager

mfranciszkiewicz:
- `BaseAbstractResourceManager` split into `ResourceCache`, `ResourceStorage` and `AbstractResourceManager`
- less IPFS-specific tests
- IPFS tests are optional


golem/resource/base/resourcesmanager.py
```diff
@@ -12,209 +12,280 @@
...
 
-        self.file_to_hash = dict()
-        self.hash_to_path = dict()
-        self.task_id_to_files = dict()
-        self.task_common_prefixes = dict()
-        self.download_queue = deque()
+def norm_path(path):
+    normpath = os.path.normpath(path)
+    if normpath in ['.', '..']:
+        return ''
```
badb: Is it a good idea to replace ".." with ""? When ".." may occur in ResourceStorage path?

mfranciszkiewicz: ~F.e. when someone uses a directory listing method which includes special "." and ".." directories~
Lone '..' should not occur here.

PR #783, opened 2017-01-23
Services...

chfast:
Simple base class to implement services. Internally uses Twisted reactor/LoopingCall.


chfast:
I will try to improve Payment tests later on.

mfranciszkiewicz (approved):

PR #796, opened 2017-01-29
tGNT payments

chfast:
Testing this is horribly difficult.


golem/ethereum/paymentprocessor.py
```diff
@@ -95,76 +99,85 @@ def check():
...
-        # Remember current check as a temporary status.
-        self.__temp_sync = check()
-        # Mark as synchronized only if previous and current status are true.
-        self.__sync = prev and self.__temp_sync
-        log.info("Synchronized: {}".format(self.__sync))
-        return self.__sync
-
-    def balance(self, refresh=False):
+        if not check():
+            self.__temp_sync = self.__sync = False
```
badb: Can you please avoid multiple assignments like this? It can be quite misleading (at least for me).

jiivan: Do you prefer
`self.__sync = False`
` self.__temp_sync = self.sync`

I disagree with that. In this case I prefer more concise version.

But I have to write that I really don't like private variables. I don't see positives and they just kill inheritance.

chfast: You suggest using protected fields instead of private? Can be. Testing might be also easier.

I still thinking about this as a prototype. Here I was experimenting with "private where possible" to check how it "feels".

Maybe it's time to setup a document with coding rules?

jiivan: Since private/protected are alien concepts in python (private just means magic name mangling) I suggest using plain names without any underscore prefix in most cases.

ad coding rules I propose using zen of python `import this` and most rules from PEP-8. There are many tools to help, and can even be included in a test to assure some level of code quality. By tools I mean pyflakes, pep8, pylint.

badb (approved):

chfast:
Are you ready to merge it?

PR #798, opened 2017-01-30
change default timeout values, first approach

badb:


apps/core/task/coretaskstate.py
```diff
@@ -22,13 +22,21 @@ class CoreTaskDefaults(object):
     def __init__(self):
         self.output_format = ""
         self.main_program_file = ""
-        self.full_task_timeout = 4 * 3600
-        self.subtask_timeout = 20 * 60
+        self._full_task_timeout = 4 * 3600
```
chfast: If this is not used in derived classes, maybe remove it and move the value to the property?

PR #801, opened 2017-01-31
Set src code

badb:
Don't allow users to use they own source code in dockers.


golem/task/tasksession.py
```diff
@@ -588,9 +592,47 @@ def send(self, msg, send_unverified=False):
...
+            return False
+
+        return True
+
+    def __check_docker_images(self, ctd, env):
+        image_found = False
+        for image in ctd.docker_images:
+            if any(env_image.cmp_name_and_tag(image) for env_image in env.docker_images):
+                ctd.docker_images = [image]
+                image_found = True
```
mfranciszkiewicz: `return True`

golem/task/tasksession.py
```diff
@@ -588,9 +592,47 @@ def send(self, msg, send_unverified=False):
...
+                break
+
+        if not image_found:
```
mfranciszkiewicz: `if` and `return`s will no longer be necessary.

golem/docker/image.py
```diff
@@ -15,6 +15,9 @@ def __init__(self, repository=None, image_id=None, tag=None):
         self.tag = tag if tag else "latest"
         self.name = "{}:{}".format(self.repository, self.tag)
 
+    def cmp_name_and_tag(self, docker_image):
+        return docker_image.name == self.name and docker_image.tag == docker_image.tag
```
mfranciszkiewicz: `... docker_image.tag == self.tag`

mfranciszkiewicz (approved):

Issue #735, opened 2016-12-30
Replace packaging with an installer


mfranciszkiewicz:
Current packaging process:
- is tightly coupled with architectural changes
- updating it is too time consuming, since the `freeze` library doesn't produce the expected distribution
- creates an isolated environment (`PATH` and `LD_LIBRARY_PATH` / `DYLD_LIBRARY_PATH`), which requires additional testing
- embedded libraries often cause trouble on different Linux distros

We could create an installer that:
- sets up a portable python distribution (Windows / OS X and Linux if the required version is not present)
- creates a virtualenv with required packages
- installs required system libraries
- places an entrypoint script / binary in system path

Related to #729

banasiakadam60:
- [x] Versioning
- [x] Tests

PR #806, opened 2017-02-02
Replace cx_freeze

banasiakadam60:


banasiakadam60:
#735

.version
```diff
@@ -0,0 +1,2 @@
+[version]
```
mfranciszkiewicz: This file uses the same format as other `.ini` files. Perhaps rename as `version.ini`?

Installer/Installer_Linux/install.sh
```diff
@@ -0,0 +1,167 @@
...
+# @brief check if dependencies (pip, Docker, IPFS and Ethereum) are installed and set proper 'global' variables
+function check_dependencies()
+{
+    # check if pip is installed
+    if [ -n "$( pip -V 2>&1 | grep 'No command' )" ]; then
+        INSTALL_PIP=1
+    fi
+
+    # Check if docker deamon exists
+    if [ -n "$( service docker status | grep 'Loaded: not-found (Reason: No such file or directory)' )" ]; then
```
mfranciszkiewicz: `service --status-all | grep -Fq 'docker'`

Installer/Installer_Linux/install.sh
```diff
@@ -0,0 +1,167 @@
...
+        * ) echo "Please answer yes or no.";;
+    esac
+}
+
+
+# @brief check if dependencies (pip, Docker, IPFS and Ethereum) are installed and set proper 'global' variables
+function check_dependencies()
+{
+    # check if pip is installed
+    if [ -n "$( pip -V 2>&1 | grep 'No command' )" ]; then
```
mfranciszkiewicz: Maybe try checking `$?` instead?

Installer/Installer_Linux/install.sh
```diff
@@ -0,0 +1,167 @@
...
+    fi
+
+    # check if geth is installed
+    if [ -z "$( dpkg -l | grep geth )" ]; then
+        ask_user "Geth not found. Do you want to install it? (y/n)"
+        INSTALL_GETH=$?
+    fi
+
+    # check if ipfs is installed
+    if [ -n "$( ipfs version 2>&1 | grep 'No command' )" ]; then
```
mfranciszkiewicz: `$?`

golem/core/common.py
```diff
@@ -40,7 +40,10 @@ def get_golem_path():
     Return path to main golem directory
     :return str: path to main golem directory
     """
-    return os.path.abspath(os.path.join(os.path.dirname(__file__), "../../"))
+    golem_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../"))
+    if golem_path.endswith('golem') or golem_path.endswith('golem/'):
+        return golem_path
+    return os.path.join(golem_path, 'golem')
```
mfranciszkiewicz: Previously, the expected path was `golem/../` instead of `golem`. Does this change work as intended?

setup/setup_commons.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def get_golem_version(increase):
+    from ConfigParser import ConfigParser
+    from golem.core.common import get_golem_path
+    from os.path import join
+    config = ConfigParser()
+    config.read(join(get_golem_path(), '.version'))
+    version = config.get('version', 'version')
+    if platform.startswith('linux') and increase:    # upgrade version only when building on Linux and building wheel
```
mfranciszkiewicz: I think it's not that easy. We should be able to automate builds using CI. Re-running tests should not change the version.

Installer/Installer_Win/install_script.iss
```diff
@@ -0,0 +1,294 @@
+; Script generated by the Inno Setup Script Wizard.
+; SEE THE DOCUMENTATION FOR DETAILS ON CREATING INNO SETUP SCRIPT FILES!
+
+#define MyAppName "Golem"
+#define MyAppPublisher "Golem Factory GmbH"
+#define MyAppURL "https://golem.network"
+#define MyAppExeName "golemapp.exe"
+; NOTE: if compilation failed, make sure that this variable are set properly and golem is installed from wheel
+; NOTE 2: make sure that you've got DockerToolbox.exe, InstallDocker.msi and python-2.7.13.msi in {#Repository}\Installer\Inetaller_Win\deps
+#define Repository "C:\golem"
```
mfranciszkiewicz: Is it possible to use current directory / a directory relative to current directory?

setup/setup_commons.py
mfranciszkiewicz: We also should use semantic versioning, but this could be a separate PR.

mfranciszkiewicz (commented):
`setup.py bdist_wheel` builds a platform-independent wheel, whereas the included `taskcollector` binary is platform specific. The wheel could be platform specific if taskcollector was built using setuptools. This issue could be also worked around by building a dummy c program and including it in the package.

.version
banasiakadam60: Done

Installer/Installer_Linux/install.sh
banasiakadam60: Done

Installer/Installer_Linux/install.sh
banasiakadam60: Done

Installer/Installer_Linux/install.sh
banasiakadam60: Done

golem/core/common.py
banasiakadam60: My bad. I thought, that 'path to main golem directory' means `/some/path/golem`, not `/some/path/golem/../`
Fixed

setup/setup_commons.py
banasiakadam60: OK, we can do it, but later, not now. For now it is enought as for me.
BTW. CI wouldn't increase version. `increase` parameter is defined as `'bdist_wheel` in argv` in `setup.py`, and as far as I know, CI doesn't launch `bdist_wheel`.

Installer/Installer_Win/install_script.iss
banasiakadam60: There is `src` constant, but I cannot find a way to use constants in `#define`

paulperegud:
@banasiakadam60 please add some documentation to this change. I don't even know where to start.

banasiakadam60:
So first of all: now we can build a wheel package (this will be improved in PR with versioning). you just need to launch `python setup.py bdist_wheel`. It works on Linux (tested on KUbuntu 16.04) and Windows (tested on win 10).
Next, there are two scripts:

- `install.sh` - this is script which allows to install Golem with all dependencies on Linux (tested on Ubuntu, should works also on Debian and Mint). At the begining it checks if golem is installed. Next check if pip/Docker/geth/IPFS is installed. If not it asks user if want to install it manually or if the script should do this for him. After that it install dependencies (if neccessary), download golem wheel from server, install via pip and remove wheel.

- `install_script.iss` - it's Inno Setup install script (and my nightmare). So... You need to have wheel package for windows in `{Repo}/dist` directory and all required `exe` files in `{Repo}\Installer\Installer_win\deps\` dir (this mean: DockerToolbox.exe, geth-windows-amd64-1.5.9-a07539fb.exe, python-2.7.13.msi, pywin32-220.win32-py2.7.exe, VCForPython27.msi. They are too big to keep them in repo). Then there is a lot of black magic, but if you compile it with Inno Setup, you will get golem installer for windows in `{Repo}\Installer\Installer_win` dir. 

From more interesting things:

- `setup.py` was rewritten. Now it contains only necessary things. All auxiliary functions were moved to `setup_common.py`. BTW. TaskCollector is automatically build for linux and OSx (I think). I have no idea how to call `vc.exe` from `cmd` on Windows, so I left this for now. 

- I've removed `scripts\packaging`, because it was used by `cx_freez`, so now, we don't need it.

To sum up:

- On Linux: you need to build wheel, and copy it on the server. `install.sh` will automatically download it and install with dependencies

- On Windows: build wheel, compile `install_script.iss` with Inno Setup and you have installer (I'm going to add building wheel to the `.iss` script)

paulperegud:
Minor stuff from testing:
```
install.sh: line 166: [: too many arguments
```
Steps to reproduce:
1) clean ubuntu 16.04
2) `bash install.sh` <- failed
3) `sudo bash install.sh`

This can be avoided by aborting if install.sh is executed without root rights. This way you can also avoid calling sudo in the script all over the place.

Installer/Installer_Linux/install.sh
```diff
@@ -0,0 +1,169 @@
...
+}
+
+# @brief Install required dependencies
+function install_dependencies()
+{
+    # @todo can we do it without sudo? We can remove the last line and tell user to install it manually,
+    # but still we need sudo when installing docker and geth, so until we find a better solution, it can be done in this way
+    sudo apt-get update
+    if [ $INSTALL_PIP -eq 1 ]; then
+        sudo apt-get install python-pip
```
paulperegud: Instead of calling `apt-get install` here, append package name to variable with list of packages. Call apt-get install $LISTOFPACKAGES at the end of the function. This way user is not asked multiple times if he wants to install packages X Y Z.

Installer/Installer_Linux/install.sh
```diff
@@ -0,0 +1,169 @@
...
+        INSTALL_IPFS=$?
+    fi
+}
+
+# @brief Install required dependencies
+function install_dependencies()
+{
+    # @todo can we do it without sudo? We can remove the last line and tell user to install it manually,
+    # but still we need sudo when installing docker and geth, so until we find a better solution, it can be done in this way
+    sudo apt-get update
```
paulperegud: `apt-get update` be called once, at the end of install_dependencies(), just before single call to `apt-get install`

Installer/Installer_Linux/install.sh
```diff
@@ -0,0 +1,169 @@
...
+}
+
+
+# @brief Download and install golem wheel
+# @return 1 if error occured, 0 otherwise
+function install_golem()
+{
+    wget $HOST$PACKAGE
+    if [ -f $PACKAGE ]; then
+        pip install $PACKAGE
```
paulperegud: `pip install $PACKAGE` fails because of lack of permission to write to `/usr/local/lib/python2.7/site-packages`. Unless script is called with root.

paulperegud (dismissed):
for install.sh:

Whole thing should check if user has root permissions and refuse to run otherwise.

Installer/Installer_Linux/install.sh
banasiakadam60: Fixed

Installer/Installer_Linux/install.sh
banasiakadam60: Fixed

Installer/Installer_Linux/install.sh
banasiakadam60: Fixed

golemapp.py
```diff
@@ -75,5 +78,24 @@ def start_crossbar_worker(unbuffered, module):
     module.run()
 
 
+def __pull_docker_images():
+    from golem.core.common import get_golem_path
+    from subprocess import call
+    from os.path import join
+    with open(join(get_golem_path(), 'apps', 'images.ini'), 'rb') as file_:
+        data = file_.read().split('\n')
+    blender_image = "{}:{}".format(data[1].split()[0], data[1].split()[2])
```
badb: Is it possible to iterate thourh images from `images.ini` and not to call specific images directly?

banasiakadam60: Fixed

mfranciszkiewicz:
On a clean Ubuntu 16.04.2 LTS installation (started with `$ golemapp`):
```
  File "/usr/local/bin/golemapp", line 11, in <module>
    sys.exit(start())
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/golemapp.py", line 55, in start
    start_app(rendering=True, **config)
  File "/usr/local/lib/python2.7/dist-packages/gui/startapp.py", line 117, in start_app
    transaction_system, **config_overrides)
  File "/usr/local/lib/python2.7/dist-packages/gui/startapp.py", line 57, in start_client
    config_logging("client", datadir=datadir)
  File "/usr/local/lib/python2.7/dist-packages/golem/core/common.py", line 103, in config_logging
    from loggingconfig import LOGGING
ImportError: No module named loggingconfig
```

mfranciszkiewicz:
Windows:
- wheel requirements fail to install; this may be related to VC env vars not present (console window closes instantly and it's really hard to tell)
- `openssl.exe` is not needed

banasiakadam60:
@mfranciszkiewicz Should be fixed

mfranciszkiewicz:
Replace line 146 in `setup/setup_commons.py` with 
`        if root != '.' and root.split(os.path.sep)[1] in excluded:`
(it stopped building on windows)

setup/setup_commons.py
```diff
@@ -0,0 +1,173 @@
...
+
+def get_files():
+    from golem.core.common import get_golem_path
+    golem_path = get_golem_path()
+    extensions = ['py', 'pyc', 'pyd', 'ini', 'template', 'dll', 'png', 'txt']
+    excluded = ['golem.egg-info', 'build', 'tests', 'Installer', '.git']
+    beginnig = "../../golem/"
+    result = []
+    for root, dirs, files in walk('.', topdown=False):
+        if root != '.' and root.split('/')[1] in excluded:
```
mfranciszkiewicz: `if root != '.' and root.split(os.path.sep)[1] in excluded:`

PR #808, opened 2017-02-03
Monitor refactoring & additional stats

jiivan:


golem/network/transport/message.py
```diff
@@ -742,17 +743,17 @@ def __init__(self, key_id=None, node_info=None, conn_id=None, super_node_info=No
...
 
     def dict_repr(self):
         return {
-            MessageSetTaskSession.KEY_ID_STR: self.key_id,
-            MessageSetTaskSession.NODE_INFO_STR: self.node_info,
-            MessageSetTaskSession.CONN_ID_STR: self.conn_id,
-            MessageSetTaskSession.SUPER_NODE_INFO_STR: self.super_node_info
+            self.KEY_ID_STR: self.key_id,
+            self.NODE_INFO_STR: self.node_info,
+            Mesself.CONN_ID_STR: self.conn_id,
```
mfranciszkiewicz: `self`

mfranciszkiewicz (changes requested):

golem/network/transport/message.py
jiivan: Thank you - fixed that.

PR #821, opened 2017-02-09
Detect render engine

badb:


apps/blender/task/blenderrendertask.py
```diff
@@ -542,12 +542,22 @@ def after_test(self, results, tmp_dir):
...
                         for w in warnings:
+                            w = u"    {}\n".format(w)
+                            if len(ret) == 0:
+                                ret.append(u"Additional data is missing:\n")
+
                             if w not in ret:
                                 ret.append(w)
+                        if warnings:
+                            ret.append(u"\nMake sure you added all required files to resources.")
+                    with open(filename, "r") as fd:
```
jiivan: Maybe `fd.seek(0)` instead of double opening? And if you're already changing this I think it's a good idea to correct misleading `fd` var name to `f`. Because it's not a file descriptor `<int>` (see: https://docs.python.org/2/library/stdtypes.html#file.fileno ) but a file object.

apps/blender/resources/scripts/blendercrop.py.template
```diff
@@ -2,6 +2,22 @@
...
+
+    def execute(self, context):
+        self.report({"ERROR"}, "Engine " + bpy.context.scene.render.engine + \
+                               " not supported by Golem")
+        return {"FINISHED"}
+
+
+bpy.utils.register_class(EngineWarning)
+engine = bpy.context.scene.render.engine
+if engine != "BLENDER_RENDER" and engine != "CYCLES":
```
jiivan: For me `if engine not in ("BLENDER_RENDER", "CYCLES"):` is more readable.

jiivan (approved):

PR #824, opened 2017-02-13
Drop a connection if no message can be serialized

mfranciszkiewicz:
Drops connections originating from older Golem clients.

Packaging script updated.


badb:
Good fix. We should also think how to add a mechanism that stops nodes from trying to connect again for certain time if connection was dropped.

PR #830, opened 2017-02-17
Verificator

badb:


apps/blender/task/blenderrendertask.py
```diff
@@ -305,13 +297,16 @@ def __init__(self,
                  return_address="",
                  return_port=0,
                  key_id="",
-                 docker_images=None):
+                 docker_images=None,
+                 verificator_class=BlenderVerificator):
```
jiivan: I suppose it wll always be the same class, so maybe do:
```python
class BlenderRenderTask(...):
    VERIFICATOR_CLASS = BlenderVerificator
```

apps/lux/task/luxrendertask.py
```diff
@@ -170,13 +173,15 @@ def __init__(self,
                  return_address="",
                  return_port=0,
                  key_id="",
-                 docker_images=None):
+                 docker_images=None,
+                 verificator_class=LuxRenderVerificator):
```
jiivan: I propose the same solution as in Blender class with class attribute/const.

tests/apps/lux/task/test_luxrendertask.py
```diff
@@ -21,25 +23,24 @@ def test_init(self):
 
 class TestLuxRenderTaskBuilder(TestDirFixture, LogTestCase):
 
-    def test_luxtask(self):
+    def __get_test_lux_task(self):
```
jiivan: why use `__` name mangling?

apps/blender/task/blenderrendertask.py
```diff
@@ -322,6 +317,12 @@ def __init__(self,
         self.preview_updater = None
         self.preview_updaters = None
 
+        self.verificator.compositing = compositing
+        self.verificator.output_format = output_format
+        self.verificator.src_code = self.src_code
+        self.verificator.docker_images = docker_images
+        self.verificator.verification_timeout = subtask_timeout
+
```
jiivan: Can this be moved to `RenderingTask` or even better `CoreTask`?

apps/core/task/verificator.py
```diff
@@ -0,0 +1,59 @@
...
+    WAITING = 1
+    PARTIALLY_VERIFIED = 2
+    VERIFIED = 3
+    WRONG_ANSWER = 4
+
+
+class CoreVerificator(object):
+    handle_key_error_for_state = HandleKeyError(state_check_log_key_error)
+
+    def __init__(self, verification_options=None, advance_verification=False):
```
jiivan: `advance*d*_verification`?

apps/core/task/verificator.py
```diff
@@ -0,0 +1,59 @@
...
+        self.ver_states = {}
+        self.advance_verification = advance_verification
+        self.verification_options = verification_options
+
+    def set_verification_options(self, verification_options):
```
jiivan: Maybe `property.setter` instead of javaish setter method?

banasiakadam60 (approved):

jiivan (changes requested):

apps/core/task/verificator.py
badb: I don't know if this makes a difference. It's not a simple setter, because it also set other option and in more complicated cases may check some restrictions, set default values.

apps/rendering/task/framerenderingtask.py
```diff
@@ -55,12 +27,14 @@ def __init__(self):
...
 
+    VERIFICATOR_CLASS = FrameRenderingVerificator
+
     ################
     # Task methods #
     ################
 
-    def __init__(self, **kwargs):
-        RenderingTask.__init__(self, **kwargs)
+    def __init__(self,  **kwargs):
```
jiivan: Whitespace before `**kwargs`.

jiivan (approved):

PR #831, opened 2017-02-17
Network message hashing

### LINKED ISSUE #829: windows / linux compatibility
Windows nodes cannot verify messages signature from Linux node and vice versa. Probably there's something wrong with serialization method or order of elements in signed message part.

mfranciszkiewicz:
Messages, before hashing, are now serialized with CBOR.

Resolves #829

### DIFF
--- a/golem/network/transport/tcpnetwork.py
+++ b/golem/network/transport/tcpnetwork.py
@@ -503,7 +503,7 @@ class BasicProtocol(SessionProtocol):
         # Drop the connection if no messages were deserialized
         elif data:
             logger.error("Deserialization of messages failed")
-            self.session.dropped()
+            # self.session.dropped()
 
     def _data_to_messages(self):
         return Message.deserialize(self.db)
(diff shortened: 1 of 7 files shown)


golem/network/transport/tcpnetwork.py
```diff
@@ -503,7 +503,7 @@ def _interpret(self, data):
         # Drop the connection if no messages were deserialized
         elif data:
             logger.error("Deserialization of messages failed")
-            self.session.dropped()
+            # self.session.dropped()
```
paulperegud, line 506: why not drop connection here?

paulperegud (approved):

golem/network/transport/tcpnetwork.py
mfranciszkiewicz: There's a bug in session protocol and connections are dropped prematurely. This will be addressed in a separate PR.

paulperegud: Please add a note about this (as a part of moving knowledge online).

PR #841, opened 2017-02-24
IPFS API client

mfranciszkiewicz:
- uses a new(er) IPFS API client library
- removes various `IPFSClient` workarounds


banasiakadam60 (approved):
LGTM

PR #858, opened 2017-03-03
devp2p homebrew compatibility

gborowik:
devp2p homebrew compatibility


gborowik:
Due to another branch that resolves the problem this one can be closed.

PR #865, opened 2017-03-07
Issue864

### LINKED ISSUE #864: Golem crashes when price per task is higher than 18.5 GNT / h
I have no idea what's the reason. There is nothing interesting in the log, no errors are generated. App simply terminates when I click "test task and the price is higher then 18.5 GNT / h. 

Update
Reason: `WAMP message serialization error: huge unsigned int`

badb:
Resolves #864


golem/task/taskbase.py
```diff
@@ -142,7 +142,7 @@ def __getstate__(self):
         state = self.__dict__.copy()
         del state['listeners']
         from pprint import pformat
-        logger.warning(pformat(state))
+        logger.debug(pformat(state))
```
jiivan: Those two lines are removed in #872 (can safe remove lines 144-145).

gui/applicationlogic.py
```diff
@@ -417,20 +417,26 @@ def on_abort():
...
 
-            self.client.run_test_task(self.build_and_serialize_task(task_state))
-
-            return True
+            try:
+                self.client.run_test_task(self.build_and_serialize_task(task_state))
+                return True
+            except Exception as ex:
+                self.test_task_computation_error(ex)
+                return False
```
mfranciszkiewicz: `False` will be returned anyway

jiivan (approved):

gui/applicationlogic.py
```diff
@@ -417,20 +417,26 @@ def on_abort():
...
 
         return False
 
     def build_and_serialize_task(self, task_state, cbk=None):
         tb = self.get_builder(task_state)
         t = Task.build_task(tb)
+        if hasattr(t, "task_definition") and hasattr(t.task_definition, "max_price"):
```
mfranciszkiewicz: This could be merged with the `if` block few lines below. `DictSerializer` converts python objects to dicts, it does not perform serialization per data type.

mfranciszkiewicz (approved):

PR #866, opened 2017-03-07
xhyve: docker VM CPU and memory settings (OS X)

mfranciszkiewicz:
- `DockerMachineManager` renamed to `DockerManager`
- VM management delegated to `VirtualboxHypervisor` and `XhyveHypervisor`
- `DockerManager` takes care of creating and starting the VM named **golem** (from _default_)
- `DockerManager` is responsible for pulling / building docker images


gborowik:
It uses active docker image, eg. 'default', but should use 'golem'

gborowik:
image

PR #867, opened 2017-03-08
Reuse geth if it is already running

paulperegud:


golem/ethereum/node.py
```diff
@@ -118,6 +132,24 @@ def stop(self):
                 log.warn("Cannot terminate node: process {} no longer exists".format(self.__ps.pid))
 
             self.__ps = None
-            self.rpcport = None
             duration = time.clock() - start_time
             log.info("Node terminated in {:.2f} s".format(duration))
+
+"""
```
chfast: The docstring must be inside the function.

golem/ethereum/node.py
```diff
@@ -73,33 +75,45 @@ def __init__(self):
...
+        if self.__ps is not None:
+            raise RuntimeError("Ethereum node already started by us")
+
+        if is_geth_listening(self.testnet):
+            the_chain = "mainnet"
+            if self.testnet:
+                the_chain = "ropsten"
+            running_chain = identify_chain(self.testnet)
+            if running_chain == the_chain:
+                log.info("Using existing Ethereum node ()".format(running_chain))
```
chfast: Shouldn't it be `{}`?

chfast (approved):
Looks good, but 2 small issues.

golem/ethereum/node.py
paulperegud: fixed

golem/ethereum/node.py
paulperegud: fixed

golem/ethereum/node.py
```diff
@@ -81,6 +81,9 @@ def __init__(self):
     def is_running(self):
         return self.__ps is not None or self.__system_geth
 
+    def is_reusing(self):
```
chfast: Let's make `system_geth` public and that's it.

paulperegud: fixed, merging

PR #871, opened 2017-03-09
Fix processing of incoming payments

paulperegud:
* fix contract address
* allow blockHash and blockNumber to be empty (because of geth --light limitations)
* switch back to geth --light

Still TODO: pairing of expected payments with incoming payments, displaying of payments in UI. Will be done as a different PR, maybe by @jiivan


chfast (approved):

golem/ethereum/paymentmonitor.py
```diff
@@ -59,10 +60,14 @@ def process_incoming_payments(self):
             if payee != self.__addr:
                 raise ValueError("Payee should be: {}, but is: {}".format(self.__addr, payee))
             value = int(l['data'], 16)
+
+            log.debug("BlockHash: {}; BlockNumber: {}".format(l['blockHash'], l['blockNumber']))
```
jiivan, line 64: Formatting log message in this way work like this:
 1. Calculate "BlockHash: {}; BlockNumber: {}".format(l['blockHash'], l['blockNumber'])
 1. (simplified) decide wether to emit this log

It can (and IMO should be) optimized to this form:
`log.debug("BlockHash: %r; BlockNumber: %r", l['blockHash'], l['blockNumber'])

What this invocation does is:
 1. Decide wether log record should emitted
 1. Only if it should it will perform string formatting

PR #901, opened 2017-03-21
Fix for Qt5 dialogs (OS X)

### LINKED ISSUE #882: OS X: GUI not clickable after showing a non-modal dialog
Mouse events are ignored after showing a non-modal dialog. It is possible to navigate the controls with keyboard.

mfranciszkiewicz:
Resolves #882


gborowik:
Works great!

PR #905, opened 2017-03-22
add variable to turn of task persistence

badb:
Add variable that turns off saving and restoring task in task manager. It's just a temporary solution before we improve task saving mechanism, ie. make sure that we delete old task and show them in GUI. 
Testnet scaling package should not have this feature turn on.


golem/task/taskmanager.py
```diff
@@ -53,6 +53,9 @@ def __init__(self, node_name, node, keys_auth, listen_address="", listen_port=0,
         self.listen_address = listen_address
         self.listen_port = listen_port
 
+        # FIXME Remove this variable and make task persistance obligatory after it is more tested
+        self.task_persistance = False
```
mfranciszkiewicz: It's persistence :)

jiivan (approved):

PR #907, opened 2017-03-23
function for formatting log text

badb:
Just a quickfix proposition for frustrating unicode errors. Other logs are going to need this fix too, so maybe someone has a better idea.


jiivan:
Resolved in #910

PR #930, opened 2017-03-31
New seed addresses + misc fixes

mfranciszkiewicz:


requirements.txt
```diff
@@ -34,7 +34,7 @@ certifi
 weakreflist
 tabulate
 PyYAML>=3.12
-crossbar
+crossbar==17.2.1
```
chfast, line 37: This enforces other dependencies (Twisted, autobahn, ipaddress), so upgrade of virtualenv used before is not possible.

mfranciszkiewicz: Unfortunately this was required for appveyor. The newest version of `crossbar` broke dependencies on windows.

PR #1004, opened 2017-04-18
Installation script and packaging update for Linux and macOS

### LINKED ISSUE #1005: Docker checksum hardcoded on install.sh didn't match with the one from http://get.docker.com
The install script fails to install docker since there is a wrong checksum hardcoded:

`declare -r docker_checksum='82e964b9a14d294268e4571f542b1508'`

Where it should be: 

```
md5sum /tmp/$docker_script | awk '{print $1}'
1f4ffc2c1884b3e499de90f614ac05a7
```

This triggered the warning:  Cannot install docker. Install it manually: https://docs.docker.com/engine/installation/

### LINKED ISSUE #1006: Linux/Install.sh: Broken paths if installed as root user.
Install.sh can't run properly by root user since /root is outside of /home. Maybe the script should stop with an error message if the user is root, as you can see:

```
root@mw-sw01:/root# echo $GOLEM_DIR
/home//golem/
```

mfranciszkiewicz:
PyInstaller command:
- builds a `tar.gz` distribution for Linux and macOS

Linux installation script:
- fetches the latest release (via github api; **max 60 requests per hour and IP address**)
- updated docker installation script md5 sum (resolves #1005)
- root installation (resolves #1006)

### DIFF
--- a/Installer/Installer_Linux/install.sh
+++ b/Installer/Installer_Linux/install.sh
@@ -3,19 +3,33 @@
 #description    :This script will install Golem and required dependencies
 #author         :Golem Team
 #email          :contact@golemnetwork.com
-#date           :20170113
-#version        :0.1
+#date           :20170418
+#version        :0.2
 #usage          :sh install.sh
 #notes          :Only for Ubuntu and Mint
 #==============================================================================
 
+function release_url()
+{
+    json=$(curl -Ls -H 'Accept: application/json' $1)
+    echo ${json} | python -c '\
+        import sys, json;                          \
+        j = json.load(sys.stdin);                  \
+        k = "browser_download_url";                \
+        print([asset[k] for entry in j             \
+                   if "assets" in entry            \
+               for asset in entry["assets"]        \
+                   if asset[k].find("linux") != -1 \
+              ][0])'
+}
+
 # CONSTANTS
 declare -r CONFIG="$HOME/.local/.golem_version"
-declare -r golem_package="https://github.com/golemfactory/golem/releases/download/0.5.0/golem-linux_x64-0.5.0.tar.gz"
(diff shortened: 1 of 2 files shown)


gborowik:
For macOS works great!

Installer/Installer_Linux/install.sh
```diff
@@ -3,20 +3,37 @@
...
+        print([asset[k] for entry in j             \
+                   if "assets" in entry            \
+               for asset in entry["assets"]        \
+                   if asset[k].find("linux") != -1 \
+              ][0])'
+}
+
 # CONSTANTS
+if [[ $EUID -eq 0 ]]; then
+    declare -r SUDO_USER=$(whoami)
```
banasiakadam60: When run with `sudo` it will set `SUDO_USER` to `root`
```
adam@adam:~ [15:26:33] $ cat test.sh 
#!/bin/bash

echo $(whoami)
adam@adam:~ [15:26:36] $ ./test.sh 
adam
adam@adam:~ [15:26:38] $ sudo ./test.sh 
root
adam@adam:~ [15:26:44] $ 
```
And we use `SUDO_USER` to get a name of user who called script (which is always launch with root privileges: https://github.com/golemfactory/golem/blob/develop/Installer/Installer_Linux/install.sh#L186)
`SUDO_USER       Set to the login of the user who invoked sudo`
So, if you want to use the user name (which always will be `root`), why don't use just `$USER`? As for me, you should remove this `if`

banasiakadam60 (changes requested):

Installer/Installer_Linux/install.sh
mfranciszkiewicz: I agree, it needs to be fixed. But it doesn't work when user is logged in as root, see https://github.com/golemfactory/golem/issues/1006

banasiakadam60 (approved):

PR #1018, opened 2017-04-21
Adds Debian compatibility without breaking Ubuntu compatibility

juanmanavella:
For your consideration, this slightly modified version of install.sh script is able to install Golem on Debian (tested on jessie/sid) without breaking Ubuntu compatibility and should be able to run in any Debian release.


juanmanavella:
I'm closing this pull request favoring an updated install.sh which includes qt5-defaults as it pulls some dependencies on Debian to fix https://github.com/golemfactory/golem/issues/1019

PR #1021, opened 2017-04-24
Measure test time and display it to user

badb:


tests/apps/core/benchmark/test_benchmarkrunner.py
```diff
@@ -34,13 +34,13 @@ def test_task_thread_getter(self):
...
+        with mock.patch.multiple(self.instance, run=mock.DEFAULT, tt=None) as values:
             self.instance.run()
-            values['start'].assert_called_once_with()
+            values['run'].assert_called_once_with()
 
-        with mock.patch.multiple(self.instance, start=mock.DEFAULT, tt=mock.DEFAULT) as values:
+        with mock.patch.multiple(self.instance, tt=mock.DEFAULT) as values:
             self.instance.run()
-            values['start'].assert_called_once_with()
+            # values['run'].assert_called_once_with()
```
jiivan: Uncomment/remove?

tests/apps/core/benchmark/test_benchmarkrunner.py
```diff
@@ -133,3 +131,68 @@ def test_task_computed(self):
         self.instance.task_computed(task_thread)
         self.benchmark.verify_result.assert_called_once_with(result_dict['data'])
         self.instance.success_callback.assert_called_once_with(mock.ANY)
+
+    def test_is_success(self):
+        task_thread = mock.MagicMock()
+        self.instance.start_time = time.time()
+        self.instance.end_time = time.time() + 4
```
jiivan: Maybe use absolute values here?

tests/apps/core/benchmark/test_benchmarkrunner.py
```diff
@@ -133,3 +131,68 @@ def test_task_computed(self):
...
+
+        # Is success
+        task_thread.result = {'data': "some data"}, 30
+        assert self.instance.is_success(task_thread)
+
+        # End time not measured
+        self.instance.end_time = None
+        assert not self.instance.is_success(task_thread)
+
+        # Start time not measure
```
jiivan: `measured`

tests/apps/core/benchmark/test_benchmarkrunner.py
```diff
@@ -133,3 +131,68 @@ def test_task_computed(self):
...
+
+class WrongTask(golem.task.taskbase.Task):
+    def query_extra_data(self, perf_index):
+        raise ValueError("Wrong task")
+
+
+class BenchmarkRunnerWrongTaskTest(TempDirFixture):
+
+    def test_run_with_error(self):
+        super(self.__class__, self).setUp()
```
jiivan: Isn't it called before by unittest?

jiivan: May I suggest a rewrite of this method? I really feel this is overuse of OO programming. :) IMO:
 * No need to use instance attributes here - never used outside of this method. Simple vars would be more readable (without `self.`)
 * No need to use classmethod. A simple lambda or inner function would do. Yet in this context I would use a `Mock()`
 * No need for `_error()` method. It's ideal case for a `Mock()`

golem/task/localcomputer.py
```diff
@@ -69,14 +74,30 @@ def get_progress(self):
...
+        self.success_callback(task_thread.result, self._get_exc_time())
+
+    def computation_failure(self, task_thread):
+        logger_msg = self.comp_failed_warning
+        if task_thread.error_msg:
+            logger_msg += " " + task_thread.error_msg
+        logger.warning(logger_msg)
+        self.error_callback(task_thread.error_msg)
+
+    def _get_exc_time(self):
```
jiivan: At first I though `exc` here was abbreviation of `exception` not `execution`. Maybe `get_time_spent()` or `get_execution_time()` would be clearer?

jiivan: The same goes for `exc_time` argument passed to success callbacks. Maybe `execution_time` or `time_spent`? Or `exec_time` if you prefer abbreviated name?

jiivan (changes requested):

gui/applicationlogic.py
```diff
@@ -476,7 +476,7 @@ def run_benchmark(self, benchmark, label, cfg_param_name):
         self.customizer.gui.setEnabled('recount', False)        # disable all 'recount' buttons
         self.progress_dialog.show()
 
-        self.br.start()
+        self.br.run()
```
mfranciszkiewicz: This operation is blocking the GUI thread (`start` and immediate `join` on a thread). Since we have the whole callback machinery in place, is this change necessary?

tests/apps/core/benchmark/test_benchmarkrunner.py
```diff
@@ -177,22 +176,14 @@ def query_extra_data(self, perf_index):
...
+        benchmark = mock.MagicMock()
+        instance = benchmarkrunner.BenchmarkRunner(
             task=WrongTask(None, None),
             root_path=self.tempdir,
-            success_callback=lambda: self._success(),
-            error_callback=lambda *args: self._error(*args),
-            benchmark=self.benchmark,
+            success_callback=mock.Mock(),
+            error_callback=mock.Mock(),
+            benchmark=benchmark,
```
jiivan: Thank you :)

jiivan (approved):
LGTM

mfranciszkiewicz (approved):

PR #1026, opened 2017-04-26
Refactoring part of Incomes Matching PR

jiivan:
Some refactorings that prepare ground for Incomes Matching.


golem/network/transport/message.py
```diff
@@ -179,6 +179,24 @@ def __repr__(self):
         return "{}".format(self.__class__)
 
 
+class MessageRedefined(Message):
```
badb: This needs another name. Like MessageWithMapping. Or just Message and we can replace all other messages with this class.

golem/task/tasksession.py
```diff
@@ -38,6 +33,17 @@ def call_task_computer_and_drop_after_attr_error(*args, **kwargs):
...
+        @functools.wraps(f)
+        def curry(self, *args, **kwargs):
+            result = f(self, *args, **kwargs)
+            self.dropped()
+            return result
+        return curry
+    return inner
+
+
 class TaskSession(MiddlemanSafeSession):
```
badb, line 47: We should increase protocol version here (if we're introducing new type of Messages)

golem/task/tasksession.py
```diff
@@ -278,34 +277,25 @@ def send_report_computed_task(self, task_result, address, port, eth_account, nod
             return
         node_name = self.task_server.get_node_name()
 
-        self.send(MessageReportComputedTask(task_result.subtask_id, task_result.result_type, task_result.computing_time,
-                                            node_name, address, port, self.task_server.get_key_id(), node_info,
-                                            eth_account, extra_data))
+        self.send(message.MessageReportComputedTask(task_result.subtask_id, task_result.result_type, task_result.computing_time, node_name, address, port, self.task_server.get_key_id(), node_info, eth_account, extra_data))
```
badb, line 280: What line length do you suggest using? Previously we've used 120, right now I'm trying to use 100, but it was never directly specified.

badb (changes requested):
Very cool, but needs more unittests, especially new Message class and tasksession/taskserver refactor functions.

golem/task/taskserver.py
```diff
@@ -518,9 +528,9 @@ def __connection_for_task_request_failure(self, conn_id, node_name, key_id, task
             pc.status = PenConnStatus.WaitingAlt
             pc.time = time.time()
 
-    def __connection_for_task_result_established(self, session, conn_id, key_id, waiting_task_result):
-        self.remove_forwarded_session_request(key_id)
-        session.key_id = key_id
+    def __connection_for_task_result_established(self, session, conn_id, waiting_task_result):
+        self.remove_forwarded_session_request(waiting_task_result.owner_key_id)
```
Elfoniok, line 532: Are you sure that key_id is always owner_key_id?

golem/task/taskserver.py
```diff
@@ -532,18 +542,17 @@ def __connection_for_task_result_established(self, session, conn_id, key_id, wai
...
 
-    def __connection_for_task_result_failure(self, conn_id, key_id, waiting_task_result):
+    def __connection_for_task_result_failure(self, conn_id, waiting_task_result):
 
         def response(session):
-            self.__connection_for_task_result_established(
-                session, conn_id, key_id, waiting_task_result)
+            self.__connection_for_task_result_established(session, conn_id, waiting_task_result)
 
         if key_id in self.response_list:
```
Elfoniok: key_id not defined here?

golem/task/taskserver.py
```diff
@@ -532,18 +542,17 @@ def __connection_for_task_result_established(self, session, conn_id, key_id, wai
...
             self.response_list[conn_id].append(response)
         else:
             self.response_list[conn_id] = deque([response])
 
-        self.client.want_to_start_task_session(key_id, self.node, conn_id)
+        self.client.want_to_start_task_session(wating_task_result, key_id, self.node, conn_id)
```
Elfoniok: key_id not defined here? ...."wating" there is no such thing! One more thing you changed passed arguments list, but i don't see modification of Client function definition

golem/task/tasksession.py
jiivan: Actually new messages will be introduced in second PR. Here the refactoring doesn't change the external representation of Msg.

Elfoniok (changes requested):
There are errors to fix

golem/task/tasksession.py
jiivan: [PEP8](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) recommends something between 72 and 100 characters. I personally disagree with those limits. I feel that all moderately modern workstations are able to properly display any line length (with or without wrapping). That's why I disabled line length checks in [style checker](https://github.com/golemfactory/golem/blob/develop/golem/testutils.py#L137).

And this diff shows why I feel single lines are easier to maintain. When I added `message` namespace I just didn't want to go ascii arting with [aligned indentation](https://www.python.org/dev/peps/pep-0008/#indentation). That's why I quickly joined all lines. This won't be a problem with simple hanging indents though.

But I'm flexible in this matter. If most of you agree with line length restrictions I will enable them.

golem/task/taskserver.py
jiivan: Yes. SEE: https://github.com/golemfactory/golem/pull/1026/files/47a21c8720eecc0be34a7f0e876e273715d67951#diff-58d817e424421f899f53335aa2effd0eL797

Elfoniok: What about call site present inside __connection_for_task_result_failure ?

jiivan: https://github.com/golemfactory/golem/pull/1026/files/47a21c8720eecc0be34a7f0e876e273715d67951#diff-58d817e424421f899f53335aa2effd0eL535

Elfoniok: OK

Elfoniok: OK

tests/golem/task/test_taskserver.py
```diff
@@ -520,6 +521,35 @@ def test_conn_final_failures(self):
         self.assertTrue(ts.remove_responses.called)
         self.assertTrue(ts.task_computer.session_timeout.called)
 
+    def test_task_result_connection_failure(self):
```
jiivan, line 524: @Radagast-red ok?

Elfoniok: Ok

Elfoniok (approved):
As a general note, i understand that you have right in general, But as at least some tools (github for example) poorly support lines longer than 120 columns. I would opt for keeping and formatting code according to this.

badb (approved):

golem/model.py
```diff
@@ -138,6 +142,9 @@ def __init__(self, *args, **kwargs):
         if self.details is None:
             self.details = {}
 
+    def __repr__(self):
+        tx = self.details.get('tx', 'NULL')
+        return "<Payment stid: %r v: %.3f s: %r tx: %s>" % (self.subtask, self.value/denoms.ether, self.status, tx)
```
mfranciszkiewicz: Maybe it's not a big deal, but `%` formatting is inconsistent with the rest of the project.

PR #1033, opened 2017-05-02
Verify img to compare size

badb:
Change name of the function to something that better describes what it does. 
Create different implementation for general rendering and for Blender and move it to external file from imgrepr.

### DIFF
--- a/tests/apps/rendering/resources/test_imgcompare.py
+++ b/tests/apps/rendering/resources/test_imgcompare.py
@@ -2,20 +2,20 @@ import os
 
 from PIL import Image
 
-from apps.rendering.resources.imgcompare import verify_img
+from apps.rendering.resources.imgcompare import check_size
 
 from golem.testutils import TempDirFixture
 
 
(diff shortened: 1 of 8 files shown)


tests/apps/rendering/resources/test_imgcompare.py
```diff
@@ -0,0 +1,21 @@
...
+
+from apps.rendering.resources.imgcompare import check_size
+
+from golem.testutils import TempDirFixture
+
+
+class TestCompareImgFunctions(TempDirFixture):
+    def test_check_size(self):
+        file1 = self.temp_file_name('img.png')
+        for y in range(100, 200):
```
ckcollab: Apologies if I'm intruding, but: why this? Why not edge cases like lower bound, middle, upper bound? Won't this just slow down tests needlessly?

badb: Thanks for pointing this out. Yes, it's unnecessary, I'll fix it.

banasiakadam60 (approved):

PR #1040, opened 2017-05-09
Run as many threads as cpu cores in Blender docker script

badb:


chfast:
Should this affect also testing of a task?

chfast:
So this should improve my new sent tasks or incoming subtasks?

PR #1041, opened 2017-05-09
Pycharm config scheme

Elfoniok:

### DIFF
--- a/golem.xml
+++ b/golem.xml
@@ -1,0 +1,3 @@
+<code_scheme name="golem">
+  <option name="RIGHT_MARGIN" value="80" />
+</code_scheme>


Elfoniok:
These need to be imported once, after project is loaded to pycharm.

banasiakadam60 (approved):

jiivan (commented):
Unfortunately I have no idea how pyCharm works. :)
But I do think this config should be somehow connected to `setup.cfg`.

badb (approved):

vicajoy:
Hi. I would like to know where the config should be posted so that PyCharm would start run the projects.

PR #1046, opened 2017-05-09
Denylist

badb:
Add support for deny list. Users should be able to create file with nodes that they don't want to work with. It's not a longterm solution but it will be very useful for further beta testing.

### DIFF
--- a/golem/task/deny.py
+++ b/golem/task/deny.py
@@ -3,12 +3,15 @@ import os
 DENY_LIST_NAME = "deny.txt"
 
 
-def get_deny_list(datadir, list_name=DENY_LIST_NAME):
+def get_deny_set(datadir, list_name=DENY_LIST_NAME):
+    deny_set = set()
+
     list_path = os.path.join(datadir, list_name)
     if not os.path.isfile(list_path):
-        return []
+        return deny_set
 
     with open(list_path) as f:
-        lines = f.readlines()
+        for line in f:
+            deny_set.add(line.strip())
 
-    return [line.strip() for line in lines]
+    return deny_set
--- a/golem/task/deny.py
+++ b/golem/task/deny.py
@@ -1,0 +1,14 @@
+import os
+
+DENY_LIST_NAME = "deny.ini"
+
+
+def get_deny_list(datadir, list_name=DENY_LIST_NAME):
+    list_path = os.path.join(datadir, list_name)
+    if not os.path.isfile(list_path):
+        return []
+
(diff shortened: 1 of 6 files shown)


golem/task/deny.py
```diff
@@ -0,0 +1,14 @@
...
+
+def get_deny_list(datadir, list_name=DENY_LIST_NAME):
+    list_path = os.path.join(datadir, list_name)
+    if not os.path.isfile(list_path):
+        return []
+
+    with open(list_path) as f:
+        lines = f.readlines()
+
+    return [line.strip() for line in lines]
```
jiivan: I suggest returning `set()`/`frozenset()` here. As it works better with bigger sets.

golem/task/deny.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+DENY_LIST_NAME = "deny.ini"
```
jiivan: `.ini` extension suggests `ini file format` -> https://docs.python.org/2/library/configparser.html

Though I think your solution with simple node id per line is better. Maybe `.lst` or `.txt` extension?

golem/task/deny.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def get_deny_list(datadir, list_name=DENY_LIST_NAME):
+    list_path = os.path.join(datadir, list_name)
+    if not os.path.isfile(list_path):
+        return []
+
+    with open(list_path) as f:
+        lines = f.readlines()
```
jiivan: For potentially bigger sets I suggest:
```python
for line in f:
    blackset.add(line.strip())
```
or `f.xreadlines()`.

jiivan (changes requested):

mfranciszkiewicz (approved):

apolikamixitos:
@badb I've just seen this PR, i checked the code and i saw that there isn't a wildcard support.

Is it possible to add this feature ? Cause i think it could be interesting for groups or companies to use their sort of 'Trusted Nodes' (Whitelisted nodes) and block all others in the meanwhile as long as the issue with shared content will be resolved . What do you think ?

badb:
@apolikamixitos Golem Ids and not names are used here, so I don't think that wild card mechanism will be useful. 
Whitelist mechanism is a good idea. 
Remember, however, that this is temporary alpha/beta Brass solution. It's not a final mechanism that will allow user to certify apps / developers in Clay

jiivan (approved):

PR #1048, opened 2017-05-10
Line length

jiivan:
Main change of this PR is enabling checking of newly agreed on line-length restriction in `golem.testutils.PEP8MixIn`. Some reformatting of existing code is needed in order to pass tests with this change.


chfast (dismissed):
Big NO.

Add description what it is about if not about code formatting.

I don't agree to reformat existing code without introducing other changes around.

And I don't like the format where every argument of a function is in its own line. Nobody has used it so far (I might be wrong about it).

jiivan:
@chfast I've updated the description. Hope we can avoid further misunderstandings now. :)

ad. reformatting code: as I mentioned in a description it is mainly to pass the new requirement in tests.
ad. single line arguments in function definitions: I think it's easiest to maintain (adding/removing/modifying arguments without worrying about line length) and still readable. But if you insist on other form of line breaking I can change this.
BUT in function invocation it really is the easiest way to maintain short line-length.

badb (approved):
I'm not convinced to one argument per line approach, but adding automatic line length checking will be useful.

PR #1052, opened 2017-05-11
Gevent

Elfoniok:
On windows gevent is used with uvent implementation.


requirements.txt
```diff
@@ -19,7 +19,7 @@ pystun==0.1.0
 pytz
 six==1.10.0
 Twisted>=17.1.0
-cffi>=1.9.1
+cffi>=1.10.0
```
mfranciszkiewicz: Please change it to `==`. `cffi==1.10.0` is an explicit dependency of crossbar.

paulperegud: done

paulperegud (approved):

badb (commented):
Can we exclude gevent-reactor and uevent from codecov?

tests/golem/test_defer.py
```diff
@@ -0,0 +1,3022 @@
+# Copyright (c) Twisted Matrix Laboratories.
+# See LICENSE for details.
```
paulperegud: License is the same one we are using?

badb: Twisted is using MIT: https://pypi.python.org/pypi/Twisted

requirements.txt
```diff
@@ -47,3 +46,7 @@ qt5reactor
 GitPython>=2.1.0
 semantic_version
 pyOpenSSL==16.2.0
+gevent>=1.2.1
+devp2p==0.9.1
```
badb: Why are we going back to devp2p in this branch?

Elfoniok: Done

PR #1065, opened 2017-05-16
RPC update: periodically publish task list

mfranciszkiewicz:


golem/client.py
```diff
@@ -114,6 +115,7 @@ def __init__(self, datadir=None, transaction_system=False, connect_to_known_host
         self.last_nss_time = time.time()
         self.last_net_check_time = time.time()
         self.last_balance_time = time.time()
+        self.last_tasks_time = time.time()
```
chfast, line 118: Why not calling `time()` single time?

mfranciszkiewicz: I just continued the style from lines above. We should probably change this.

PR #1067, opened 2017-05-17
Taskpreset

badb:
Add functionality of saving / loading / removing presets for tasks. 
GUI for this part is absolutely terrible, but it's just to help us during tests and new GUI preparation. It has been commented out to not confuse users.


golem/rpc/mapping/aliases.py
```diff
@@ -74,6 +74,9 @@ class Task(object):
     tasks_stats             = 'comp.tasks.stats'
     tasks_known             = 'comp.tasks.known'
     tasks_known_delete      = 'comp.tasks.known.delete'
+    tasks_save_preset       = 'comp.tasks.save.preset'
+    tasks_load_presets      = 'comp.tasks.load.presets'
+    tasks_remove_preset     = 'comp.tasks.remove.presets'
```
mfranciszkiewicz: For continuity, maybe `comp.tasks.preset.[remove|save]`?

golem/task/taskpreset.py
```diff
@@ -0,0 +1,43 @@
...
+                      for task_preset in task_presets}
+    return proper_presets
+
+
+def remove_task_preset(task_type, name):
+    try:
+        query = TaskPreset.delete().where(_is_same_task_preset(task_type, name))
+        query.execute()
+    except Exception:
+        print "NO EXCEPTION"
```
mfranciszkiewicz: I don't think this is true. Or is it?

Issue #971, opened 2017-04-06
Huge increase of usb error codes [Golem 0.4.0] [MacOS 10.12.4]


apolikamixitos:
```
I0407 00:49:44.942492 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:49:54.942804 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:49:57.939926 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:50:10.942422 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:50:23.943716 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:50:24.941162 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:50:36.974703 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:50:49.942457 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:50:54.940353 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:51:02.940006 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:51:16.062512 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:51:24.942395 usb.go:93] handle_events: error: libusb: interrupted [code -10]
I0407 00:51:28.939552 usb.go:93] handle_events: error: libusb: interrupted [code -10]
```

badb:
Bug connecte with geth synchronization.

chfast:
This is caused by geth searching for hardware wallets.

chfast:
geth 1.6.1 is released, so we can upgrade and add `--nousb` option.

badb:
Resolved by #1071

PR #1071, opened 2017-05-18
Rinkeby

chfast:


chfast:
Fixes https://github.com/golemfactory/golem/issues/971.

paulperegud:
@chfast Since we don't reuse geth instances because of golem's specific --datadir param, what do you think about running geth on non-standard network port to not interfere with other geth instances on user's machine?

Apart from that, LGTM.

chfast:
This is planned, but not done here.
I will do this in very next PR.

golem/testutils.py
```diff
@@ -24,16 +24,7 @@ class TempDirFixture(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         logging.basicConfig(level=logging.DEBUG)
-
-        if is_windows():
-            import win32api
-            tmppath = win32api.GetLongPathName(tempfile.gettempdir())
```
badb, line 30: I'm pretty sure that this method was there for a reason and removing it will cause some local disasters.

chfast: I believe it was there because `gettempdir()` returned "short path name". Now we use Golem's datadir which comes from other source and is in "long path" format already.

chfast:
So?

mfranciszkiewicz (approved):

chfast:
@badb ?

badb (approved):

Issue #836, opened 2017-02-21
Use light Ethereum protocol


chfast:
Just add `--light` flag to `geth`.

Blocked by https://github.com/ethereum/go-ethereum/issues/3639.

paulperegud:
Workaround: use IPC instead of RPC. Both switch to light protocol and switch to IPC are implemented in #832

paulperegud:
Blocked by https://github.com/ethereum/go-ethereum/issues/3752

This affects our ability to distinguish between mined and not-yet-mined transactions.

PR #1073, opened 2017-05-19
Incomes matching/messaging

### LINKED ISSUE #504: Connect income with task
Replace get_incoming_payment with get_incomes_list. Check if get_incomes_list is working as it's supposed to. 
It should analyse history of payment from given node. If the sum of received incomes from blockchain is bigger than a sum of confirmed incomes from database, than new waiting incomes should be settled.

### LINKED ISSUE #574: Incomes don't come
Some machines don't show the incomes for computed tasks and account balance doesn't increase, even if many tasks were computed.

Reported by: marcusstair (Windows 7 64 bit). I can also observe it by myself on Ubuntu 14.04.

jiivan:
Adds matching of incoming payments from blockchain to subtasks.

It's achieved by adding two additional message types: `MessageSubtaskPayment` and `MessageSubtaskPaymentRequest`. Payment is sent by requestor and PaymentRequest by provider.

We can't use `geth --light` mode here, because it doesn't return `block_number` and other transaction information in logs. See also: #836 

It partially fixes: #574 (without GUI)
It fixes: #504


golem/ethereum/scripts/eth.py
```diff
@@ -130,19 +133,41 @@ def multi(o, payments):
 def history(o):
     log_id = '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'
     my_addr = '0x' + zpad(o.me.address, 32).encode('hex')
-    outgoing = o.eth.get_logs(from_block='earliest', topics=[log_id, my_addr])
-    incomming = o.eth.get_logs(from_block='earliest', topics=[log_id, None, my_addr])
-
-    balance = o.eth.get_balance('0x' + o.me.address.encode('hex'))
+    my_addr = '0x' + zpad('b3d01938429ad68c821f9623c05b556ad500ff8c'.decode('hex'), 32).encode('hex')  # colo de menin
+    my_addr = '0x' + zpad('ba088df72148e6ed972b0b160cb4754840f5bec9'.decode('hex'), 32).encode('hex')  # a luz de tieta
```
paulperegud: remnants of testing process?

golem/task/taskserver.py
```diff
@@ -805,6 +888,70 @@ def __remove_old_sessions(self):
...
+
+            logger.debug('_send_waiting() len(sessions):%r', len(sessions))
+            if not sessions:
+                p2p_node = p2p_node_getter(elem)
+                if p2p_node is None:
+                    logger.debug('Empty node info in %r', elem)
+                self._add_pending_request(
+                    req_type=req_type,
+                    task_owner=p2p_node,
+                    port=p2p_node.prv_port,
```
badb: Will fail if p2p_node is None

golem/ethereum/scripts/eth.py
```diff
@@ -130,19 +133,47 @@ def multi(o, payments):
 def history(o):
     log_id = '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'
     my_addr = '0x' + zpad(o.me.address, 32).encode('hex')
-    outgoing = o.eth.get_logs(from_block='earliest', topics=[log_id, my_addr])
-    incomming = o.eth.get_logs(from_block='earliest', topics=[log_id, None, my_addr])
 
-    balance = o.eth.get_balance('0x' + o.me.address.encode('hex'))
+    def get_logs_step(**kwargs):
```
badb, line 137: Can you add some comments what does this function do?

golem/ethereum/node.py
```diff
@@ -154,7 +154,7 @@ def start(self):
         if identified_chain != chain:
             raise OSError("Wrong '{}' Ethereum chain".format(identified_chain))
 
-        log.info("Node started in {} s".format(wait_time))
+        log.info("Node started in %ss: `%s`", wait_time, " ".join(args))
```
chfast, line 157: You don't like the `.format` variant?

golem/ethereum/paymentprocessor.py
```diff
@@ -123,15 +123,21 @@ def check():
 
         return True
 
+    def eth_address(self, zpad=True):
```
chfast, line 126: Instead of passing `zpad` as an arg, let's try to guess it by checking the length. There is also `normalize_address()` helper function in `pyethereum`, but it works the other way around (from hex to bytes), but we can use it as a first step, then covert back to hex.

chfast (approved):
I'm fine with the Ethereum / transactions changes, with comments:
- Stop reformatting untouched code.
- I don't believe in this system where you pass additional offchain messages to confirm onchain changes. In the and you cannot trust in any of these messages and have to validate them against the blockchain. So the messages are not very useful.

golem/ethereum/node.py
jiivan: I don't but it doesn't matter. `logging` doesn't support `.format()` formatting (yet?)

I explained it here https://gist.github.com/jiivan/eedb505574ccb2503af48fa2e3043036 pls have a look

golem/ethereum/paymentprocessor.py
jiivan: Isn't it against _Zen of Python_, namely `Explicit is better than implicit.`? But I'm no expert in ethereum logic and I'll gladly see what you mean, maybe in another PR?

mfranciszkiewicz (approved):

PR #1080, opened 2017-05-24
Load incomplete payments on startup

### LINKED ISSUE #266: Load incomplete payments on startup

jiivan:
fixes: #266


golem/ethereum/paymentprocessor.py
```diff
@@ -166,9 +168,18 @@ def _gnt_reserved(self):
...
+        with db.atomic():
+            for awaiting_payment in Payment\
+                    .select()\
+                    .where(Payment.status == PaymentStatus.awaiting):
+                self.add(awaiting_payment)
+
     def add(self, payment, deadline=DEFAULT_DEADLINE):
         if payment.status is not PaymentStatus.awaiting:
             raise RuntimeError("Invalid payment status: {}".format(payment.status))
+        if payment in self._awaiting:
```
chfast: When this can happen?

chfast (commented):
Looks good apart one concern raised.

But in general we should load all non-confirmed payments, not only awaiting?

golem/ethereum/paymentprocessor.py
jiivan: Hopefully it won't happen, but IMHO it's an easy check, worth adding.

chfast: For me it is exactly the opposite. The defensive code like this usually hides problems and means we have not idea what our code is doing. I can accept exception here, but not silent break.

chfast (changes requested):

chfast (approved):

chfast:
:shipit:

PR #1081, opened 2017-05-24
After test

badb:


gui/applicationlogic.py
```diff
@@ -534,6 +534,7 @@ def test_task_computation_success(self, results, est_mem, time_spent,
         # disable 'abort' button
         self.progress_dialog_customizer.enable_abort_button(False)
 
+        logger.debug("Information after test: %s" % after_test_data)
```
chfast: 1. `.format`?
2. Maybe just `"After test: "`?

chfast (approved):

PR #1085, opened 2017-05-27
geth random port + temp dir for tests

chfast:
1. Random port for geth.
2. Bounded get IPC path length (by placing the socket in system temp dir directly).
3. Restore temp dir for tests.

### DIFF
--- a/golem/testutils.py
+++ b/golem/testutils.py
@@ -10,10 +10,8 @@ from time import sleep
 
 from mock import MagicMock
 
-from golem.core.common import get_golem_path
-from golem.core.common import is_windows
+from golem.core.common import get_golem_path, is_windows
 
-from golem.core.simpleenv import get_local_datadir
 from golem.model import Database
 from golem.ethereum import Client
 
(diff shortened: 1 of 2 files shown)


badb (changes requested):
Some blender benchmark & verification tests fail on this branch on windows. Give me some time to investigate this issue.

golem/testutils.py
```diff
@@ -24,7 +22,7 @@ class TempDirFixture(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         logging.basicConfig(level=logging.DEBUG)
-        cls.__root_dir = get_local_datadir('tests')
+        cls.__root_dir = tempfile.mkdtemp(prefix='golem-tests-')
```
badb:
```
if is_windows():
    import win32api
     cls.__root_dir = win32api.GetLongPathName(cls.__root_dir)
``` 
should solve the issue

Elfoniok:
Wouldn't it be better to provide this option from config file?

paulperegud (approved):

paulperegud:
works on Linux; haven't managed to test it on Windows

mfranciszkiewicz (approved):
`test_opt_node` hangs on Windows 10 x64, but merging with `develop` seems to do the trick.

PR #1104, opened 2017-06-05
RPC update: task previews and frames

mfranciszkiewicz:


badb (changes requested):
(1) Calculate total returns wrong value, 
(2) We should change default "use_frames" to True, but then calculate_total crashes ("num_subtasks" is set to 0)

mfranciszkiewicz:
Could you provide some more info? Calculate total is not covered by unit tests.

badb (approved):

PR #1117, opened 2017-06-12
Overflow handling in incomes keeper

jiivan:
Contrary to sqlite3 documentation BigIntegerField overflows (in sum() aggregate) over signed word (int16).


golem/transactions/ethereum/ethereumincomeskeeper.py
```diff
@@ -29,7 +34,7 @@ def received(self, sender_node_id, task_id, subtask_id, transaction_id,
         # signed int
         spent_tokens = model.Income.select(peewee.fn.sum(model.Income.value))\
             .where(model.Income.transaction == transaction_id)\
-            .scalar()
+            .scalar(convert=True)
```
badb, line 37: This will fail if database was empty. Also - is it really needed?

Issue #1124, opened 2017-06-16
Linux installation: upgrade dependencies when updating Golem


mfranciszkiewicz:
When updating an existing Golem installation force an update for depdendencies (geth, hyperg)

PR #1129, opened 2017-06-19
Upgrade dependencies when updating Golem

banasiakadam60:

### DIFF
--- a/Installer/Installer_Linux/install.sh
+++ b/Installer/Installer_Linux/install.sh
@@ -34,8 +34,8 @@ declare -r hyperg=$(release_url "https://api.github.com/repos/mfranciszkiewicz/g
 declare -r hyperg_pack=/tmp/hyperg.tar.gz
 
 # Questions
-declare -i INSTALL_DOCKER=0
-declare -i INSTALL_GETH=0
+declare -i INSTALL_DOCKER=1
+declare -i INSTALL_GETH=1
 # declare -i INSTALL_IPFS=0 # to restore IPFS revert this commit
 declare -i reinstall=0
 
@@ -76,7 +76,7 @@ function ask_user()
 {
     while [ 1 ]; do
         read -p "$@ " yn
-        case $yn in
+        case ${yn} in
             y|Y ) return 1;;
             n|N ) return 0;;
             * ) warning_msg "Please answer yes or no.";;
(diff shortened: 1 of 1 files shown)


mfranciszkiewicz:
The installation script gets stuck on dependency installation. I type the password for sudo and then nothing happens.

Also, docker installation script warns:
```
If you already have Docker installed, this script can cause trouble, which is
why we're displaying this warning and provide the opportunity to cancel the
installation.
```
so we should probably skip updating Docker.

mfranciszkiewicz (dismissed):
Please address the issues from my previous comment

banasiakadam60:
@mfranciszkiewicz 
OK, `https://get.docker.com` does
```
$sh_c "mkdir -p /etc/apt/sources.list.d"
$sh_c "echo deb \[arch=$(dpkg --print-architecture)\] ${apt_url}/repo ${lsb_dist}-${dist_version} ${repo} > /etc/apt/sources.list.d/docker.list"
$sh_c 'sleep 3; apt-get update; apt-get install -y -q docker-engine'
```
So I'll add `docker-engine` to this line https://github.com/golemfactory/golem/blob/update_installer/Installer/Installer_Linux/install.sh#L149

Installer/Installer_Linux/install.sh
```diff
@@ -173,18 +175,18 @@ function install_golem()
 {
     download_package
     result=$?
-    if [[ $result -eq 1 ]]; then
+    if [[ ${result} -eq 1 ]]; then
         return 1
     fi
 
-    tar -zxvf /tmp/$PACKAGE
```
paulperegud, line 180: @banasiakadam60 Next time please split changes 'that makes things better' from core of your change. Otherwise commits are filled with noise and harder than necessary to read. In this particular case changing the way local variables are expanded should be made in different commit from change of the way dependencies are upgraded.

paulperegud (approved):
LGTM (although I haven't managed to test it properly yet)

banasiakadam60:
#1124

banasiakadam60:
@paulperegud It would be great if you could test it tomorrow. Eg. launch script, downgrade `docker` and `geth` (`ethereum` and `docker-engine` packages) and rerun this script.

banasiakadam60:
Tested, it works for me

PR #1136, opened 2017-06-20
Revert reactor switch

Elfoniok:


Elfoniok:
There are to much concerns regarding performance of gevent reactor, these need more investigation and due to architecture change it might be not needed. Since we are not benefiting from it since pydevp2p are on hold, it would be better to merge this.

badb (approved):

PR #1138, opened 2017-06-20
Increase the preview resolution

mfranciszkiewicz:


Elfoniok (changes requested):
There is problem with preview, switch for larger mode disappeared, and very large content is presented in too small original preview window.

mfranciszkiewicz:
@Radagast-red That is an issue with the new interface and the way it handles previews, not the increased preview resolution.

Elfoniok (approved):

PR #1144, opened 2017-06-21
Allow user to specifiy geth port

badb:


jiivan (approved):

chfast:
No tests for it?

This is quite patchy. geth is actually using TCP and UDP ports, but this way you changed only TCP one.

badb:
Are  you sure it's only TCP? There is only this one option in default geth command line settings + I've always have geth listening on specified port both on TCP and UDP.

chfast (approved):

banasiakadam60:
@badb What's with this PR? Should it be merged? It was accepted some time ago, and still is hanging?

PR #1147, opened 2017-06-21
Hyperg info in log error

Elfoniok:

### DIFF
--- a/golem/task/taskserver.py
+++ b/golem/task/taskserver.py
@@ -629,7 +629,8 @@ class TaskServer(PendingConnectionsServer):
         self._mark_connected(conn_id, session.address, session.port)
         self.task_sessions[subtask_id] = session
         session.send_hello()
-        session.send_task_failure(subtask_id, err_msg)
+        session.send_task_failure(subtask_id,
+                                  self.client.daemon_manager.version + err_msg)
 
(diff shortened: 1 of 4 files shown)


golem/task/taskserver.py
```diff
@@ -629,7 +629,8 @@ def __connection_for_task_failure_established(self, session, conn_id, key_id, su
         self._mark_connected(conn_id, session.address, session.port)
         self.task_sessions[subtask_id] = session
         session.send_hello()
-        session.send_task_failure(subtask_id, err_msg)
+        session.send_task_failure(subtask_id,
+                                  self.client.daemon_manager.version + err_msg)
```
chfast, line 633: I'm not sure it is useful, but at least it should be something like `Hyperg version: ...\n`.

Elfoniok: Actually \n is already there. But i agree that it should be more descriptive, i see this as a quick solution for upcoming release. In longer term i would like to send more data about environment in case of fail, but this require rethinking how to send this and present.

badb:
@Radagast-red Are you going to work on this?

Elfoniok:
I will after we move our code base to p3

badb:
I'm closing it for now. Feel free to reopen it when youl continue to work on it.

PR #1148, opened 2017-06-21
issue#852: Made output extensions for Lux uppercase like blender.

maaktweluit:
When investigating issue#852 i have compared Lux vs Blender task creation.

There i saw Blender uses uppercase extensions while Lux uses lowercase.
After changing the extensions used by Lux to uppercase the interface works as expected.

Is there anything else i can add or test to make this better then i found it?

Maaktweluit


badb (approved):
LGTM, thanks your contribution!

PR #1186, opened 2017-06-27
Quickhack for scaling trust config values on backend

badb:


badb:
I've moved this issue to frontend.

PR #1191, opened 2017-06-29
Add Rebbu seed

banasiakadam60:


mfranciszkiewicz:
Are we removing ruggu altogether?

banasiakadam60:
@mfranciszkiewicz Ruggu is (or should be) 0.6.0 seed, so, as for me, it can be removed and replaced with Rebbu. @badb ?

PR #1194, opened 2017-06-30
Logger improvments

Elfoniok:
*Fix mixing up Slogger and Logger objects
*Add PythonLoggingObserver for logs from Twisted loggers (included other processes)
*Setting debug level in crossbar

### DIFF
--- a/gui/startapp.py
+++ b/gui/startapp.py
@@ -65,11 +65,6 @@ def start_client(start_ranking, datadir=None,
                  reactor=None, **config_overrides):
 
     config_logging("client", datadir=datadir)
-    from ethereum import slogging
-    slogging.configure(u':debug')
-    from twisted.python import log
-    observer = log.PythonLoggingObserver(loggerName='twisted')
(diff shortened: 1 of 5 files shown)


Elfoniok:
Monkey patch is not nice thing, I am waiting for some comments here
https://github.com/ethereum/pyethereum/issues/760
at least it fix are issues for now, since many loggers were disabled from level configuration because of this.

gui/startapp.py
```diff
@@ -65,6 +65,11 @@ def start_client(start_ranking, datadir=None,
                  reactor=None, **config_overrides):
 
     config_logging("client", datadir=datadir)
+    from ethereum import slogging
+    slogging.configure(u':debug')
+    from twisted.python import log
+    observer = log.PythonLoggingObserver(loggerName='twisted')
+    observer.start()
```
jiivan: Maybe lines `68-72` could be moved to `config_logging()` function?

PR #1207, opened 2017-07-05
Task are available immediately

Elfoniok:


Elfoniok:
So the motivation is to sent an event as soon as task is created.

Because now task it sent after uploading resources, and rechecking external addresses for node which both are time consuming operation.

However i think we should refactor task/sub tasks management in further development, this change will improve user UX for now.

mfranciszkiewicz:
Alternative implementation: https://github.com/golemfactory/golem/pull/1212

PR #1210, opened 2017-07-06
Preview API call for framesshould be dict and not list

badb:


apps/core/task/coretask.py
```diff
@@ -71,8 +71,9 @@ def get_preview(cls, task, single=False):
     @staticmethod
     def _preview_result(result, single=False):
         if result is not None:
-            return result if single else [result]
-        return None if single else []
+            return result if single or isinstance(result, dict) else \
```
jiivan: I think multiline `if` will be much more readable than this powerful oneliner. :)

jiivan (approved):

PR #1213, opened 2017-07-06
TaskThread: force-stop computation on timeout

### LINKED ISSUE #1192: Docker vm not stopped on timeout
Hey Golem,

When testing on Ubuntu server i have noticed that spawned dockers keep running when a timeout occurs. `tasks stats` shows a timeout has happened but `docker ps` still shows the running job. When killing the docker the log show: `No subtask with id ...`.

Looking at the code i found that timeout handling (golem/task/taskthread.py:48) never checks if the VM is running, [so i've added this](https://github.com/golemfactory/golem/compare/develop...maaktweluit:kill_on_timeout?expand=1).
This code is running for 20h now and no stale dockers detected, so i think its the fix is correct. Though maybe incomplete.
Should the end_date also be set when a timeout occurs? then it would be better to call `self.end_comp()`.

If there is anything else i can add or improve please let me know!

Regards,

Maaktweluit

mfranciszkiewicz:
Resolves #1192

### DIFF
--- a/golem/task/taskthread.py
+++ b/golem/task/taskthread.py
@@ -78,6 +78,12 @@ class TaskThread(Thread):
             self.vm.end_comp()
 
     def _fail(self, error_obj):
+        # Preserves the original cause of failure
+        if self.error:
+            return
+        # Terminate computation (if any)
+        self.end_comp()
+
         logger.error("Task computing error: {}".format(error_obj))
         self.error = True
         self.error_msg = str(error_obj)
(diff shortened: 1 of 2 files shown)


maaktweluit:
Updated my node `mwu-vps/1(v0.7.1)` to this branch rebased on develop.
Current stats are: `308 (err:1 tout:13)`
Most timeouts where from before the bug.

banasiakadam60:
@mfranciszkiewicz Can we merge it?

badb (approved):

PR #1215, opened 2017-07-06
Don't remove PATH when uninstalling on Windows

banasiakadam60:
Signed-off-by: banasiakadam60 <banasiakadam60@gmail.com>


banasiakadam60:
+verified

Issue #1204, opened 2017-07-04
Golem doesn't abort if HyperG is not installed


marmistrz:

mfranciszkiewicz:
Resolved in #1193

PR #1193, opened 2017-06-29
Publish started / exception / done events for long-running methods

mfranciszkiewicz:

PR #1219, opened 2017-07-10
[WIP] Fix Golem doesn't abort if HyperG is not found

matthewsedam:
Issue #1204. This is a work in progress and should not be reviewed just yet. This is also my first pull request for Golem, so any help and guidance is appreciated.


marmistrz (changes requested):
I guess it needs at least a comment why `sys.exit` is not enough. At first sight the current master version looks perfectly correct (while, as we know, it isn't)

matthewsedam:
That's what I'm looking into. I'll add what I find. I'm on vacation at the moment, so I may not respond until next week.

mfranciszkiewicz:
Resolved in #1193

Issue #1232, opened 2017-07-18
The linux Install script needs to install docker-ce instead of docker-engine.


throwaway951:
Just wanted to bring to your attention a name-change that needs updating:
currently, the script throws an error:
`E: Package 'docker-engine' has no installation candidate`

this is mostly a result of the name change:
```
mendu@mineworker:~$ sudo apt-get install docker-engine
[sudo] password for mendu: 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package docker-engine is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  docker-ce
E: Package 'docker-engine' has no installation candidate
```

PR #1233, opened 2017-07-18
Replace docker-engine with docker-ce

snario:
Fixes https://github.com/golemfactory/golem/issues/1232 because `docker-engine` has been replaced with `docker-ce`.

### DIFF
--- a/Installer/Installer_Linux/install.sh
+++ b/Installer/Installer_Linux/install.sh
@@ -161,7 +161,7 @@ function install_dependencies()
     declare -a packages=( openssl pkg-config libjpeg-dev libopenexr-dev \
                libssl-dev autoconf libgmp-dev libtool qt5-default libffi-dev \
                libgtk2.0-0 libxss1 libgconf-2-4 libnss3 libasound2 \
-               ethereum docker-engine )
+               ethereum docker-ce )
     echo -e "\e[91m"
     for package in ${packages[*]}; do
         sudo apt-get install -q -y ${package} >/dev/null


Installer/Installer_Linux/install.sh
```diff
@@ -161,7 +161,7 @@ function install_dependencies()
     declare -a packages=( openssl pkg-config libjpeg-dev libopenexr-dev \
                libssl-dev autoconf libgmp-dev libtool qt5-default libffi-dev \
                libgtk2.0-0 libxss1 libgconf-2-4 libnss3 libasound2 \
-               ethereum docker-engine )
+               ethereum docker-ce )
```
mfranciszkiewicz, line 164: @banasiakadam60 Don't we install docker with the official installation script (https://get.docker.com)? IMO this should be removed.

badb: @banasiakadam60 ?

banasiakadam60: @mfranciszkiewicz Yes we does: https://github.com/golemfactory/golem/blob/develop/Installer/Installer_Linux/install.sh#L128
But it was added just to update `docker`

banasiakadam60 (approved):

banasiakadam60:
@snario Please sign CLA. Without this, we cannot merge this PR.
And thanks for this fix! :+1:

badb (approved):

banasiakadam60:
Fixed in #1397

PR #1234, opened 2017-07-19
Fix path problems

banasiakadam60:
Fix path:
1) Add everything on install
2) Remove golem on uninstall (+ remove from backup)
3) Do not add if key already in path


banasiakadam60:
I've checked this few times, but it would be great if someone find a moment to check it ones again

badb (approved):
Much better! However there's still some room for improvements, ie. Path is not removed and potentially added many times if user specify non-standard installation path ("C:\Program Files\mygolem" for example).

banasiakadam60:
Golem should be removed anyway from `Path`, location doesn't change anything. I've checked that. Also, at the beginning we check if Golem is already in `Path`, if there is anything containing the word `Golem` in the `Path` nothing will be added. And if you already have Golem installed, InnoSetup will not ask you about location and will automatically use the previous one, so there is no way, that `Path` will be outdated

PR #1248, opened 2017-08-03
Reputation cleanup

ggruszczynski:


ggruszczynski:
main changes:

1) negative trust would cause negative weights which can sum up to 0 later on 
--> (sum_weight += weight).

As for quick fix i changed the trust to be from 0 to 1
__neighbour_weight(local_trust):
    return NEIGHBOUR_WEIGHT_BASE ** (NEIGHBOUR_WEIGHT_POWER * local_trust)

2) it seems that there was a mismatch in def computed_neighbours_rank(node_id, neighbours) and
def requested_neighbours_rank(node_id, neighbours) in usage of requested/computed weights

gborowik (approved):

gborowik:
approved!!!

banasiakadam60:
@ggruszczynski What's the status? It was approved 12 days ago, can it be merged? Or you need to change sth in this PR?

ggruszczynski:
to prevent mess, it was scheduled to be merged after release

golem/ranking/helper/min_max_utility.py
```diff
@@ -9,9 +9,24 @@
 logger = logging.getLogger(__name__)
 
 
+# def count_trust(pos, neg):
```
badb: Can you remove commented out code?

golem/ranking/helper/trust_const.py
```diff
@@ -1,6 +1,8 @@
 UNKNOWN_TRUST = 0.0
+NEUTRAL_TRUST = 0.0
+
 MAX_TRUST = 1.0
-MIN_TRUST = -1.0
+MIN_TRUST = 0.0
```
badb: If I understand this properly and we changed trust range from [-1, 1] to [0, 1] then we should change default values in appconfig from -1 to 0 & change the way we scale this value in GUI (old & new). Correct me, if that's not needed.

badb (changes requested):

badb:
Merge together with: https://github.com/golemfactory/golem-electron/tree/trust_range

PR #1254, opened 2017-08-10
Cleanup jsonpickle

marmistrz:
This whole PR would have taken 10x less time if we were using a language which has normal type checking and not duck-typing everywhere :/

- [x] `golem/core/compress.py`
  * The load and save functions seem to be used only in the tests
- [x] `golem/task/taskpreset.py`
  * dead import
- [x] `golem/core/simpleserializer.py`
  * This one is used in `golem/core/prochelper.py` to save snapshots containing the information about all application instance processes. The values stored are: `float` (`timestamp`), `int` and `NoneType`. This can be safely serialized using the `json` module instead.
- [x] `golem/model.py `
  * Used for python_value and db_value of JsonField. The field is used in the following classes:
    * `Payment`, field `dialog`, which is a dictionary containing the following keys
       * `'fee'` - of type `int`
       * `'block_hash'` - of type `str`
       * `'block_number'` - of type `int`
       * `'check'` - of type `bool`
    * `ExpectedIncome`, field `sender_node_details` of class `golem.network.p2p.node.Node` - this needs proper serialization
    * `TaskPreset`, field `data`
- [x] `golem/diag/service.py`
Here only dumps is used, so no risk. But why don't we simply use json?
- [x] `golem/interface/client/tasks.py`
Here a loads in while reading a task from file.


marmistrz:
All the automatic tests pass on Circle and should pass on the other CIs. 

What's left is fixing the linter messages (hence the fail status on Circle), including 71 (sic!) "line-too-longs", mostly from legacy code.

tests/golem/docker/test_docker_luxrender_task.py
```diff
@@ -76,7 +76,7 @@ def set_root_dir(p, new_root_dir=golem_dir):
 
         return task_def
 
-    def _test_task(self):# -> LuxTask():
+    def _test_task(self) -> LuxTask:
```
mfranciszkiewicz, line 79: This function returns a tuple

circle.yml
```diff
@@ -43,7 +43,7 @@ test:
     - pip3 install coverage codecov
   override:
     - ./lintdiff.sh pylint --disable=R apps golem gui scripts setup_util '*.py'
-    - ./lintdiff.sh pylint --disable=R,protected-access tests
+    - ./lintdiff.sh pylint --disable=R,protected-access,too-many-lines tests
```
mfranciszkiewicz: This change belongs to a separate PR.

tests/golem/test_client.py
```diff
@@ -638,12 +732,18 @@ def test_enqueue_new_task_dict(self, *_):
                 '/Users/user/Desktop/folder/model.mesh',
                 '/Users/user/Desktop/folder/stylized_levi.blend'
             ],
-            'name': 'Golem Task 17:41:45 GMT+0200 (CEST)',
-            'type': 'blender',
-            'timeout': '09:25:00',
-            'subtasks': '6',
-            'subtask_timeout': '4:10:00',
-            'bid': '0.000032',
+            'name':
```
mfranciszkiewicz: After the change, this part is (IMO) much less readable. Please do not put values in a new line.

mfranciszkiewicz (commented):
Please avoid making changes to code parts unrelated to the issue you are working on (i.e. reformatting, annotating).

tests/golem/docker/test_docker_luxrender_task.py
marmistrz: I added `@enforce.runtime_validation` and all the tests pass.

tests/golem/test_client.py
marmistrz: I missed that, thanks!

PR #1260, opened 2017-08-14
Fix ethereum address

### LINKED ISSUE #1243: Ethereum address of requestor is unknown
@Radagast-red suspects some message is parsed incorrectly. We need a test showing the problem.

CC @ggruszczynski.

paulperegud:
fixes #1243

### DIFF
--- a/golem/transactions/ethereum/ethereumtransactionsystem.py
+++ b/golem/transactions/ethereum/ethereumtransactionsystem.py
@@ -80,11 +80,7 @@ class EthereumTransactionSystem(TransactionSystem):
         syncing = True
         while syncing:
             try:
-                # syncing = self.__eth_node.is_syncing() # old one
-
-                syncing = self.incomes_keeper.processor.\
-                    _PaymentProcessor__client.is_syncing()  # todo GG
-                # syncing = self.incomes_keeper.processor.synchronized()
+                syncing = self.incomes_keeper.processor.synchronized()
             except Exception as e:
                 log.error("IPC error: {}".format(e))
                 syncing = False
--- a/golem/transactions/ethereum/ethereumtransactionsystem.py
+++ b/golem/transactions/ethereum/ethereumtransactionsystem.py
@@ -6,6 +6,8 @@ from ethereum.utils import privtoaddr
 from golem.ethereum import Client
 from golem.ethereum.paymentprocessor import PaymentProcessor
 from golem.report import report_calls, Component
+from golem.transactions.ethereum.ethereumpaymentskeeper \
+    import EthereumAddress
 from golem.transactions.ethereum.ethereumincomeskeeper \
     import EthereumIncomesKeeper
 from golem.transactions.transactionsystem import TransactionSystem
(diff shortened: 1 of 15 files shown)


golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -27,10 +27,10 @@ def __init__(self, datadir, node_priv_key):
 
         # FIXME: Passing private key all around might be a security issue.
         #        Proper account managment is needed.
-        if not isinstance(node_priv_key, str)\
-                or len(node_priv_key) != 64:
-            raise ValueError("Invalid private key: {}".format(node_priv_key))
-        self.__node_address = b'0x' + privtoaddr(node_priv_key)
+        try:
+            self.__node_address_raw = privtoaddr(node_priv_key)
```
chfast: Don't use `_raw` suffix. We should keep addresses as raw bytes everywhere and encode on the fly.

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -58,7 +58,7 @@ def add_payment_info(self, *args, **kwargs):
 
     def get_payment_address(self):
         """ Human readable Ethereum address for incoming payments."""
-        return '0x' + encode_hex(self.__node_address)
+        return '0x' + encode_hex(self.__node_address_raw)
```
chfast: Is this method needed? If only in UI, let UI to encode it.

chfast (changes requested):

golem/transactions/ethereum/ethereumtransactionsystem.py
paulperegud: No, we also use it in our network code.
E.g. here https://github.com/golemfactory/golem/blob/develop/golem/task/taskserver.py#L618

chfast (approved):

golem/transactions/ethereum/ethereumtransactionsystem.py
ggruszczynski: why dont use the utility class - EthereumAddress

paulperegud: It does not know how to handle private keys

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -21,59 +23,64 @@ def __init__(self, datadir, node_priv_key):
...
+            raise ValueError("Invalid Ethereum address constructed '{}'"
+                             .format(node_address))
+
         log.info("Node Ethereum address: " + self.get_payment_address())
 
-        self.__eth_node = self.incomes_keeper.eth_node = Client(datadir)
         payment_processor = PaymentProcessor(
-            self.__eth_node,
-            node_priv_key,
+            client=Client(datadir),
```
badb, line 43: Be careful, during merge you should add "port" here

tests/golem/docker/test_docker_luxrender_task.py
```diff
@@ -220,8 +220,8 @@ def test_luxrender_real_task_png(self):
     def test_luxrender_real_task_exr(self):
         task = self._test_task()
         task.output_format = "exr"
-        task.res_y = 200
-        task.res_x = 200
+        task.res_y = 300
```
badb, line 223: Why is this change in this PR? @ggruszczynski Can you please do the merge, because I'm not sure if the version from this PR or from develop should stay.

badb:
Moved to https://github.com/golemfactory/golem/pull/1316

PR #1262, opened 2017-08-15
CircleCI: replaced FreeImage manual install with apt.

maaktweluit:
Not sure if this will work.
But since building FreeImage adds 3 minutes to each build it is worth a try.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -19,7 +19,7 @@ dependencies:
   pre:
     - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
     - sudo add-apt-repository -y ppa:ethereum/ethereum && sudo apt-get update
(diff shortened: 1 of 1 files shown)


maaktweluit:
Works as expected, test success :D 
https://circleci.com/gh/golemfactory/golem/4961

badb (approved):

Elfoniok:
Lets merge this, there is no reason i think why Freeimage was installed from source? This can be also done on Travis i think?

Just  today Circle crashed on building Freeimage so using prebuilt binaries would be a good change.

Elfoniok (approved):

maaktweluit:
Thanks for the reply!

To confirm this, i have compared the source of the manual package vs the official github released version and it shows only whitespace changes.

Edit: travis uses a package as well it seems
```
- brew install python3 qt5 openexr freeimage
```

Issue #896, opened 2017-03-20
Remove from preview for tasks in frames


badb:
In `remove_from_preview` function in `renderingtask` there is a line that check if the preview file path is a list. If it is (ie. the task is frame rendering task) then preview is not changed. 
1) This checking should be to framerenderingtask. Rendering task should not know that preview may be a list of previews
2) Function for removing previews from frames should be implemented.

yashodeepm:
I am new to contributing. Can you help me take this issue?

badb:
When subtask is restarted, then the part of image that it produces should be removed from preview. 
For single frame task in Blender it is done by just painting the right part of the preview with black color. This procedure is not done however when there are many frames - and that's something that should be implemented. 
As you noticed properly [this part](https://github.com/golemfactory/golem/blob/develop/apps/rendering/task/renderingtask.py#L167)  of the code should be removed, but only after a proper method in framerenderingtask is implemented. We need `_remove_from_preview` function implemented in framerenderingtask that runs parent method if task is not using multiple frames and it’s own removal method in other case.

PR #1268, opened 2017-08-16
Remove from preview check removed from renderingtask.py(#896)

yashodeepm:
remove_from_preview removed the check if the preview file path was a list as in issue #896.


badb:
It is a good first step. However as you see from test suite, without adding proper implementation of this method in framerenderingtask it's just going to break the code.

PR #1271, opened 2017-08-17
Log error in case of fail to initialize geth

Elfoniok:
If geth is not available the application will throw without a notice...


golem/ethereum/node.py
```diff
@@ -122,6 +122,8 @@ def start(self):
                                       'init', init_file], **pipes)
         init_subp.wait()
         if init_subp.returncode != 0:
+            log.error(
+                "geth init failed with code {}".format(init_subp.returncode))
```
chfast: Construct the message once.

chfast (changes requested):

chfast (approved):

PR #1296, opened 2017-08-24
Performance refactor

badb:


badb:
Don't merge before proper change on GUI site are not tested & merged as well.

golem/task/taskcomputer.py
```diff
@@ -432,6 +393,17 @@ def quit(self):
         for t in self.current_computations:
             t.end_comp()
 
+    @staticmethod
+    def ___check_if_benchmark_is_needed(benchmarks):
```
jiivan: Why triple underscore?

To me this method looks unrelated to this Class (it doesn't use any of class functionality). For this kind of code I suggest making module level functions instead.

golem/task/taskcomputer.py
```diff
@@ -432,6 +393,17 @@ def quit(self):
         for t in self.current_computations:
             t.end_comp()
 
+    @staticmethod
+    def ___check_if_benchmark_is_needed(benchmarks):
+        run_benchmarks = False
+        if benchmarks:
+            for env_id in benchmarks.keys():
+                try:
+                    _ = Performance.get(environment_id=env_id)
```
jiivan: Probably it won't change much, but is it possible to make a single query with something like `select count(*) where environment_id__in=benchmarks.keys()` and then check if `count` equals keys length?

apps/blender/blenderenvironment.py
```diff
@@ -3,6 +3,7 @@
 from golem.core.common import get_golem_path
 from golem.docker.image import DockerImage
 from golem.docker.environment import DockerEnvironment
+from golem.model import Performance
```
banasiakadam60: Is it used anywhere?

banasiakadam60 (approved):
LGTM

golem/client.py
```diff
@@ -858,36 +859,25 @@ def get_environments(self):
...
+                                                       deferred.callback,
+                                                       deferred.errback)
+            result = yield deferred
+            returnValue(result)
         else:
-            raise Exception("Unknown environment: {}".format(env_id))
-
-        result = yield deferred
-        returnValue(result)
+            performance = DefaultEnvironment.run_default_benchmark(
```
mfranciszkiewicz: If I'm reading it correctly, this benchmark is blocking the main thread. Can we execute it in background?

badb: Should be fixed.

PR #1278, opened 2017-08-18
Dummy task

inexxt:
This PR: 
 - creates dummy task, using old dummy task concept (PoW computations)
 - in particular, dummy task makes it easier to handle additional code and data 
 - turns off old dummy task tests
 - refactors `Task`s structure a little, creating some interfaces and allowing for easier task creation
 - annotates types in a few places
 - adds `enforce` dependency, for strict types checking (it is not enabled by default, and currently only works in the `DummyTask` class, but because of type annotations can easily be used everywhere)
 - <del> removes randomness from luxrender tests</del> done in #1286

Known bugs:
 - During task creation, temporary directory is created under /tmp (or equivalent), which is not removed aftrewards. Partial solution to this will be creation of the directory under temp inside task dir, but it has to be done in the Task itself, not in the TaskDefinition.

PR #1297, opened 2017-08-24
Provider requestor communication

inexxt:
This PR:
 - adds support for Requestor-Provider communication during subtask execution. The communication is done by files. Provider saves messages files, which have to be valid jsons, in the `MESSAGES_OUT_DIR`. Then `TaskServer` periodically checks for new messages, and if there are any, loads them to `MessageSubtaskProvReq` structure and sends it to Requestor, which then answers by function `respond_to_message` (function declaration is added on the `Task` level). The answer is packed in the appropriate Message structure, send back, and saved in the `MESSAGES_IN_DIR` in the Providers docker container. The Requestor part is synchronous, the Provider is free to implement messaging in any way he or she wishes.
Currently only `provider asks->requestor answers` mode is supported, because
   - `Task` is not running in lopp
   - For now, there is no usecase for the reversed order of communication (`requestor asks->provider answers`)
 - adds `MESSAGES_IN_DIR, MESSAGES_OUT_DIR, WORK_DIR, RESOURCES_DIR, OUTPUT_DIR` variables to `params.py`. At the same time, these variables are hardcoded both in the `DockerJob` and all `docker_*.py` scripts. It needs refactoring, but it is not so trivial, because a lot of tests depend on these.

<del>This PR is continuation of the #1278 so it has to be merged first.</del> **merged**


Elfoniok:
As a general note.

You mix code cleanup/refactoring with features. It is hard for review such pull request, think if you could divide this in more pull request. First you make commit with cleanup/refactor then you add new features.

apps/core/task/coretask.py
```diff
@@ -4,7 +4,7 @@
 import os
 import uuid
 from enum import Enum
-from typing import Type
+from typing import Type, Dict
```
Elfoniok: Is it use anywhere?

golem/network/transport/message.py
```diff
@@ -1141,6 +1153,62 @@ def __init__(self, subtask_id=None, **kwargs):
         super(MessageSubtaskPaymentRequest, self).__init__(**kwargs)
 
 
+class MessageSubtaskReqToProv(Message):
```
Elfoniok, line 1156: Are you sure you need two exactly same messages just with different names?

apps/dummy/resources/scripts/docker_dummytask.py
```diff
@@ -1,14 +1,19 @@
 from __future__ import print_function
 
 import imp
+import json
```
Elfoniok: Not used

apps/dummy/resources/scripts/docker_dummytask.py
```diff
@@ -1,14 +1,19 @@
...
 import os
 
 import params  # This module is generated before this script is run
+import time
```
Elfoniok: Not used, i can see that you use it in commented code, but first off all this testing code should be provide as a snippet, not in PR, second import for this should be grouped together with code.

apps/dummy/task/dummytask.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import os
 import random
+from typing import Dict
```
Elfoniok: Same as above

apps/dummy/task/dummytask.py
```diff
@@ -18,6 +19,7 @@
 logger = logging.getLogger("apps.dummy")
 
 
+@enforce.runtime_validation(group="dummy")
```
Elfoniok, line 21: We don't want this in out code currently, why would we?

golem/docker/job.py
```diff
@@ -4,10 +4,13 @@
 import posixpath
 import threading
 from os import path
+from typing import List, Dict
```
Elfoniok: List is not used

golem/docker/task_thread.py
```diff
@@ -1,7 +1,11 @@
+import hashlib
 import logging
 import os
+from typing import List, Dict
 
 import requests
+from jsonpickle import json
```
Elfoniok: jsonpickle should  not be used anymore

golem/task/taskbase.py
```diff
@@ -423,3 +423,11 @@ def get_output_states(self) -> List:
...
         """
         return []
+
+    def react_to_message(self, subtask_id, data: Dict) -> Dict:
+        """ Reacts to a message from provider
+        :param subtask_id:
+        :param data: Message data in dict
+        :return: Response in dict
+        """
+        return {}
```
Elfoniok: Please add a new line on each of the files.

golem/network/transport/message.py
inexxt: I think it's clearer to do that this way (each message type has its own response function, see dict in the `tasksession.py/__set_msg_interpretations`) than including a flag indicating which way is the message going, then checking for that flag in the response function and calling appropriate function on that.

apps/dummy/resources/scripts/docker_dummytask.py
inexxt: Yes, I didn't know where to put it.
Should I create a gist for that and include the link in the code, or create an issue with "messaging integration test" and include it there?
I left this code because if we were to write integration tests in the near future (and I think something like messaging should be tested in that way), this code may be used for that.

apps/dummy/task/dummytask.py
inexxt: Full type checker is disabled at default (see `enforce.config({'groups': {'set': {'dummy': False}}})` at the end of the file), but it can be useful during tests. Type checker like `mypy` can't rule out all the bugs in the types in the dynamic typed language like Python, so it is useful to do a test run before release/merging PR/etc with all types checked in the runtime, see if everything works ok, and then disable it in the production.

PR #1299, opened 2017-08-24
TaskServer.__remove_old_sessions fix

mfranciszkiewicz:


marmistrz (changes requested):
I think the commit message should be more meaningful, something like: "Fix the task_session field being destructively modified..." or "Always copy the field in ..." 

Other than than, lgtm

badb (approved):

PR #1300, opened 2017-08-24
Slow and fast tests

maaktweluit:
Split the testing into 2 suites:
- Fast 
- Slow

The fast tests are run by default, they should contain: code style, compilation, type and unit tests.
The slow tests run only when a  `--runslow` argument is given. This triggers all tests to run, adding integration tests to the list above.

Any test can be added to the "slow" suite by attaching the `@pytest.mark.slow` decorator:
```
import pytest

@pytest.mark.slow
def test_hello_world:
# ...
```

All CI platforms are configured to run the fast test on pull requests with < 2 approvals. All branches and approved pull requests get to endure the slow test.

Known issues:
- Does not re-trigger builds after 2 approvals, leaving the light green

Possible solutions:
- [x] Investigate if a web-hook on 2 approvals can be configured ( each approval also possible )
- [x] If above solution does not work: the build can always fail when testing fast

Known issues:
- Due to github request limits the agents can not always get the approvals

Possible solutions:
- [x] Add API key for github requests #1314
- [x] Use private build agents / workers


maaktweluit:
[Found a way](https://gist.github.com/bjhess/2726012) to trigger a webhook on each pull request event.
We should make an endpoint to filter the requests and trigger the build accordingly.

I propose to implement the exit code for now, it can even skip coverage in fast mode.
For the hooks we can make an issue or wait for buildbot to make it better.

maaktweluit:
Made all agents fail when only fast test ran. Looks like only circle is able to query github today, added a known issue in the main comment.

Also picked #1310 fix to this branch to get proper linting results.

maaktweluit:
Ready for review!

Builds will only show proper coverage-diff and full green statuses after 2 approvals

maaktweluit:
Marked some more tests as slow.

On circle this removed 400s compared to the previous build.
On appveyor it only removed 60s, probably due to `@ci_skip`.

Please note 'checks' are red until 2 approvals, unless #1314 happens.

marmistrz:
The status checks being red has one intrinsic issue. The status checks failing or succeeding makes it easier to find out if my change is not broken. If the status check is always red, people will start to ignore it, finding out that the tests fail only when they're *ready* with a change.

Besides, Circle tests failing will clutter the inbox.

I understand that your intent is to block changes with not enough approvals?

maaktweluit:
Agree with all points above.
As described in the OP there where 2 solves, use github (statuses and webhooks) OR fail when ran slow.

Also i am not sure if coverage show properly when only parts of the test run.

I propose to leave this pull as is and make a follow up issue to migrate the exit 0 to a new status / webhook call.

marmistrz:
My idea would be to make the approval check a separate status check, just as the CLA check is.

We could just mark it somewhere that the coverage should be ignored until the full tests run.

maaktweluit:
the approval check is already native to github, the current problem is using the same status for slow and fast runs.

> Does not re-trigger builds after 2 approvals, leaving the light green

This problem was solved with `exit 1`, and could be solved with a separate status or webhook.

@badb should I change this in this issue or make a follow up?

badb (approved):

.travis.yml
```diff
@@ -21,21 +21,28 @@ cache:
...
+  # requirements
   - sudo -HE pip3 install --upgrade pip setuptools six virtualenv pytest-cov codecov
   - sudo -HE pip3 install -r requirements.txt
   - sudo -HE pip3 install -r requirements-test.txt
   - sudo -HE python3 setup.py develop
 
 script:
-  - python3 -m pytest --cov=golem
+  - python3 -m pytest --cov=golem${ARG_RUNSLOW}
+  - if [[ -z "${ARG_RUNSLOW}" ]]; then echo "Not enough approvals."; exit 1; fi
```
marmistrz: I still believe that failing the tests just because there are not enough approvals is not the right way to do this. This should be properly done using the GitHub hooks.

If the fast CI suite will always fails, people will start ignoring the failing status checks on Circle/Travis. Moreover, people with a CircleCI account (as me, I had a CI there for another project) get an e-mail message whenever a test fails. With the current approach - there'll be a mail flood here.

Additionally, it's sometimes useful to see which is the last commit to pass the test, especially when developing something in a hurry (because it's a blocker)

circle.yml
```diff
@@ -17,32 +17,42 @@ dependencies:
...
     - python3 setup.py develop
   pre:
+    # geth
     - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
     - sudo add-apt-repository -y ppa:ethereum/ethereum && sudo apt-get update
-    - sudo apt-get install -y qt5-default qtbase5-dev-tools geth unzip gcc-5 g++-5 libfreeimage3 libfreeimage-dev
+    - sudo apt-get install -y qt5-default qtbase5-dev-tools geth
+
+    # taskcollector
+    - sudo apt-get install -y gcc-5 g++-5 libfreeimage3 libfreeimage-dev
```
marmistrz, line 29: Separating the dependency requirements is basically a good idea, but will require `apt` to parse the dependency tree twice. I'm not sure if this won't slow down the CI a little (but this may not be a great deal)

circle.yml
```diff
@@ -17,32 +17,42 @@ dependencies:
...
     - /usr/local/bin/ipfs config --json SupernodeRouting.Servers "[]"
     - /usr/local/bin/ipfs config --json Addresses.Swarm '["/ip6/::/tcp/4001", "/ip6/::/udp/4002/utp", "/ip4/0.0.0.0/udp/4002/utp"]'
     - /usr/local/bin/ipfs daemon:
         background: true
 
-    - wget https://github.com/mfranciszkiewicz/golem-hyperdrive/releases/download/v0.2.1/hyperg_0.2.1_linux-x64.tar.gz
-    - tar xvf hyperg_0.2.1_linux-x64.tar.gz
-    - ./hyperg/hyperg:
-        background: true
+    # Slow tests only on branches and PRs with >= 2 approvals
```
marmistrz: This is out of date.

conftest.py
```diff
@@ -0,0 +1,18 @@
...
+import _pytest
+
+
+def pytest_addoption(parser: _pytest.config.Parser) -> None:
+    parser.addoption("--runslow", action="store_true",
+                     default=False, help="run slow tests")
+
+
+def pytest_collection_modifyitems(config: _pytest.config.Config,
+                                  items: list) -> None:
```
marmistrz: Do you know the type of the items contained in the `items`? If yes, then you may annotate it as `List[T]`, where `T` is the type.

scripts/get-slow-argument.py
```diff
@@ -0,0 +1,41 @@
...
+
+# When build is not a PR the input is: "" or "false"
+if pull_request_id not in ["", "false"]:
+    base_url = "https://api.github.com/" \
+        "repos/golemfactory/golem/pulls/{}/reviews"
+    url = base_url.format(pull_request_id)
+
+    try:
+        # Github API requires user agent.
+        req = urllib.request.Request(url, headers={'User-Agent': 'build-bot'})
```
marmistrz: You may want to use the famous `requests` module, especially that Golem is already using it, it's a little less verbose.

marmistrz (changes requested):

.travis.yml
maaktweluit: Fully agree, we can fix this in #1365

circle.yml
maaktweluit: good thought!

was thinking here the taskcollector should be a single binary installation

I can fix it if you want, but would propose to leave it.

circle.yml
maaktweluit: fixed 3x

conftest.py
maaktweluit: nope, but would love to find out.
Looked it up and seems like `List[_pytest.Item]` from [source](https://github.com/pytest-dev/pytest/blob/e57cc5571979009f0dd7ec447891d7ab761a93d5/_pytest/main.py#L534)
Do you know an easy reference for looking types up?
Also i see currently the `list` is without capital, does this matter for python typing?

scripts/get-slow-argument.py
maaktweluit: thanks for the tip, replaced it super easily.

maaktweluit:
ready for another round of reviews.

travis fails in a weird way, can not reproduce on my mac.
will look into this and the next comments tonight.

conftest.py
marmistrz: If you are not sure, you can always try `enforce`. `List` (capital) is simply a generic version of `list`. By specifying the contained type you can get extra checks, but in a helper script it's not a big deal.

As for looking types up.... welcome to Python! :) There's basically no such way, since the essence of Python is dynamic duck-typing. What you can do is:
* inspect the code
* use `print(type(foo))` to find out what the type is in runtime
* use `enforce` to validate your hypotheses
* ask colleagues

circle.yml
marmistrz: leaving this up to you, it shouldn't be a big deal either. If we're going to move to our own Docker image, then we can leave it.

CircleCI 2.0 uses Docker anyway, the image circleci/classic:latest by default, I don't know about 1.0 (which we're using right now)

marmistrz:
From travis:
```
  File "scripts/get-slow-argument.py", line 27, in <module>
    result = [a for a in json_data if a["state"] == "APPROVED"]
  File "scripts/get-slow-argument.py", line 27, in <listcomp>
    result = [a for a in json_data if a["state"] == "APPROVED"]
TypeError: string indices must be integers
```
Looks like some type mismatch, it may be a good idea to print the response.

Note that some tests may fail due to #1278 having been merged.

conftest.py
maaktweluit: Thanks for the great tips!

Updated it and it seems to work when running and linting it.

maaktweluit:
Lowered the required approvals to 1 and fixed travis!

badb (commented):
Test

badb (changes requested):
Test

maaktweluit:
removing the approval keeps it on the API:
https://api.github.com/repos/golemfactory/golem/pulls/1300/reviews

We should make the script smarter to only look at the last entry per user/reviewer.

badb (approved):

Wolf480pl (dismissed):
I want these slow tests split ASAP so accept.

badb:
@maaktweluit Hm, it seems now that slow tests are always skipped, doesn't matter how many approvals they have. Can you check this out?

maaktweluit:
@badb it looks like the old approval remains in the list, so the un-approval does not override it.
https://api.github.com/repos/golemfactory/golem/pulls/1300/reviews

Will have a look if i can remove duplicates ( per user ) before checking the approval count

maaktweluit:
@badb resolved in latest commit, it only takes the last "vote" of each user now

love python!

badb:
Ok, but still slow (dummy) tests are skipped, even though there are 2 approvals.

maaktweluit:
@badb they are skipped for a different reason:
```
SKIP [1] tests/golem/task/dummy/test_dummy_task.py:9: It doesn't work, using new dummy task instead
```

Want me to remove these skips? not sure what the impact will be, the message states it does not work.

badb (changes requested):
Test

badb (approved):

PR #1307, opened 2017-08-25
make setup.py pyinstaller output tarball path configurable

pstiasny:
@banasiakadam60  @mfranciszkiewicz

### DIFF
--- a/setup_util/setup_commons.py
+++ b/setup_util/setup_commons.py
@@ -39,16 +39,17 @@ class PyTest(test):
 
 class PyInstaller(Command):
     description = "run pyinstaller and packaging actions"
-    user_options = []
+    user_options = [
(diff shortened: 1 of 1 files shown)


setup_util/setup_commons.py
```diff
@@ -39,16 +39,17 @@ def run_tests(self):
...
+    ]
 
     def initialize_options(self):
-        pass
+        self.package_path = None
 
     def finalize_options(self):
         pass
 
-    @classmethod
```
banasiakadam60, line 50: @mfranciszkiewicz Can it be removed? Or it will break something?

banasiakadam60 (approved):
If @mfranciszkiewicz will approve removing `@classmethod` it can be merged

gborowik (approved):

Issue #823, opened 2017-02-10
Change workflow: task testing


badb:
Current workflow:

1. Requestor defines all task parameters
2. Task is tested. Only task that passes the test can be added to Golem Network. 
3. If user change any parameters after step 2, the test must be run again. 

New workflow idea:
1. Requestor defines only basic task parameters (eg. files, scene)
2. Task is tested. Only task that passes the test can be added to Golem Network. 
3. Additional parameters values are guessed based on test results, eg. timeouts, output name (?), number of subtasks, etc. 
4. User can change default values of additional parameters. It's not necessary to run again the test. 
5. If user change one of the basic parameters, eg. files, than it's required to re-run the test.

maaktweluit:
Started working on this from the CLI side.
Added a method "create" and want to add "test" so this flow change can be build on top.

For now all code is in the `client.py`, but want to move it to a proper re-usable location.
Was thinking of the following locations ( pseudo-code ), please advise:
- `TaskDef().save( file_name )`
- `TaskFactory().save( file_name)`

The new `TaskFactory` could be responsible for saving, loading and testing tasks. This way all logic can be abstracted to `BaseTask` level and grouped in `apps/core`.

Made a [first draft](https://github.com/golemfactory/golem/compare/develop...maaktweluit:cli_create_task?expand=1) but was not able to test it yet, will continue tonight or tomorrow.

maaktweluit:
Updated the branch with `tasks test`.

Next step is to go fishing for test result data, curious what I will find.

maaktweluit:
Made a final update before my holiday, will be AFK for a while, please continue this issue without me if needed.

Scooped up some data from the test and parsed it for some simple validations.
[This is a working POC](https://github.com/golemfactory/golem/compare/develop...maaktweluit:cli_create_task?expand=1) where `output_file` and `frames` are validated and defaulted. ( ubuntu desktop 16.04 )

- Where should all this logic go to be compatible with the rpc / gui?

Most steps I could think of next are marked as `TODO` in the code.

Please note: All this POC does is run `golemcli tasks test` on save files from the old GUI. To test 'empty' values they are set to `""` in the `test.gt` file manually.

Will continue when i'm back from holiday!

PR #1313, opened 2017-08-28
[WIP] Cli test task and update task save file with results

maaktweluit:
Working on #823 
This was assisted by #1352 
This issue should fix testing a task save file and storing the results / errors.

This PR contains a set of useful functions waiting to be put on the proper spot in the project:
- [ ] load file from default saves location
- [ ] check if file is writable
- [ ] test and default blender task resolution
- [ ] test and default blender task frames
- [ ] test and default (blender) task output name
- [ ] save tested task_def to file ( needs better UX )

Probably some of these are already there, please advise where to put the rest, if any.

### DIFF
--- a/apps/blender/resources/blenderdefinitionupdater.py
+++ b/apps/blender/resources/blenderdefinitionupdater.py
@@ -1,0 +1,135 @@
+import os
+import copy
+
+
+def propose_fixed_definition(task_def, data):
+    # print("Hello")
+    # print(task_def)
+    # print(data)
+
+    data["proposed_def"] = _get_proposed_definition(task_def, data)
+
+
+def _get_proposed_definition(task_def, data):
+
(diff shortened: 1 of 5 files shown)


marmistrz:
Can we close it due to #1331?

maaktweluit:
@marmistrz we can remove the "create" part from this issue and only keep the "test and save default values to task file".
Will update the title to be more clear.

When #1331 is finished i'll rebase and continue on this PR

marmistrz:
Anyway, this commit is a no-go even if you remove the *create* part due to usage of jsonpickle, which is possibly dangerous and was removed #1254.

maaktweluit:
closing this PR because it is stale for a year.

We can use the branch as reference when we pick up #823 again

PR #1325, opened 2017-08-29
Skip peer discovery in HyperG

mfranciszkiewicz:


golem/resource/hyperdrive/resourcesmanager.py
```diff
@@ -85,3 +90,56 @@ def _wrap_resource(self, resource, task_id=None):
...
+            self._peers[node.key] = peer
+
+
+class HyperdrivePeerManager(HyperDriveMetadataManager):
+
+    def __init__(self, daemon_address):
+        super(HyperdrivePeerManager, self).__init__(daemon_address)
+        self._tasks = dict()
+
+    def add(self, task_id, key_id):
```
Wolf480pl, line 124: Is this method user anywhere?
Is there any other place where you are adding stuff to `self._tasks`?

golem/task/taskserver.py
```diff
@@ -27,7 +27,70 @@
 tmp_cycler = itertools.cycle(list(range(550)))
 
 
-class TaskServer(PendingConnectionsServer):
+class TaskResourcesMixin(object):
+
+    def add_resource_peer(self, node_name, addr, port, key_id, node_info):
```
Wolf480pl, line 34: Is this called only from dead code?

golem/task/tasksession.py
```diff
@@ -581,7 +583,7 @@ def _react_to_task_result_hash(self, msg):
         secret = msg.secret
         multihash = msg.multihash
         subtask_id = msg.subtask_id
-        client_options = msg.options
```
Wolf480pl: We're making the protocol incompatible with 0.8, right? So can we remove `options` from this message, if it's not used anywhere?

Wolf480pl (changes requested):
To me it looks like the per-task_id list of peers will always be empty, because `HyperdrivePeerManager.add(...)` appears to be never called.

golem/task/taskserver.py
mfranciszkiewicz: It's been moved from TaskServer, I'm not removing dead code here

golem/resource/hyperdrive/resourcesmanager.py
mfranciszkiewicz: Not *yet*

Wolf480pl: Oh, the requestor adds its own address in `get_for_task` and sends the list to the provider. So it always pulls from requestor. Ok.

golem/task/taskserver.py
Wolf480pl: mmk

Wolf480pl (approved):

Issue #231, opened 2016-03-29
Keep track of the reason the task isn't supported.


badb:
Golem Taskeeper keeps information about all of tasks in the network. Some of them are supported and some not. In the future, user may look at the list of all tasks available and see what should she do to support that type of a task, ie. install new software, lower the price or update application.

Wolf480pl:
Should I expose this information by RPC, or is a public method on taskkeeper enough?

badb:
We should also write this information in log (INFO severity), RPC call may also be a nice addition for debugging process.

maaktweluit:
@Wolf480pl the related PR is merged, can this issue be closed?

Wolf480pl:
@maaktweluit dunno. AFAIK this information is passed around but still not exposed anywhere. Ask @badb whether actually using the information somewhere is part of the issue.

maaktweluit:
checked and can indeed be closed

PR #1328, opened 2017-08-30
Keep track of the reason the task isn't supported.

Wolf480pl:
This implements #231


golem/docker/environment.py
```diff
@@ -15,11 +15,17 @@ def __init__(self, docker_images):
...
 
-    def check_docker_images(self):
-        return any(img.is_available() for img in self.docker_images)
+    def check_docker_images(self) -> SupportStatus:
+        if any(img.is_available() for img in self.docker_images):
+            return SupportStatus.ok()
 
-    def supported(self):
-        return self.check_docker_images() and Environment.supported(self)
+        return SupportStatus.err({'environment_unsupported': {
```
mfranciszkiewicz: Could you create an Enum for these error keys (`environment_unsupported`, `environment_not_accepting_tasks`, etc.)?

golem/environments/environment.py
```diff
@@ -2,6 +2,35 @@
 from os import path
 
 
+class SupportStatus(object):
+    def __init__(self, desc, ok=False) -> None:
```
mfranciszkiewicz: It seems that `ok` is always set manually and we're not taking advantage of the default value. Plus, `desc` could be initialized with a default value, e.g. 
```
def __init__(self, ok, desc=None) -> None:
    self.desc = desc or dict()
    ...
```

golem/environments/environmentsmanager.py
```diff
@@ -1,13 +1,13 @@
...
-
+from .environment import SupportStatus
 logger = logging.getLogger(__name__)
 
 
 class EnvironmentsManager(object):
     """ Manage known environments. Allow user to choose accepted environment, keep track of supported environments """
     def __init__(self):
-        self.supported_environments = set()
+        self.support_status = {}
```
mfranciszkiewicz: Since this is not a single instance of `SupportStatus`, please rename

golem/environments/environment.py
```diff
@@ -2,6 +2,35 @@
...
 
+class SupportStatus(object):
+    def __init__(self, desc, ok=False) -> None:
+        self.desc = desc
+        self._ok = ok
+
+    def is_ok(self) -> bool:
+        return self._ok
+
+    def __bool__(self) -> bool:
```
mfranciszkiewicz, line 14: If I'm not mistaken, this cast is used exactly once. Maybe use `is_ok` instead?

golem/docker/environment.py
```diff
@@ -15,11 +15,17 @@ def __init__(self, docker_images):
...
+            'env_id': self.get_id(),
+            'docker_images_missing_any': self.docker_images,
+        }})
+
+    def supported(self) -> SupportStatus:
```
mfranciszkiewicz: The name strongly suggests a `bool` return value

golem/task/taskkeeper.py
```diff
@@ -185,18 +188,18 @@ def __init__(
         self.removed_task_timeout = remove_task_timeout
         self.environments_manager = environments_manager
 
-    def is_supported(self, th_dict_repr):
+    def is_supported(self, th_dict_repr) -> SupportStatus:
```
mfranciszkiewicz: Judging by the name, I'd expect a `bool` return value

golem/task/taskkeeper.py
```diff
@@ -185,18 +188,18 @@ def __init__(
...
+        :return SupportStatus: ok() if this node may compute a task
         """
         supported = self.check_environment(th_dict_repr)
-        supported = supported and self.check_price(th_dict_repr)
-        return supported and self.check_version(th_dict_repr)
+        supported = supported.join(self.check_price(th_dict_repr))
+        return supported.join(self.check_version(th_dict_repr))
 
     @staticmethod
     def is_correct(th_dict_repr):
```
mfranciszkiewicz: This could also take advantage of `SupportStatus`-like class

Wolf480pl:
I have no idea why the test is failing - it worked ok locally.

golem/environments/environment.py
Wolf480pl: This was for backwards compat, in case I miss some places that still expect bool.
Also, I think it's kinda convenient and Pythonish, similarly to how you check for empty collections.

golem/docker/environment.py
Wolf480pl: I made them not-enum on purpose, so that apps can easily define their own reasons.

mfranciszkiewicz: And they will still be able to. But not in core.

tests/golem/task/test_taskkeeper.py
```diff
@@ -263,7 +270,7 @@ def test_init(self):
         self.assertIsInstance(csi, CompSubtaskInfo)
 
 
-class TestCompTaskKeeper(LogTestCase, PEP8MixIn, TempDirFixture):
+class TestCompTaskKeeper(LogTestCase, TempDirFixture):
     PEP8_FILES = [
```
badb: If you are removing PEP8MixIn, then remove also PEP8_FILES. I would suggest however to keep it for now .

golem/docker/environment.py
```diff
@@ -15,11 +15,17 @@ def __init__(self, docker_images):
...
+        return self.check_docker_images().join(Environment.supported(self))
```
badb: It fails on Travis and AppVeyor because they use `ci_patch` decorator, that set check_docker_images return value to bool. If you change this patch to your SUpportStatus it should be fine.

golem/docker/environment.py
Wolf480pl: What should I use as dict keys: enum members, or their string values?

golem/task/taskkeeper.py
Wolf480pl: Not sure if it's in scope of this issue

tests/golem/task/test_taskkeeper.py
Wolf480pl: Ok, will keep it for now.

badb (approved):

mfranciszkiewicz (approved):

Issue #324, opened 2016-05-04
Add verification for task timeout


badb:
~~Computer should start to compute task only if current time + subtask_ttl < task_timeout~~
Please make sure that task computer sets minimum from task deadline and subtask deadline as it's timeout instead.

Wolf480pl:
You mean `task_timeout` argument to `__compute_task()`, which is calculated from subtask deadline, or do you mean task deadline from task header?

maaktweluit:
@Wolf480pl is this solved fully through #1350 ?

Wolf480pl:
@maaktweluit AFAIK yes.

maaktweluit:
Checked and can indeed be closed

PR #1336, opened 2017-08-31
Check subtask deadline

Wolf480pl:
Make TaskComputer verify deadline of subtasks by checkig if it's not later than the task's deadline.

This implements #324

### DIFF
--- a/golem/task/taskcomputer.py
+++ b/golem/task/taskcomputer.py
@@ -117,10 +117,9 @@ class TaskComputer(object):
             subtask_id = self.task_to_subtask_mapping[task_id]
             if subtask_id in self.assigned_subtasks:
                 subtask = self.assigned_subtasks[subtask_id]
-                timeout = deadline_to_timeout(subtask.deadline)
                 self.__compute_task(subtask_id, subtask.docker_images,
                                     subtask.src_code, subtask.extra_data,
-                                    subtask.short_description, timeout)
+                                    subtask.short_description, subtask.deadline)
                 self.waiting_for_task = None
                 return True
             else:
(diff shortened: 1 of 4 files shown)


golem/task/taskcomputer.py
```diff
@@ -227,36 +255,44 @@ def run(self):
...
         self.compute_tasks = config_desc.accept_tasks
         self.change_docker_config(config_desc, run_benchmarks, in_background)
-    
+
     def _validate_task_state(self, task_state):
         td = task_state.definition
         if not os.path.exists(td.main_program_file):
-            logger.error("Main program file does not exist: {}".format(td.main_program_file))
+            logger.error("Main program file does not exist: {}"
+                         .format(td.main_program_file))
```
jiivan: While reformatting please also use `%` substisution.

golem/task/taskcomputer.py
```diff
@@ -384,25 +422,39 @@ def __request_task(self):
...
         task_id = self.assigned_subtasks[subtask_id].task_id
         working_dir = self.assigned_subtasks[subtask_id].working_directory
         unique_str = str(uuid.uuid4())
 
+        task_header = self.task_server.task_keeper.task_headers[task_id]
+        task_deadline = task_header.deadline
+
+        if subtask_deadline > task_deadline:
+            self.subtask_failure(task_id,
+                                 "Subtask deadline is after task deadline: "
```
jiivan: Maybe something like: `"Subtask deadline reached: "?

tests/golem/docker/test_docker_luxrender_task.py
```diff
@@ -210,9 +215,10 @@ def test_luxrender_real_task_png(self):
         task.res_x = 200
         task.haltspp = 20
         # 1) to make it deterministic,
-        # 2) depending on the kernel, small cropwindow can generate darker img, this is a know issue in lux:
+        # 2) depending on the kernel, small cropwindow can generate darker img,
+        #    this is a know issue in lux:
```
jiivan: know**n** issue

jiivan (approved):
Besides my little remarks LGTM :+1:

golem/task/taskcomputer.py
Wolf480pl: No, because it's not reached yet.

Both deadlines are (most likely) in the future.
The thing is, a subtask, which is part of a bigger task, should be finished before the big task is finished.
If the deadlines say eg. "yeah, you still have 10 hours to do this subtask" but the parent task's deadline is in 2 hours, then something is wrong.

golem/task/taskcomputer.py
Wolf480pl: Really? I thought `%` was legacy/deprecated.

jiivan: It is but it's the only substitution that works with python logging library.
See: https://gist.github.com/jiivan/eedb505574ccb2503af48fa2e3043036

golem/task/taskcomputer.py
jiivan: Oh, now I understand. So maybe: "Subtask deadline shouldn't be after..." So it would be more obvious?

Wolf480pl:
No idea why circle failed :/

badb:
This PR was absolutely fine, there was done what should be done. However after discussion with Grzegorz we decided that it shouldn't be merged now. This will be useful when we add an estimated time that connected with performance may given an estimation if provider is able to compute something or not. 
Without that it may slow down computations for requestor. Please check the issue, for new task description. 
Check the issue for new task description

To sum up: this branch should be closed but not removed - kept for reference with further development process.

PR #1341, opened 2017-09-06
Fix linter warnings in localcomputer.py

marmistrz:
A little cleanup before I get down to localcomputer

### DIFF
--- a/golem/task/localcomputer.py
+++ b/golem/task/localcomputer.py
@@ -3,6 +3,7 @@ import os
 import shutil
 from threading import Lock
 import time
+from typing import Callable
 
 from golem.core.common import to_unicode
 from golem.docker.task_thread import DockerTaskThread
@@ -13,15 +14,27 @@ from golem.task.taskbase import Task, resource_types
 logger = logging.getLogger("golem.task")
 
 
-class LocalComputer(object):
+class LocalComputer:
     DEFAULT_WARNING = "Computation failed"
     DEFAULT_SUCCESS = "Task computation success!"
 
-    def __init__(self, task, root_path, success_callback, error_callback, get_compute_task_def, check_mem=False,
-                 comp_failed_warning=DEFAULT_WARNING, comp_success_message=DEFAULT_SUCCESS, use_task_resources=True,
-                 additional_resources=None):
+    def __init__(
+        self,
+        task: Task,
+        root_path: str,
+        success_callback: Callable,
+        error_callback: Callable,
+        get_compute_task_def: Callable,
+        check_mem: bool = False,
+        comp_failed_warning: str = DEFAULT_WARNING,
+        comp_success_message: str = DEFAULT_SUCCESS,
+        use_task_resources: bool = True,
+        additional_resources=None
+    ) -> None:
         if not isinstance(task, Task):
-            raise TypeError("Incorrect task type: {}. Should be: Task".format(type(task)))
+            raise TypeError(
+                "Incorrect task type: {}. Should be: Task".format(type(task))
+            )
         self.task = task
         self.res_path = None
         self.tmp_dir = None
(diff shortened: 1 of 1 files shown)


golem/task/localcomputer.py
```diff
@@ -13,15 +14,27 @@
...
+        root_path: str,
+        success_callback: Callable,
+        error_callback: Callable,
+        get_compute_task_def: Callable,
+        check_mem: bool = False,
+        comp_failed_warning: str = DEFAULT_WARNING,
+        comp_success_message: str = DEFAULT_SUCCESS,
+        use_task_resources: bool = True,
+        additional_resources=None
+    ) -> None:
```
jiivan, line 33: Let me start with a citation from [PEP484](https://www.python.org/dev/peps/pep-0484/#non-goals):

> Python will remain a dynamically typed language, and the authors have no desire to ever make type hints mandatory, even by convention.

With this in mind I think that annotating/typing arguments and methods should only be used when there is confusion or lack of clarity in code, to help developers and linters understand the code. Defining return type of `__init__` function is useless, because the `None` is obvious. Also annotating _every_ argument imho doesn't serve the purpose of readability. For example isn't it obvious that `*_callback` args are callables? Also `comp_failed_warning` and `comp_success_message` use strings defined one line above... The same probably goes for bool args with default boolean values. Does annotating argument with default value `False` add anything to readability?

golem/task/localcomputer.py
```diff
@@ -58,7 +72,7 @@ def run(self):
             logger.warning("{}: {}".format(self.comp_failed_warning, exc))
             self.error_callback(to_unicode(exc))
 
-    def end_comp(self):
+    def end_comp(self) -> bool:
```
jiivan, line 75: I think this a _good_ example where annotating actually adds clarity.

golem/task/localcomputer.py
```diff
@@ -36,18 +49,19 @@ def __init__(self, task, root_path, success_callback, error_callback, get_comput
...
+        self.test_task_res_path = ""
         if additional_resources is None:
             additional_resources = []
         self.additional_resources = additional_resources
 
         self.start_time = None
         self.end_time = None
 
-    def run(self):
+    def run(self) -> None:
```
jiivan, line 60: Does the name `run()` suggest any other return type than `None`? IMO it should be annotated only when it returns something other than that.

golem/task/localcomputer.py
marmistrz: Let me quote the very same PEP-484:

> (Note that the return type of `__init__` ought to be annotated with `-> None`. The reason for this is subtle. If `__init__` assumed a return annotation of `-> None`, would that mean that an argument-less, un-annotated `__init__` method should still be type-checked? Rather than leaving this ambiguous or introducing an exception to the exception, we simply say that `__init__` ought to have a return annotation; the default behavior is thus the same as for other methods.)

Now, recall that tools such as `mypy` check only what appears in a so-called *typed context*. By default unannotated functions definitions are considered untyped context, which means this is the only way to enable constructor type-checks.

This is very far from being obvious. It may be `bool`. But you may expect `Optional[bool]` instead. Or a `Union[bool, int, NoneType]` or `Union[bool, List[int]]`, since both `int` and `list` can appear in a logical context. To quote PEP-20, named *The Zen of Python*:

> Explicit is better than implicit.

While the fact that a `task` is an instance of `Task` is obvious to you, it is far from being obvious to an automatic type-checker. In a 70k LOC project it's really easy to make a human mistake, mix up the order of arguments or their type (use an integer where you ought to use a class constructed from an integer).

But you reminded me that I ought to annotate the precise type of the `get_compute_task_def`, which is `Callable[[], ComputeTaskDef]`. I forgot about it.

marmistrz: Oh, @jiivan, that's exactly the reason why annotations even in obvious places can be really, really useful!

https://github.com/golemfactory/golem/pull/1351/files

If `self.subs` (or maybe `event_names` too) were annotated, such an error would not stay undetected.

badb:
Merge / rebase needed.

badb:
Continued in https://github.com/golemfactory/golem/pull/1670

Issue #1338, opened 2017-09-05
Tests for CLI with patching get segfault 


badb:
It was observed on Linux with Python 3.5.3. Tests from `tests/golem/interface/test_cli.py` gets segfault if they are called together with the rest of the suite. If only tests from this one file are called then everything works fine. 
Ask @banasiakadam60 for further details.

Wolf480pl:
Each of these is enough to trigger the segfault:
`python -m pytest tests/apps/core/gui/controller/test_newtaskdialogcustomizer.py tests/golem/interface/test_cli.py`
`python -m pytest -v tests/apps/core/gui/controller/test_newtaskdialogcustomizer.py tests/golem/interface/test_cli.py -k 'test_execute_interactive or test_customizer'`

Wolf480pl:
Adding `import readline` to `test_newtaskdialogcustomizer.py` right after `customizer._show_add_resource_dialog()` causes it to segfault without `test_cli.py`, with just one test: `test_customizer`.

Wolf480pl:
Happens when you `QDialog.open()` and then `import readline` without importing readline before. Quick workaround: add `import readline` at the beginning of `test_newtaskdialogcustomizer.py`.

Wolf480pl:
Minimal script reproducing the segfault:
```
from PyQt5.QtWidgets import QDialog, QMainWindow, QApplication

app = QApplication([])
window = QMainWindow()
QDialog(window).open()
import readline
```

PR #1346, opened 2017-09-07
Workaround for SEGFAULT during tests

Wolf480pl:
This is a workaround for issue #1338

Apparently importing readline after opening a QDialog causes a segfault,
but not if readline was already imported before.

This should prevent this from happening during tests by importing readline at the beginning of `test_newtaskdialogcustomizer.py`, before any QDialogs are created.


banasiakadam60 (approved):
LGTM

banasiakadam60:
@mfranciszkiewicz Do you agree to merge this?

mfranciszkiewicz (approved):

Issue #665, opened 2016-11-15
Limit the number of saved task headers


badb:
Especially don't save too many tasks from the same task requestor. The limit can be arbitrarily set or dynamically adjusted to the global number of tasks in the network. 

Which tasks should have priorities? 

- The oldest
- The ones with closest timeout

Wolf480pl:
Do we want a limit per requestor public key? What if someone has many public keys?

Wolf480pl:
As for priorities, I think it requires some consideration from security point of view.
I asked @paulperegud and he had some ideas, but he's not sure for which protocol version this is for.
@badb is it for gossip or broadcast?

paulperegud:
This one is for broadcast. Please limit yourself to limiting the amount of tasks from single author.

paulperegud:
This one is for broadcast. Please limit yourself to limiting the amount of tasks from single author.

Wolf480pl:
@paulperegud which ones should I drop? oldest? newest? cheapest? random?

badb:
I vote for removing the newest. Maybe @paulperegud has different idea, but this one should work well. If there are no more subtasks in some old tasks it's going to be removed from the list anyway, so there's a place for the new task to be added.

Wolf480pl:
A task header doesn't even have a creation time, so I guess it'd be by receive time.

Wolf480pl:
If we drop a task because we reached the limit, do we want to trigger the post-removal cooldown (the thing that prevents the task from being re-added too quickly) ?

PR #1347, opened 2017-09-07
Limit the number of task headers per requestor in taskkeeper

Wolf480pl:
This implements (part of) #665

### DIFF
--- a/tests/golem/task/test_taskkeeper.py
+++ b/tests/golem/task/test_taskkeeper.py
@@ -237,10 +237,54 @@ class TestTaskHeaderKeeper(LogTestCase):
         assert not correct
         assert err == "Subtask timeout is less than 0"
 
+    def test_task_limit(self):
+        tk = TaskHeaderKeeper(EnvironmentsManager(), 10)
+        limit = tk.max_tasks_per_requestor
 
-def get_dict_task_header():
+        thd = get_dict_task_header("ta0")
+        thd["deadline"] = timeout_to_deadline(0.1)
+        tk.add_task_header(thd)
+
+        for i in range(1, limit):
+            thd = get_dict_task_header("ta%d" % i)
+            tk.add_task_header(thd)
+
+        for i in range(limit):
+            self.assertIn("ta%d" % i, tk.task_headers)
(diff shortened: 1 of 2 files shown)


tests/golem/task/test_taskkeeper.py
```diff
@@ -231,10 +231,58 @@ def test_is_correct(self):
         assert not correct
         assert err == "Subtask timeout is less than 0"
 
+    def test_task_limit(self):
```
marmistrz: If you add `-> None`, you get mypy type-checks for free.

tests/golem/task/test_taskkeeper.py
```diff
@@ -251,9 +299,11 @@ def get_dict_task_header():
 
 def get_task_header():
```
marmistrz: How about adding `-> TaskHeader` to get free typechecks?

tests/golem/task/test_taskkeeper.py
Wolf480pl: Even in tests?

tests/golem/task/test_taskkeeper.py
Wolf480pl: Even in tests?

marmistrz: Oh, I didn't notice these are tests. It's up to you then, but I prefer to have free checks in tests too. A common reason for failing tests is buggy tests :)

Wolf480pl: At least tests don't fail silently.

gborowik:
@Wolf480pl Could you please refactor the code by providing functions with as little code as possible as develop a separate test for the each of them? Such a process makes it human readable and usable in future. Thanks.

Wolf480pl:
@gborowik is it better now?

gborowik:
For sure. Please update the tests too.

Wolf480pl:
@gborowik I added some tests

gborowik (approved):

mfranciszkiewicz (approved):

PR #1351, opened 2017-09-11
Fixed typo

maaktweluit:


marmistrz (commented):
Nice one! If you know the type of `self.subs` (which is certainly of form `Dict[T, U]`, can you annotate it?

A more meaningful commit message would be useful too.

PR #1331, opened 2017-08-31
Quick and dirty creating tasks from CLI

marmistrz:
Linter report:
```
Comparing 78597ff2...31591663
pylint main...      OK
pylint tests...     OK
pycodestyle...      OK
mypy...             OK
```

Need to be merged first:
- [x] #1254 should be certainly merged first.
- [x] #1329 - I'm using the `to_dict` method

PR #1352, opened 2017-09-11
Quick and dirty creating tasks from CLI (with task_id changed)

badb:


badb:
Same as https://github.com/golemfactory/golem/pull/1331, but with one small additional change.

PR #1359, opened 2017-09-12
Warn that the UUID is discarded while loading a preset.

marmistrz:


golem/interface/client/tasks.py
```diff
@@ -189,6 +189,8 @@ def __progress_str(progress):
     def create_from_json(self, jsondata: str) -> Any:
         dictionary = json.loads(jsondata)
         # FIXME CHANGE TASKI ID
+        if 'id' in dictionary:
+            print("Warning: discarding the UUID from the preset")
```
badb, line 193: Why do you think this requires a warning? There should be just an info.

marmistrz: Maybe the user has changed the UUID on his own...? I can change it to `Note: discarding the UUID from the preset`

PR #1360, opened 2017-09-12
Use python3 by default in setup_commons.py

marmistrz:


mfranciszkiewicz:
Why not `sys.executable`?

marmistrz:
Mainly because I didn't know about it :P Nice catch, fixed that.

PR #1366, opened 2017-09-13
A little of refactor in task manager and task keeper

marmistrz:

PR #1368, opened 2017-09-13
Taskkeeper nopersist

marmistrz:
Depends on:
* [ ] #1366


marmistrz:
Oh, I missed a thing, that doesn't make any sense.

PR #1379, opened 2017-09-14
Remove from preview for tasks in frames #896

### LINKED ISSUE #896: Remove from preview for tasks in frames

jiivan:
fixes: #896


apps/rendering/task/framerenderingtask.py
```diff
@@ -175,6 +175,16 @@ def to_dictionary(self):
     # Specific task methods #
     #########################
 
+    @CoreTask.handle_key_error
+    def _remove_from_preview(self, subtask_id):
```
marmistrz, line 179: It may be a good idea to annotate the type of `subtask_id`. It's very far from being obvious.

gborowik:
@jiivan Why circleci and travis did fail?

jiivan:
@gborowik CircleCI passed after merging with develop. Travis failed because of premature dependency removal in `requirements-test.txt`. Fixed.

gborowik (approved):

apps/rendering/task/framerenderingtask.py
```diff
@@ -175,6 +175,16 @@ def to_dictionary(self):
     # Specific task methods #
     #########################
 
+    @CoreTask.handle_key_error
+    def _remove_from_preview(self, subtask_id):
+        if not isinstance(self.preview_file_path, list):
```
Elfoniok: Does isinstance isn't inherently wrong? what if someone change this to set? for example.

golem/interface/cli.py
```diff
@@ -120,11 +120,17 @@ def process(self, args):
...
+
+        if not hasattr(clean, 'callback'):
+            return "No callback found", output
+
         try:
 
-            namespace = self.parser.parse_args(args)
-            clean = self._clean_namespace(namespace)
-            formatter = self.get_formatter(clean)
+            print(namespace)
```
Elfoniok: What for print statments?

golem/interface/cli.py
```diff
@@ -120,11 +120,17 @@ def process(self, args):
         output = sys.stderr
         started = time.time()
 
+        namespace = self.parser.parse_args(args)
+        clean = self._clean_namespace(namespace)
+        formatter = self.get_formatter(clean)
```
Elfoniok: Why this is taken out of try block?

golem/interface/cli.py
```diff
@@ -120,11 +120,17 @@ def process(self, args):
...
+
+        if not hasattr(clean, 'callback'):
```
Elfoniok: I don't thin it is necessary, is it? Some previous bug caused that some function threw exception here but it will be fixed after transferring some fixes from devp2p branch.

golem/interface/cli.py
```diff
@@ -144,6 +150,8 @@ def process(self, args):
             result = ExecutionException("Command timed out", " ".join(args), started)
 
         except Exception as exc:
+            import traceback
+            traceback.print_exc()
```
Elfoniok: why print it here?

Elfoniok (changes requested):

apps/rendering/task/framerenderingtask.py
jiivan: In this case it has to be ordered sequence (ordering matters). Added check for `tuple`.

jiivan:
@Radagast-red `golem/interface/cli.py` was added by accident. Thanks for pointing this out.

Elfoniok (approved):

PR #1374, opened 2017-09-14
Persistence types

marmistrz:
Depends on #1366.

PR #1382, opened 2017-09-14
Fix conflicting saves for persistent tasks

marmistrz:
It seems to work when enabled.

Depends on:
* [ ] #1366 
* [ ] #1374 

It'd be nice to add some extra tests for dumping/restoring tasks, but it's out of the scope of this PR.
I can do it once it's merged.


jiivan:
Closed in favor of #1581

PR #1388, opened 2017-09-15
Fix installer

banasiakadam60:
Since we're packing binaries also on Windows, it's path has change from `%REPO_PATH%/dist` to `%REPO_PATH/dist/golem-%version%`, so it had to be changed in `install_script`. Without that `golemapp` and `golemcli` wasn't added to the `path`.


jiivan (approved):
LGTM

PR #1391, opened 2017-09-15
Fixing duplication of task temp directory creation

inexxt:
Task temp dir was created in two places: in `initialize()` in `Task` and in `add_new_task` in `TaskManager`.

This PR removes the `TaskManager` one.


apps/core/task/coretask.py
```diff
@@ -186,6 +186,7 @@ def is_docker_task(self):
                and len(self.header.docker_images) > 0
 
     def initialize(self, dir_manager):
+        dir_manager.clear_temporary(self.header.task_id)
         self.tmp_dir = dir_manager.get_task_temporary_dir(self.header.task_id, create=True)
```
marmistrz: Can you fix the line-too-long warning here, since you're touching this code anyway?
And check if `dir_manager` is of class `DirManager`, if yes - annotate the types. This should be trivial using `enforce`.

golem/task/taskmanager.py
```diff
@@ -137,10 +137,6 @@ def add_new_task(self, task):
         task.header.task_owner = self.node
         task.header.signature = self.sign_task_header(task.header)
 
-        self.dir_manager.clear_temporary(task.header.task_id)
-        self.dir_manager.get_task_temporary_dir(task.header.task_id,
-                                                create=True)
-
```
marmistrz: Since you're touchting `add_new_task`, can you fix the `pycodestyle` warning at line 151?

marmistrz (changes requested):

golem/task/taskmanager.py
inexxt: Done

marmistrz:
I'd squash those two style commits into one, so that there is a total of 2 commits.

maaktweluit (approved):
LGTM

badb (approved):

PR #1406, opened 2017-09-21
Remove old mock dependency

jiivan:
Since python3 `mock` is included in the standard library under `unittest.mock` and is no longer maintained as a separate package.

### DIFF
--- a/tests/apps/lux/task/test_luxverificator.py
+++ b/tests/apps/lux/task/test_luxverificator.py
@@ -1,6 +1,5 @@
 import os
 import unittest.mock as mock
-import unittest.mock as mock
 
 from golem.testutils import PEP8MixIn, TempDirFixture
 from golem.tools.assertlogs import LogTestCase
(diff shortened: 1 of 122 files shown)


tests/apps/lux/task/test_luxverificator.py
```diff
@@ -1,5 +1,6 @@
 import os
-from mock import patch, Mock
+import unittest.mock as mock
+import unittest.mock as mock
```
Elfoniok: ? Double declaration?

Elfoniok (commented):
As a general note, wouldn't be easier to import Mock class? not mock namespace?

tests/apps/lux/task/test_luxverificator.py
jiivan: Fixed

jiivan:
I like to avoid namespace pollution by importing modules instead of single classes/functions.

Elfoniok (approved):

PR #1410, opened 2017-09-21
Add a possibility to export keys

marmistrz:
Depends on:
* [ ] #1395 

Closes #741.

PR #1422, opened 2017-09-26
Mypy changes

chfast:
This is cherry-pick from #1410.

### DIFF
--- a/golem/rpc/mapping/aliases.py
+++ b/golem/rpc/mapping/aliases.py
@@ -1,10 +1,10 @@
-class Golem(object):
+class Golem:
     status                  = 'golem.status'
 
     evt_golem_status        = 'evt.golem.status'
 
 
-class Environment(object):
+class Environment:
 
     opts                    = 'env.opts'
     opts_update             = 'env.opts.update'
@@ -28,14 +28,14 @@ class Environment(object):
     evt_opts_changed        = 'evt.env.opts'
 
 
-class Crypto(object):
+class Crypto:
 
     key_id                  = 'crypto.keys.id'
     pub_key                 = 'crypto.keys.pub'
     difficulty              = 'crypto.difficulty'
 
 
-class Network(object):
+class Network:
 
     ident                   = 'net.ident'
     ident_key               = 'net.ident.key'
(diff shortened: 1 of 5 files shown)


golem/rpc/mapping/aliases.py
```diff
@@ -1,10 +1,10 @@
-class Golem(object):
```
Wolf480pl: why?

golem/rpc/mapping/aliases.py
```diff
@@ -1,10 +1,10 @@
...
+class Golem:
```
Wolf480pl, line 1: Why? What did that poor `(object)` do wrong, @marmistrz ....

chfast: I don't know. https://github.com/golemfactory/golem/pull/1410#discussion_r140773038
@marmistrz ?

mfranciszkiewicz (approved):

golem/rpc/mapping/aliases.py
marmistrz: TL;DR: It's superfluous in Python 3, it's how the official docs do.

The historical reasons follow. Initially in Python 2 you did `class Foo:`. At some point they decided to change the behavior of classes, and `class Foo:` meant an old-style class, `class Foo(object):` - a new-style class, since they didn't want to break backwards compatibility. 

Now, in Python 3, they discontinued the old-style classes altogether, so `class Foo:` gives a new-style class. Thus, writing `(object)` is superfluous, even the Python 3 documentation uses `class Foo:`.

Wolf480pl: Explicit is better than implicit, so IMO `class Foo(object):` is better.
Also, this is [not quite equivalent][0], so I think it's better not to change it.

[0]: https://stackoverflow.com/questions/1238606/is-it-necessary-or-useful-to-inherit-from-pythons-object-in-python-3-x/45893772#45893772

marmistrz: If you need monkey-patching `builtins.object`, you're probably doing something wrong

Issue #1416, opened 2017-09-25
Elaborated logs about docker issues


jiivan:
Currently `golem.core.processmonitor.ProcessMonitor` only reports `Subprocess XXX exited with code N`. It's problematic because we won't see STDERR which can give clues what is wrong. Also there are standard problems with docker that can be reported together with hints how to solve them:

- [x] Can't download docker image (https certificate issues)
- [x] Can't connect to docker server. Probably docker env not set. May set env manually or restart needed (on MsWin).
- [x] Can't start docker image

Ideally those hints would be propagated as dialog box in frontend.

jiivan:
0a12ad0 
TODO: Propagate logs to frontend

PR #1432, opened 2017-10-02
Elaborated logs about docker issues

jiivan:
fixes: #1416

### DIFF
--- a/golem/core/processmonitor.py
+++ b/golem/core/processmonitor.py
@@ -35,6 +35,10 @@ class ProcessMonitor(Thread):
                     if self.working:
                         self.run_callbacks(process)
                     self._child_processes.pop(i)
+                    if isinstance(process, (psutil.Popen, subprocess.Popen)):
+                        stdout_data, stderr_data = process.communicate()
+                        print("STDOUT: {}\nSTDERR: {}"
+                            .format(stdout_data, stderr_data))
 
             time.sleep(0.5)
 
--- a/golem/docker/manager.py
+++ b/golem/docker/manager.py
@@ -409,6 +409,18 @@ class DockerManager(DockerConfigManager):
             if not retried:
                 self.command('regenerate_certs', self.docker_machine)
                 return self._set_docker_machine_env(retried=True)
+            typical_solution_s = """It seems there is a  problem with your Docker installation.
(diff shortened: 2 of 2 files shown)


maaktweluit:
Relates to #1571

golem/core/processmonitor.py
```diff
@@ -35,6 +35,10 @@ def _start(self):
                     if self.working:
                         self.run_callbacks(process)
                     self._child_processes.pop(i)
+                    if isinstance(process, (psutil.Popen, subprocess.Popen)):
+                        stdout_data, stderr_data = process.communicate()
+                        print("STDOUT: {}\nSTDERR: {}"
```
maaktweluit, line 40: maybe use `logger` over `print`

`logger` has native formatting, works faster when the log level is turned off. Samples in #1571

golem/docker/manager.py
```diff
@@ -409,6 +409,18 @@ def _set_docker_machine_env(self, retried=False):
...
+Ensure that you try the following before reporting an issue:
+
+ 1. The virtualization of Intel VT-x/EPT or AMD-V/RVI is enabled in BIOS
+    or virtual machine settings.
+ 2. The proper environment is set. On windows powershell please run:
+    & "C:\Program Files\Docker Toolbox\docker-machine.exe" env golem | Invoke-Expression
+ 3. virtualbox driver is available:
+    docker-machine.exe create --driver virtualbox golem
+ 4. Restart Windows machine"""
+            logger.error(typical_solution_s)
```
maaktweluit, line 422: Is this problem also on macs? in that case we should also add a similar block there.

I can help test on mac if you have none available.

maaktweluit (changes requested):
Good start! small comments

golem/docker/manager.py
jiivan: Windows instances are most problematic in this issue, but I will be glad if you could test it on mac, or have experienced any similar problems (usually on "fresh" installs).

jiivan:
Closed in favor of #1588

golem/docker/manager.py
maaktweluit: Ok, will test it next week when back from holiday :) This is already a great addition to help the end users.

PR #1437, opened 2017-10-03
Added durations and maxfail to each unit test platform

maaktweluit:
After investigation on all possible `pytest` arguments these where picked as handy low hanging fruit.

Added:
- Durations -> shows the slowest X tests when finished
- Maxfail -> exits test when the Xth test failed, to free up the queue
- Verbose skipping and failing


maaktweluit:
builds only failed on not enough approvals.

does show the extra information nicely!

jiivan (approved):

PR #1441, opened 2017-10-05
Allow batched payments

ggruszczynski:


apps/rendering/resources/ImgVerificator.py
```diff
@@ -119,10 +119,10 @@ def crop_img_relative(self, img, crop_window):
         p.load_from_pil_object(cropped_img, img.get_name())
         return p
 
-    def get_random_crop_window(self, coverage=0.2, window=(0, 1, 0, 1)):
+    def get_random_crop_window(self, coverage=0.33, window=(0, 1, 0, 1)):
```
badb, line 122: Please, try to avoid adding changes that are not connected to the PR topic.

ggruszczynski: try avoid means 'revert' or 'dont do this in the future' ?

badb: Don't do this in the future.

lint.sh
```diff
@@ -34,6 +33,11 @@ GREEN=$(
 ) 2>/dev/null
 RESET=$(tput sgr0) 2>/dev/null
 
+if ! git diff-index --quiet HEAD --; then
```
badb: Put please revert this. Put this in separate PR if you really need this one (but  please let us know why).

badb (approved):
It can be merged after this linter changes are reverted or very well argumented.

lint.sh
ggruszczynski: reverted

PR #1446, opened 2017-10-06
Resource handshake

### LINKED ISSUE #1434: HyperG handshake for golem p2p
Currently we are using two types of networks - twisted-based golem p2p network & hyperg for resource management. They are using different port and different connection mechanism, eg. there's simple reverse connection in golem and sometimes happen that nodes are able to connect via golem network, but are not able to exchange resources. If we combine two protocols and try to send a very small file after the handshake, than the node would be sure that it's only working with nodes that with which it's able to exchage resources. 

On task network, after first hello method, but before exchanging WantToComputeTask method, nodes should exchange hashes of small files and continue cooperation only after both files are succesfully dowloaded.

mfranciszkiewicz:
Resolves #1434


gborowik:
Travis failed:
"/usr/local/Homebrew/Library/Homebrew/brew.rb:12:in `<main>': Homebrew must be run under Ruby 2.3! (RuntimeError)
The command "brew install python3 qt5 openexr freeimage" failed and exited with 1 during .
Your build has been stopped.
/Users/travis/.travis/job_stages: line 166: shell_session_update: command not found"

mfranciszkiewicz:
@gborowik Yes, this issue is affecting all PRs

gborowik (approved):

golem/core/variables.py
```diff
@@ -5,7 +5,7 @@
 #       VARIABLES          #
 ############################
 APP_NAME = "Brass Golem"
-APP_VERSION = "0.8.0"
+APP_VERSION = "0.8.1"
```
jiivan, line 8: If `TASK_PROTOCOL_ID` changes we should bump minor version not patch.

golem/resource/resourcehandshake.py
```diff
@@ -0,0 +1,333 @@
...
+
+        handshake = ResourceHandshake(self.task_server.node.key,
+                                      self._task_request_message)
+        directory = self.resource_manager.storage.get_dir(self.NONCE_TASK)
+
+        try:
+            handshake.start(directory)
+        except Exception as err:
+            import traceback
+            traceback.print_exc()
```
jiivan: Isn't it enough to just use logger in `_handshake_error()`?

golem/task/tasksession.py
```diff
@@ -298,38 +300,6 @@ def _reject_subtask_result(self, subtask_id):
         self.task_server.reject_result(subtask_id, self.result_owner)
         self.send_result_rejected(subtask_id)
 
-    def request_task(
```
jiivan, line 301: Isn't it still referenced in `golem/task/taskserver.py`?

tests/golem/resource/test_resourcehandshake.py
```diff
@@ -0,0 +1,646 @@
+import os
+import types
+import uuid
+
+from mock import Mock, patch, ANY
```
jiivan: Since python3.3 `mock` is part of standard library in `unittest.mock`. We shouldn't use external package.

tests/golem/resource/test_resourcehandshake.py
```diff
@@ -0,0 +1,646 @@
...
+        handshake.start(self.tempdir)
+
+        session._set_handshake(session.key_id, handshake)
+        session._share_handshake_nonce(session.key_id)
+
+        assert session.send.called
+        msg = session.send.call_args[0][0]
+        assert isinstance(msg, MessageResourceHandshakeStart)
+
+    def __create_task_server(self, session):
```
jiivan, line 541: Why using the double underscore here?

tests/golem/resource/test_resourcehandshake.py
```diff
@@ -0,0 +1,646 @@
...
+        self.key_id = str(uuid.uuid4())
+        self.message = dict(
+            node_name='test node',
+            task_id=str(uuid.uuid4()),
+            perf_index=4000,
+            price=5,
+            max_resource_size=10 * 10 ** 8,
+            max_memory_size=10 * 10 ** 8,
+            num_cores=10
+        )
```
jiivan, line 98: maybe `session = MockTaskSession(self.tempdir)` and following initialization could be put here in `setUp()`? It is used in all tests in this TestCase.

jiivan (changes requested):
My main issue is the usage of obsolete external `mock` package.

golem/core/variables.py
mfranciszkiewicz: This updates the version according to the latest tag, for some reason `develop` is still in `0.8.0`. I will abstain from bumping the version above the limit here.

golem/task/tasksession.py
mfranciszkiewicz: It's been moved to the parent (mixin) class.

golem/resource/resourcehandshake.py
mfranciszkiewicz: Probably, in most cases, yes.

tests/golem/resource/test_resourcehandshake.py
mfranciszkiewicz: I don't see why this cannot be a private(ish) function.

tests/golem/resource/test_resourcehandshake.py
mfranciszkiewicz: But sometimes with different parameters (emulating successful resource downloads / uploads).

badb (approved):

jiivan (approved):

Issue #1430, opened 2017-10-02
A mechanism checking whether a new version of Golem is currently available on GitHub


gborowik:
Due to the incompatibility of Golem versions, it would be nice to check and inform the user whether the new version of Golem is available on GitHub and should be downloaded.

badb:
It will be inconvenient to use github API for that (limits!). We should check this somewhere else where it's not causing this problems. Monitor might not be the best idea, so let's add this check in seed nodes, ie. Golem node should compare it's version with seeds version.

PR #1447, opened 2017-10-06
Check version updates

mdtanrikulu:
#1430

### DIFF
--- a/golem/version.py
+++ b/golem/version.py
@@ -1,23 +1,17 @@
 import requests
 import json
 import logging
-from golem.core.variables import APP_VERSION
 from semantic_version import Version
+from golem.core.variables import APP_VERSION
+from golem.utils import OrderedClassMembers
 
 log = logging.getLogger("golem.version")
 
 
-class Importance(object):
-
-    def __init__(self):
-        self.MAJOR = "MAJOR"
-        self.MINOR = "MINOR"
-        self.PATCH = "PATCH"
-
-    def __iter__(self):
-        # about dict order @see https://www.python.org/dev/peps/pep-0520/
-        for count, (_, value) in enumerate(self.__dict__.items(), 0):
-            yield count, value
+class Importance(metaclass=OrderedClassMembers):
+    MAJOR = "MAJOR"
+    MINOR = "MINOR"
+    PATCH = "PATCH"
 
 
 def check_update():
@@ -33,11 +27,10 @@ def check_update():
     latest_release = latest_release['tag_name']
     latest_version = Version(latest_release, partial=True)
     current_version = Version(APP_VERSION, partial=True)
-    importance = dict(Importance())
 
     if latest_version > current_version:
         level = [i for i in range(len(list(latest_version))) if list(latest_version)[
             i] != list(current_version)[i]]
-        return json.dumps({'version': latest_release, 'importance': importance[level[0]]})
+        return json.dumps({'version': latest_release, 'importance': Importance.__ordered__[level[0]]})
 
     return True
--- a/golem/version.py
+++ b/golem/version.py
@@ -1,0 +1,43 @@
+import requests
+import json
+import logging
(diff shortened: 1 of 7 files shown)


golem/version.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+def check_update():
+    GITHUB_RELEASE_URL = "https://api.github.com/repos/golemfactory/golem/releases"
+    response = requests.get(GITHUB_RELEASE_URL)
+
+    if response.status_code != 200:
+        log.error("Github release check error code {}".format(
+            response.status_code))
+        response.raise_for_status()
```
Elfoniok, line 40: Where you catch exception threw here?

golem/version.py
```diff
@@ -0,0 +1,36 @@
...
+
+    latest_release = max(response.json(), key=lambda ev: ev['tag_name'])
```
Radagast-red: Don't you get strings from here?  If so max can not work well

golem/version.py
```diff
@@ -0,0 +1,36 @@
...
+    latest_release = latest_release['tag_name']
+    latest_version = Version(latest_release, partial=True)
+    current_version = Version(APP_VERSION, partial=True)
+
+    if latest_version > current_version:
```
Elfoniok: Same here if you compare strings 10 version will be lower than 9

golem/version.py
```diff
@@ -0,0 +1,36 @@
...
+        level = [i for i in range(len(list(latest_version))) if list(latest_version)[
+            i] != list(current_version)[i]]
+        return json.dumps({'version': latest_release, 'importance': Importance.__ordered__[level[0]]})
+
+    return True
```
Elfoniok, line 51: What does it mean to return True here?

Elfoniok (changes requested):

golem/version.py
mdtanrikulu: This is the feature of [semantic_version](https://pypi.python.org/pypi/semantic_version/) package. `Version` classes easily comparable with simple comparison operators.

golem/version.py
mdtanrikulu: `response.raise_for_status()`
Requests(v2.8.0 >=) library has it's own exception handling, if status code is not one of the success codes 2xx , it will raise exception.

golem/version.py
Elfoniok: Ok great

golem/version.py
Elfoniok: Yes so? where you are going to catch those exceptions? I can see you are catching this in tests, buy what for client? or even RPC call?

mdtanrikulu: On RPC call there's an exception handling but for CLI there's no, so I'll fix it soon.

golem/version.py
mdtanrikulu: If the current version is latest one, just returned boolean instead of sending json message about it, and I'll check the type on client-side if message is boolean or not.

Elfoniok: Can you add a comment to this function?

mdtanrikulu: Sure.

Elfoniok (approved):

golemcli.py
```diff
@@ -23,6 +25,22 @@
...
+            print()
+            print('#######################################################################################################')
+            print()
+            print('\t\x1b[0;33;40m' + "We released new version of Golem Brass (v{}). Update your instance for best results!".format(
+                is_updated['version']) + '\x1b[0m')
+            print()
+            print('#######################################################################################################')
+            print()
+    except Exception as e:
+        print(e)
```
jiivan, line 42: `traceback.print_exc()`

maaktweluit:
@jiivan @mdtanrikulu this looks like a duplicate of PR #1479 can this be closed or is it different?

mdtanrikulu:
Yeap duplicate, we can close it.

Issue #221, opened 2016-03-21
Multiple confirmations of Ethereum payments


chfast:
Add support for waiting for multiple confirmations of Ethereum payment. The number of confirmations should be configurable. Ethereum Wallet uses 12.

globatum:
@chfast this one is labeled for brass 0.4 but we are at 0.6. can we close this?

paulperegud:
#planned

paulperegud:
related task: #1291

PR #1466, opened 2017-10-10
Wait for minimal number of confirmation for Ethereum payments #221

Krigpl:
As per description of the issue #221

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -69,6 +69,9 @@ class PaymentProcessor(Service):
 
     SYNC_CHECK_INTERVAL = 10
 
+    # Minimal number of confirmations before we treat transactions as done
+    REQUIRED_CONFIRMATIONS = 12
+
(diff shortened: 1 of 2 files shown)


golem/ethereum/paymentprocessor.py
```diff
@@ -310,6 +313,8 @@ def monitor_progress(self):
                         "block hash length should be 64, but is: {}".format(
                             len(block_hash)))
                 block_number = receipt['blockNumber']
+                if block_number < self.REQUIRED_CONFIRMATIONS:
+                    continue
```
chfast, line 317: Can add (or confirm one exists) a test where there are some confirmations but not 12?

Krigpl: Do you mean something more fancy than lines 399-402 in `test_paymentprocessor.py` added in this commit?

chfast: Oh no. I missed the test has actually 2 stages.

chfast (approved):

PR #1472, opened 2017-10-11
Connect only to one bootstrap node at a time

### LINKED ISSUE #1459: Golem should connect to only 1 random bootstrap node
Currently Golem tries to connect to every hardcoded seed node and they get overloaded. Make sure that Golem tries to connect to only one randomly chosen and try again only if that fails.

jiivan:
fixes: #1459


golem/network/p2p/p2pservice.py
```diff
@@ -140,13 +141,15 @@ def connect_to_seeds(self):
...
 
-        for ip_address, port in self.seeds:
+        for ip_address, port in random.sample(self.seeds, k=len(self.seeds)):
             try:
                 socket_address = tcpnetwork.SocketAddress(ip_address, port)
                 self.connect(socket_address)
             except Exception as exc:
-                logger.error("Cannot connect to seed {}:{}: {}"
-                             .format(ip_address, port, exc))
+                logger.error("Cannot connect to seed %s:%s: %s",
```
maaktweluit, line 147: When trying again after failure should this be an `error` or `warning`?

maaktweluit (approved):
LGTM!

One small question, nothing to stop the merge.

golem/network/p2p/p2pservice.py
jiivan: Hmm that's very interesting question. :) We don't have definite guidelines for loglevels so it's a very subjective matter. For sure connecting to seed nodes is a crucial step in connecting to p2p network and failure at this point is an important event (possibly prohibiting the node from any meaningful activity).

maaktweluit: maybe a single seed failing can be a warning, and ALL seeds failing an error?

In the first case the user can continue, and the latter is a blocking issue.

badb (approved):

PR #1473, opened 2017-10-11
Allow to define starting seeds in external config file

### LINKED ISSUE #1461: Allow to define starting seeds in external config file
Right now Golem has:
* set of hardcoded bootstrap nodes to which it tries to connect after the start and this should stay like that. 
* set of previously known good nodes to which it tries to connect on start and this should stay like that
* one single seed node, which can be specifi in `appconfig.ini` to which Golem can try to connect on start.

jiivan:
fixes: #1461


golem/network/p2p/p2pservice.py
```diff
@@ -1037,6 +1037,19 @@ def __sync_seeds(self, known_hosts=None):
...
+        for config_seed in self.config_desc.seeds.split(None):
+            try:
+                ip_address, port = config_seed.split(':', 1)
+                port = int(port)
+                # Verify that ip_address is valid.
+                # Throws ipaddress.AddressValueError.
+                ipaddress.ip_address(ip_address)
+            except ValueError:
+                logger.info("Invalid seed from config: %r. Ignoring.",
+                    config_seed)
```
maaktweluit: wrong continuation, should match with `(`

tests/golem/core/test_simpleconfig.py
```diff
@@ -1,61 +0,0 @@
-from golem.appconfig import CommonConfig
```
maaktweluit, line 1: Why is this file removed?

maaktweluit (changes requested):
LGTM!

Small lint fix and one question

tests/golem/core/test_simpleconfig.py
jiivan: CommonConfig is unused.

maaktweluit (approved):

PR #1474, opened 2017-10-11
Fix off-by-one errors in PeerKeeper

Krigpl:
Interval represented by bucket was supposed to be right-open but that was not reflected in the code.
We were losing track of the midpoint value during split


Krigpl:
Looks like I made some mess with git, let me clean it up

badb (approved):
Unittests for peerkeeper must be added, but this may be done later.

PR #1476, opened 2017-10-11
Changes to message serialization, signing and verification

mfranciszkiewicz:
- signature verification does not require message re-serialization
- signature verification is no longer affected by the order of properties inside objects

Addresses #1457


golem/network/transport/message.py
```diff
@@ -11,123 +11,117 @@
...
-                encrypted = False
-            except Exception as err:
-                logger.info("Failed to decrypt message {}".format(str(err)))
-                continue
+        11 bytes in total
 
-            m = cls.deserialize_message(msg)
+        :return: serialized header
+        """
+        return struct.pack('!HQ?', self.TYPE,
```
etam: E: golem/network/transport/message.py:102:35: Instance of 'Message' has no 'TYPE' member (no-member)

golem/core/databuffer.py
```diff
@@ -21,19 +18,12 @@ def append_ulong(self, num):
...
-        :param check_size: keep buffer size below MAX_BUFFER_SIZE
-        :param overflow_prefix: string to prepend on overflow
         :param str data: string to append
         """
-        new_size = self.data_size() + len(data)
-        if check_size and new_size > MAX_BUFFER_SIZE:
-            self.buffered_data = b"".join([overflow_prefix or '', data])
-        else:
-            self.buffered_data = b"".join([self.buffered_data, data])
+        self.buffered_data = b"".join([self.buffered_data, data])
```
etam: Is this bytes or strings?

```python
>>> b''.join([b'123', "abc"])
  File "<stdin>", line 1, in <module>
TypeError: sequence item 1: expected a bytes-like object, str found
```

If it's all bytes, then this should be fine:
```python
self.buffered_data += data
```

golem/core/databuffer.py
```diff
@@ -1,9 +1,6 @@
 import struct
-from golem.core.common import to_unicode
 
-from .variables import LONG_STANDARD_SIZE
-
-MAX_BUFFER_SIZE = 2 * 1024 * 1024
+from .variables import LONG_STANDARD_SIZE
 
 
 class DataBuffer:
```
etam: Comments in this class suggests, that it operates on strings, but instead it operates on bytes.

golem/core/variables.py
```diff
@@ -5,7 +5,7 @@
 #       VARIABLES          #
 ############################
 APP_NAME = "Brass Golem"
-APP_VERSION = "0.8.0"
```
etam, line 8: Is this supposed to go in here?

golem/network/transport/message.py
```diff
@@ -11,123 +11,117 @@
...
-        :param SafeServer server: server that is able to decrypt data
-        :return list: list of decrypted and deserialized messages
-        """
-        if not isinstance(db_, DataBuffer):
-            raise TypeError(
-                "Incorrect db type: {}. Should be: DataBuffer".format(db_)
+            if sign_func:
+                self.sig = sign_func(self.get_short_hash())
+            else:
+                self.sig = b'0' * 65
```
etam: Magic constant

tests/golem/network/transport/test_network.py
```diff
@@ -27,8 +27,7 @@ def interpret(self, msg):
         self.msgs.append(msg)
 
     def sign(self, msg):
-        msg.sig = "ASessionSign"
-        return msg
+        return b'1' * 65
```
etam: Magic constant

etam (changes requested):

golem/network/transport/message.py
```diff
@@ -11,123 +11,117 @@
...
         self.sig = sig
-        if timestamp is None:
-            timestamp = time.time()
-        self.timestamp = timestamp
-        self.encrypted = False  # inform if message was encrypted
 
-        self.load_dict_repr(dict_repr)
+        # Encoded data
+        self._payload = payload  # child's payload only (may be encrypted)
+        self._raw = raw  # whole message
```
Krigpl: I might be missing something but is it used at all?

golem/network/transport/message.py
```diff
@@ -11,123 +11,117 @@
...
+        self.encrypted = encrypted
         self.sig = sig
-        if timestamp is None:
-            timestamp = time.time()
-        self.timestamp = timestamp
-        self.encrypted = False  # inform if message was encrypted
 
-        self.load_dict_repr(dict_repr)
+        # Encoded data
+        self._payload = payload  # child's payload only (may be encrypted)
```
Krigpl, line 51: `self.encrypted`, `self.sig` and `self._payload` don't need to exist.
They are only set and used during serialization, we can just used them there locally.

golem/network/transport/message.py
Krigpl, line 52: I might be missing something but is it used?

golem/core/databuffer.py
mfranciszkiewicz: Comments weren't changed while porting from Python 2 to Python 3.

golem/core/databuffer.py
mfranciszkiewicz: Yes, it should be fine. `''.join` is just faster.

golem/network/transport/message.py
```diff
@@ -11,123 +11,117 @@
...
+        self._raw = raw  # whole message
+
+    @property
+    def raw(self):
+        """Returns a raw copy of the message"""
+        return self._raw[:]
 
     def get_short_hash(self):
```
Krigpl, line 59: I believe this doesn't have to be exposed anymore.
And by making it an internal method we can get rid of the properties I mentioned in the other comment

golem/network/transport/message.py
mfranciszkiewicz: Actually I've tried replacing all byte counts with constants and the code was IMO less readable. I'd like to hear another opinion.

golem/core/variables.py
mfranciszkiewicz: The setup script automatically updates the version and it's pointless to fight it with every PR.

tests/golem/network/transport/test_network.py
mfranciszkiewicz: I believe that this is ok in tests.

golem/network/transport/message.py
mfranciszkiewicz: The base `Message` class is not instantiated in the application. Where is that message from?

Krigpl (changes requested):
Overall looks good, just a couple of comments to make the Message class even a bit simpler.

golem/network/transport/message.py
mfranciszkiewicz: This will be used in the near future

golem/network/transport/message.py
mfranciszkiewicz: I agree on hiding the method. EDIT: actually, it's still needed outside the class.

golem/core/databuffer.py
etam:
```
$ python3 -m timeit 'b1 = b"0"*100; b2 = b"f"*100; b1 = b"".join([b1, b2])'
10000000 loops, best of 3: 0.19 usec per loop
$ python3 -m timeit 'b1 = b"0"*100; b2 = b"f"*100; b1 += b2'
10000000 loops, best of 3: 0.126 usec per loop
```
It looks like `+=` is faster.

golem/network/transport/message.py
etam: From pylint, run by shippable.

golem/core/databuffer.py
mfranciszkiewicz: Though 100 bytes is by orders of magnitude too low for this test, it seems that `b''.join` is not faster in Python 3.

mfranciszkiewicz: I'll revert changes all changes made in databuffer and address them in a separate PR.

etam: At 10000 bytes they become equal

golem/network/transport/message.py
mfranciszkiewicz: Without more refactoring, these are still needed for signature verification, done outside the class.

golem/network/transport/message.py
Krigpl: It's used three times and in one place you put a comment explaining what it is so might be worth using an actual constant.
Something like `SIG_LEN = 65` shouldn't hurt legibility too much

golem/network/transport/message.py
```diff
@@ -1,8 +1,8 @@
-import collections
 import logging
+import struct
 import time
 
-from golem.core.common import to_unicode
+
 from golem.core.databuffer import DataBuffer
```
Krigpl: I think it'd be nice if we'd get rid of DataBuffer here. That would mean removing `deserialize` method and leaving only `deserialize_message` which doesn't depend on DataBuffer. `deserialize` is only used in `tcpnetwork.py` and that's where DataBuffer logic should live. You already did pretty much the same with `decrypt_and_deserialize` so removing DataBuffer dependency altogether would be nice.

Krigpl (approved):
I'm fine with the current version, just one clean up suggestion.
Approving, but double check with @etam whether he has any more comments.

golem/network/transport/message.py
mfranciszkiewicz: I've added `TYPE = None` to make the linter happy, but this part will fail (0 and up are reserved, negative value is not really possible). Overall, it doesn't make much sense.

etam: Without TYPE it will fail anyway, so it's absolutely fine.

etam (approved):

golem/network/transport/message.py
mfranciszkiewicz: The difference is that `decrypt_and_deserialize` was only used in tests, so I did that freely. I suggest leaving this for the DataBuffer-related PR.

mfranciszkiewicz:
@Krigpl @etam I've removed the `DataBuffer` dependency from message.py, in preparation for the upcoming `DataBuffer` changes.

etam (approved):
LGTM

Krigpl (approved):

PR #1480, opened 2017-10-12
Implement PoW for public keys

etam:
This should make it harder to perform a Sybil attack on Golem network.


golem/core/keysauth.py
```diff
@@ -399,18 +413,31 @@ class EllipticalKeysAuth(KeysAuth):
...
+            public key. Works with floats too.
+            Value in range <0, 256>. 0 is not difficult.
+            Maximum is impossible.
         """
-        KeysAuth.__init__(self, datadir, private_key_name, public_key_name)
+        KeysAuth.__init__(self, datadir, private_key_name, public_key_name,
+                          difficulty)
+
+        if not self.is_difficult(difficulty):
+            logging.warning("Current key is not difficult enough. Creating new one.")
```
maaktweluit: `logger` vs `logging`? what is the difference?

maaktweluit (commented):
Not enough knowledge to approve, just one small q.

golem/core/keysauth.py
etam: https://docs.python.org/3/howto/logging.html
TL;DR: Typo

golem/core/keysauth.py
```diff
@@ -75,19 +79,29 @@ def get_random_float():
...
+        self.difficulty = difficulty
         self.get_keys_dir(datadir)
         self.private_key_name = private_key_name
         self.public_key_name = public_key_name
         self._private_key = self._load_private_key()
         self.public_key = self._load_public_key()
         self.key_id = self.cnt_key_id(self.public_key)
 
+    @classmethod
+    def is_pubkey_difficult(cls, pub_key: Union[bytes, str],
```
ggruszczynski: signature mismatch with baseclass:
``` python
@staticmethod
def is_pubkey_difficult(pub_key: Union[bytes, str],
                            difficulty: IntFloatT) -> bool:
```

golem/core/keysauth.py
```diff
@@ -472,21 +499,57 @@ def verify(self, sig, data, public_key=None):
...
+
+    @staticmethod
+    def _is_pubkey_difficult(pub_key: bytes, max_hash: int) -> bool:
+        return sha2(pub_key) < max_hash
+
+    @staticmethod
+    def is_pubkey_difficult(pub_key: Union[bytes, str],
+                            difficulty: IntFloatT) -> bool:
+        if isinstance(pub_key, str):
+            pub_key = decode_hex(pub_key)
```
ggruszczynski: this is not covered by unit tests:
``` python
if isinstance(pub_key, str):
     pub_key = decode_hex(pub_key)
```
Do we use it anywhere? What happens if someone send a key in a bad format?

golem/core/keysauth.py
```diff
@@ -399,18 +413,31 @@ class EllipticalKeysAuth(KeysAuth):
...
+            Maximum is impossible.
         """
-        KeysAuth.__init__(self, datadir, private_key_name, public_key_name)
+        KeysAuth.__init__(self, datadir, private_key_name, public_key_name,
+                          difficulty)
+
+        if not self.is_difficult(difficulty):
+            logger.warning("Current key is not difficult enough. Creating new one.")
+            self.generate_new(difficulty)
+
```
ggruszczynski: It is quite confusing to get the logic, because there are 3 methods with similar purpose:
Merge/clean them please.

``` python
def generate_new(self, difficulty: IntFloatT) -> None:

@classmethod
def _generate_keys(cls, private_key_loc, public_key_loc, difficulty: IntFloatT):

@staticmethod
def _generate_new_keys(difficulty: IntFloatT) -> (bytes, bytes):
```

golem/core/keysauth.py
```diff
@@ -548,15 +612,16 @@ def _load_public_key(self):
...
         return key
 
-    @staticmethod
-    def _generate_keys(private_key_loc, public_key_loc):
-        key = mk_privkey(str(get_random_float()))
-        pub_key = privtopub(key)
+    @classmethod
+    def _generate_keys(cls, private_key_loc, public_key_loc,
+                       difficulty: IntFloatT):
+        key, pub_key = cls._generate_new_keys(difficulty)
```
ggruszczynski: Please keep nomenclature `priv_key, pub_key = cls._generate_new_keys(difficulty)` ... just in case ;)

golem/core/keysauth.py
```diff
@@ -472,21 +499,57 @@ def verify(self, sig, data, public_key=None):
...
 
+    def get_difficulty(self, key_id: Optional[str] = None) -> float:
+        """
+        :param key_id: *Default: None* count difficulty of given key.
+                       If key_id is None then use default key_id
+        :return: key_id difficulty
+        """
+        pub_key = decode_hex(key_id) if key_id else self.public_key
+        return 256 - math.log2(sha2(pub_key))
+
```
ggruszczynski: It seems, that `get_difficulty` has a lot of common with `is_***_difficult`, I think that they should be merged like:

```python
def is_difficult_enought(some_key, difficulty):
    return get_difficulty(some_key) > difficulty
```

ggruszczynski (changes requested):
God job, however some code cleanup would be beneficial.

golem/core/keysauth.py
etam: The thing is that `get_difficulty` takes much more time to compute, than current `is_difficult_enough`. And it's crucial when generating new keys.

etam:
ReReview please. I've made a lot of cleanups and refactoring around KeysAuth to untangle the spaghetti there was.

golem/core/keysauth.py
```diff
@@ -72,388 +69,172 @@ def get_random_float():
...
 
-        try:
-            with open(private_key_loc, 'wb') as f:
-                f.write(self._private_key.exportKey('PEM'))
-            with open(public_key_loc, 'wb') as f:
-                f.write(self.public_key.exportKey())
-                return True
-        except IOError:
-            return False
+    def get_public_key(self) -> bytes:
```
maaktweluit: why does this have a getter when `public_key` is a public attribute?

golem/core/keysauth.py
```diff
@@ -72,388 +69,172 @@ def get_random_float():
...
-    def _load_public_key(self):
-        private_key_loc = RSAKeysAuth._get_private_key_loc(self.private_key_name)
-        public_key_loc = RSAKeysAuth._get_public_key_loc(self.public_key_name)
-        if not os.path.isfile(private_key_loc) or not os.path.isfile(public_key_loc):
-            RSAKeysAuth._generate_keys(private_key_loc, public_key_loc)
-        with open(public_key_loc) as f:
-            key = f.read()
-        key = RSA.importKey(key)
-        return key
+    def get_key_id(self) -> str:
```
maaktweluit: same q as above

maaktweluit (approved):
RE checked the separate commits and the full change set, 2 small questions.

Besides those LGTM!

golem/core/keysauth.py
etam: fixed

golem/core/keysauth.py
etam: fixed

golem/core/keysauth.py
etam: > this is not covered by unit tests

will fix that

> Do we use it anywhere?

We use it here: https://github.com/golemfactory/golem/blob/2c82f8e7a1af783ef59ee8cda472c1fe768b6259/golem/network/p2p/peersession.py#L390

> What happens if someone send a key in a bad format?

Shit happens ;) But seriously, if such event happens, it should be detected at lower layer of networking stack. And if it reaches here, then decode_hex will throw an exception.

etam: fixed (added test)

golem/core/keysauth.py
```diff
@@ -72,388 +69,164 @@ def get_random_float():
...
 
-        try:
-            with open(private_key_loc, 'wb') as f:
-                f.write(self._private_key.exportKey('PEM'))
-            with open(public_key_loc, 'wb') as f:
-                f.write(self.public_key.exportKey())
-                return True
-        except IOError:
-            return False
+        return (priv_key, pub_key)
```
ggruszczynski: The method shall also check whether pubkey matches the privkey...
or store only the priv_key.

golem/core/keysauth.py
```diff
@@ -72,388 +69,164 @@ def get_random_float():
...
-        if not os.path.isfile(private_key_loc) or not os.path.isfile(public_key_loc):
-            RSAKeysAuth._generate_keys(private_key_loc, public_key_loc)
-        with open(public_key_loc) as f:
-            key = f.read()
-        key = RSA.importKey(key)
-        return key
+    def _set_keys(self, priv_key: bytes, pub_key: bytes) -> None:
+        self._private_key = priv_key
+        self.public_key = pub_key
+        self.key_id = self._cnt_key_id(pub_key)
```
ggruszczynski: I think that:
```python
 self.key_id = encode_hex(pub_key)
```
is enought, because `_def _cnt_key_id(public_key: bytes)` isn't used anywhere else (only in tests).
Moreover the `cnt` shorcut is strange.

golem/core/keysauth.py
```diff
@@ -472,42 +245,85 @@ def verify(self, sig, data, public_key=None):
...
+
+    @staticmethod
+    def is_pubkey_difficult(pub_key: Union[bytes, str],
+                            difficulty: IntFloatT) -> bool:
+        if isinstance(pub_key, str):
+            pub_key = decode_hex(pub_key)
+        max_hash = EllipticalKeysAuth._count_max_hash(difficulty)
+        return EllipticalKeysAuth._is_pubkey_difficult(pub_key, max_hash)
+
+    def is_difficult(self, difficulty: IntFloatT) -> bool:
```
ggruszczynski: the `def is_difficult(self, difficulty: IntFloatT)` is a dead code. Appears only in tests.

golem/core/keysauth.py
```diff
@@ -472,42 +245,85 @@ def verify(self, sig, data, public_key=None):
...
+        """
+        pub_key = decode_hex(key_id) if key_id else self.public_key
+        return 256 - math.log2(sha2(pub_key))
 
-    def load_from_file(self, file_name):
-        """ Load private key from given file. If it's proper key, then generate public key and
-        save both in default files
-        :param str file_name: file containing private key
-        :return bool: information if keys have been changed
+    def load_from_file(self, file_name: str) -> bool:
```
ggruszczynski: I think that we shall always use the safer method: _load_and_check_keys(...)
BTW, maybe it is enough to store just the priv_key?

golem/core/keysauth.py
```diff
@@ -72,388 +69,164 @@ def get_random_float():
...
-            return True
+        if not len(pub_key) == EllipticalKeysAuth.PUB_KEY_LEN:
+            logger.warning("Unexpected public key size: %d. "
+                           "Will create new keys.", len(pub_key))
+            return None
 
-        if not (make_dir(private_key_loc) and make_dir(public_key_loc)):
-            return False
+        if not EllipticalKeysAuth.is_pubkey_difficult(pub_key, difficulty):
+            logger.warning("Current key is not difficult enough. "
```
ggruszczynski: What will happen when GolemFactory will decide to increase the difficulty?
I am afraid that silent overwriting of the previous keys may result in unexpected lost of identity, founds etc.
For now, let the `self._save_keys()` just create a new file "keys_no1-DD-MM-YYYY)" then "keys_no1-DD-MM-YYYY".

ggruszczynski (changes requested):

golem/core/keysauth.py
etam: done

golem/core/keysauth.py
etam: Well, but it might be useful. And it makes the interface complete.

golem/core/keysauth.py
etam: I asked a question https://github.com/imapp-pl/golem_rd/issues/132#issuecomment-345200716 
no reply so far

etam: But anyway, I've made it use _load_and_check_keys

golem/core/keysauth.py
etam: done

golem/core/keysauth.py
etam: done

golem/core/keysauth.py
```diff
@@ -72,388 +70,184 @@ def get_random_float():
...
-            logger.error("Cannot verify signature: {}".format(exc))
-        return False
+            with open(private_key_loc, 'rb') as f:
+                priv_key = f.read()
+            pub_key = None
+            if public_key_loc:
+                with open(public_key_loc, 'rb') as f:
+                    pub_key = f.read()
+        except FileNotFoundError:
+            return None
```
ggruszczynski: Please, avoid silent try-except block.
Please, add a logger in the except block.

ggruszczynski: Please, avoid silent try-except block.
Please, add a logger in the except block.

golem/core/keysauth.py
```diff
@@ -472,42 +266,87 @@ def verify(self, sig, data, public_key=None):
...
+
+        :param key_id: *Default: None* count difficulty of given key.
+                       If key_id is None then use default key_id
+        :return: key_id difficulty
+        """
+        pub_key = decode_hex(key_id) if key_id else self.public_key
+        return 256 - math.log2(sha2(pub_key))
+
+    def load_from_file(self, file_name: str) -> bool:
+        """ Load private key from given file. If it's proper key, then generate
```
ggruszczynski: `def load_from_file` is a deadcode together with 
`def load_keys_from_file(self, file_name):` from golem/client.py
Remove please.

golem/core/keysauth.py
```diff
@@ -472,42 +266,87 @@ def verify(self, sig, data, public_key=None):
...
-        self._set_and_save(priv_key, pub_key)
+
+        self._set_keys(*keys)
+        EllipticalKeysAuth._backup_keys(
+            self._private_key_loc, self._public_key_loc)
+        self._save_keys()
         return True
 
     def save_to_files(self, private_key_loc: str, public_key_loc: str) -> bool:
         """ Save current pair of keys in given locations
```
ggruszczynski: `def save_to_files(self, private_key_loc: str, public_key_loc: str) -> bool: deadcode`
Please remove together with
`def save_keys_to_files(self, private_key_path, public_key_path)` from golem/client.py

golem/core/keysauth.py
ggruszczynski: There is to much features (and dead code) which might be useful in golem. 
It is enough to keep the live code in develop branch.
Please remove this function.

ggruszczynski (approved):
Minor dead-code cleanup left.
Please do not git-rebase large amount of code. 
It overrides commit history making re-review from previous-review impossible.
Good job, thank you!

golem/core/keysauth.py
tworec: according to what I have ack from @badb, as for Golem Brass we will accept that node will re-generate key when difficulty is increased.

requirements.txt
```diff
@@ -49,4 +49,5 @@ numpy==1.13.1
 scipy==0.19.1
 pyssim
 enforce==0.3.4
--e git://github.com/golemfactory/golem-messages.git@v1.1.0#egg=golem_messages
+# TODO: revert this before merging
```
tworec: why?

tworec:
closes imapp-pl/golem_rd#132

requirements.txt
etam: https://github.com/golemfactory/golem-messages/pull/25 was merged

tworec: I've realized that already but thanks

golem/core/keysauth.py
```diff
@@ -54,462 +50,180 @@ def get_random_float():
...
-
-    @abstractmethod
-    def decrypt(self, data):
-        """ Decrypt given data with default private key
-        :param str data: encrypted data
-        :return str: decrypted data
+class KeysAuth:
+    """
+    Elliptical curves cryptographic authorization manager. Generates
+    private and public keys based on ECC (curve secp256k1). Keeps only private.
```
etam: This "Keeps only private" is confusing and, I think, unneeded.

tworec: thx @etam, please check it again now

etam:
@tworec Github doesn't allow me to approve this PR, because I created it. But I approve.

etam:
🎉

PR #1489, opened 2017-10-16
Limit TaskHeaders sent

Elfoniok:


golem/network/p2p/peersession.py
```diff
@@ -438,7 +439,16 @@ def _react_to_peers(self, msg):
...
     def _react_to_get_tasks(self, msg):
         tasks = self.p2p_service.get_tasks_headers()
-        self.send(message.MessageTasks(tasks))
+        max_ind = len(tasks) - 1
+        if max_ind < 0:
+            return
+        tasks_to_send = []
+        if len(tasks) > TASK_HEADERS_LIMIT:
+            for i in range(0, TASK_HEADERS_LIMIT):
+                tasks_to_send.append(tasks[random.randint(0, max_ind)])
```
ggruszczynski: random.randint(0, max_ind) can return the same number several times.
Please, use drawing without repetition.

ggruszczynski (changes requested):
Please:
1) use drawing without repetition.
2) to keep a good test coverage add a simple unit test 
Have a look at test_peerseession and test_taskserver.test_add_task_header --> get_example_task_header.

golem/network/p2p/peersession.py
mfranciszkiewicz: In Python 3.6, there is a `random.choices(collection)` function. We can backport it to 3.5 from [here](https://github.com/python/cpython/blob/master/Lib/random.py#L342)

Elfoniok: But choices is with repetition, no? I want to try random.sample is this all right?
https://docs.python.org/3/library/random.html#random.sample

tests/golem/network/p2p/test_peersession.py
```diff
@@ -186,6 +187,30 @@ def test_interpret(self):
...
+        peer_session._react_to_get_tasks(Mock())
+        assert not peer_session.send.called
+
+        peer_session.p2p_service.get_tasks_headers.return_value = list(
+            range(0, 100))
+        peer_session._react_to_get_tasks(Mock())
+
+        sent_tasks = peer_session.send.call_args_list[0][0][0].tasks_array
+        assert len(sent_tasks) <= TASK_HEADERS_LIMIT
+        assert len(sent_tasks) == len(set(sent_tasks))
```
ggruszczynski: what is the purpose of  `assert len(sent_tasks) == len(set(sent_tasks))` ?
I think that `assert len(sent_tasks) == TASK_HEADERS_LIMIT` should be enought

Elfoniok: I would but it's better to check if all entries are unique, as you pointed out without repetition.

ggruszczynski (approved):

golem/network/p2p/peersession.py
```diff
@@ -441,7 +442,14 @@ def _react_to_peers(self, msg):
 
     def _react_to_get_tasks(self, msg):
         tasks = self.p2p_service.get_tasks_headers()
-        self.send(message.MessageTasks(tasks))
+        max_ind = len(tasks) - 1
+        if max_ind < 0:
```
jiivan: Maybe this would be more readable?
```python
if not tasks:
    return
```

jiivan (approved):
Small comment, not a stopper.

maaktweluit:
@Elfoniok This PR looks ready for merge, anything left to fix? If so, can i help?

PR #1491, opened 2017-10-16
Limit the number of peers sent in MessagePeers

mfranciszkiewicz:


tests/golem/network/p2p/test_peerkeeper.py
```diff
@@ -0,0 +1,74 @@
...
+        # Sort keys by distance to self.key
+        distances = {p.key: node_id_distance(p, self.key_num) for p in peers}
+        ordered = sorted(distances.items(), key=operator.itemgetter(1))
+        ordered = [o[0] for o in ordered]
+
+        # Default count
+        expected_n = CONCURRENCY
+        nodes = self.peer_keeper.neighbours(self.key_num)
+        assert len(nodes) == expected_n
+        assert all(node.key in ordered[:16] for node in nodes)
```
Krigpl, line 50: I believe `ordered[:expected_n]` should be used here, right?
Same for the line 56.
Does it mean you added some buffer because otherwise it doesn't work? Could you check whether it works with fixes from #1474? Otherwise this is something to investigate

mfranciszkiewicz: That is the case, it doesn't (always) return closest neighbours. Maybe this test should be moved to #1474?

Krigpl: Feel free to commit it as it is and I'll update it in #1474.

Krigpl (approved):

PR #1494, opened 2017-10-16
DisableMessageDegree

Elfoniok:


badb:
Resolves: 
https://github.com/golemfactory/golem/issues/1502

gborowik (approved):

maaktweluit (approved):
LGTM!

Quick search showed some more references, are they un-related?
https://github.com/golemfactory/golem/search?utf8=%E2%9C%93&q=MessageDegree&type=

Elfoniok:
The idea was to disable sending/receiving. The message with its structure and sending/reacting method stays. Call sites of this methods was removed since they was broken anyway (producing so much noise), receiving code was unlinked to drop any Degree messages if someone will post them.

maaktweluit:
@Radagast-red Thank you for the clear reply, double approved now :)

gborowik:
@maaktweluit So it is what I discussed with @Radagast-red. The main idea was to disable sending and receiving. The other stuff was left.

mfranciszkiewicz:
> receiving code was unlinked to drop any Degree messages if someone will post them

@Radagast-red Golem will actually drop that peer https://github.com/golemfactory/golem/blob/dfab211a4c936da72e3b97cb268e83655805da5f/golem/network/transport/session.py#L92-L96

Elfoniok:
Ok this is a good point. Should i change protocol version then? Or should i react to MessageDegree simply and do nothing?

mfranciszkiewicz:
@Radagast-red I think changing protocol version is enough for now. `react_to_degree` doesn't really do much, so most of the time is spent on deserializing the message, etc.

Issue #1505, opened 2017-10-16
Check why TaskHeaders that are after deadline are received after restart


badb:
Sometimes when you turn off Golem and turn it on again, your node receives several TaskHeaders that has already hit the deadline. Try to reproduce this issue and figure out what is the reason for it. There are several theories:
* It comes from node which don't have syncronized clock
* Deadlines are wrongly computed (utc time vs time.time()?)
* Removing old tasks is not called properly?
Fix the issue, when you understand the problem.

ggruszczynski:
NOTES:
the message is:
`WARNING  [golem.task.taskkeeper              ] Wrong task header received Deadline already passed`

REPRODUCTION:
Set task deadline to 1s :)

ggruszczynski:
Perhaps a new buggy issue:
Hang (keep on breakpoint for a while) for example at peersession._react_to_get_tasks causes ==> Golem crashes.

```
WARNING  [twisted                            ] dropping connection to peer process:11261 with abort=True: WebSocket ping timeout (peer did not respond with pong in time)
WARNING  [twisted                            ] dropping connection to peer tcp4:127.0.0.1:61000 with abort=True: WebSocket ping timeout (peer did not respond with pong in time)
WARNING  [golem.task.tasksession             ] Fail to decrypt message Fail to verify data
WARNING  [golem.network.transport.tcpnetwork ] Decryption of message failed
WARNING  [golem.task.tasksession             ] Fail to decrypt message Fail to verify data
WARNING  [golem.network.transport.tcpnetwork ] Decryption of message failed
WARNING  [golem.task.tasksession             ] Fail to decrypt message Fail to verify data
WARNING  [golem.network.transport.tcpnetwork ] Decryption of message failed
WARNING  [golem.task.tasksession             ] Fail to decrypt message Fail to verify data
WARNING  [golem.network.transport.tcpnetwork ] Decryption of message failed
WARNING  [golem.task.tasksession             ] Fail to decrypt message Fail to verify data
WARNING  [golem.network.transport.tcpnetwork ] Decryption of message failed
WARNING  [twisted                            ] session closed with reason wamp.close.transport_lost [()]
INFO     [twisted                            ] Scheduling retry 1 to connect <twisted.internet.endpoints.TCP4ClientEndpoint object at 0x7f2fb4072208> in 2.2225069136706552 seconds.
WARNING  [golem.rpc                          ] RPC: Cannot publish 'evt.net.connection', session is not yet established
[...]

PR #1490, opened 2017-10-16
Updated rinkeby.json (Byzantium)

mfranciszkiewicz:

PR #1522, opened 2017-10-19
Taskheaders after deadline 1505

ggruszczynski:
Solves:
Check why TaskHeaders that are after deadline are received after restart #1505

Already merged with:
Protocol id 1380 #1509
Updated rinkeby.json (Byzantium) #1490


golem/task/taskmanager.py
```diff
@@ -439,7 +445,7 @@ def check_timeouts(self):
             cur_time = get_timestamp_utc()
             if cur_time > th.deadline:
                 logger.info("Task {} dies".format(th.task_id))
-                t.task_stats = TaskStatus.timeout
+                t.task_status = TaskStatus.timeout
```
Elfoniok, line 448: So this is actually a reason?

golem/task/taskserver.py
```diff
@@ -280,9 +280,10 @@ def disconnect(self):
                 logger.error("Error closing incoming session: %s", exc)
 
     def get_tasks_headers(self):
-        ths = self.task_keeper.get_all_tasks() + \
-              self.task_manager.get_tasks_headers()
-        return [th.to_dict() for th in ths]
+        ths_tk = self.task_keeper.get_all_tasks()
+        ths_tm = self.task_manager.get_tasks_headers()
+        ret  = [th.to_dict() for th in ths_tk + ths_tm]
```
Elfoniok, line 285: Why are you creating unnecessary local vars?

golem/task/taskmanager.py
ggruszczynski: i think so :)

PR #1524, opened 2017-10-19
Geth: connect to nodes with public RPC interface

mfranciszkiewicz:
- by default, connect to known public geth nodes
- optionally, start a local node with the `--start-geth` command line argument

TODO:
- populate the known public node list with production nodes


golem/ethereum/node.py
```diff
@@ -105,22 +110,71 @@ def start(self, port=None):
...
+        self.web3 = Web3(provider)
+        atexit.register(lambda: self.stop())
+
+        CHECK_PERIOD = 0.1
+        wait_time = 0
+        while not self.web3.isConnected():
+            if wait_time > self.CONNECTION_TIMEOUT:
+                if not self.start_node and self.public_nodes:
+                    return self.start(port)
+                self.public_nodes = random_public_nodes()
```
Elfoniok: What is the point assigning something just before throwing exception?

golem/ethereum/node.py
```diff
@@ -105,22 +110,71 @@ def start(self, port=None):
...
+                raise OSError("Cannot connect to geth at {}".format(provider))
```
Elfoniok: So in case of remote geth we are blocking and trying forever, and in case of local geth we are trying once and crashing if this failed?

golem/ethereum/node.py
mfranciszkiewicz: We reset the state in case someone wants to retry starting the node.

golem/ethereum/node.py
mfranciszkiewicz: We're not blocking and trying forever. `public_nodes` is a shallow copy of `random_public_nodes()`, and each node is popped from the list on connection.

golem/ethereum/node.py
```diff
@@ -26,6 +27,18 @@
 log = logging.getLogger('golem.ethereum')
 
 
+PUBLIC_RPC_NODES = [
+    'http://188.165.227.180:55555',
```
chfast: Also add the node to https://stats.rinkeby.io by `--ethstats='Golem Name:Respect my authoritah!@stats.rinkeby.io'`.
There is also one running on pebble on the same port 55555.
I also disabled web3 module of RPC.

mfranciszkiewicz: Will do

badb (approved):

Issue #1380, opened 2017-09-14
Run golem nodes in sub-network


ggruszczynski:
Run golem with additional console argument to restrict network size.
With subnetwork activated, nodes will be allowed to connect only inside it.
Very useful for tests and (semi)private networks.

`$ golemapp --subnetworkID=123
`

then tasksession.py and peersession.py will accept this value:

P2P_PROTOCOL_ID = subnetworkID
TASK_PROTOCOL_ID = subnetworkID

ggruszczynski:
For devp2p -> Protocol id devp2p #1433

ggruszczynski:
Implemented in https://github.com/golemfactory/golem/pull/1526

PR #1526, opened 2017-10-19
Protovol_id_v2

ggruszczynski:
Solves:
Run golem nodes in sub-network #1380

Signed-off-by: ggruszczynski <ggruszczynski@golem.network>


golem/core/variables.py
```diff
@@ -24,22 +24,41 @@
...
-# SESSION VARIABLES #
-#####################
+
+
+###############
+# PROTOCOL ID #
+###############
+class PROTOCOL_ID(object):
+    """
+    https://docs.python.org/2/faq/programming.html#how-do-i-share-global-variables-across-modules
```
etam: `docs.python.org/3`

golem/core/variables.py
```diff
@@ -24,22 +24,41 @@
...
+    https://bytes.com/topic/python/answers/19859-accessing-updating-global-variables-among-several-modules
```
etam: Actually, this might be not needed.

If everywhere there was
```python
import golem.core.variables as vars

[...]

vars.P2P_PROTOCOL_ID
```
then it would work fine.

But
```python
from golem.core.variables import PROTOCOL_ID
```
creates a local copy of the variable, which is surprising.

Well, it wouldn't be surprising if anyone would care about what pylint and flake8 have to say about our code.

ggruszczynski: That's right, however I decided to introduce a class just in case that someone could use it the other way around  ;-)

etam:
One more thing: `PROTOCOL_ID.P2P_ID` there's too much `_ID`. It'd be better `PROTOCOL.P2P_ID` or `PROTOCOL_ID.P2P`.

tests/test_golemapp.py
```diff
@@ -3,20 +3,29 @@
...
+@ci_skip
+class TestGolemApp(TempDirFixture, PEP8MixIn):
+    PEP8_FILES = [
+        "golemapp.py",
+    ]
+    def setUp(self):
+        super(TestGolemApp, self).setUp()
+
+    def tearDown(self):
+        super(TestGolemApp, self).tearDown()
```
etam:
```
W: tests/test_golemapp.py:17:4: Useless super delegation in method 'setUp' (useless-super-delegation)
W: tests/test_golemapp.py:20:4: Useless super delegation in method 'tearDown' (useless-super-delegation)
```

Are those methods needed?

etam (changes requested):

etam (approved):

Elfoniok (approved):

PR #1527, opened 2017-10-19
Coincurve-based crypto

mfranciszkiewicz:
Related issue: #1457


badb (approved):

golem/core/crypto.py
```diff
@@ -0,0 +1,308 @@
...
+                    bitcoin.get_privkey_format(self.raw_privkey)  # failed for some keys
+                valid_priv_key = True
+            except AssertionError:
+                valid_priv_key = False
+            if len(self.raw_pubkey) == 64 and valid_priv_key:
+                break
+            elif raw_privkey or raw_pubkey:
+                raise Exception('invalid priv or pubkey')
+
+        assert len(self.raw_pubkey) == 64
```
jiivan, line 82: `assert` shouldn't be used outside of testing/debugging code. It won't work in optimized mode. See: https://docs.python.org/3/library/constants.html#__debug__

golem/core/crypto.py
```diff
@@ -0,0 +1,308 @@
...
+
+    @property
+    def raw_pubkey(self):
+        if self.pubkey_x and self.pubkey_y:
+            return str_to_bytes(self.pubkey_x + self.pubkey_y)
```
jiivan, line 87: Any docstring here? In what scenario `self.pubkey_x and self.pubkey_y` returns False?

PR #1531, opened 2017-10-20
Improve p2p handshake protocol

Krigpl:
After connection server sends MessageHello to the client, then client responds with MessageHello and RandVal|Challange, then server responds with RandVal|Challange. If everything goes right peers are verified after that process.


jiivan (approved):

golem/core/variables.py
```diff
@@ -44,7 +44,7 @@ class PROTOCOL_CONST(object):
     https://docs.python.org/3/faq/programming.html#how-do-i-share-global-variables-across-modules #noqa
     https://bytes.com/topic/python/answers/19859-accessing-updating-global-variables-among-several-modules #noqa
     """
-    P2P_ID = 17
+    P2P_ID = 18
     TASK_ID = 18
```
jiivan, line 48: It just occured to me that it's a great opportunity to write
```python
P2P_ID = 18
TASK_ID = P2P_ID
```
It'll be a good first step of #1692

mfranciszkiewicz (approved):

golem/core/variables.py
Krigpl: It is, but I'm not gonna change it in this PR since this is quite unrelated.

PR #1535, opened 2017-10-22
add config scripts for a windows buildbot worker

pstiasny:


scripts/ci/requirements.txt
```diff
@@ -0,0 +1,3 @@
+git+https://github.com/ansible/ansible.git@60e98a363548454e03bedd09636e236c07b1666c#egg=ansible
```
etam: This git id is meaningless. I don't know if or when it should be updated and why. It would be better if it was a tag or branch.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    - name: visualstudio2017-workload-nativedesktop 
+      params: "--includeOptional"
+    - name: docker-toolbox
+      params: '/TASKS="modifypath"'
+
+- name: install pypiwin32
+  win_shell: py -2 -m pip install pypiwin32
+  args:
+    creates: C:\Python27\Scripts\pywin32_postinstall.py
+  notify: pywin32 installed
```
etam, line 23: Handlers are run at the end of play. And this looks like something that should be run before next commands (or am I wrong? IDK)

If that's the case, then add `register: pywin32_installed` and change the handler as a next task with `when: pywin32_installed|changed`.

etam (changes requested):

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
maaktweluit: good suggestion!

Another option would be to call `flush_handlers` if it is better to keep it as handler

scripts/ci/site.yml
```diff
@@ -23,3 +23,19 @@
...
+  pre_tasks:
+    - name: get windows user password
+      local_action:
+        module: ec2_win_password
+        instance_id: "{{hostvars[inventory_hostname]['ec2_id']}}"
+        region: us-east-1
+        key_file: "~/.ssh/{{hostvars[inventory_hostname]['ec2_key_name']}}"
+      register: win_password
+    - set_fact:
+        ansible_password: "{{win_password['win_password']}}"
```
maaktweluit: does this automatically apply `no_log` for hiding the password?

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+      C:\BuildbotWorker 
+      {{hostvars[groups['master'][0]]['ec2_private_ip_address']}}
+      windows_server_2016 
+      {{site_settings.worker_pass}}
+  args:
+    creates: C:\BuildbotWorker
+
+- name: download resources for the build tasks
+  win_get_url:
+    url: 'https://s3.amazonaws.com/golem-buildbot-build-resources/pywin32-221-cp36-cp36m-win_amd64.whl'
```
maaktweluit: why is this loaded from a private source? is there no public trusted source available?

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
```diff
@@ -0,0 +1,45 @@
...
+    params: "{{ item.params|default('') }}"
+  with_items:
+    - name: python2
+    - name: python3
+    - name: git
+      params: '"/GitAndUnixToolsOnPath"'
+    - name: visualstudio2017community
+    - name: visualstudio2017-workload-nativedesktop 
+      params: "--includeOptional"
+    - name: docker-toolbox
```
maaktweluit: in all my tests there was no need for docker tool-box, yet.
Did not get it to work on AWS due to VirtualBox limitations.

maaktweluit (changes requested):
Only the password comment is important.

LGTM!

Will you still add the service or is this something i can pick up?

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
pstiasny: Yes, unfortunately VirtualBox won't run on AWS. It will be necessary to move to a different platform if we want to test with it.
I'll remove the package.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
pstiasny: This is mirrored from http://www.lfd.uci.edu/~gohlke/pythonlibs/#pywin32. That package was recommended by @mfranciszkiewicz IIRC.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
pstiasny: This is not required for any of the current commands, although it will be required before initializing a windows service.

pstiasny:
Thanks for your comments. I pushed changes to address them.

pstiasny:
@maaktweluit Feel free to take over for the windows service.

In addition to the post I linked in the issue, if you haven't already, you might want to look into [the ansible user right module](https://docs.ansible.com/ansible/devel/win_user_right_module.html) for setting "SeServiceLogonRight" ("log on as a service", required by buildbot).

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
maaktweluit: Discussed with @mfranciszkiewicz and we can use the public source too:
https://sourceforge.net/projects/pywin32/files/pywin32/Build%20221/pywin32-221.win-amd64-py3.6.exe

wget works for me ( must follow redirects )

maaktweluit (changes requested):
LGTM!

While testing i got quite some timeouts, to solve them i've added:


```
  set_fact:
    ansible_winrm_read_timeout_sec: 60
```
and 
```
- win_chocolatey:
    timeout: 5400
```

did you have those too or is it my local connection?

Besides this and the small q its approved!

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
pstiasny: Are you sure about this? I feel having own mirror would be more reliable than sourceforge.

pstiasny:
@maaktweluit I had no problems with timeouts, but feel free to tweak to whatever works for you. Although wouldn't it be more appropriate to put `ansible_winrm_read_timeout_sec` in group_vars instead of using set_fact?

maaktweluit (approved):

maaktweluit:
good one! will add those when testing from a clean machine, when running the second time it goes through fine.

Will discuss if we trust Sourceforge enough to use the "official" project URL, for now the internal one is fine, just for my own AWS it will not work.

pstiasny:
@maaktweluit the file is publicly available, it should work for you just fine (unless you need a different version).

maaktweluit:
My bad! i assumed it was internal only :$ then we can leave it for sure!

maaktweluit:
@etam are all your requested changes fixed? just finished the service and want to merge this PR before making a new PR.

etam:
@maaktweluit My comment about git id in requirements is without any answear

maaktweluit:
@etam Thanks!

@pstiasny Can you look at this last comment? or want me to include it in my service PR?

pstiasny:
@etam, I addressed that issue in adf28e4029863897dbc082b2ace2e345dc99eac4

etam (approved):

PR #1537, opened 2017-10-23
Improvements to test_peerkeeper.py

### LINKED ISSUE #1436: Make sure that p2p2 behaves as expected
* [ ] number of peers send as a reponse to send_peers should be constant (and probably less than dozen) 
* [ ] peer send as a reponse to send_peers shold be known closest neighbours of node that was asking
* [ ] make sure that old nodes are disconnected after some time
* [ ] make sure that nodes with "better" addresses, ie. closest in xor distance are prefered

There should be automate tests created for those points and they should be fixed in case of problems.

Krigpl:
- It checks pep8 for peerkeeper.py
- Checks whether returned neighbours are the closest
- Checks whether old peers are being replaced

(Also pep8 changes to peerkeeper.py)

Resolves #1436

### DIFF
--- a/golem/network/p2p/peerkeeper.py
+++ b/golem/network/p2p/peerkeeper.py
@@ -4,6 +4,7 @@ import random
 import operator
 from collections import deque
 import math
+import itertools
 
 logger = logging.getLogger("golem.network.p2p.peerkeeper")
 
@@ -171,16 +172,12 @@ class PeerKeeper(object):
         if not alpha:
             alpha = self.concurrency
 
-        neigh = []
-        for bucket in self.buckets_by_id_distance(key_num):
-            if len(neigh) == alpha:
-                break
-            for peer in bucket.peers_by_id_distance(key_num):
-                if int(peer.key, 16) != key_num:
-                    neigh.append(peer)
-                    if len(neigh) == alpha:
-                        break
-        return neigh
+        def gen_neigh():
+            for bucket in self.buckets_by_id_distance(key_num):
+                for peer in bucket.peers_by_id_distance(key_num):
+                    if int(peer.key, 16) != key_num:
+                        yield peer
+        return list(itertools.islice(gen_neigh(), alpha))
 
     def buckets_by_id_distance(self, key_num):
         """
--- a/golem/network/p2p/peerkeeper.py
+++ b/golem/network/p2p/peerkeeper.py
@@ -176,12 +176,14 @@ class PeerKeeper(object):
 
         neigh = []
         for bucket in self.buckets_by_id_distance(key_num):
+            if len(neigh) == alpha:
+                break
             for peer in bucket.peers_by_id_distance(key_num):
                 if int(peer.key, 16) != key_num:
                     neigh.append(peer)
-                    if len(neigh) == alpha * 2:
+                    if len(neigh) == alpha:
                         break
-        return sorted(neigh, key=lambda p: node_id_distance(p, key_num))[:alpha]
+        return neigh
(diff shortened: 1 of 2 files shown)


mfranciszkiewicz (approved):

golem/network/p2p/peerkeeper.py
```diff
@@ -157,21 +173,24 @@ def neighbours(self, key_num, alpha=None):
...
+            if len(neigh) == alpha:
+                break
             for peer in bucket.peers_by_id_distance(key_num):
                 if int(peer.key, 16) != key_num:
                     neigh.append(peer)
-                    if len(neigh) == alpha * 2:
+                    if len(neigh) == alpha:
                         break
-        return sorted(neigh, key=lambda p: node_id_distance(p, key_num))[:alpha]
+        return neigh
```
etam: What do you think about this?
```python
return itertools.islice(
    (peer for peer in bucket.peers_by_id_distance(key_num) if int(peer.key, 16) != key_num
          for bucket in self.buckets_by_id_distance(key_num))
    , alpha)
```

tests/golem/network/p2p/test_peerkeeper.py
```diff
@@ -35,8 +46,8 @@ def test_neighbours(self):
 
         for k in keys:
             peer = MockPeer(k)
-            peers.add(peer)
-            self.peer_keeper.add_peer(peer)
+            if self.peer_keeper.add_peer(peer) is None:
```
etam, line 49: `if not self.peer_keeper.add_peer(peer):`

tests/golem/network/p2p/test_peerkeeper.py
```diff
@@ -47,21 +58,50 @@ def test_neighbours(self):
...
+        not_added_peer = None
+        peer_to_remove = None
+
+        while not_added_peer is None:
+            k = random_key(self.n_bytes)
+            if k == self.key:
+                continue
+            peer = MockPeer(k)
+            peer_to_remove = self.peer_keeper.add_peer(peer)
+            if peer_to_remove is not None:
```
etam, line 84: `if peer_to_remove:`

etam (changes requested):

golem/network/p2p/peerkeeper.py
Krigpl: More compact, much less legible, less efficient (alpha is usually small). Imo not worth it.

tests/golem/network/p2p/test_peerkeeper.py
Krigpl: This is Python and I feel safer checking against `None` explicitly

tests/golem/network/p2p/test_peerkeeper.py
Krigpl: ditto

golem/network/p2p/peerkeeper.py
etam: > More compact

True

> much less legible

Disputable

For me checking twice `len(neigh) == alpha` makes it not much legible either.

> less efficient

Questionable

What about this?
```python
def neigh():
    for bucket in self.buckets_by_id_distance(key_num):
        for peer in bucket.peers_by_id_distance(key_num):
            if int(peer.key, 16) != key_num:
                yield peer
return list(itertools.islice(neigh(), alpha))

maaktweluit:
@etam are the changes/replies good to approve?

etam (approved):

PR #1538, opened 2017-10-23
DataBuffer cleanup

### LINKED ISSUE #1504: Message buffer - better management & optimization
P2P2 is using this databuffer for accepting messages. https://github.com/golemfactory/golem/blob/develop/golem/core/databuffer.py
Propose better implementation of this class. Especially take into account that we:
* Don't want to show warnings if the receive part of the message and hit the current one batch limit
* Would not want to accept too long messages and will disconnect if someone tries to send something that hit the MAX_MESSAGE_SIZE limits. 

Consult @mfranciszkiewicz for further details.

Krigpl:
Resolves #1504
In particular
- doesn't show warning messages on short data
- removes DataBuffer from ResourceManager since it wasn't used there
- DataBuffer doesn't have buffer size limit now, the limit is checked by the client (currently only tcpnetwork)
- tcpnetwork closes connection on too long messages
- test for the above

### DIFF
--- a/golem/core/databuffer.py
+++ b/golem/core/databuffer.py
@@ -3,8 +3,6 @@ from golem.core.common import to_unicode
 
 from .variables import LONG_STANDARD_SIZE
 
-MAX_BUFFER_SIZE = 2 * 1024 * 1024
-
 
 class DataBuffer:
     """ Data buffer that helps with network communication. """
@@ -23,17 +21,11 @@ class DataBuffer:
         self.buffered_data = b"".join([self.buffered_data, str_num_rep])
         return str_num_rep
 
-    def append_string(self, data, check_size=True, overflow_prefix=None):
+    def append_string(self, data):
         """ Append given string to data buffer
-        :param check_size: keep buffer size below MAX_BUFFER_SIZE
-        :param overflow_prefix: string to prepend on overflow
         :param str data: string to append
         """
-        new_size = self.data_size() + len(data)
-        if check_size and new_size > MAX_BUFFER_SIZE:
-            self.buffered_data = b"".join([overflow_prefix or '', data])
-        else:
-            self.buffered_data = b"".join([self.buffered_data, data])
+        self.buffered_data = b"".join([self.buffered_data, data])
 
     def data_size(self):
         """ Return size of data in buffer
@@ -114,8 +106,8 @@ class DataBuffer:
         """ Append length of a given data and then given data to the buffer
         :param str data: data to append
         """
-        prefix = self.append_ulong(len(data))
-        self.append_string(data, overflow_prefix=prefix)
+        self.append_ulong(len(data))
+        self.append_string(data)
 
     def clear_buffer(self):
         """ Remove all data from the buffer """
--- a/golem/core/databuffer.py
+++ b/golem/core/databuffer.py
@@ -37,11 +37,10 @@ class DataBuffer:
     def peek_ulong(self):
         """
         Check long number that is located at the beginning of this data buffer
-        :return long: number at the beginning of the buffer
+        :return (long|None): number at the beginning of the buffer if it's there
         """
         if len(self.buffered_data) < LONG_STANDARD_SIZE:
-            raise ValueError(
-                "buffer_data is shorter than {}".format(LONG_STANDARD_SIZE))
+            return None
 
         (ret_val,) = \
             struct.unpack("!L", self.buffered_data[0:LONG_STANDARD_SIZE])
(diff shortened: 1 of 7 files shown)


etam:
Please rebase onto current 'develop' branch, so the tests could successfully run.

golem/core/databuffer.py
```diff
@@ -23,17 +21,11 @@ def append_ulong(self, num):
         self.buffered_data = b"".join([self.buffered_data, str_num_rep])
         return str_num_rep
 
-    def append_string(self, data, check_size=True, overflow_prefix=None):
+    def append_string(self, data):
```
etam: `data` is not `str`. It's `bytes`

golem/core/databuffer.py
```diff
@@ -23,17 +21,11 @@ def append_ulong(self, num):
...
-        :param check_size: keep buffer size below MAX_BUFFER_SIZE
-        :param overflow_prefix: string to prepend on overflow
         :param str data: string to append
         """
-        new_size = self.data_size() + len(data)
-        if check_size and new_size > MAX_BUFFER_SIZE:
-            self.buffered_data = b"".join([overflow_prefix or '', data])
-        else:
-            self.buffered_data = b"".join([self.buffered_data, data])
+        self.buffered_data = b"".join([self.buffered_data, data])
```
etam: `self.buffered_data += data`

golem/core/databuffer.py
etam: It should be updated in whole class, not just here.

etam (changes requested):

golem/network/transport/tcpnetwork.py
```diff
@@ -491,28 +493,28 @@ def _prepare_msg_to_send(self, msg):
...
             mess = self._data_to_messages()
 
-        # Interpret messages
-        if mess:
-            for m in mess:
-                self.session.interpret(m)
-        elif data:
-            logger.debug("Deserialization of messages from {}:{} failed, maybe it's still "
-                        "too short?".format(self.session.address, self.session.port))
+        if self.db.data_size() > MAX_MESSAGE_SIZE:
```
mfranciszkiewicz: You should probably check this before calling `self._data_to_messages()`

Krigpl: I dunno, if `MAX_MESSAGE_SIZE` means we want to reject massages longer than this then it can happen that we will receive half of message of length `3/4*MAX` then the other half + another message of length `1/2*MAX`. Both messages are under `MAX` but total number of bytes in the buffer is larger than `MAX`.
We can of course change it so we won't allow buffer to be larger than some maximum size, but current approach is fine as well imo.

golem/core/databuffer.py
```diff
@@ -20,20 +18,14 @@ def append_ulong(self, num):
...
         return str_num_rep
 
-    def append_string(self, data, check_size=True, overflow_prefix=None):
-        """ Append given string to data buffer
-        :param check_size: keep buffer size below MAX_BUFFER_SIZE
-        :param overflow_prefix: string to prepend on overflow
-        :param str data: string to append
+    def append_bytes(self, data):
+        """ Append given bytes to data buffer
+        :param str data: bytes to append
```
etam: `:param bytes data: ...`

etam (changes requested):

golem/core/databuffer.py
etam: In whole class, not just here, ofc.

etam (approved):

golem/network/transport/tcpnetwork.py
mfranciszkiewicz: If we receive a message of size `MAX + 1`, we append it to the buffer and read it in the next step. The buffer is now empty.

Krigpl: Good point, in that case probably peeking `len` and checking against `MAX` is the way to go

tests/golem/network/transport/test_tcpnetwork.py
```diff
@@ -204,8 +205,7 @@ def test_dataReceived(self):
         protocol.opened = True
         self.assertIsNone(protocol.dataReceived(data))
         protocol.session = MagicMock()
```
maaktweluit, line 207: maybe update to `unittest.mock` now this file is touched.

maaktweluit (approved):
LGTM!

One small cleanup possible replacing `mock` with native `unittest.mock`

maaktweluit:
Does this resolve #1504 ? in this case its better to change the initial comment to "resolves" over "addresses" since github will auto-close

tests/golem/network/transport/test_tcpnetwork.py
Krigpl: There are 14 instances of `MagickMock` in this file, I'd rather not add noise in this commit by changing them

Krigpl:
> Does this resolve #1504 ? in this case its better to change the initial comment to "resolves" over "addresses" since github will auto-close

TIL

PR #1546, opened 2017-10-24
Cleanup taskcollector building on Linux

### LINKED ISSUE #941: Building taskcollector on Linux does not work
These linker flags are not correct for Linux: `LDFLAGS=-static -L /usr/lib/libfreeimage.so -lfreeimage`.
The /usr/lib/libfreeimage.so path does not seem to be right in general. And you cannot link the static library.

etam:
Fixes #941

### DIFF
--- a/Installer/Installer_Linux/install.sh
+++ b/Installer/Installer_Linux/install.sh
@@ -119,7 +119,7 @@ function install_dependencies()
     declare -a packages=( openssl pkg-config libjpeg-dev libopenexr-dev \
                libssl-dev autoconf libgmp-dev libtool libffi-dev \
                libgtk2.0-0 libxss1 libgconf-2-4 libnss3 libasound2 \
-               ethereum )
+               ethereum libfreeimage3 )
 
     if [[ ${INSTALL_GETH} -eq 1 ]]; then
         info_msg "INSTALLING GETH"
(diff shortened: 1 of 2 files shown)


Elfoniok (approved):

Elfoniok:
Task collector will not work, on Linux on fresh machine without freeimage.

etam (commented):
why can't I delete a comment?

Installer/Installer_Linux/install.sh
```diff
@@ -168,7 +168,7 @@ function install_dependencies()
         wget -qO- ${hyperg} > ${hyperg_pack}
         [[ -d $HOME/hyperg ]] && rm -rf $HOME/hyperg
         tar -xvf ${hyperg_pack} >/dev/null
-        mv hyperg $HOME/
+        mv hyperg $HOME/ 2>dev/null
```
etam: 1. it's not related to building taskcollector
2. maybe something like `[[ "$PWD" != "$HOME" ]] && mv hyperg $HOME/`?

Elfoniok: 1. No its not but its not worth PR for such a change.
2. Maybe good idea, but don't want to change behavior just before release.

etam: But this might also swallow other error messages. I think we should leave as it was before, or do it properly.

(And I think attaching small changes to other unrelated PRs is not a good idea)

maaktweluit (approved):
LGTM!

Forced a new build on appveyor through the api, should be green soon and ready to merge

PR #1547, opened 2017-10-24
use python3 in linux installer

### LINKED ISSUE #1279: Supported Ubuntu version?
The install script notes that it is for Ubuntu and Mint, but doesn't mention any version numbers. I'm getting a bunch of error messages running the install script on Ubuntu Server 17.04 and didn't know if I should be using a different version.

This is the output so far. It has been sitting at this point for a while and looks like it may have locked up. The last line that fails on the symbolic link is because I tried running this script once already with the same results.

------@------:~$ ./install.sh
./install.sh: line 15: python: command not found
./install.sh: line 15: python: command not found
./install.sh: line 15: python: command not found
Docker not found. Do you want to install it? (y/n) y
Geth not found. Do you want to install it? (y/n) y

INSTALLING GOLEM DEPENDENCIES

[sudo] password for -----:

INSTALLING GETH

gpg: keybox '/tmp/tmp420gyk7t/pubring.gpg' created
gpg: /tmp/tmp420gyk7t/trustdb.gpg: trustdb created
gpg: key 1C52189C923F6CA9: public key "Launchpad PPA for Ethereum" imported
gpg: Total number processed: 1
gpg:               imported: 1
Warning: apt-key output should not be parsed (stdout is not a terminal)

INSTALLING DOCKER

Cannot install docker. Install it manually: https://docs.docker.com/engine/installation/

Installing HyperG

wget: missing URL
Usage: wget [OPTION]... [URL]...

Try `wget --help' for more options.
tar: This does not look like a tar archive

gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now
mv: cannot stat 'hyperg': No such file or directory
ln: failed to create symbolic link '/usr/local/bin/hyperg': File exists

etam:
Fresh vagrant image of ubuntu xenial (16.04) does not have 'python' command, only 'python3'.
We've migrated golem code to python3, so we should also use it here.

Fixes #1279


etam:
https://wiki.ubuntu.com/XenialXerus/ReleaseNotes/

> Python2 is not installed anymore by default on the server, cloud and the touch images, long live Python3!

Elfoniok:
Ok even though python is on desktop machines python3 is by default as well, so probably a good idea.

Elfoniok (approved):

maaktweluit (approved):
LGTM!

PR #1550, opened 2017-10-24
Start geth if cannot connect to public RPC nodes

mfranciszkiewicz:
- spawn a rinkeby geth node if connection to public RPC failed
- catch `AssertionError`s in `web3.isConnected()` caused by nodes without required APIs exposed


badb (approved):

badb:
Will be merged from b0.9

PR #1558, opened 2017-10-24
Several bugfixes

badb:


golem/client.py
```diff
@@ -656,23 +657,27 @@ def get_task_stats(self) -> Dict[str, int]:
...
+        return self.get_task_computer_stat('tasks_with_errors')
 
     def get_payment_address(self):
         address = self.transaction_system.get_payment_address()
         return str(address) if address else None
 
+    def get_task_computer_stat(self, name):
+        if self.task_server and self.task_server.task_computer:
+            return self.task_server.task_computer.stats.get_stats(name)
+        return 0
```
jiivan, line 679: For consistency sake should it return None instead of 0 as `get_task_count` and `get_computed_task_count` etc?

golem/client.py
```diff
@@ -619,7 +619,8 @@ def get_task_server_port(self):
         return self.task_server.cur_port
 
     def get_task_count(self):
-        return len(self.task_server.task_keeper.get_all_tasks())
+        if self.task_server and self.task_server.task_keeper:
+            return len(self.task_server.task_keeper.get_all_tasks())
```
Krigpl, line 624: `else return 0` ? Or is `None` expected?

golem/client.py
```diff
@@ -656,23 +657,27 @@ def get_task_stats(self) -> Dict[str, int]:
         }
 
     def get_supported_task_count(self) -> int:
-        return len(self.task_server.task_keeper.supported_tasks)
+        if self.task_server and self.task_server.task_keeper:
+            return len(self.task_server.task_keeper.supported_tasks)
```
Krigpl, line 662: `else return 0`, to comply with type annotation

golem/client.py
Krigpl: I'd say others should return 0 as well, but I'm not sure whether that's the intention here

Krigpl (approved):
lgtm, my nits aren't crucial

maaktweluit:
yay for semver, it saved my OCD ^^

PR #1563, opened 2017-10-25
Simple test for APP_VERSION

Krigpl:


Elfoniok (approved):

tests/golem/core/test_variables.py
```diff
@@ -0,0 +1,12 @@
...
+from golem.core.variables import APP_VERSION
+
+
+class TestVariables(TestCase):
+
+    def test_app_version(self):
+        v = APP_VERSION.split('.')
+        assert len(v) == 3
+        v = [str(int(i)) for i in v]
+        assert '.'.join(v) == APP_VERSION
```
maaktweluit, line 12: maybe also assert if the semver module is happy with the value?

maaktweluit (approved):
Good as is, but small addition possible

maaktweluit:
Double approved!

PR #1566, opened 2017-10-25
Hardening logging startup

jiivan:
Failure to create log file would crash golem startup. In this PR a warning is issued to STDERR instead. Additional check added to verify whether datadir is writable.

Also this PR fixes logs dir location issue. It will try to determine datadir whether it was given as command line option `-d|--datadir` or not.


golem/core/common.py
```diff
@@ -168,23 +168,28 @@ def config_logging(suffix='', datadir=None):
...
 
-    logdir_path = Path('logs')
-    if datadir is not None:
-        logdir_path = Path(datadir) / logdir_path
-        datadir += '/'
-    else:
-        datadir = ''
-    if not logdir_path.exists():
-        logdir_path.mkdir(parents=True)
+    if datadir is None:
```
etam, line 172: `if not datadir:`

golem/core/common.py
```diff
@@ -168,23 +168,28 @@ def config_logging(suffix='', datadir=None):
...
+        from golem.core.simpleenv import get_local_datadir
```
etam: Put all imports at the top of the file.

etam (changes requested):

golem/core/common.py
jiivan: It will also catch `''` witch is not desired.

etam (approved):

PR #1568, opened 2017-10-26
Handshake timeout

jiivan:
Handle timeout during handshake.


golem/network/p2p/p2pservice.py
```diff
@@ -272,6 +273,8 @@ def add_peer(self, key_id, peer):
         with self._peer_lock:
             self.peers[key_id] = peer
             self.peer_order.append(key_id)
+        # Timeouts of this session/peer will be hanled in sync_network()
+        self.pending_sessions.remove(peer)
```
ggruszczynski: shouldn't `self.pending_sessions.remove(peer)` be inside the with block?

Elfoniok (approved):
Lets try it out :)

golem/network/p2p/p2pservice.py
```diff
@@ -215,6 +215,7 @@ def sync_network(self):
         """Get information about new tasks and new peers in the network.
```
ggruszczynski, line 215: there should be `def sync_network(self, timeout=1.0):` to match base method signature.

ggruszczynski (approved):

PR #1569, opened 2017-10-26
 Updated payment confirmation msg

ggruszczynski:
1 We don't care what is the value of reward in the message since byzantine node can send whatever it likes.
2 Updated msg info-logs

Signed-off-by: ggruszczynski <ggruszczynski@golem.network>


jiivan (approved):
LGTM

Issue #1448, opened 2017-10-06
Unsupported task stats


badb:
Main purpose:
Save persistence Golem statistics about unsupported stats and make it available to user via API

Introduction:
In Golem there are two main roles, that user can have:
Requestor - request task computation, send requests to the network
Provider - computing node, share its computer power in the network. 
Golem node’s that belong to providers monitor p2p network to get information about new tasks. Those tasks are save in [TaskHeaderKeeper` instance](https://github.com/golemfactory/golem/blob/develop/golem/task/taskkeeper.py#L164) with information whether they are “supported” or not (and why not). 
We would like to display a report created from this information to the users. Using this report the provider would be able to find out if his Golem instance was not computing anything because he set too restrictive parameters (and which one) or because there are no tasks in the network. Without this information the user must check all the settings if they have default values. 

The task: 
Right now there’s an RPC method, that returns to user information how many task are currently in the network (visible ot this node) and how many of them are supported. 
We want to have additional method that returns information what reasons for not supporting task were observed. User should be able to call this method from CLI:
Basic reasons are defined [here](https://github.com/golemfactory/golem/blob/develop/golem/environments/environment.py#L40) in `UnsupportedReason`. 
We want to add reasons that are checked in function [request_task](https://github.com/golemfactory/golem/blob/develop/golem/task/taskserver.py#L121) 
For each category we would like to display information how many tasks were not supported because of this reason (at current moment and in history).
The most popular reason should be displayed at the top. 
For values for which it makes sense the more specific value should be saved, eg. for price there should be displayed the average price per task observed in the network, etc.

PR #1574, opened 2017-10-26
Unsupported task stats #1448.

RDProjekt:
A new command "tasks unsupport" has been added to client that shows statistics for unsupported tasks that the node currently sees in the network. When this command is followed with a number of days,
the statistics are shown for all tasks seen by this node during the last number of calendar days. The statistics are persistent.

### DIFF
--- a/golem/task/taskserver.py
+++ b/golem/task/taskserver.py
@@ -17,6 +17,7 @@ from golem.task.benchmarkmanager import BenchmarkManager
 from golem.task.deny import get_deny_set
 from golem.task.taskbase import TaskHeader
 from golem.task.taskconnectionshelper import TaskConnectionsHelper
+from golem.environments.environment import SupportStatus, UnsupportReason
 from .taskcomputer import TaskComputer
 from .taskkeeper import TaskHeaderKeeper
 from .taskmanager import TaskManager
@@ -98,13 +99,17 @@ class TaskServer(PendingConnectionsServer, TaskResourcesMixin):
                  keys_auth,
                  client,
                  use_ipv6=False,
-                 use_docker_machine_manager=True):
+                 use_docker_machine_manager=True,
+                 task_archiver=None):
         self.client = client
         self.keys_auth = keys_auth
         self.config_desc = config_desc
 
         self.node = node
-        self.task_keeper = TaskHeaderKeeper(client.environments_manager, min_price=config_desc.min_price)
+        self.task_archiver = task_archiver
+        self.task_keeper = TaskHeaderKeeper(client.environments_manager,
+                                            min_price=config_desc.min_price,
+                                            task_archiver=task_archiver)
         self.task_manager = TaskManager(config_desc.node_name, self.node, self.keys_auth,
                                         root_path=TaskServer.__get_task_manager_root(client.datadir),
                                         use_distributed_resources=config_desc.use_distributed_resource_management,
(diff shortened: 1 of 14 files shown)


golem/task/taskserver.py
```diff
@@ -192,10 +197,14 @@ def request_task(self):
...
                 performance = 0.0
-            is_requestor_accepted = self.should_accept_requestor(
-                theader.task_owner_key_id)
-            is_price_accepted = self.config_desc.min_price < theader.max_price
-            if is_requestor_accepted and is_price_accepted:
+            supported = self.should_accept_requestor(theader.task_owner_key_id)
+            if self.config_desc.min_price > theader.max_price:
+                supported = supported.join(SupportStatus.err({
+                    UnsupportReason.MAX_PRICE: theader.max_price}))
+            if not supported.is_ok() and self.task_archiver:
```
Krigpl: If task is not supported but there isn't `task_archiver` then we'll fall into `else` branch and add the task which isn't right

golem/task/taskserver.py
```diff
@@ -192,10 +197,14 @@ def request_task(self):
                 performance = env.get_performance()
             else:
                 performance = 0.0
-            is_requestor_accepted = self.should_accept_requestor(
-                theader.task_owner_key_id)
-            is_price_accepted = self.config_desc.min_price < theader.max_price
-            if is_requestor_accepted and is_price_accepted:
+            supported = self.should_accept_requestor(theader.task_owner_key_id)
```
Krigpl, line 201: I'd say blacklisted requestor should have higher priority than max price threshold. So it shouldn't be overridden in the next `if` block

tests/golem/task/test_taskarchiver.py
```diff
@@ -0,0 +1,186 @@
...
+        if not deadline:
+            deadline = timeout_to_deadline(36000)
+        ret = TaskHeader("ABC", str(uuid4()), "10.10.10.10", 10101, "key",
+                         "DEFAULT", max_price=max_price, deadline=deadline,
+                         min_version=min_version)
+        if last_checking:
+            ret.last_checking = last_checking
+        return ret
+
+    def getRow(self, reasonsReport, unsupportReason):
```
Krigpl: `camel_case`

tests/golem/task/test_taskserver.py
```diff
@@ -107,7 +161,7 @@ def test_send_results(self, trust):
         n = Node()
         ka = EllipticalKeysAuth(self.path)
         ts = TaskServer(n, ccd, ka, self.client,
-                        use_docker_machine_manager=False)
+                        use_docker_machine_manager=False, task_archiver=None)
```
Krigpl: No need for `task_achiever=None`, we can just use the default (in the whole file)

golem/task/taskserver.py
RDProjekt: Blacklisted requestor is joined with max price threshold rather than overridden by it. In the resulting unsupported task stats, both unsupport reasons (deny_list, max_price) will have increased counts in such case.

mfranciszkiewicz:
@tomasz-l Could you resolve merge conflicts, please?

Krigpl:
FWIW I resolved the conflicts in #1665

tomasz-l:
OK thanks.

PR #1581, opened 2017-10-27
Task persistance a.k.a. Code for saving task manager state

### LINKED ISSUE #1339: Check code for saving task manager state
Check the code for dumping task manager state & task keeper state, do any necessary refactoring & changes.

jiivan:
fixes: #1339


golem/task/taskmanager.py
```diff
@@ -208,27 +209,28 @@ def remove_dump(self, task_id: str):
...
                 try:
                     task, state = pickle.load(f)
                     self.tasks[task.header.task_id] = task
                     self.tasks_states[task.header.task_id] = state
+
+                    dispatcher.send(
+                        signal='golem.taskmanager',
+                        event='task_status_updated',
+                        task_id=task.header.task_id
+                    )
```
jiivan: I think that moving `dispatcher.send` inside `try` and `with` block is a mistake. File should be closed before dispatching, and try block handles unpickling related errors.

ggruszczynski (approved):

golem/task/taskmanager.py
```diff
@@ -168,57 +171,85 @@ def start_task(self, task_id):
...
+            logger.info("Task %s added" % task.header.task_id)
             self.notice_task_updated(task.header.task_id)
 
-    def dump_task(self, task_id):
-        logger.debug('DUMP TASK')
+    def _dump_filepath(self, task_id):
+        return self.tasks_dir / ('%s.pickle' % (task_id,))
+
+    def dump_task(self, task_id: str) -> None:
+        logger.debug('DUMP TASK CALLED')
```
Elfoniok: Please add task id to logs

golem/task/taskmanager.py
```diff
@@ -72,32 +70,39 @@ def __init__(self, node_name, node, keys_auth, listen_address="",
...
         self.listen_address = listen_address
         self.listen_port = listen_port
 
-        # FIXME Remove this variable and make task persistance obligatory after it is more tested
-        # Remember to also remove it from init params
         self.task_persistence = task_persistence
 
-        self.tasks_dir = Path(tasks_dir)
+        tasks_dir = Path(tasks_dir)
+        self.tasks_dir = tasks_dir / "tmanager"
```
Elfoniok, line 83: Can you get this dir from config?

Elfoniok (changes requested):
Please revert changes that delete task.state updates. There were done very poorly (only some random set instructions were removed) moreover, i think its bad idea to remove state from task. Actually this is task state. Its true tat it looks like those two (task_states dict and state in task) should be unified. But this should be done in separate PR. Probably it would be better to remove dict itself.

Elfoniok (approved):

golem/task/taskmanager.py
etam: Shouldn't this be managed by DirManager? It already does something similar for other directories.

PR #1585, opened 2017-10-27
Removing __make_node_state_snapshot

### LINKED ISSUE #1424: Check if node state snapshot is still used somewhere
`__make_node_state_snapshot` function is changed in devp2p branch. We should check if it's still really used somewhere. If yes, than it should be refactored. If no, than it should be removed.

maaktweluit:
Resolves #1424 

Was not used anywhere so removed it.
My laptop is running a long test so fixed it on github, will check tests and pystyle asap.


Krigpl (commented):
After these changes `NodeStateSnapshot` class is only used in tests, so maybe worth getting rid of it completely

maaktweluit:
@Krigpl thanks, removed it!

jiivan (approved):

Krigpl (approved):

PR #1588, opened 2017-10-27
Elaborated logs about docker issues #1416

### LINKED ISSUE #1416: Elaborated logs about docker issues

jiivan:
A lot changed since the issue was created. Docker no longer uses ProcessMonitor and hyperg uses it in a different way. So only small portion of #1432 remains. Also #1432 was based on discarded `devp2p` branch.

fixes: #1416

### DIFF
--- a/golem/docker/manager.py
+++ b/golem/docker/manager.py
@@ -409,6 +409,18 @@ class DockerManager(DockerConfigManager):
             if not retried:
                 self.command('regenerate_certs', self.docker_machine)
                 return self._set_docker_machine_env(retried=True)
+            typical_solution_s = """It seems there is a  problem with your Docker installation.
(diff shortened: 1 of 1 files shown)


maaktweluit (approved):
LGTM

PR #1590, opened 2017-10-29
misc cleanups to all CI platforms

maaktweluit:


shippable.yml
```diff
@@ -4,13 +4,8 @@ python:
   - 3.5
 
 install:
-  - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
-  - sudo add-apt-repository -y ppa:ethereum/ethereum && sudo apt-get update
-  - sudo apt-get install -y qt5-default qtbase5-dev-tools geth unzip gcc-5 g++-5 libfreeimage3 libfreeimage-dev
-  - pip3 install six
   - pip3 install -r requirements.txt
   - pip3 install -r requirements-lint.txt
-  - make -C apps/rendering/resources/taskcollector
```
etam, line 13: I think that building taskcollector should be tested (it requires libfreeimage-dev)

etam (changes requested):

shippable.yml
Elfoniok: Hmm shipable probably just don't use taskcollector, so it can be skipped?

circle.yml
```diff
@@ -44,14 +45,10 @@ dependencies:
     - /usr/local/bin/ipfs daemon:
         background: true
```
etam, line 46: Is ipfs still needed? AFAIK we're not using it anymore, so this could be deleted

maaktweluit: nope, no longer needed, but removed in #1620 ( also up for review and merge :) )

shippable.yml
maaktweluit: i think linters do not use it at all, the light is green on the PR so "assumed" it was good like this.

Elfoniok (approved):

shippable.yml
etam: Do we have any tests for taskcollector? If not at least we could test if it builds.

etam: Well, on shippable we only run linters, not full tests, so this can be dropped from here.

etam (approved):

PR #1595, opened 2017-10-31
Cleaning up rpc method names

Elfoniok:


Krigpl (approved):

etam:
It's a good practice to rename files in one commit and change them in second. This way git better tracks renames. (In this case I need to use "git diff -M10%" to see what changed in renamed files)

golem/rpc/mapping/rpcmethodnames.py
```diff
@@ -0,0 +1,88 @@
+CORE_METHOD_MAP = dict(
+    get_golem_version=      'golem.version',
```
etam, line 2: Why not aliases, like it was before?

This way it's easier to make mistake. You'll spend hours debugging if there's mismatch in strings. You'll get instant error if object does not exist.

etam: Yep. You removed 'golem.version' from the other file.

etam (changes requested):

jiivan (approved):
Great! Maybe in a next step change meaningless classes in `golem/rpc/mapping/rpceventnames.py` to plain dicts?

etam:
If we have constant set of methods, then using dict is not a good idea. Better use enums, or leave it as is.

golem/rpc/mapping/rpcmethodnames.py
Elfoniok: You will not, there is not point for using those in code, that is why events, are left as aliases, but method names are passed to crossbar ruter only. You don't need them unless you are using debug class, but i guess then you know what are you doing and can use those correctly.

etam (approved):

PR #1615, opened 2017-11-03
Move messages to external module

### LINKED ISSUE #1: Initialize message module
Move code connected with messages (format, parser, objects, etc.) from golem to this repository. Create module and integrate Golem with it.
There should be an interface that exposes two simple methods: parse and format.

jiivan:
fixes: https://github.com/golemfactory/golem-messages/issues/1


tests/golem/network/transport/test_message.py
```diff
@@ -27,9 +24,11 @@ def slots(self):
         raise Exception()
 
 
-class TestMessages(unittest.TestCase, PEP8MixIn):
-    PEP8_FILES = ['golem/network/transport/message.py', ]
+fake_sign = lambda x: b'\000'*65
+fake_decrypt = lambda x: x
```
Elfoniok, line 28: What problem this is solving?

golem/network/transport/tcpnetwork.py
```diff
@@ -488,7 +489,7 @@ def connectionLost(self, reason=connectionDone):
 
     # Protected functions
     def _prepare_msg_to_send(self, msg):
-        ser_msg = msg.serialize()
+        ser_msg = msg.serialize(lambda x: b'\000'*message.Message.SIG_LEN)
```
Elfoniok, line 492: Why this is needed?

jiivan: I made `sign` argument mandatory, as it is always used in golem code. It was omitted only in tests.

tests/golem/network/transport/test_message.py
jiivan: Same as above. This test checks messaging not encryption, but arguments are mandatory.

Elfoniok (approved):

PR #1634, opened 2017-11-08
Clear old dirs

### LINKED ISSUE #1560: Automatically clear old computation related directories
Currently all the data that are used in Golem to generate and verify images are stored on a user hdd in specific directories. It's important for us to debug computation, however they dir size can grow up relatively quickly when Golem is intensly used. 
In the old interface there was a button that allows user to clean the directories. There is still a CLI command that allows to do it. There is no such options in the new GUI and it shouldn't be reintroduce. Instead golem data directories should be cleaned up automatically - data that are older than X should be removed. It's not defined how big should be X, however 3 days seems to be a safe option. 

Remember that this must work on all three OS.

etam:
Fixes #1560 

Also introduces "services" in golem client.

Reading all changes at once might be a bit overwhelming. I suggest reading each commit independently.


tests/golem/test_client.py
```diff
@@ -555,6 +411,166 @@ def test_start_stop(self, load_config, connect_to_network, *_):
...
+        do_work_service._run()
+
+        assert c.p2pservice.ping_peers.called
+        assert log.exception.call_count == 5
+
+
+class TestMonitoringPublisherService(TestWithReactor):
+    @patch('golem.client.log')
+    @patch('golem.client.dispatcher.send')
+    def test_run(self, send, log, *_):
```
maaktweluit: for what arguments is `*_` present?

golem/client.py
```diff
@@ -1122,3 +1086,142 @@ def _unlock_datadir(self):
...
+                 transaction_system: EthereumTransactionSystem):
+        super().__init__(interval_seconds=int(PUBLISH_BALANCE_INTERVAL))
+        self._rpc_publisher = rpc_publisher
+        self._transaction_system = transaction_system
+
+    def _run(self):
+        try:
+            gnt, av_gnt, eth = self._transaction_system.get_balance()
+        except Exception as exc:
+            log.debug('Error retrieving balance: {}'.format(exc))
```
maaktweluit: is the severity of this error so low it should be debug?

maaktweluit (commented):
LGTM!

Too much things i do not know so can not approve.

jiivan (approved):
LGTM

golem/client.py
jiivan: If you want to change this line, pls also consider removing bad `.format()` code.  https://gist.github.com/jiivan/eedb505574ccb2503af48fa2e3043036

etam: Dunno. It's just moved this code from other place.

tests/golem/test_client.py
etam: Probably for ignoring any additional arguments. I just copied how other tests are done. I'll check if it's needed.

jiivan: For mocks.

golem/client.py
etam: Removed `.format()`

tests/golem/test_client.py
etam: Removed `*_`

etam:
ReReview plz

maaktweluit (approved):
LGTM

Issue #1400, opened 2017-09-19
Integrate unittest with buildbots


badb:
Currently we have 7 separate CI tools integrated. Buildbots that produces packages for 3 OS, CircleCI to run unittest on Linux, Travis to run unittest on Mac, AppVeyor to run unittest on Windows and Shippable to run linters. 
If we are able to move unittest on buildbots (and create build only if all tests pass) then we'll significantly simplify CI architecture. 

# Main tasks
- [x] Run unit test on linux POC
- [x] Add automated installer on linux #1455 
- [x] Run unit test on mac POC
- [x] Run unit test on windows POC
- [x] Add automated installer on mac 
- [x] Run windows as service and own user #1636 
- [x] Add automated installer on windows
- [x] Add controlling build, report status
- [x] Deploy
- [x] Write nightly upload build

# Tasks from reviews and findings
- [x] Remove python 2
- [x] Update `hyperg` to `0.2.3`
- [x] remove `ipfs` after #1548 
- [x] decide on hosting
- [x] upload coverage for all builds

maaktweluit:
Two down, one to go:

![screen shot 2017-11-14 at 13 50 56](https://user-images.githubusercontent.com/10008353/32780813-f169d994-c942-11e7-93f5-11dd6e22b163.png)

![screen shot 2017-11-14 at 13 51 08](https://user-images.githubusercontent.com/10008353/32780816-f4a0e4cc-c942-11e7-9506-06e084292920.png)

Next steps before releasing:
- python 2 to python 3
- windows unit test
- grouped build ( optional )
- decide on hosting

Added those to OP

maaktweluit:
From the PR #1455, 3 tasks left to pickup on next update:
- Update `hyperg` to `0.2.3`
- Remove ipfs after #1548 
- upload codecov to s3

Added those to OP

maaktweluit:
Windows blockers solved :D

![screen shot 2017-11-21 at 16 50 02](https://user-images.githubusercontent.com/10008353/33082117-4739a9fa-cedc-11e7-9788-dcc687f76699.png)

Will group all open branches and:
- Remove python2
- Remove ipfs
- Add guiding build
- Check re-build protection to not build all 3 when one fails.

maaktweluit:
Removed IPFS and python2
Also added inno to build installer after binaries

Next steps are:
- guiding builds
- re-build protection
- coverage
- nightly repo uploads.

maaktweluit:
[...]

PR #1636, opened 2017-11-08
Buildbot Windows service

maaktweluit:
Another step as part of #1400 

Added buildbot as a windows service with automatic startup.

- [x] Requires #1535 to be merged first

### DIFF
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -29,6 +29,14 @@ To apply the configuration:
 
     ansible-playbook --key-file=~/.ssh/your_ssh_key -i inventory/ site.yml
 
+This will always fail the first time settings up the service since the user has no acces to run as service. To enable this go to:
+
+* Administrative Tools
+* Local Security Policy
+* Local Policies > User Rights Management > Login as Service
+* Add 'buildbot-worker' to the list and save
+* Run the command above again to setup the service.
+
 
 # Note about HTTPS
 
--- a/scripts/ci/README.md
+++ b/scripts/ci/README.md
@@ -29,15 +29,6 @@ To apply the configuration:
 
     ansible-playbook --key-file=~/.ssh/your_ssh_key -i inventory/ site.yml
 
-This will always fail the first time settings up the service since the user has no acces to run as service. To enable this go to:
-
-* Administrative Tools
-* Local Security Policy
-* Local Policies > User Rights Management > Login as Service
-* Add 'buildbot-worker' to the list and save
-* Run the command above again to setup the service.
-
-
 # Note about HTTPS
 
 TLS certificate handling is not automated at the moment.  You will need
--- a/scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
+++ b/scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
@@ -6,13 +6,14 @@
 - win_chocolatey:
     name: "{{ item.name }}"
     params: "{{ item.params|default('') }}"
+    timeout: 5400
   with_items:
     - name: python2
     - name: python3
     - name: git
       params: '"/GitAndUnixToolsOnPath"'
     - name: visualstudio2017community
-    - name: visualstudio2017-workload-nativedesktop 
+    - name: visualstudio2017-workload-nativedesktop
       params: "--includeOptional"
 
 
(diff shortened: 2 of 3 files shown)


scripts/ci/README.md
```diff
@@ -29,6 +29,14 @@ To apply the configuration:
...
     ansible-playbook --key-file=~/.ssh/your_ssh_key -i inventory/ site.yml
 
+This will always fail the first time settings up the service since the user has no acces to run as service. To enable this go to:
+
+* Administrative Tools
+* Local Security Policy
+* Local Policies > User Rights Management > Login as Service
+* Add 'buildbot-worker' to the list and save
+* Run the command above again to setup the service.
+
```
pstiasny: In https://github.com/golemfactory/golem/pull/1535#issuecomment-339801915 I mentioned it should be possible to set appropriate user right via the [user rights module](https://docs.ansible.com/ansible/devel/win_user_right_module.html#examples). Did you try it?

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
```diff
@@ -21,7 +22,14 @@
   args:
     creates: C:\Python27\Scripts\pywin32_postinstall.py
   notify: pywin32 installed
-  
+
+- name: setup buildbot-worker user
+  win_user:
+    name: buildbot-worker
+    password: "{{site_settings.worker_pass}}"
```
pstiasny: I'm not sure it's a good idea to reuse the worker pass (for the master-worker protocol) as a Windows password. Seems unnecessary and weakens security. It would be best to generate a random one, or at least use a separate variable.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
```diff
@@ -57,17 +57,48 @@
...
+    data: C:\BuildbotWorker
+    type: String
+
+
+- name: start service
   win_service:
-    name: buildbot-worker
-    path: 'C:\Windows\py.exe -2 C:\Python27\Scripts\buildbot-worker.exe start C:\BuildbotWorker'
+    name: BuildBot
+    path: 'C:\Python27\lib\site-packages\win32\PythonService.exe'
```
pstiasny: Would it be possible to omit `path`? Won't `buildbot_worker_windows_service.exe` already take care of configuring that?

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
```diff
@@ -57,17 +57,48 @@
...
   win_get_url:
-    url: 'https://s3.amazonaws.com/golem-buildbot-build-resources/pywin32-221-cp36-cp36m-win_amd64.whl' 
+    url: 'https://s3.amazonaws.com/golem-buildbot-build-resources/pywin32-221-cp36-cp36m-win_amd64.whl'
     dest: 'C:\BuildResources\pywin32-221-cp36-cp36m-win_amd64.whl'
     force: no
 
+- name: check if service is created
+  win_shell: Get-Service -Name BuildBot -ErrorAction SilentlyContinue
+  register: service_exists
+  ignore_errors: yes
```
pstiasny: Looking at `win_service` examples, it looks like it should be possible to use it to query service state without going through the shell:
```
- name: check if a service is installed
  win_service:
    name: service name
  register: service_info
```

pstiasny (changes requested):

scripts/ci/README.md
maaktweluit: great one! forgot about it, will try it right away!

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
maaktweluit: will try this, its only to start the service after configuring, maybe the name alone is enough

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
maaktweluit: fully agree, will generate a password and store it in the local secret file.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
maaktweluit: i do not know, this string comes from `buildbot_worker_service.exe` so hopefully it can start the service without, making it less error prone on changes. will test it

maaktweluit:
@pstiasny please re-review, all comments are fixed and tested.

pstiasny (approved):

maaktweluit:
@badb or @mfranciszkiewicz could you please review this PR?

mfranciszkiewicz (approved):

PR #1642, opened 2017-11-08
first draft on Concent client

### LINKED ISSUE #1627: Concent connector
Create generic component to communicate with Concent. It should contain http client and discovery of Concent service. Address of Concent service should be a part of configuration and it varies depending on environment (devel, test, main). It should expose simple method that takes formatted message as parameter and returns formatted message from response, null if response has no body but has status 200 (OK), exception in other case. The component should be easy to mock.

maaktweluit:
resolves #1627

### DIFF
--- a/golem/network/concent/client.py
+++ b/golem/network/concent/client.py
@@ -18,7 +18,7 @@ class ConcentClient:
 
     def message(self, message):
 
-        logger.error('message')
+        print('message')
         if not self.__can_call_concent():
             raise Exception("5 minute failure grace time")
 
@@ -36,9 +36,9 @@ class ConcentClient:
                 return None
 
         # TODO: only unavailable on certain errors
-        logger.error('call_time')
+        print('call_time')
         cur_time = time.time()
-        logger.error(cur_time)
+        print(cur_time)
 
         self._last_available_check = cur_time
         self._is_available = False
@@ -50,12 +50,12 @@ class ConcentClient:
 
     def __can_call_concent(self):
         if self._last_available_check is None:
-            logger.error('is none')
+            print('is none')
             return True
 
-        logger.error('call_time')
+        print('call_time')
         cur_time = time.time()
-        logger.error(cur_time)
+        print(cur_time)
 
         if self._last_available_check < cur_time - retry_time:
             return True
--- a/golem/network/concent/client.py
+++ b/golem/network/concent/client.py
@@ -18,7 +18,6 @@ class ConcentClient:
 
     def message(self, message):
 
-        print('message')
         if not self.__can_call_concent():
             raise Exception("5 minute failure grace time")
 
@@ -36,11 +35,7 @@ class ConcentClient:
                 return None
 
         # TODO: only unavailable on certain errors
-        print('call_time')
-        cur_time = time.time()
-        print(cur_time)
-
-        self._last_available_check = cur_time
+        self._last_available_check = time.time()
         self._is_available = False
 
         raise Exception("Failed to call concent")
@@ -50,14 +45,9 @@ class ConcentClient:
 
     def __can_call_concent(self):
         if self._last_available_check is None:
-            print('is none')
             return True
 
-        print('call_time')
-        cur_time = time.time()
-        print(cur_time)
-
-        if self._last_available_check < cur_time - retry_time:
+        if self._last_available_check < time.time() - retry_time:
             return True
 
         return self._is_available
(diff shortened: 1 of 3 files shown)


golem/network/concent/client.py
```diff
@@ -0,0 +1,96 @@
...
+        :rtype: String|None
+        """
+        if not self.__can_call_concent():
+            raise ConcentGraceException("5 minute failure grace time")
+
+        response = None
+        try:
+            response = requests.post(CONCENT_URL, data=message)
+        except requests.exceptions.RequestException as e:
+            if e.response:
```
lukasz-glen: the exception should be logged anyway, to not loose information regardless of the code below

lukasz-glen (changes requested):

golem/network/concent/client.py
maaktweluit: also logged exception thrown here, please check again

lukasz-glen (approved):

golem/network/concent/client.py
```diff
@@ -0,0 +1,91 @@
+import logging
+import time
+
+import requests
```
jiivan, line 4: A tiny remark here. I don't that keeping stdlib and custom libs import separated is necessary. It really doesn't matter, and just keeping them alphabetically sorted is enough.

golem/network/concent/client.py
```diff
@@ -0,0 +1,91 @@
...
+    pass
+
+
+class ConcentClient:
+
+    def __init__(self):
+        self._is_available = None
+        self._last_available_check = None
+
+    def message(self, message):
```
jiivan: I have a gut feeling that naming this method `send()` would let us avoid problems. :)

golem/network/concent/client.py
```diff
@@ -0,0 +1,91 @@
...
+        if not self.__can_call_concent():
+            raise ConcentGraceException("5 minute failure grace time")
+
+        response = None
+        try:
+            response = requests.post(CONCENT_URL, data=message)
+        except requests.exceptions.RequestException as e:
+            logger.warning('Concent RequestException %r', e)
+            if e.response:
+                response = e.response
```
jiivan, line 59: If I may suggest: I prefer handling failed cases in ifs and returning in them, and handling positive logical path afterwards.

golem/network/concent/client.py
```diff
@@ -0,0 +1,91 @@
...
+        if response:
+            logger.warning('request failed with status %d and body: %r',
+                           response.status_code, response.text)
+
+        self._last_available_check = time.time()
+        self._is_available = False
+
+        raise ConcentUnavailableException("Failed to call concent")
+
+    def is_available(self):
```
jiivan: Accessors are really not necessary in python. Maybe just use `self.is_available` attr instead of `self._is_available`?

jiivan (approved):
Looks great. Just added my personal suggestions (`cause you asked ;) )

golem/network/concent/client.py
maaktweluit: good one! updated it :)

golem/network/concent/client.py
maaktweluit: pylint errors when trying to sort them alphabetically, leaving as is for now to keep the alphabetical per list.

golem/network/concent/client.py
maaktweluit: swapped all if's top facilitate this, good one!

golem/network/concent/client.py
maaktweluit: pylint does not allow classes with only one public function, maybe we can add `reset_counter()` and make this a public attr?

jiivan: `__can_call_concent()` can easily be public. I think that pylint intention (however this sounds :) ) is to avoid small classes which could be implemented with simple functions. Is this the case here? Well it's worth a second thought.

maaktweluit: Solved :D

tests/golem/network/concent/test_concent_client.py
```diff
@@ -0,0 +1,127 @@
...
+# Exception mock
+def connection_error(*args, **kwargs):
+    response = mock.MagicMock()
+    response.status_code = 500
+    response.text = "ERROR"
+    kwargs['response'] = response
+    raise RequestException(args, kwargs)
+
+
+class TestConcentClient(TestCase):
```
jiivan, line 40: One last thought (sorry that it's late, just popped in): consider adding `golem.testutils.PEP8MixIn`.

maaktweluit: Good thought! Made #1664 to align it with the linters running on shippable.

PR #1660, opened 2017-11-14
Added loglevel argument

maaktweluit:
Enabled `--loglevel` as argument and replace the following objects with the new level:
- ROOT logger
- handlers
- loggers

Possible values are `WARNING`, `INFO` and `DEBUG`

Also fixed some linting errors along the way

### DIFF
--- a/golem/core/common.py
+++ b/golem/core/common.py
@@ -214,6 +214,10 @@ def config_logging(suffix='', datadir=None, loglevel=None):
 
     crossbar_log_lvl = logging.getLevelName(
         logging.getLogger('golem.rpc.crossbar').level).lower()
+    # Fix inconsistency in log levels, only warn affected
+    if crossbar_log_lvl == 'warning':
+        crossbar_log_lvl = 'warn'
+
     txaio.set_global_log_level(crossbar_log_lvl)
 
 
--- a/golem/core/common.py
+++ b/golem/core/common.py
@@ -5,10 +5,12 @@ import logging.config
 from multiprocessing import cpu_count
 import os
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


Krigpl (approved):

golem/core/common.py
```diff
@@ -171,18 +171,30 @@ def config_logging(suffix='', datadir=None):
...
+        if loglevel:
+            handler['level'] = loglevel
         if 'filename' in handler:
             handler['filename'] %= {
                 'logdir': str(logdir_path),
                 'suffix': suffix,
             }
 
+    if loglevel:
+        for _logger in list(LOGGING.get('loggers', {}).values()):
```
mfranciszkiewicz: Conversion to `list` is superfluous

golem/core/common.py
```diff
@@ -4,12 +4,13 @@
...
 from multiprocessing import cpu_count
 import os
-from pathlib import Path
-import pytz
 import sys
+import pytz
 
 from golem.core import simpleenv
 
+logger = logging.getLogger(__name__)
```
mfranciszkiewicz: Could we avoid using loggers prior to configuring them?

mfranciszkiewicz (changes requested):

golem/core/common.py
maaktweluit: Moved log message to after the config is applied, good catch!

golem/core/common.py
maaktweluit: Indeed, removed it and everything seems to still work :)

mfranciszkiewicz (approved):

Issue #856, opened 2017-03-02
testGetHostAddress fails for IPv6 addresses with percent sign 


badb:
`ip_address` function doesn't parse ipv6 notation with percent sign, eg. `fe80::71a3:2b00:ddd3:753f%16`

Therefore `testGetHostAddress` function from `test_hostaddress` can fail

```
>       self.assertTrue(is_ip_address(get_host_address(None, True)))
E       AssertionError: False is not true
```

zeeshanabid94:
Hey! I am interested in this task. I would like to contribute. But I am only a beginner in contributing to the open source community. I can work on this function and try to fix it. Do you want it to parse the ipv6 notation properly?

ujjwalgulecha:
Do we still have this issue? @badb

PR #1677, opened 2017-11-19
Fixes issue #856

SLongofono:
The ipaddress module doesn't appear to support the local scope addressing.  As elsewhere in the code, this considers only the address itself, ignoring anything at or beyond the '%'.  Also checks the convention for local-only addresses of using the 'fe80' prefix.

### DIFF
--- a/tests/golem/core/test_hostaddress.py
+++ b/tests/golem/core/test_hostaddress.py
@@ -44,7 +44,8 @@ def is_ip_address(address):
     except (ValueError, AddressValueError):
         try:
             # Check for local scope prefix and designator
-            ip_address(s.split('%')[0])
+            s = s.split('%')
+            ip_address(s[0])
             return (s[0][:4].lower().startswith('fe80'))
         except (ValueError, AddressValueError):
             pass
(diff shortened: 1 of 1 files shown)


badb:
Resolves #856

tests/golem/core/test_hostaddress.py
```diff
@@ -38,9 +38,16 @@ def is_ip_address(address):
...
         # will raise error in case of incorrect address
-        ip_address(str(address))
+        s = str(address)
+        ip_address(s)
         return True
     except (ValueError, AddressValueError):
+        try:
+            # Check for local scope prefix and designator
+            ip_address(s.split('%')[0])
+            return (s[0][:4].lower().startswith('fe80'))
```
badb: s is still a string, so `s[0]` is the first char in the string.

badb (changes requested):
@SLongofono Thank you for contributing! There are however still some issues with this PR:

1) Fixing just is_ip_address method in tests won't fix problem with golem (when ip_address method fails). 
2) There should be some tests added. 
3) Current implemention is not working properly.

tests/golem/core/test_hostaddress.py
SLongofono: str.split() returns a list, so that should work as expected, passing this string less the scoping or the unmodified string into ip_address.

As for the tests, forgive my ignorance, where should I submit the ones I verified my code with?

SLongofono: Fixed.  Added tests, please let me know if they belong elsewhere.

mplebanski:
Can we close this?

PR #1681, opened 2017-11-20
Kill docker & try to install vbox driver

badb:


maaktweluit (approved):
LGTM

maaktweluit:
@badb can i merge this or would you like one more approval?

badb:
We can:
1) Use only x64 location from @mfranciszkiewicz version
2) Use x64 and remove VM installer (mixed version)
3) Use this version with additional script with killing vm and installing driver script. 

I think that version 2 seems to be the safest and less problematic. @mfranciszkiewicz @maaktweluit what do you think?

maaktweluit:
i think we can go for version 2 for now.
Most people removed all old versions before 0.9
When the new 0.10.0 needs the driver fix we can add it again, but hope we can build it to clean up by itself.

mfranciszkiewicz:
I'm also for option 2.

~BTW, we could upgrade golem's VM just like in [Docker Toolbox installer](https://github.com/docker/toolbox/blob/master/windows/Toolbox.iss#L312).~
Only boot2docker.iso image is updated.

badb:
Version 2 will be implemented in #1683

PR #1685, opened 2017-11-20
RPC router shutdown method changed

mfranciszkiewicz:
Replaces #1582.

RPC router will shut down only when requested. Currently, rare ping timeouts cause clients to disconnect and in turn, the router shuts down.


etam:
If I run golemapp from command line, pressing ctrl+c causes a clean shutdown. Will it still work this way?
(I'm totally incompetent in the golem <-> crossbar relationship)

mfranciszkiewicz:
@etam Yes, after pressing ctrl+c you will see a message "Node shutdown requested". The point is to not close the router after all clients disconnect.

etam (approved):

PR #1695, opened 2017-11-22
Income query fix

mfranciszkiewicz:
Previous query tried to join `ExpectedIncome`s with `Income`s without the knowledge that `ExpectedIncome`s are removed when a matching `Income` is detected.


mdtanrikulu:
Related with golemfactory/golem-electron#155

ggruszczynski (approved):

PR #1696, opened 2017-11-23
move pyreadline to requirements-win.txt

### LINKED ISSUE #1478: pyreadline should be installed only on windows
As written here: http://ipython.org/pyreadline.html

Installing it on linux breaks pdb like this: https://stackoverflow.com/questions/32608232/valueerror-type-v-not-supported-error-after-installing-pyreadline

etam:
Fixes #1478

### DIFF
--- a/requirements-win.txt
+++ b/requirements-win.txt
@@ -2,3 +2,4 @@
 pywin32
 vboxapi
 pyvbox
+pyreadline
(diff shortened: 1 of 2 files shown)


maaktweluit (approved):
LGTM!

Great cleanup :)

PR #1699, opened 2017-11-24
Concent: MessageForceReportComputedTask

mfranciszkiewicz:
Addresses https://github.com/golemfactory/golem-messages/issues/16

Please submit your feedback on `ConcentClientService` and `ConcentClient`.

`ConcentClientService`:
- separate service thread
- processes a queue of `ConcentRequest`s
- handles Concent HTTP request errors
- requests can be deferred (and cancelled)
- requests have a predefined lifetime (maximum queue time)

`ConcentClient` changes:
- no longer handles the grace period; now it's a responsibility of `ConcentClientService`
- differentiation of client HTTP request and server errors

Possible further changes:
- `ConcentRequest` callbacks on success / errbacks on error, timeout


requirements.txt
```diff
@@ -48,4 +48,4 @@ numpy==1.13.1
 scipy==0.19.1
 pyssim
 enforce==0.3.4
--e git://github.com/golemfactory/golem-messages.git@v1.1.1#egg=golem_messages
+-e git://github.com/golemfactory/golem-messages.git@message-service-refused#egg=golem_messages
```
maaktweluit: TODO: release before merge?

golem/network/concent/client.py
```diff
@@ -1,83 +1,249 @@
...
+        elif response.status_code % 400 < 100:
+            logger.warning('Concent request failed with status %d and '
+                           'response: %r', response.status_code, response.text)
 
+            raise ConcentRequestException("Concent request exception ({}): {}"
+                                          .format(response.status_code,
+                                                  response.text))
 
-class ConcentClient:
+        return response.content or None
```
maaktweluit, line 60: initial requirement was to return `None` instead of empty string.

Does this work like javascript and do that?

golem/network/concent/client.py
```diff
@@ -1,83 +1,249 @@
...
-            if response:
-                logger.warning('request failed with status %d and body: %r',
-                               response.status_code, response.text)
 
-            self._last_available_check = time.time()
-            self.is_available = False
+class ConcentClientService(threading.Thread):
 
-            raise ConcentUnavailableException("Failed to call concent")
+    MIN_GRACE_TIME = 5.  # s
```
maaktweluit, line 127: nice addition!

golem/network/concent/client.py
```diff
@@ -1,83 +1,249 @@
...
-        self.is_available = True
-        if not response.text or response.text == "":
-            return None
-        return response.text
+    QUEUE_TIMEOUT = 5  # s
 
-    def can_call_concent(self):
-        if self._last_available_check is None:
-            return True
+    def __init__(self, enabled=False):
```
maaktweluit: what is the future default value of `enabled`?

In case of true it might be easier to clean up if you switch the default.

golem/network/concent/client.py
```diff
@@ -1,83 +1,249 @@
...
+            self._loop()
+
+    def stop(self) -> None:
+        self._stop_event.set()
+
+    def submit(self,
+               key: Hashable,
+               msg: message.Message,
+               url: Optional[str] = None,
+               delay: Optional[float] = None) -> None:
```
maaktweluit: I think adding a delay is more likely then adding a custom URL, maybe you can switch the order to make it callable without `None` url

golem/network/concent/client.py
```diff
@@ -1,83 +1,249 @@
...
+        """
+        Submit a message to Concent.
+
+        :param key: Request identifier
+        :param msg: Message to send
+        :param url: Concent API endpoint URL
+        :param delay: Time to wait before sending the message
+        :return: None
+        """
+        from twisted.internet import reactor
```
maaktweluit: any reason to not import on top of the file?

golem/network/concent/client.py
```diff
@@ -1,83 +1,249 @@
...
+            req = self._queue.get(True, self.QUEUE_TIMEOUT)
+        except queue.Empty:
+            return
+
+        # FIXME: remove
+        if not self._enabled:
+            self._queued.pop(req.key, None)
+            return
+
+        if req.deadline_ts < time.time():
```
maaktweluit: call `time.time()` 2x quickly in a row, when it was one more a variable would be preferred

golem/network/concent/client.py
```diff
@@ -1,83 +1,249 @@
...
+        self._grace_time = min(self._grace_time * self.GRACE_FACTOR,
+                               self.MAX_GRACE_TIME)
+
+        logger.debug('Concent grace time: %r', self._grace_time)
+        time.sleep(self._grace_time)
+
+    def _enqueue(self, req):
+        req.status = ConcentRequestStatus.Queued
+        self._delayed.pop(req.key, None)
+        self._queued[req.key] = req
```
maaktweluit: this looks like the `req` is added to `self._queued` twice, or am i reading something wrong?

tests/golem/network/concent/test_concent_client.py
```diff
@@ -43,85 +56,259 @@ class TestConcentClient(TestCase):
...
+        concent_service._loop = mock.MagicMock()
+
+        concent_service.start()
+        time.sleep(1.)
+        concent_service.stop()
+        concent_service.join(timeout=3)
+
+        assert concent_service._loop.called
+
+    def test_submit(self, *_):
```
maaktweluit, line 156: maybe this can assert the `MockClient` got called

maaktweluit (changes requested):
LGTM!

Sorry for the list of questions.
Some things that need attention, but all could be my mistake, will approve on reply!

golem/network/concent/client.py
mfranciszkiewicz: Yes, it's returning `None` in case the response evaluates to `False` (empty string, collections, None).

golem/network/concent/client.py
mfranciszkiewicz: `enabled` is currently for enabling / disabling concent requests. It will probably be removed soon.

golem/network/concent/client.py
mfranciszkiewicz: Yes, reactor defines global values and installs signal handlers. We do not want that on import.

golem/network/concent/client.py
maaktweluit: should the default be `true` in that case so its better prepared for the future?

golem/network/concent/client.py
mfranciszkiewicz: It's queued instantly if there is no delay (and put in `_queued`), otherwise a deferred call to `_enqueue` is saved in `_waiting`. It doesn't seem that it's queued twice.

tests/golem/network/concent/test_concent_client.py
mfranciszkiewicz: There are no calls to `ConcentClient` here, but I've added explicit checks to `test_loop*` methods.

golem/network/concent/client.py
mfranciszkiewicz: Done

golem/network/concent/client.py
mfranciszkiewicz: Done

requirements.txt
mfranciszkiewicz: FIXME added

golem/network/concent/client.py
```diff
@@ -1,83 +1,251 @@
...
-        if not response.text or response.text == "":
-            return None
-        return response.text
+    QUEUE_TIMEOUT = 5  # s
 
-    def can_call_concent(self):
-        if self._last_available_check is None:
-            return True
+    def __init__(self, enabled=True):
+        super(ConcentClientService, self).__init__(daemon=True)
```
jiivan: Simplified form of just calling `super().__init__(daemon=True)` could be more readable.

golem/network/concent/client.py
```diff
@@ -1,83 +1,251 @@
...
+                                          .format(response.status_code,
+                                                  response.text))
 
-class ConcentClient:
+        return response.content or None
+
+
+class ConcentRequestStatus(Enum):
+
+    Initial = 0
```
jiivan: If values of this "enum" are not important than maybe use `object()` (or `auto()` in python3.6) instead of manual series of ints?

golem/task/tasksession.py
```diff
@@ -357,6 +361,18 @@ def send_report_computed_task(
             eth_account=eth_account,
             extra_data=extra_data))
 
+        # FIXME: MessageForceReportComputedTask is going to be updated
+        concent_msg = message.MessageForceReportComputedTask(
+            task_result.subtask_id)
+        # FIXME: create a property setter for messages._raw
+        concent_msg._raw = concent_msg.serialize(self.sign)
```
jiivan: Why use a cache in `_raw` attribute instead of just calling `serialize()` where it's needed? Is this cache invalidated every time the message contents change?

golem/network/concent/client.py
```diff
@@ -1,83 +1,251 @@
...
 
+            raise ConcentRequestException("Concent request exception ({}): {}"
+                                          .format(response.status_code,
+                                                  response.text))
 
-class ConcentClient:
+        return response.content or None
+
+
+class ConcentRequestStatus(Enum):
```
jiivan, line 63: If this class is only used for `ConcentRequest` class maybe it would be cleaner to put it's definition inside `ConcentRequest` and just call it `STATUS(Enum)`?

mfranciszkiewicz: It’s also meant for external usage. In that case ConcentRequest.STATUS.Initial doesn’t seem appealing.

golem/task/tasksession.py
mfranciszkiewicz: Serialize requires a signing function (+keys), which isn’t available further in the call stack.

jiivan: Maybe then either pass keys, or serialized message to the call stack?

mfranciszkiewicz: I don't think passing private keys around is a good practice. Currently, message type information is needed and further deserialization is an overkill. Maybe passing both is a good idea.

jiivan: If it's really needed then passing both is a better idea than introducing cache attribute IMO.

mfranciszkiewicz: `_raw` is already present in `Message` but only populated for ingress messages

mfranciszkiewicz: Done

golem/network/concent/client.py
mfranciszkiewicz: Done

golem/network/concent/client.py
mfranciszkiewicz: Done (using `object`)

jiivan (approved):

maaktweluit (approved):

maaktweluit:
Thanks for checking, fixing and replying to all review comments: All have been resolved.

Please let me know when parts have changed requiring re-review

PR #1701, opened 2017-11-27
Remove old tasks

### LINKED ISSUE #1667: Remove tasks older than ... from task manager
When task persistence is reintroduce, tasks are staying in task manager forever. User has to remove them one by one. We should add an option to the configuration that defines how long tasks should be saved in task manager after the timeout. By default it should be set to three days. If the option is equal to zero, tasks should be never automatically removed from task manager.

etam:
Fixes #1667

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -18,7 +18,7 @@ from golem.appconfig import (AppConfig, PUBLISH_BALANCE_INTERVAL,
 from golem.clientconfigdescriptor import ClientConfigDescriptor, ConfigApprover
 from golem.config.presets import HardwarePresetsMixin
 from golem.core.async import AsyncRequest, async_run
-from golem.core.common import to_unicode
+from golem.core.common import to_unicode, string_to_timeout
 from golem.core.fileshelper import du
 from golem.core.hardware import HardwarePresets
 from golem.core.keysauth import EllipticalKeysAuth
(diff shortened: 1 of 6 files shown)


golem/client.py
```diff
@@ -1210,3 +1220,23 @@ def _run(self):
...
+                 interval_seconds: int,
+                 older_than_seconds: int):
+        super().__init__(interval_seconds)
+        self._client = client
+        self.older_than_seconds = older_than_seconds
+
+    def _run(self):
+        now = time.time()
+        for task in self._client.get_tasks():
+            if task['time_started'] + self.older_than_seconds <= now:
```
jiivan: What if task timeout is greater than `CLEAN_TASKS_OLDER_THAN_SECONDS`? Maybe it will be better to check against `time_finished`? It will probably be `time_started + time_elapsed if remaining_time == 0`?

jiivan (changes requested):
Looks great and proves that your `Service` architecture is neat! Only one question below:

golem/client.py
etam: If tasks never starts at all, there will be no `time_finished` nor `time_elapsed`.
Maybe just `time_started + timeout + older_than_seconds`?

jiivan: Seems reasonable

jiivan (approved):

ggruszczynski (approved):
LGTM

Issue #1492, opened 2017-10-16
Send requestor-oriented stats to monitor


badb:
Right now stats that are sent to moniotor and can be observed on https://stats.golem.network/show are mostly provider-oriented, ie. stats about successfuly computed tasks, about timeouts and errors comes only from the providers. It doesn’t depict correctly requestor UX, for example the task that had been computed correclty, but requestor was not able dowload the results will be still counted as a success in the stats page. 

Collecting and presenting requestor events will significantly improve measurement of the network health. 

The purpose of this task is to send information to monitor about:
* number of requested subtasks from this node
* number of currently requested subtasks from this node
* number of successfully collected results 
* number of successfully verified results
* number of subtasks that fail verification
* number of subtasks that got timeout
* number of subtasks that was announced as computed, but it was not possible to download the results

lukasz-glen:
Information sent to the monitor includes:

- total time to complete task (starts from broadcast, ends with completion of all subtasks or cancellation of task or failure of task) divided into tasks that failed, succeeded but contained failures or timeouts, succeeded without failures and timeouts

- number of tasks that failed, succeeded but contained failures or timeouts, succeeded without failures and timeouts

- number of received messages MessageWantToComputeTask (as a measurement of response to 'task announcement')

Issue #1664, opened 2017-11-15
Align linter implementations, shippable vs PEP8MixIn


maaktweluit:
Currently we run shippable but it is not mandatory.
We used to have `golem.testutils.PEP8MixIn` which is mandatory.

Lets check the rules and files for both tests and align them so we can ensure each changed line from now on is conform to all code styles,

PR #1702, opened 2017-11-27
Requestor-oriented stats #1492

RDProjekt:
Statistics about requested tasks are gathered using pydispatch signals sent from `TaskManager` objects. They are transmitted to the monitor in `MonitoringPublisherService`'s `_run` method. The statistics are collected per-session and are not stored on application's exit.

### DIFF
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -13,7 +13,7 @@ from golem.manager.nodestatesnapshot import LocalTaskStateSnapshot
 from golem.network.transport.tcpnetwork import SocketAddress
 from golem.resource.dirmanager import DirManager
 from golem.resource.hyperdrive.resourcesmanager import \
-    HyperdriveResourceManager  # noqa
+    HyperdriveResourceManager
 from golem.task.result.resultmanager import EncryptedResultPackageManager
 from golem.task.taskbase import TaskEventListener, Task, \
     ResourceType
@@ -838,6 +838,30 @@ class TaskManager(TaskEventListener):
     @handle_task_key_error
     def notice_task_updated(self, task_id, subtask_id=None, task_op=None,
                             persist=True):
+        """Called when a task is modified, saves the task and
+        propagates information
+
+        Whenever task is changed `notice_task_updated` should be called
+        to save the task - if the change is save-worthy, as specified
+        by the `persist` parameter - and propagate information about
+        changed task to other parts of the system.
+
+        Most of the calls are save-worthy, but a minority is not: for
+        instance when the work offer is received, the task does not
+        change so saving it does not make sense, but it still makes
+        sense to let other parts of the system know about the change.
+        Also, when a number of minor changes are always followed by a
+        major one, as it is with restarting a frame task, it does not
+        make sense to store all the partial changes, so only the
+        final one is considered save-worthy.
+
+        :param str task_id: id of the updated task
+        :param Optional[str] subtask_id: if the operation done on the
+          task is related to a subtask, id of that subtask
+        :param Optional[TaskOp] task_op: description of the performed
+          operation
+        :param bool persist: should the task be persisted now
+        """
         # self.save_state()
         if persist and self.task_persistence:
             self.dump_task(task_id)
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -19,6 +19,7 @@ from golem.task.taskbase import TaskEventListener, Task, \
     ResourceType
 
 from golem.task.taskkeeper import CompTaskKeeper, compute_subtask_value
+from golem.task.taskrequestorstats import RequestorTaskStatsManager
 
 from golem.task.taskstate import TaskState, TaskStatus, SubtaskStatus, \
     SubtaskState, TaskOp
@@ -104,6 +105,8 @@ class TaskManager(TaskEventListener):
             tasks_dir,
             persist=self.task_persistence,
         )
+
+        self.requestor_stats_manager = RequestorTaskStatsManager()
 
         if self.task_persistence:
             self.restore_tasks()
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -12,7 +12,8 @@ from golem.core.common import HandleKeyError, get_timestamp_utc, \
 from golem.manager.nodestatesnapshot import LocalTaskStateSnapshot
 from golem.network.transport.tcpnetwork import SocketAddress
 from golem.resource.dirmanager import DirManager
-from golem.resource.hyperdrive.resourcesmanager import HyperdriveResourceManager  # noqa
+from golem.resource.hyperdrive.resourcesmanager import \
+    HyperdriveResourceManager  # noqa
 from golem.task.result.resultmanager import EncryptedResultPackageManager
 from golem.task.taskbase import TaskEventListener, Task, \
     ResourceType
@@ -20,7 +21,7 @@ from golem.task.taskbase import TaskEventListener, Task, \
 from golem.task.taskkeeper import CompTaskKeeper, compute_subtask_value
 
 from golem.task.taskstate import TaskState, TaskStatus, SubtaskStatus, \
-    SubtaskState
+    SubtaskState, TaskOp
 
 logger = logging.getLogger(__name__)
 
@@ -164,6 +165,10 @@ class TaskManager(TaskEventListener):
         self.tasks[task.header.task_id] = task
         self.tasks_states[task.header.task_id] = ts
 
+        self.notice_task_updated(task.header.task_id,
+                                 task_op=TaskOp.TASK_CREATED,
+                                 persist=False)
+
     @handle_task_key_error
     def start_task(self, task_id):
         task = self.tasks[task_id]
@@ -180,7 +185,8 @@ class TaskManager(TaskEventListener):
         if self.task_persistence:
             self.dump_task(task.header.task_id)
             logger.info("Task %s added" % task.header.task_id)
-            self.notice_task_updated(task.header.task_id)
+        self.notice_task_updated(task.header.task_id,
+                                 task_op=TaskOp.TASK_STARTED)
 
     def _dump_filepath(self, task_id):
         return self.tasks_dir / ('%s.pickle' % (task_id,))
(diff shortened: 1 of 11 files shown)


golem/task/taskmanager.py
```diff
@@ -12,15 +12,17 @@
 from golem.manager.nodestatesnapshot import LocalTaskStateSnapshot
 from golem.network.transport.tcpnetwork import SocketAddress
 from golem.resource.dirmanager import DirManager
-from golem.resource.hyperdrive.resourcesmanager import HyperdriveResourceManager  # noqa
+from golem.resource.hyperdrive.resourcesmanager import \
+    HyperdriveResourceManager  # noqa
```
maaktweluit: Can the `# noqa` be removed now the line is properly split?

golem/task/taskrequestorstats.py
```diff
@@ -0,0 +1,544 @@
...
+    ("not_downloadable_subtasks_cnt", int),
+    ("failed_subtasks_cnt", int),
+    ("work_offers_cnt", int)])
+
+EMPTY_CURRENT_STATS = CurrentStats(0, 0, 0, 0, 0, 0, 0, 0, 0)
+
+
+def update_current_stats_with_task(
+        current: CurrentStats,
+        old: Optional[TaskStats],
```
maaktweluit, line 267: Should optional arguments be always last?

golem/task/taskrequestorstats.py
```diff
@@ -0,0 +1,544 @@
...
+        old_task_stats = None
+        if task_id in self.tasks:
+            old_task_stats = self.get_task_stats(task_id)
+
+        if task_op == TaskOp.WORK_OFFER_RECEIVED:
+            self.tasks[task_id].got_want_to_compute()
+
+        elif task_op == TaskOp.TASK_RESTORED:
+            if task_state.status in TASK_COMPLETED_STATUSES:
+                logger.info("Skipping completed task {}".format(task_id))
```
maaktweluit: should this be info or debug? does it have use to read for end user too or only for development?

golem/task/taskrequestorstats.py
```diff
@@ -0,0 +1,544 @@
...
+                TaskMsg(ts=time.time(), op=task_op),
+                task_state.subtask_states[subtask_id].subtask_status)
+
+        elif task_op in self.UNNOTEWORTHY_OPS:
+            # these are not interesting and are not stored
+            pass
+
+        else:
+            # Unknown task_op, log problem
+            logger.info("Unknown TaskOp {}".format(task_op.name))
```
maaktweluit: how bad is this? maybe upgrade to warning?

maaktweluit: Also please use native logger formatters `logger.info('message: %r', variable)`
`.format()` is a bad practice since it will always parse the string, while the native logger ignores parsing when the level is turned off

tests/golem/task/test_taskmanager.py
```diff
@@ -119,11 +121,54 @@ def _get_task_mock(self, task_id="xyz", subtask_id="xxyyzz", timeout=120.0,
...
+
     def test_start_task(self):
         task_mock = self._get_task_mock()
 
+        (h,c) = self._connect_signal_handler()
         self.tm.add_new_task(task_mock)
         self.tm.start_task(task_mock.header.task_id)
+        c([("xyz", None, TaskOp.TASK_CREATED),
+           ("xyz", None, TaskOp.TASK_STARTED)])
+        del h
```
maaktweluit, line 171: should this tests assert some things, like for instance mock called?

maaktweluit: More tests in this file have the same question

tests/golem/task/test_taskrequestorstats.py
```diff
@@ -0,0 +1,751 @@
...
+    RequestorTaskStats, logger, CurrentStats, TaskStats, EMPTY_TASK_STATS, \
+    FinishedTasksStats, FinishedTasksSummary, RequestorTaskStatsManager, \
+    EMPTY_CURRENT_STATS, EMPTY_FINISHED_STATS
+from golem.task.taskstate import TaskStatus, TaskOp, SubtaskStatus, TaskState, \
+    SubtaskState
+from golem.tools.assertlogs import LogTestCase
+
+
+class TestTaskInfo(TestCase, testutils.PEP8MixIn):
+    PEP8_FILES = [
```
maaktweluit, line 16: FYI, the PEP8 mix in is in progress of being replaced with mandatory linter tests on CI #1664

maaktweluit (changes requested):
LGTM!

One thing i think needs changing, the log message formatter.
The other comments are all optional, will approve if replied to :)

golem/task/taskmanager.py
```diff
@@ -784,12 +836,16 @@ def notify_update_task(self, task_id):
...
 
     @handle_task_key_error
-    def notice_task_updated(self, task_id):
+    def notice_task_updated(self, task_id, subtask_id=None, task_op=None,
+                            persist=True):
         # self.save_state()
-        if self.task_persistence:
+        if persist and self.task_persistence:
             self.dump_task(task_id)
         dispatcher.send(
```
ggruszczynski, line 868: self.task_persistence is a bool being true by default.
We dont need another `persist=True` flag.

golem/task/taskrequestorstats.py
```diff
@@ -0,0 +1,544 @@
...
+
+Names of fields are mostly self-explanatory.
+``not_downloaded_subtasks_cnt`` is the number of tasks that were
+announced as done by the computing node but were not received.
+"""
+
+EMPTY_TASK_STATS = TaskStats(False, 0.0, False, False, 0, 0, 0, 0, 0, 0, 0)
+
+CurrentStats = NamedTuple("CurrentStats", [
+    ("tasks_cnt", int),
```
ggruszczynski, line 247: What is the difference between TaskStats and CurrentStats?
Please, add some documentation.

golem/task/taskstate.py
```diff
@@ -104,3 +106,24 @@ class TaskTestStatus(object):
     started = 'Started'
     success = 'Success'
     error = 'Error'
+
+
+class TaskOp(Enum):
```
ggruszczynski, line 111: Do we really need another `TaskOp` class/enum containing similar fields that classes above?

golem/task/taskmanager.py
RDProjekt: Absolutely, done.

golem/task/taskrequestorstats.py
RDProjekt: This one is `Optional` as in "can be `None`", but the parameter needs to be always provided. I think it would be weird to pass `new` first and `old` after it.

golem/task/taskrequestorstats.py
RDProjekt: End user can't do much about it and it does not really affect her. Demoted to debug.

golem/task/taskrequestorstats.py
RDProjekt: Changed all new logger calls to use logger formatters. If this ever happens it means that someone changed `TaskOp` enum but did not update the stats collecting code. End user can't do anything about it, so instead of upgrading demoted it to debug.

tests/golem/task/test_taskmanager.py
RDProjekt: `c()` does that, first thing it checks is if the handler has been called.

golem/task/taskmanager.py
RDProjekt: This is kind of opposite of `self.task_persistence`, which describes if the system has persistence enabled. `persist` parameter says if the particular change that called `notice_task_updated` was important enough to save the task. I have added a comment describing this in more details.

golem/task/taskrequestorstats.py
RDProjekt: They have different fields. `TaskStats` are about a single task, `CurrentStats` is a summary for all tasks in the session. Added documentation to `CurrentStats`, `TaskStats` already had it.

golem/task/taskstate.py
RDProjekt: Yes, other structures describe current state of the task and `TaskOp` describes the _change_ of the state.

maaktweluit (approved):

golem/task/taskmanager.py
```diff
@@ -248,6 +254,24 @@ def resources_send(self, task_id):
...
+
+        :param str task_id: id of the task in the offer
+        :param key_id: id of the node offering computations
+        :param node_name: name of the node offering computations
+        :return: Nothing
+        :rtype: None
+        """
+        if task_id in self.tasks:
+            self.notice_task_updated(task_id,
+                                     task_op=TaskOp.WORK_OFFER_RECEIVED)
```
tworec, line 273: according to doc in `notice_task_updated` (line 850) this op should be notified with `persist=False`

golem/task/taskrequestorstats.py
```diff
@@ -0,0 +1,549 @@
+import logging
+import time
+from collections import defaultdict
+from typing import (  # pylint: disable=unused-import
+    NamedTuple, List, Optional, DefaultDict)
```
tworec, line 5: ugly: `DefaultDict` import is only for type annotation comments

golem/task/taskstate.py
tworec: This enum is a mess. It mixes TASK, SUBTASK and others.

tests/golem/task/test_taskmanager.py
tworec: `h()` and `c()` are meaningless. They should be  fully named.

golem/task/taskrequestorstats.py
```diff
@@ -0,0 +1,549 @@
...
+are tasks that did not succeed.
+"""
+
+EMPTY_FINISHED_STATS = FinishedTasksStats(
+    EMPTY_FINISHED_SUMMARY,
+    EMPTY_FINISHED_SUMMARY,
+    EMPTY_FINISHED_SUMMARY)
+
+
+def update_finished_stats_with_task(
```
tworec, line 337: code in this method looks awful/boilerplate because of using immutable tuples instead of mutable class. But I'll leave it as is, to not be burried into marginal style fixes any more

tworec (changes requested):
Few comments from me.

PR #1704, opened 2017-11-28
fix pytest version

etam:


requirements.txt
```diff
@@ -34,7 +34,8 @@ pydispatcher
 Crypto
 scrypt
 packaging
-pytest>=3.0.5
+# pytest 3.3.0 requires newer attrs, which conflicts with crossbar 17.3.1
+pytest>=3.0.5,<3.3.0
```
maaktweluit: is this required to run the app normally? think the only one that should stay is in `-test.txt`

maaktweluit (changes requested):

requirements.txt
etam: It's used in setup.py.

etam: It's used indirectly. It allows to run `python setup.py test`, but it seems broken.

maaktweluit (approved):

mfranciszkiewicz (approved):

etam:
@mfranciszkiewicz Please don't squash my commits.

mfranciszkiewicz:
@etam Your commits specifically? Please elaborate why.

etam:
@mfranciszkiewicz Because they are supposed to be separate, with their own commit messages. Now it makes no sense, that commit titled 'fix pytest version' does also something different. It's not a big issue in this case, because the squashed commit is small, but I really hate those huge commits that changes bazillion things at once.
I also use various git features (like rebase interactive) to keep the history nice and clean.

mfranciszkiewicz:
@etam If "fix pytest version" does more than it says, then this PR's title is simply wrong. The rest is a matter of personal preference, but I will keep that in mind.

etam:
@mfranciszkiewicz Yes, I agree, that the second commit goes beyond what this PR was about. I should have created a second one for it.

PR #1705, opened 2017-11-29
test dirmanager symlink_or_copy

### LINKED ISSUE #1644: E: golem/resource/dirmanager.py:22:12: No name 'dir_util' in module 'distutils'
https://github.com/golemfactory/golem/blob/66a18964ce43df8c3aa8833c0bf28651611f5ba0/golem/resource/dirmanager.py#L21

etam:
I was supposed to fix #1644, but it seems the import works fine.

It's nice to have more tests anyway.


tests/golem/resource/test_dirmanager.py
```diff
@@ -2,13 +2,60 @@
...
+        target_path = os.path.join(self.path, 'target')
+
+        with open(source_path, 'w') as f:
+            f.write('source')
+        with open(target_path, 'w') as f:
+            f.write('target')
+
+        def raise_OSError(*_):
+            raise OSError
+        os.symlink = raise_OSError
```
jiivan: This overwrites `os.symlink` for all other tests. Better use `unittest.mock` with `side_effect`.

tests/golem/resource/test_dirmanager.py
```diff
@@ -2,13 +2,60 @@
...
+        source_file_path = os.path.join(source_dir_path, 'file')
+        target_path = os.path.join(self.path, 'target')
+
+        os.mkdir(source_dir_path)
+        with open(source_file_path, 'w') as f:
+            f.write('source')
+
+        def raise_OSError(*_):
+            raise OSError
+        os.symlink = raise_OSError
```
jiivan: `unittest.mock` with side effect

jiivan (changes requested):
Have you also seen pathlib? It may simplify your code (but of course it's just a matter of taste).

tests/golem/resource/test_dirmanager.py
etam: I tried
```python
    @patch('os.symlink')
    def test_conflict_file(self, os_symlink):
        os_symlink = Mock(side_effect=OSError)
```
But flake8 says `F841 local variable 'os_symlink' is assigned to but never used` and it seems it doesn't change anything. And I don't want to patch whole `os` module.

etam: I've found a way to patch os.symlink just for tested function.

jiivan (approved):

Krigpl (approved):

PR #1709, opened 2017-11-30
Task persistence: fixes, sharing restored task resources with HyperG

mfranciszkiewicz:
- registers a task event listener in unplicked tasks
- restores the `TaskManager.subtask2task_mapping` for unplicked tasks
- restores / adds resources to HyperG for unplicked tasks
- stores information on shared resource hash in `TaskState`
- removes the redundant `Task.task_status` property

### DIFF
--- a/golem/resource/hyperdrive/resourcesmanager.py
+++ b/golem/resource/hyperdrive/resourcesmanager.py
@@ -47,7 +47,7 @@ class HyperdriveResourceManager(ClientHandler, AbstractResourceManager):
 
         return results
 
-    def add_files(self, files, task_id,
+    def add_files(self, files, task_id, resource_hash=None,
                   absolute_path=False, client=None, client_options=None):
 
         if not files:
@@ -59,6 +59,7 @@ class HyperdriveResourceManager(ClientHandler, AbstractResourceManager):
                  for path in files}
 
         return self._add_files(files, task_id,
+                               resource_hash=resource_hash,
                                client=client,
                                client_options=client_options)
 
@@ -71,22 +72,25 @@ class HyperdriveResourceManager(ClientHandler, AbstractResourceManager):
                                client=client,
                                client_options=client_options)
 
-    def _add_files(self, files, task_id,
+    def _add_files(self, files, task_id, resource_hash=None,
                    client=None, client_options=None):
 
-        for f in files.keys():
-            if not os.path.exists(f):
-                logger.error("Resource manager: file '{}' does not exist"
-                             .format(f))
-                return
+        if not all(os.path.exists(f) for f in files.keys()):
+            logger.error("Resource manager: missing files (task: %r)", task_id)
+            return
 
         client = client or self.new_client()
-        response = self._handle_retries(client.add,
-                                        self.commands.add,
-                                        files,
+
+        if resource_hash:
+            args = (client.restore, self.commands.restore, resource_hash)
+        else:
+            args = (client.add, self.commands.add, files)
+
+        response = self._handle_retries(*args,
                                         id=task_id,
                                         client_options=client_options,
-                                        obj_id=str(uuid.uuid4()))
+                                        obj_id=str(uuid.uuid4()),
+                                        raise_exc=bool(resource_hash))
 
         file_list = list(files.values())
         self._cache_response(file_list, response, task_id)
--- a/golem/resource/hyperdrive/resourcesmanager.py
+++ b/golem/resource/hyperdrive/resourcesmanager.py
@@ -75,7 +75,7 @@ class HyperdriveResourceManager(ClientHandler, AbstractResourceManager):
     def _add_files(self, files, task_id, resource_hash=None,
                    client=None, client_options=None):
 
-        if not all(os.path.exists(f) for f in files.keys()):
+        if not all(os.path.exists(f) for f in files):
             logger.error("Resource manager: missing files (task: %r)", task_id)
             return
 
--- a/golem/resource/hyperdrive/resourcesmanager.py
+++ b/golem/resource/hyperdrive/resourcesmanager.py
@@ -75,8 +75,12 @@ class HyperdriveResourceManager(ClientHandler, AbstractResourceManager):
     def _add_files(self, files, task_id, resource_hash=None,
                    client=None, client_options=None):
 
-        if not all(os.path.exists(f) for f in files):
-            logger.error("Resource manager: missing files (task: %r)", task_id)
+        checked = {f: os.path.exists(f) for f in files}
+
+        if not all(checked.values()):
+            missing = [f for f, exists in files.items() if not exists]
+            logger.error("Resource manager: missing files (task: %r):\n%s",
+                         task_id, missing)
             return
 
         client = client or self.new_client()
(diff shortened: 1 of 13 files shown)


golem/resource/hyperdrive/resourcesmanager.py
```diff
@@ -71,22 +72,25 @@ def add_file(self, path, task_id,
...
-    def _add_files(self, files, task_id,
+    def _add_files(self, files, task_id, resource_hash=None,
                    client=None, client_options=None):
 
-        for f in files.keys():
-            if not os.path.exists(f):
-                logger.error("Resource manager: file '{}' does not exist"
-                             .format(f))
-                return
+        if not all(os.path.exists(f) for f in files.keys()):
```
jiivan: `.keys()` is not needed. Iterating over dicts always iterates over its keys.

golem/task/taskmanager.py
```diff
@@ -210,7 +207,7 @@ def remove_dump(self, task_id: str):
             filepath.unlink()
             logger.debug('TASK DUMP with id %s REMOVED from %r',
                          task_id, filepath)
-        except OSError as e:
+        except (FileNotFoundError, OSError) as e:
             logger.warning("Couldn't remove dump file: %s - %s", e)
```
jiivan: I just noticed missing parameter `filepath` in this log.

tests/golem/resource/hyperdrive/test_hyperdrive_resourcemanager.py
```diff
@@ -55,8 +58,44 @@ def test(self):
         assert len(peer_manager.get_for_task(task_id)) == 1
 
 
+class TestHyperdriveResourceManager(TempDirFixture):
+
+    def test_add_files(self):
```
jiivan: This method should be split into separate cases i.e.:
 * `test_add_files_invalid_paths()`
 * `test_add_files_empty_resource_hash()`
etc.

tests/golem/task/test_taskserver.py
```diff
@@ -872,3 +875,95 @@ def _get_config_desc(self):
...
+        for i in range(count):
+            task_id = str(uuid.uuid4())
+            task_server.task_manager.tasks[task_id] = Mock()
+            task_server.task_manager.tasks_states[task_id] = TaskState()
+
+    @staticmethod
+    def _build_raise(cls):
+        def _raise(*_a, **_kw):
+            raise cls()
+        return Mock(side_effect=_raise)
```
jiivan: ` side_effect` can be an exception (class or instance). What's the benefit of `_build_raise()` method?

https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect

jiivan (changes requested):
LGTM, only minor remarks

jiivan (approved):
Thanks!

golem/resource/hyperdrive/resourcesmanager.py
```diff
@@ -71,22 +72,25 @@ def add_file(self, path, task_id,
...
+    def _add_files(self, files, task_id, resource_hash=None,
                    client=None, client_options=None):
 
-        for f in files.keys():
-            if not os.path.exists(f):
-                logger.error("Resource manager: file '{}' does not exist"
-                             .format(f))
-                return
+        if not all(os.path.exists(f) for f in files):
+            logger.error("Resource manager: missing files (task: %r)", task_id)
```
maaktweluit: is task_id enough to investigate?
if not maybe a list of files in a debug log next to it would be helpful

golem/task/taskmanager.py
```diff
@@ -143,44 +144,40 @@ def add_new_task(self, task):
...
         ts.outputs = task.get_output_names()
         ts.total_subtasks = task.get_total_tasks()
         ts.time_started = time.time()
 
-        self.tasks[task.header.task_id] = task
-        self.tasks_states[task.header.task_id] = ts
+        self.tasks[task_id] = task
+        self.tasks_states[task_id] = ts
+        self.notice_task_updated(task_id)
+        logger.info("Task %s added", task_id)
```
maaktweluit, line 168: nice cleanup!

golem/task/taskmanager.py
```diff
@@ -143,44 +144,40 @@ def add_new_task(self, task):
                                                self.listen_port):
             raise IOError("Incorrect socket address")
 
-        task.task_status = TaskStatus.notStarted
```
maaktweluit, line 146: is still still the default status now its gone here?

golem/task/taskmanager.py
```diff
@@ -543,12 +542,11 @@ def get_resources(
 
     @handle_task_key_error
     def restart_task(self, task_id):
-        logger.info("restarting task")
+        logger.info("Restarting task %s", task_id)
```
maaktweluit: Log not consistent with other info's in the class, `Task %s restarting`.
Super small, can be ignored...

golem/task/taskserver.py
```diff
@@ -58,6 +63,47 @@ def get_resources(self, task_id):
...
+                resource_hash=resource_hash,
+                async=False
+            )
+        except ConnectionError:
+            logger.error("Cannot restore resources for task: %r", task_id)
+            self.task_manager.delete_task(task_id)
+        except HTTPError:
+            if resource_hash:
+                return self._restore_resources(files, task_id)
+            logger.error("Cannot restore resources for task: %r", task_id)
```
maaktweluit: is it worth noting the difference between `ConnectionError` and `HTTPError` in the logged message?

tests/golem/resource/hyperdrive/test_hyperdrive_resourcemanager.py
```diff
@@ -55,8 +58,44 @@ def test(self):
...
+        super().setUp()
+
+        self.task_id = str(uuid.uuid4())
+        self.dir_manager = DirManager(self.tempdir)
+        self.resource_manager = HyperdriveResourceManager(self.dir_manager)
+        self.resource_manager._handle_retries = Mock()
+
+        file_name = 'test_file'
+        file_path = os.path.join(self.tempdir, file_name)
+        open(file_path, 'w').close()
```
maaktweluit: looks like a raw method to ensure file, is there a native function for that?

maaktweluit (approved):
LGTM!

Small comments, replies are enough :)

tests/golem/resource/hyperdrive/test_hyperdrive_resourcemanager.py
jiivan: https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch

golem/task/taskserver.py
mfranciszkiewicz: Yes, `ConnectionError` means that there were problems connecting to HyperG, whereas `HTTPError` means that there was a problem with the request. In this case, `resource_hash` is not known by HyperG, so we should add those files again.

golem/task/taskmanager.py
mfranciszkiewicz: `task.task_status` is not used outside this file, so I've removed it in favor of `TaskState.status`

golem/task/taskserver.py
maaktweluit: maybe this can be added to the log.error messages? 
For example: `No conection` for the top one, `no resource_hash` for the bottom one
It would help finding a problem in this region when it occurs

maaktweluit:
Thanks for the fixes and replies!

Replied to the last open thread

PR #1712, opened 2017-11-30
Remove Message prefix

jiivan:
realted to: golemfactory/golem-messages#20


requirements.txt
```diff
@@ -47,4 +47,4 @@ numpy==1.13.1
 scipy==0.19.1
 pyssim
 enforce==0.3.4
--e git://github.com/golemfactory/golem-messages.git@v1.1.1#egg=golem_messages
+-e git://github.com/golemfactory/golem-messages.git@20_msgprefix#egg=golem_messages
```
Krigpl: Why won't make you a release instead of this custom `20_msgprefix` tag?

jiivan: I want to make a release when this PR is accepted, and update requirements before merging.

Krigpl: What stops you from making the release before this PR is accepted?

jiivan: Because it would stop every other PR related to golem-messages

Krigpl (approved):

PR #1718, opened 2017-12-01
TaskSession message history (Concent)

mfranciszkiewicz:
- Facility to save / remove `Message`s to / from database in background
- `Session` decorators and interface for annotating logged `react_to_*` methods
- `Message`s in database have a fixed lifetime and are swept periodically

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -36,6 +36,7 @@ from golem.monitor.model.nodemetadatamodel import NodeMetadataModel
 from golem.monitor.monitor import SystemMonitor
 from golem.monitorconfig import MONITOR_CONFIG
 from golem.network.concent.client import ConcentClientService
+from golem.network import history
 from golem.network.hyperdrive.daemon_manager import HyperdriveDaemonManager
 from golem.network.p2p.node import Node
 from golem.network.p2p.p2pservice import P2PService
@@ -143,7 +144,8 @@ class Client(HardwarePresetsMixin):
         self._services = [
             NetworkConnectionPublisherService(
                 self,
-                int(self.config_desc.network_check_interval))
+                int(self.config_desc.network_check_interval)),
+            history.install_service()
         ]
 
         clean_resources_older_than = \
@@ -249,6 +251,7 @@ class Client(HardwarePresetsMixin):
     @report_calls(Component.client, 'stop', stage=Stage.post)
     def stop(self):
         self.stop_network()
+
         for service in self._services:
             if service.running:
                 service.stop()
--- a/golem/client.py
+++ b/golem/client.py
@@ -22,7 +22,7 @@ from golem.core.common import to_unicode, string_to_timeout
 from golem.core.fileshelper import du
 from golem.core.hardware import HardwarePresets
 from golem.core.keysauth import EllipticalKeysAuth
-from golem.core.service import IService, LoopingCallService
+from golem.core.service import LoopingCallService
 from golem.core.simpleenv import get_local_datadir
 from golem.core.simpleserializer import DictSerializer
 from golem.core.threads import callback_wrapper
--- a/golem/client.py
+++ b/golem/client.py
@@ -36,7 +36,7 @@ from golem.monitor.model.nodemetadatamodel import NodeMetadataModel
 from golem.monitor.monitor import SystemMonitor
 from golem.monitorconfig import MONITOR_CONFIG
 from golem.network.concent.client import ConcentClientService
-from golem.network import history
+from golem.network.history import MessageHistoryService
 from golem.network.hyperdrive.daemon_manager import HyperdriveDaemonManager
 from golem.network.p2p.node import Node
 from golem.network.p2p.p2pservice import P2PService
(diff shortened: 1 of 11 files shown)


golem/client.py
```diff
@@ -143,7 +144,8 @@ def __init__(
         self._services = [
             NetworkConnectionPublisherService(
                 self,
-                int(self.config_desc.network_check_interval))
+                int(self.config_desc.network_check_interval)),
+            MessageHistoryService()
```
jiivan: Maybe a hanging comma so further additions won't modify previous lines? Just an aesthetic :)

golem/model.py
```diff
@@ -404,11 +405,35 @@ class Meta:
...
             perf = Performance(environment_id=env_id, value=performance)
             perf.save()
+
+
+##################
+# MESSAGE MODELS #
+##################
+
+
+class Actor(Enum):
```
jiivan, line 423: Would it be possible to use string values, to make debugging with external tools easier?

golem/model.py
```diff
@@ -404,11 +405,35 @@ class Meta:
...
+    Provider = 2
+
+
+class NetworkMessage(BaseModel):
+    local_role = EnumField(Actor, null=False)
+    remote_role = EnumField(Actor, null=False)
+
+    task = CharField(null=False, index=True)
+    subtask = CharField(index=True)
+    node = CharField()
```
jiivan: Is this a recipient node or a sender node? I'm confused.

golem/network/history.py
```diff
@@ -0,0 +1,243 @@
...
+
+from peewee import (PeeweeException, DataError, ProgrammingError,
+                    NotSupportedError, Field)
+
+from golem.model import NetworkMessage, Actor
+
+logger = logging.getLogger('golem.network.history')
+
+
+class MessageHistoryService(threading.Thread):
```
jiivan: Would it be possible to inherit from `golem.core.service.Service`?

tests/golem/network/test_history.py
```diff
@@ -0,0 +1,308 @@
...
+from peewee import DataError, PeeweeException
+
+from golem.model import NetworkMessage, Actor
+from golem.network.history import MessageHistoryService, record_history, \
+    IMessageHistoryProvider, requestor_history, provider_history
+from golem.testutils import DatabaseFixture
+
+
+def message_count():
+    return len(NetworkMessage.select().execute())
```
jiivan: Why not `.select().count()`?

mfranciszkiewicz: Thanks, I forgot to change this - at one point there was a `where` clause

golem/network/history.py
mfranciszkiewicz: Unfortunately `Service` uses a fixed-interval looping call. This service locks on a queue and saves messages asap (apart from sweeping old messages, etc.).

golem/model.py
mfranciszkiewicz: Well, it's the "other" node. Depending on context (local / remote role, message class) it may be a receiver or a sender.

golem/client.py
mfranciszkiewicz: Ok :)

golem/model.py
mfranciszkiewicz: Will do

golem/model.py
jiivan: Could you put this comment in code?

golem/network/history.py
jiivan: Yes, I thought that sweep method fits in a `core.service` structure. Thanks for claryfing. Could you also put this comment in code (maybe class docstring)?

mfranciszkiewicz: Done

tests/golem/network/test_history.py
```diff
@@ -0,0 +1,308 @@
...
+    def setUp(self):
+        super().setUp()
+        self.service = MessageHistoryService()
+
+    def tearDown(self):
+        super().tearDown()
+        MessageHistoryService.instance = None
+
+    @staticmethod
+    def _build_raise(cls):
```
jiivan: `side_effect` (also fixes a warning from Shippable)

jiivan (commented):
There are some shippable errors worth fixing. Mostly in tests, but still.

jiivan (approved):

golem/client.py
```diff
@@ -143,7 +144,8 @@ def __init__(
         self._services = [
             NetworkConnectionPublisherService(
                 self,
-                int(self.config_desc.network_check_interval))
+                int(self.config_desc.network_check_interval)),
+            MessageHistoryService(),
```
etam, line 148: The type annotation of `_services` is `List[Service]` (with `from golem.core.service import Service`).
With this new type of service, I think it should be changed that they all implement a common metaclass.

etam (changes requested):
Added one comment.

Thanks for nice type annotations.

golem/client.py
jiivan: Are you sure you meant a metaclass https://docs.python.org/3/reference/datamodel.html?highlight=metaclass#metaclasses ? :)

A common mixin class may or may not be a good addition though.

etam: Maybe not directly a metaclass, but an Abstract Base Class.

mfranciszkiewicz: I've created a common interface for `Service`s (called `IService`, derived from `ABC`). Also, `Service` became ambiguous, and it's now named `LoopingCallService`.

etam (approved):

golem/client.py
etam: Nice

PR #1723, opened 2017-12-05
Fix checking for number of confirmations

Krigpl:
This is the right way of doing #1466
I was mislead by this thread https://www.reddit.com/r/ethereum/comments/43wj2p/how_do_i_get_the_number_of_confirmations_on_a/


chfast (approved):
Funny :)

jiivan (approved):

PR #1725, opened 2017-12-06
Partially fix race condition when detecting duplicated peers

Krigpl:
Don't add anything to peerkeeper until connection is verified.
Also checking whether there's a duplicate and adding new peer should be atomic - it isn't now, but it's better then it was before.
This is a follow up to #1537

### DIFF
--- a/golem/network/p2p/peersession.py
+++ b/golem/network/p2p/peersession.py
@@ -68,6 +68,7 @@ class PeerSession(BasicSafeSession):
         self.client_ver = None
         self.listen_port = None
         self.conn_id = None
+        self.metadata = None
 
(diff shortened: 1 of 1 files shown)


jiivan (commented):
Does this change affect how Timeout is handled? Will timeout be still possible during period between establishing connection and adding to peer keeper?

Krigpl:
@jiivan As far as I can tell we have two timeout checks: one for pending (non verified) connection in `tcpserver` and second in `p2p_service` for verified connections. So we should be covered.

jiivan (approved):

mfranciszkiewicz (approved):

Issue #1217, opened 2017-07-07
Replace pyelliptic


mfranciszkiewicz:
We should look for a replacement for`pyelliptic` since it is no longer maintained.

Proposed replacement: https://github.com/pyca/cryptography

Wolf480pl:
AFAIK this issue reduces to ethereum/pydevp2p#58 cause I don't see pyelliptic used anywhere in golem code.

Wolf480pl:
`cryptography` doesn't call (or provide a way to call) OpenSSL's `check_key` :(

mfranciszkiewicz:
Initial 1.1-compatible version of `pyelliptic`: https://github.com/mfranciszkiewicz/pyelliptic

maaktweluit:
Yay! maybe we can also resolve some of the loading / path hacks at the top of https://github.com/golemfactory/golem-messages/blob/master/golem_messages/cryptography.py

badb:
@mfranciszkiewicz Can we close this issue?

mfranciszkiewicz:
@badb I believe so. Closing.

PR #1727, opened 2017-12-06
use pyelliptic compatible with openssl 1.1

etam:
pyelliptic is officially deprecated https://github.com/yann2192/pyelliptic/issues/50 and we want to replace it (#1217). Until then, a patched version is required.


etam:
Should I move pyelliptic repo to golemfactory?

etam:
It's not working with mac yet. I'll fix that.

etam:
It's not working well. I'll reopen this if I get something better.

PR #1729, opened 2017-12-06
Avoid operations on floats

Krigpl:
Because floats are evil.
```
>>> math.ceil(10 ** 18 * 600 / 3600)
166666666666666656
>>> (10 ** 18 * 600 + 3599) // 3600
166666666666666667
```

### DIFF
--- a/golem/task/taskkeeper.py
+++ b/golem/task/taskkeeper.py
@@ -18,6 +18,13 @@ logger = logging.getLogger('golem.task.taskkeeper')
 
 
 def compute_subtask_value(price, computation_time):
+    """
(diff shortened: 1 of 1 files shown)


etam (changes requested):
Add a comment that thescribes the issue

Krigpl:
@etam floats are evil in general, I dunno an explicit comment here is valuable

etam (approved):
Maybe you're right

mfranciszkiewicz (approved):
Please add the comment, because someone might be tempted to refactor this back to `math.ceil` in the future.

PR #1731, opened 2017-12-07
Proper detection whether a connection is incoming or outgoing

Krigpl:
The check against Client/Server instances was a quick hack in order to avoid all this boilerplate. Turns out on Windows they use different implementation that don't share common interface, thus it doesn't work. 
Boilerplate code it is.


jiivan (approved):
LGTM

mfranciszkiewicz (approved):

PR #1732, opened 2017-12-07
avoid potential conflict with sql keyword

### LINKED ISSUE #1691: "Income.transaction" column name can cause of conflicts
'transaction' itself is SQLite command, It's better to rename for the potential conflicts.
https://github.com/golemfactory/golem/blob/c17f2ba6491fef0db5a5a5c887edc0145238b95d/golem/model.py#L259

etam:
fixes #1691

### DIFF
--- a/golem/model.py
+++ b/golem/model.py
@@ -27,7 +27,7 @@ db = SqliteDatabase(None, threadlocals=True,
 
 class Database:
     # Database user schema version, bump to recreate the database
-    SCHEMA_VERSION = 5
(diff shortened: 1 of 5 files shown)


mfranciszkiewicz:
If we really need to rename it, how about `tx` instead?

etam:
> If we really need to rename it, how about `tx` instead?

I'm not sure. A bit too abbreviated for me. Requires about 2 or 3 additional cycles of brain processing unit to parse.

PR #1733, opened 2017-12-07
Round robin over all seeds

Krigpl:
This means that we will try to connect to all seeds before attempting to reconnect with previous seed.
This helps in a scenario where our seed nodes are using different protocol ids (e.g. protocol got updated and we would like to support both versions until next release) and we would like to eventually connect to the one with compatible protocol with ours. Because until now if we were unlucky it could take some longer period of time.

### DIFF
--- a/golem/network/p2p/p2pservice.py
+++ b/golem/network/p2p/p2pservice.py
@@ -97,6 +97,7 @@ class P2PService(tcpserver.PendingConnectionsServer, DiagnosticsProvider):
         self.free_peers = []  # peers to which we're not connected
         self.resource_peers = {}
         self.seeds = set()
+        self.used_seeds = set()
 
         self._peer_lock = Lock()
 
@@ -140,7 +141,8 @@ class P2PService(tcpserver.PendingConnectionsServer, DiagnosticsProvider):
         if not self.connect_to_known_hosts:
             return
 
-        for ip_address, port in random.sample(self.seeds, k=len(self.seeds)):
+        for _ in range(len(self.seeds)):
+            ip_address, port = self._get_next_random_seed()
             try:
                 socket_address = tcpnetwork.SocketAddress(ip_address, port)
                 self.connect(socket_address)
@@ -1023,6 +1025,15 @@ class P2PService(tcpserver.PendingConnectionsServer, DiagnosticsProvider):
                 )
                 continue
             self.seeds.add((ip_address, port))
+
+    def _get_next_random_seed(self):
+        # this loop won't execute more than twice
+        while True:
+            for seed in random.sample(self.seeds, k=len(self.seeds)):
+                if seed not in self.used_seeds:
+                    self.used_seeds.add(seed)
+                    return seed
+            self.used_seeds = set()
 
     def __remove_sessions_to_end_from_peer_keeper(self):
         for node in self.peer_keeper.sessions_to_end:
--- a/tests/golem/network/p2p/test_p2pservice.py
+++ b/tests/golem/network/p2p/test_p2pservice.py
@@ -406,3 +406,14 @@ class TestP2PService(testutils.DatabaseFixture):
         self.service.peers = {'peer_id': mock.Mock()}
         self.service.disconnect()
         assert self.service.peers['peer_id'].dropped.called
+
+    def test_round_robin_seeds(self):
+        SEEDS_NUM = 10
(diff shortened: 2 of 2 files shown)


golem/network/p2p/p2pservice.py
```diff
@@ -1024,6 +1026,15 @@ def __sync_seeds(self, known_hosts=None):
                 continue
             self.seeds.add((ip_address, port))
 
+    def _get_next_random_seed(self):
```
etam, line 1029: How about this?
```python
def _get_next_random_seed(self):
    if not self.unused_seeds:
         self.unused_seeds = random.sample(self.seeds, k=len(self.seeds))
    return self.unused_seeds.pop()
```

etam (changes requested):

golem/network/p2p/p2pservice.py
```diff
@@ -140,7 +141,8 @@ def connect_to_seeds(self):
         if not self.connect_to_known_hosts:
             return
 
-        for ip_address, port in random.sample(self.seeds, k=len(self.seeds)):
+        for _ in range(len(self.seeds)):
+            ip_address, port = self._get_next_random_seed()
```
jiivan, line 145: May I suggest using a generator instead?

It would look smth like this:
```python

current_seed = next(self.seeds_round_robin())
try:
   # ...
   self.connect(...)
except Exception:
   logger
   return
return

def seeds_round_robin(self):
    """Endlessly yield seeds ensuring every item is yielded in each iteration"""
    while True:
        current_sample = random.sample(...)
        for seed in current_sample:
            yield seed
```

No added instance attributes, cleaner (IMHO) logic of `connect_to_seeds()`

jiivan (changes requested):
May I suggest using a generator instead?

golem/network/p2p/p2pservice.py
etam: But here you're creating a new generator every time `seeds_round_robin` is called. This generator must be stored in `self` for reuse.

etam:
```python
def random_round_robin(items):
    while True:
        sample = random.sample(items, k=len(items))
        for item in sample:
            yield item

class P2PService:
    def __init__(self):
        # ...
        self.seeds_round_robin = random_round_robin(self.seeds)

    def connect_to_seed(self):
        # ...
        for ip_address, port in self.seeds_round_robin:
```

golem/network/p2p/p2pservice.py
etam: The idea with generator looks even better.

golem/network/p2p/p2pservice.py
jiivan: Yeah, you're right. We could also use `itertools.chain()`, but there is still problem with `__sync_seeds()` where self.seeds is updated. Maybe we should restart the generator there.

Krigpl: Yeah, I tried using generator but came with those exact conclusions. And I don't think storing generator is more legible than the current version.

golem/network/p2p/p2pservice.py
Krigpl: I like it but it's slightly inferior to the above because `self.seeds` can change between calls to `_get_next_random_seed` and you'll end up connecting to a seed which you know you shouldn't.

golem/network/p2p/p2pservice.py
jiivan: Or... `self.seeds` could be a custom set that implements randomized iterator AND resets iterator when it's modified.

https://docs.python.org/3/library/stdtypes.html#frozenset.update

```python

class SampleSet(set):
    def __iter__(self):
        # ...
    def add()
    def update()
   def remove() etc:
      reset generator
```

Krigpl: :(

etam: But when set is slightly changed (like one seed added), then you're resetting whole thing. The current solution remembers used seeds.

etam (approved):
Given that `self.seeds` can change I think this is a good solution.

jiivan (approved):
Thanks for comments. Last remark (not critical) is that if for some reason `self.seeds` would be empty the `while` would loop infinitely. Some check could be helpful.

PR #1734, opened 2017-12-07
Convert price from UI to an int, not float

Krigpl:
We should prefer integers everywhere where possible over floats.

### DIFF
--- a/apps/core/task/coretask.py
+++ b/apps/core/task/coretask.py
@@ -571,7 +571,7 @@ class CoreTaskBuilder(TaskBuilder):
     def build_full_definition(cls, task_type: CoreTaskTypeInfo, dictionary):
         definition = cls.build_minimal_definition(task_type, dictionary)
         definition.task_name = dictionary['name']
-        definition.max_price = float(dictionary['bid']) * denoms.ether
+        definition.max_price = int(float(dictionary['bid']) * denoms.ether)
 
         definition.full_task_timeout = string_to_timeout(
             dictionary['timeout'])
--- a/apps/core/task/coretask.py
+++ b/apps/core/task/coretask.py
@@ -1,5 +1,6 @@
 import abc
 import copy
+import decimal
 import golem_messages.message
 import logging
 import os
@@ -571,7 +572,8 @@ class CoreTaskBuilder(TaskBuilder):
     def build_full_definition(cls, task_type: CoreTaskTypeInfo, dictionary):
         definition = cls.build_minimal_definition(task_type, dictionary)
         definition.task_name = dictionary['name']
-        definition.max_price = int(float(dictionary['bid']) * denoms.ether)
+        definition.max_price = \
+            int(decimal.Decimal(dictionary['bid']) * denoms.ether)
 
         definition.full_task_timeout = string_to_timeout(
             dictionary['timeout'])
--- a/golem/task/taskbase.py
+++ b/golem/task/taskbase.py
@@ -62,11 +62,11 @@ class TaskHeader(object):
                  resource_size=0,
                  estimated_memory=0,
                  min_version=APP_VERSION,
-                 max_price=0.0,
+                 max_price: int=0,
                  docker_images=None,
                  signature=None):
         """
-        :param float max_price: maximum price that this (requestor) node may
+        :param int max_price: maximum price that this (requestor) node may
         pay for an hour of computation
         :param docker_images: docker image specification
         """
--- a/golem/task/taskbase.py
+++ b/golem/task/taskbase.py
@@ -66,7 +66,7 @@ class TaskHeader(object):
                  docker_images=None,
                  signature=None):
         """
-        :param int max_price: maximum price that this (requestor) node may
+        :param max_price: maximum price that this (requestor) node may
         pay for an hour of computation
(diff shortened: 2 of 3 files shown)


golem/task/taskbase.py
```diff
@@ -62,11 +62,11 @@ def __init__(self,
                  resource_size=0,
                  estimated_memory=0,
                  min_version=APP_VERSION,
-                 max_price=0.0,
+                 max_price: int=0,
```
etam, line 65: In what unit is this `max_price`?

apps/core/task/coretask.py
```diff
@@ -571,7 +571,7 @@ def build_definition(cls, task_type: CoreTaskTypeInfo, dictionary, minimal=False
     def build_full_definition(cls, task_type: CoreTaskTypeInfo, dictionary):
         definition = cls.build_minimal_definition(task_type, dictionary)
         definition.task_name = dictionary['name']
-        definition.max_price = float(dictionary['bid']) * denoms.ether
+        definition.max_price = int(float(dictionary['bid']) * denoms.ether)
```
etam: Which type is under `dictionary['bid']`? Which unit?

(ah, the joys of dynamic typing)

golem/task/taskbase.py
Krigpl: In whichever you imagine since this is Python. But seriously it is in wei equivalent for GNT

apps/core/task/coretask.py
Krigpl: `string`

Krigpl: `bid = 1` means 1 GNT

golem/task/taskbase.py
```diff
@@ -62,11 +62,11 @@ def __init__(self,
                  resource_size=0,
                  estimated_memory=0,
                  min_version=APP_VERSION,
-                 max_price=0.0,
+                 max_price: int=0,
                  docker_images=None,
                  signature=None):
         """
-        :param float max_price: maximum price that this (requestor) node may
+        :param int max_price: maximum price that this (requestor) node may
```
etam: There's no need to repeat type here.

golem/task/taskkeeper.py
```diff
@@ -17,7 +17,7 @@
 logger = logging.getLogger('golem.task.taskkeeper')
 
 
-def compute_subtask_value(price, computation_time):
+def compute_subtask_value(price: int, computation_time: int):
```
etam, line 20: In what unit is computation_time?

The documentation of `ReportComputedTask` in golem_messages says it's type is float.

etam (changes requested):

golem/task/taskkeeper.py
Krigpl: `computation_time` is in seconds so is supposed to be an `int`. I'll check that, but I wouldn't trust documentation much tbh.

Krigpl: `computation_time` originally comes from the the UI in the form of "00:10:00" for 10minutes.
Then it gets converted to seconds using `golem.core.common.string_to_timeout` which returns an integer. Should be fine.

etam:
\<offtop\>
What if I told you that unit mismatch could be a compilation error?
http://www.boost.org/doc/libs/1_65_1/libs/units/example/quantity.cpp
\</offtop\>

golem/task/taskbase.py
Krigpl: True, but is it really necessary to request changes for such minor thing?

etam: If you're already touching it, then why not?

golem/task/taskkeeper.py
etam: In `TaskHeader` fields `subtask_timeout` and `deadline` should also be ints.

golem/task/taskbase.py
etam: The comment about `ReportComputedTask` is also a request for changes, but in another repo.

apps/core/task/coretask.py
jiivan: @Krigpl `decimal.Decimal` instead of `float`?

golem/task/taskbase.py
Krigpl: We probably have different conventions but I request changes when I'm not fine with committing the changes the way they are, and would just approve them if my only concerns are not impactful in any way like this one (it's a harmless comment after all).
Context switching for responding to such a minor is usually not worth it.

golem/task/taskkeeper.py
etam: And while at it, you could add comments that those ints are seconds.

Krigpl: True, but since this is a fix for a bug in current master branch I'd rather not make irrelevant changes to it.

jiivan (approved):

etam (approved):

golem/task/taskkeeper.py
Krigpl: Don't get me wrong, I agree with your points and would usually apply them, but for small concrete bug fixes like this one I'm reluctant to do such things.

PR #1742, opened 2017-12-11
Fill ForceReportComputedTask with contents

jiivan:
Utilises addition of MessageHistoryService.

### DIFF
--- a/tests/golem/task/test_tasksession.py
+++ b/tests/golem/task/test_tasksession.py
@@ -1,3 +1,4 @@
+import datetime
 import os
 import pathlib
 import pickle
@@ -16,6 +17,7 @@ from golem.core.keysauth import KeysAuth, EllipticalKeysAuth
 from golem.core.variables import PROTOCOL_CONST
 from golem.docker.environment import DockerEnvironment
 from golem.docker.image import DockerImage
+from golem.network import history
 from golem.network.p2p.node import Node
 from golem.network.transport.tcpnetwork import BasicProtocol
 from golem.resource.client import ClientOptions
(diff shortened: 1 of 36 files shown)


tests/golem/task/test_tasksession.py
```diff
@@ -644,6 +650,66 @@ def test_react_to_subtask_payment_request(self, inform_mock) -> None:
...
+    def test_send_report_computed_task_concent_no_service(self):
+        history.MessageHistoryService.instance = None
+        ts = TaskSession(Mock())
+        ts.sign = lambda x: b'\0' * message.Message.SIG_LEN
+        ts.verified = True
+        n = Node()
+        wtr = WaitingTaskResult("xyz", "xxyyzz", "result", ResultType.DATA,
+                                13190, 10, 0, "10.10.10.10",
+                                30102, "key1", n)
+        with patch('golem.network.history.MessageHistoryService.get_sync')\
```
Krigpl: As I understand this a way to check and we returned early from the function and didn't send anything to Concent.
This check depends on this particular implementation, more natural check would be a check against `self.concent_service.submit` not being called, since this is the side effect that should not happen.

tests/golem/task/test_tasksession.py
```diff
@@ -644,6 +650,66 @@ def test_react_to_subtask_payment_request(self, inform_mock) -> None:
...
+        n = Node()
+        wtr = WaitingTaskResult("xyz", "xxyyzz", "result", ResultType.DATA,
+                                13190, 10, 0, "10.10.10.10",
+                                30102, "key1", n)
+        try:
+            history.MessageHistoryService()
+            ts.send_report_computed_task(wtr, "10.10.10.10", 30102, "0x00", n)
+            ts.concent_service.submit.assert_not_called()
+        finally:
+            history.MessageHistoryService.instance = None
```
Krigpl: Why is this needed ?

tests/golem/task/test_tasksession.py
```diff
@@ -644,6 +650,66 @@ def test_react_to_subtask_payment_request(self, inform_mock) -> None:
...
+
+    def test_send_report_computed_task_concent_succes(self):
```
Krigpl: `s/succes/success/`

Krigpl (changes requested):

tests/golem/task/test_tasksession.py
jiivan: Moved to `tearDown()`. It resets `MessageHistoryService` state to it's default state.

tests/golem/task/test_tasksession.py
jiivan: Thanks for  pointing this out! It was a mistake. `get_sync` doesn't depend on `instance`.

Krigpl (approved):

tests/golem/task/test_tasksession.py
Krigpl: I don't know the internals of how python tests work, but I understand that each test is not run in a separate process and we need to perform cleanup of the global state after each test?

PR #1743, opened 2017-12-11
Add required files to develop before buildbot can run unit tests

maaktweluit:
Required step for #1400 

After this PR the unit tests are able to run on develop for the final production setup.

### DIFF
--- a/scripts/test-daemon-start.ps1
+++ b/scripts/test-daemon-start.ps1
@@ -1,0 +1,14 @@
+
+"Ensure hyperg is started"
+
+$HyperdriveProcess = Get-Process hyperg -ErrorAction SilentlyContinue
+
+if (!$HyperdriveProcess)
+{
+  "Starting..."
+  $HyperdriveProcess = Start-Process C:\BuildResources\hyperg\hyperg.exe -WorkingDirectory C:\Users\buildbot-worker -PassThru
+}
+
+$HyperdriveProcess
+
+$HyperdriveProcess.Id | Out-File .test-daemon-hyperg.pid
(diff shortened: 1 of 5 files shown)


scripts/test-daemon-start.ps1
```diff
@@ -0,0 +1,14 @@
+
+"Ensure hyperg is started"
+
+$HyperdriveProcess = Get-Process hyperg -ErrorAction SilentlyContinue
```
etam, line 4: Is it OK if one is already running?

maaktweluit: Yes, even when its started by another user it worked. It was only not able to kill at the end.

This check is in there so when a user ( or disconnect ) stopped the build it wont break the next time.

Looking at it now it might be better to kill and re-start in this case to be safer, will put it on the list of improvements as follow up issue.

maaktweluit:
@etam @Elfoniok added one more file i missed, please check again

scripts/test-daemon-start.ps1
maaktweluit: Added it as possible improvement to #1754

etam (approved):
I have no experience with *.ps1 nor *.iss files, but the changes look reasonable.

scripts/test-daemon-start.ps1
```diff
@@ -0,0 +1,14 @@
...
+
+if (!$HyperdriveProcess)
+{
+  "Starting..."
+  $HyperdriveProcess = Start-Process C:\BuildResources\hyperg\hyperg.exe -WorkingDirectory C:\Users\buildbot-worker -PassThru
```
mfranciszkiewicz, line 9: Doesn't this create a `Hyperg.db` directory in `C:\BuildResources\hyperg` and bundle it with the installer?
You can change hyperg's directory with a `--db` param.

maaktweluit: good catch! I thought it would be created in "working dir" will check and change if required :)

maaktweluit: Checked on the worker and it looks like the working-dir is used, so no change needed.

mfranciszkiewicz: That's good to hear

mfranciszkiewicz (approved):

PR #1745, opened 2017-12-11
Fixed: restartable MessageHistoryService

mfranciszkiewicz:
`Service`s can be restarted via `Client.pause/resume`. It was not possible with `MessageHistoryService`, which was derived from `threading.Thread`.

### DIFF
--- a/golem/network/history.py
+++ b/golem/network/history.py
@@ -43,16 +43,16 @@ class MessageHistoryService(IService, threading.Thread):
 
     def __init__(self):
         IService.__init__(self)
-        threading.Thread.__init__(self, daemon=True)
 
         if self.__class__.instance is None:
             self.__class__.instance = self
 
+        self._started = None  # set in start by Thread.__init__
+        self._queue_timeout = None  # set in start
         self._stop_event = threading.Event()
         self._save_queue = queue.Queue()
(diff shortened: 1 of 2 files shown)


etam:
Maybe it would be cleaner to use composition, instead of inheriting from Thread? This delayed initialization of Thread class looks like a dirty hack.

mfranciszkiewicz:
@etam Either way I'd have to re-create the thread. Unless "stop" doesn't stop the service.

etam:
@mfranciszkiewicz recreating Thread object is much cleaner solution.

etam (approved):

golem/network/history.py
```diff
@@ -43,33 +43,41 @@ class MessageHistoryService(IService, threading.Thread):
...
-        return self._started.is_set() and not self._stop_event.is_set()
+        return (
+            self._thread and
+            self._thread.is_alive() and
+            not self._stop_event.is_set()
+        )
 
     def start(self) -> None:
-        threading.Thread.start(self)
+        if not self.running:
```
jiivan:
```python
if self.running:
    return
```
?

jiivan (approved):
LGTM. One suggestion.

PR #1746, opened 2017-12-11
Fix batch payment cost estimation

chfast:
Consider also the cost of subtracting payee balance in the end of batch payment.

This is a hotfix. The review of gas cost estimations should be done afterwards. See #1746.


Krigpl (approved):

Krigpl:
I'll let myself merge it

PR #1749, opened 2017-12-11
Message history fix: message.TaskToCompute and NetworkMessage constraints

mfranciszkiewicz:
Subtask to task mapping does not exist when receiving a `TaskToCompute` message.

Additionally, try to save all (annotated) messages retrieved from the network. Constraint check may fail on save and will be logged.

### DIFF
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -186,11 +186,7 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin,
     ###################################
 
     def message_to_model(self, msg, local_role, remote_role):
-        task = getattr(msg, 'task_id', None)
-        subtask = getattr(msg, 'subtask_id', None)
-
-        if subtask and not task:
-            task = self._subtask_to_task(subtask, local_role)
+        task, subtask = self._task_subtask_from_message(msg, local_role)
 
         return dict(
             task=task,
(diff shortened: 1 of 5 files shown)


golem/task/tasksession.py
```diff
@@ -201,7 +197,22 @@ def message_to_model(self, msg, local_role, remote_role):
...
             remote_role=remote_role,
-        ) if task else None
+        )
+
+    def _task_subtask_from_message(self, msg, local_role):
+        task, subtask = None, None
+
+        if isinstance(msg, message.TaskToCompute):
+            definition = msg.compute_task_def
+            if isinstance(definition, dict):
```
jiivan:
```python
if definition is None:
    return
```
Maybe this is more explicit? Since golem-messagesv1.2.0 `compute_task_def` will always be an instance of `ComputeTaskDef`. https://github.com/golemfactory/golem-messages/blob/master/golem_messages/message.py#L725

tests/golem/task/test_tasksession.py
```diff
@@ -326,7 +327,7 @@ def __reset_mocks():
             conn.reset_mock()
 
         # msg.ctd is None -> failure
-        msg = message.TaskToCompute()
+        msg = message.TaskToCompute(raw=b'\0')
```
jiivan, line 330: Is this `raw` parameter really needed? Maybe it would be better to adjust `golem-messages` to handle it automatically?

jiivan (dismissed):
LGTM. small issues

tests/golem/task/test_tasksession.py
mfranciszkiewicz: Sure, let's address this issue in a new PR.

Elfoniok (approved):

PR #1750, opened 2017-12-11
Message history fix: database locking

mfranciszkiewicz:
`NetworkMessage` objects created in threads other than `MessageHistoryService`'s may cause a database lock-up.

### DIFF
--- a/golem/network/history.py
+++ b/golem/network/history.py
@@ -98,13 +98,13 @@ class MessageHistoryService(IService, threading.Thread):
 
         return list(result)
 
-    def add(self, msg: NetworkMessage) -> None:
+    def add(self, msg_dict: dict) -> None:
         """
-        Appends the message to the save queue.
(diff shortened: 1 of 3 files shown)


Krigpl (approved):

golem/network/history.py
```diff
@@ -229,13 +230,13 @@ class IMessageHistoryProvider(ABC):
     @abstractmethod
     def message_to_model(self, msg: 'golem_messages.message.Message',
                          local_role: Actor,
-                         remote_role: Actor) -> NetworkMessage:
+                         remote_role: Actor) -> dict:
         """
         Convert a message to its database model representation.
```
etam: A comment about why it's not a NetworkMessage object here yet would be helpful.

etam (changes requested):

etam (approved):

PR #1753, opened 2017-12-12
Hold ethereum-utils version

Elfoniok:
Fixes this issue:
https://github.com/golemfactory/golem/issues/1755

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,6 +20,8 @@ cffi==1.10.0
 web3==3.8.0
 ethereum==1.6.1
 ethereum-abi-utils==0.4.0
+#Newer version contains no code. Only deprecation warning.
+#https://pypi.python.org/pypi/ethereum-utils
 ethereum-utils==0.5.1
 autobahn==0.18.1
 cbor2==3.0.4
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,6 +20,7 @@ cffi==1.10.0
 web3==3.8.0
 ethereum==1.6.1
 ethereum-abi-utils==0.4.0
+ethereum-utils==0.5.1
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -20,6 +20,7 @@ cffi==1.10.0
 web3==3.8.0
 ethereum==1.6.1
 ethereum-abi-utils==0.4.0
+ethereum-utils==0.5.1
```
maaktweluit: Would be nice to add a comment why its fixed
Maybe you can link the issue to upgrade the package, is that made?

maaktweluit (changes requested):

requirements.txt
Elfoniok: Comment in requirement file?

etam: Like in requirements-test.txt. Start with `#`.

Elfoniok: That is a new custom but ok

maaktweluit (approved):

maaktweluit:
Thanks for the update!

Last weeks we had some breaking frozen packages and found out future self will be happy with these comments. This way we can upgrade more safely and quickly if we know the root cause.

requirements.txt
```diff
@@ -20,6 +20,8 @@ cffi==1.10.0
 web3==3.8.0
 ethereum==1.6.1
 ethereum-abi-utils==0.4.0
+#Package changed name
```
etam: That doesn't explain anything.
That would be better:
`Newer version contains no code. Only deprecation warning.`

etam (changes requested):

mfranciszkiewicz (approved):

PR #1757, opened 2017-12-12
Task persistence fix: proper resource restore edge-case handling

mfranciszkiewicz:
- fixes an edge case of re-adding resources to HyperG
- encapsulates HyperG daemon's error responses in `HTTPError`s

### DIFF
--- a/golem/resource/hyperdrive/resourcesmanager.py
+++ b/golem/resource/hyperdrive/resourcesmanager.py
@@ -94,7 +94,7 @@ class HyperdriveResourceManager(ClientHandler, AbstractResourceManager):
                                         id=task_id,
                                         client_options=client_options,
                                         obj_id=str(uuid.uuid4()),
-                                        raise_exc=bool(resource_hash))
+                                        raise_exc=True)
 
         file_list = list(files.values())
         self._cache_response(file_list, response, task_id)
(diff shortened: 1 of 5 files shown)


golem/resource/hyperdrive/resourcesmanager.py
```diff
@@ -94,7 +94,7 @@ def _add_files(self, files, task_id, resource_hash=None,
                                         id=task_id,
                                         client_options=client_options,
                                         obj_id=str(uuid.uuid4()),
-                                        raise_exc=bool(resource_hash))
+                                        raise_exc=True)
```
maaktweluit, line 97: is this flag still used elsewhere? if all change to `True` it can become default behaviour

maaktweluit (approved):
LGTM

golem/resource/hyperdrive/resourcesmanager.py
mfranciszkiewicz: Yes, it is used in other places. BTW this part of resource management is going to be refactored.

Elfoniok:
That solves problem discovered yesterday with some deserialized tasks crashing applications.

jiivan (approved):
LGTM

Elfoniok (approved):

PR #1758, opened 2017-12-13
Prevent overwriting issue with incremental filename 

mdtanrikulu:
Solution for (golemfactory/golem-electron#134)


apps/core/task/coretask.py
```diff
@@ -598,4 +606,48 @@ def get_output_path(cls, dictionary, definition):
...
+
+        path_with_ext = os.path.join(path, name + extension)
+
+        if not os.path.exists(path_with_ext):
+            return fname_path
+
+        i = 1
+        new_fname = "{}({})".format(fname_path, i)
+
+        while os.path.exists(os.path.join(path, name + "(" + str(i) + ")" + extension)):
```
mfranciszkiewicz: Maybe
`while os.path.exists(new_fname + extension):`
instead?

mdtanrikulu: yeap, make sense. 👍

mfranciszkiewicz (approved):

badb (approved):
We can merge, but keep in mind that it's not solving the issue completely. It works only for single-frame scenes. And will not work for mulit-frames scenes. Making this work for multiframes will be much more complicated and quite tricky.

PR #1759, opened 2017-12-13
Fixes

Elfoniok:
  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/home/elfoniok/golem/golem/network/history.py", line 62, in run
    self._loop()
  File "/home/elfoniok/golem/golem/network/history.py", line 202, in _loop
    msg = NetworkMessage(**msg_dict)
TypeError: BaseModel object argument after ** must be a mapping, not NetworkMessage

### DIFF
--- a/golem/network/history.py
+++ b/golem/network/history.py
@@ -106,21 +106,23 @@ class MessageHistoryService(IService, threading.Thread):
         if msg_dict:
             self._save_queue.put(msg_dict)
 
-    def add_sync(self, msg: NetworkMessage) -> None:
+    def add_sync(self, msg: NetworkMessage) -> bool:
         """
         Saves a message in the database.
         :param msg: Message to save
         """
         try:
             msg.save()
+            return True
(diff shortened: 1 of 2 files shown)


golem/network/history.py
```diff
@@ -106,21 +106,23 @@ def add(self, msg_dict: dict) -> None:
         if msg_dict:
             self._save_queue.put(msg_dict)
 
-    def add_sync(self, msg: NetworkMessage) -> None:
+    def add_sync(self, msg: NetworkMessage) -> bool:
```
mfranciszkiewicz: Could we use a msg_dict and create a NetworkMessage instance here? Returning True on error is a bit misleading.

mfranciszkiewicz (changes requested):
Thanks for fixing, I definitely missed this part. I have one comment though

jiivan (approved):

mfranciszkiewicz (approved):

PR #1771, opened 2017-12-15
Cleanup gas and reserved eth estimations

Krigpl:
Remove some TODOs and FIXMEs
The value of `eth_reserved` is now always at least the value needed for a single `batchTransfer`.

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -49,17 +49,15 @@ class PaymentProcessor(LoopingCallService):
     # Default deadline in seconds for new payments.
     DEFAULT_DEADLINE = 10 * 60
 
-    # Gas price: 20 shannons, Homestead suggested gas price.
+    # Gas price: 20 gwei, Homestead suggested gas price.
     GAS_PRICE = 20 * 10 ** 9
 
-    # Max gas cost for a single payment. Estimated in tests.
-    SINGLE_PAYMENT_GAS_COST = 60000
-
-    SINGLE_PAYMENT_ETH_COST = GAS_PRICE * SINGLE_PAYMENT_GAS_COST
-
-    # Gas reservation for performing single batch payment.
-    # TODO: Adjust this value later and add MAX_PAYMENTS limit.
-    GAS_RESERVATION = 21000 + 1000 * 50000
+    # Total gas for a batchTransfer is BASE + len(payments) * PER_PAYMENT
+    GAS_PER_PAYMENT = 30000
+    ETH_PER_PAYMENT = GAS_PRICE * GAS_PER_PAYMENT
+    # tx: 21000, balance substract: 5000, arithmetics < 800
+    GAS_BATCH_PAYMENT_BASE = 21000 + 800 + 5000
+    ETH_BATCH_PAYMENT_BASE = GAS_PRICE * GAS_BATCH_PAYMENT_BASE
 
     # Time required to reset the current balance when errors occur
     BALANCE_RESET_TIMEOUT = 30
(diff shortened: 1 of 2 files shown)


golem/ethereum/paymentprocessor.py
```diff
@@ -52,14 +52,11 @@ class PaymentProcessor(LoopingCallService):
...
-
-    SINGLE_PAYMENT_ETH_COST = GAS_PRICE * SINGLE_PAYMENT_GAS_COST
-
-    # Gas reservation for performing single batch payment.
-    # TODO: Adjust this value later and add MAX_PAYMENTS limit.
-    GAS_RESERVATION = 21000 + 1000 * 50000
+    # Total gas for a batchTransfer is BASE + len(payments) * PER_PAYMENT
+    GAS_PER_PAYMENT = 30000
+    ETH_PER_PAYMENT = GAS_PRICE * GAS_PER_PAYMENT
+    GAS_BATCH_PAYMENT_BASE = 21000 + 800 + 5000
```
chfast: Add comment: `tx: 21000, balance subtract: 5000, arithmetic: <800`

golem/ethereum/paymentprocessor.py
```diff
@@ -52,14 +52,11 @@ class PaymentProcessor(LoopingCallService):
     # Gas price: 20 shannons, Homestead suggested gas price.
     GAS_PRICE = 20 * 10 ** 9
```
Elfoniok: Could you change this also to be expressed with denoms.ether

Elfoniok (approved):

golem/ethereum/paymentprocessor.py
chfast: I think using Gwei is fine, but you can update the comment above.

tests/golem/transactions/ethereum/test_paymentprocessor.py
```diff
@@ -377,21 +377,24 @@ def test_monitor_progress(self):
...
         assert tx.hash in inprogress
         assert inprogress[tx.hash] == [p]
         assert self.pp.gnt_balance(True) == balance_gnt - gnt_value
+        assert self.pp.eth_balance(True) == balance_eth_after_sendout
         assert self.pp._gnt_reserved() == 0
         assert self.pp._gnt_available() == balance_gnt - gnt_value
-        assert self.pp._eth_reserved() == PaymentProcessor.SINGLE_PAYMENT_ETH_COST
-        assert self.pp._eth_available() == balance_eth - PaymentProcessor.SINGLE_PAYMENT_ETH_COST
+        assert self.pp._eth_reserved() == PaymentProcessor.ETH_BATCH_PAYMENT_BASE  # noqa
+        assert self.pp._eth_available() == balance_eth_after_sendout - PaymentProcessor.ETH_BATCH_PAYMENT_BASE  # noqa
```
chfast: Is it worth it to put `# noqa` everywhere instead of making the expressions shorter?

golem/ethereum/paymentprocessor.py
Elfoniok: why not express this in 20 * denoms.shannon ? This is what i meant (ether was just example)

Elfoniok: Ok i just learned that all fractional units of ethers are obsolete. I would argue for creating then units that are already using like Gwei. And using this in our code.

tests/golem/transactions/ethereum/test_paymentprocessor.py
Krigpl: I'd rather not reduce var names' legibility in favor of shorter lines, but I'm open to suggestions here

chfast:
```
eth_available = balance_eth_after_sendout - PaymentProcessor.ETH_BATCH_PAYMENT_BASE
assert self.pp._eth_available() == eth_available
```

PR #1772, opened 2017-12-15
Use pending block for transaction count

Krigpl:
It may happen we will try to send more than one transaction before a block is mined in which case we will use incorrect nonce for all of them except the first one.
Currently it may happen for example when we are requesting GNT from the faucet - we do it every 13 seconds or so by sending a transaction to the faucet and since the time between mined block is slightly higher it may happen those transactions will share nonce.

### DIFF
--- a/golem/ethereum/client.py
+++ b/golem/ethereum/client.py
@@ -62,12 +62,15 @@ class Client(object):
 
     def get_transaction_count(self, address):
         """
-        Returns the number of transactions
-        that have been sent from account
+        Returns the number of transactions that have been sent from account.
+        Use `pending` block to account the transactions that haven't been mined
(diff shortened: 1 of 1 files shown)


jiivan:
Is it possible to mock some test that simulates this scenario?

Krigpl:
@jiivan Depends on what you want to test. We mock geth client in the tests anyway, so the only thing that seems like sort of worth testing is whether `block=pending` is being passed to `getTransactionCount`. I can add it but I don't think is has great positive value :)

jiivan:
Then maybe it's a good idea to add comment why `'pending'` parameter is so important.

chfast (approved):
I second adding a comment in code.

PR #1774, opened 2017-12-18
Fixed version of ipaddress dependency

mfranciszkiewicz:


etam (dismissed):
Please add a comment, why fixed version is required.

Elfoniok (approved):

requirements.txt
```diff
@@ -3,7 +3,8 @@ setuptools==34.3.3
 appdirs>=1.4
 click>=4.0
 docker-py==1.7.2
-ipaddress>=1.0.18
+# Fixed version of crossbar dependency.
+ipaddress==1.0.18
```
chfast, line 7: The `ipaddress` is a part of Python 3.3+, there should not be a reason to install it with pip.

chfast (changes requested):

chfast:
Moreover, you do you need 2 commits to do this single change?

mfranciszkiewicz:
@chfast The point is to maintain change history (i.e. request for a comment). PRs can be squashed.

PR #1778, opened 2017-12-19
Fix for NodeProcess.identify_chain startup crash (geth)

mfranciszkiewicz:


golem/ethereum/node.py
```diff
@@ -177,11 +181,23 @@ def identify_chain(self):
...
-        genesis = self.web3.eth.getBlock(0)['hash']
+        genesis = genesis_block['hash']
         chain = GENESES.get(genesis, 'unknown')
         log.info("{} chain ({})".format(chain, genesis))
         return chain
 
+    def get_genesis_block(self):
+        try:
+            return self.web3.eth.getBlock(0)
+        except Exception:  # pylint:disable=broad-except
```
chfast, line 192: Can you narrow the exception type?

chfast (approved):

golem/ethereum/node.py
mfranciszkiewicz: Unfortunately each web3 provider throws different types of exceptions (ipc socket, http from requests, value errors) and error handling varies in each. The assumption here was that we are connected to the client and these exceptions are temporary. I did not want to assume that we're using the http provider only.

golem/ethereum/node.py
```diff
@@ -167,7 +171,7 @@ def is_connected(self):
         except AssertionError:  # thrown if not all required APIs are available
             return False
 
-    def identify_chain(self):
+    def identify_chain(self, genesis_block):
```
etam, line 174: this can be a `@staticmethod`

etam (approved):

PR #1779, opened 2017-12-19
Proposal: update eth/gnt balance on error only if some timeout has passed

mfranciszkiewicz:
This addresses the rapidly changing eth / gnt balance bug

### DIFF
--- a/golem/ethereum/client.py
+++ b/golem/ethereum/client.py
@@ -92,7 +92,7 @@ class Client(object):
             return self.web3.eth.getBalance(account, block)
         except ValueError as e:
             log.error("Ethereum RPC: {}".format(e))
-            return 0
+            return None
 
     def call(self, _from=None, to=None, gas=90000, gas_price=3000, value=0,
              data=None, nonce=0, block=None):
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -61,6 +61,9 @@ class PaymentProcessor(LoopingCallService):
     # TODO: Adjust this value later and add MAX_PAYMENTS limit.
     GAS_RESERVATION = 21000 + 1000 * 50000
 
+    # Time required to reset the current balance when errors occur
+    BALANCE_RESET_TIMEOUT = 10
+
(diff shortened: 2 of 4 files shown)


golem/ethereum/client.py
```diff
@@ -92,7 +92,7 @@ def get_balance(self, account, block=None):
             return self.web3.eth.getBalance(account, block)
         except ValueError as e:
             log.error("Ethereum RPC: {}".format(e))
```
chfast, line 94: What kind of error happens here? Using local or remote geth?

chfast (commented):
I'm ok with this hotfix.

In longer run, we should merge ethereum.Node and ethereum.Client and handle geth quirks in more generic way there. I don't want to end up with a ton of different timeout params for communication with Ethereum.

golem/ethereum/paymentprocessor.py
```diff
@@ -61,6 +61,9 @@ class PaymentProcessor(LoopingCallService):
     # TODO: Adjust this value later and add MAX_PAYMENTS limit.
     GAS_RESERVATION = 21000 + 1000 * 50000
 
+    # Time required to reset the current balance when errors occur
+    BALANCE_RESET_TIMEOUT = 10
```
Krigpl: I think we update the balances every 13 seconds or so. A value of 30 seconds might be more appropriate I guess.

Krigpl (approved):
Should be fine for now.

golem/ethereum/client.py
mfranciszkiewicz: As far as I can tell, `ValueError`s are thrown when the json response contains an `error` property. This is specific to the HTTP provider though.

tests/golem/transactions/ethereum/test_paymentprocessor.py
```diff
@@ -635,3 +635,14 @@ def test_no_gnt_available(self):
 
         q = Payment.create(subtask='F', payee=payee, value=value)
         assert not self.pp.add(q)
+
+    def test_balance_value(self):
```
etam, line 639: you may want to use `freezegun` package to avoid any timing issues.

etam (approved):
As a hot-fix it looks OK.

PR #1780, opened 2017-12-19
Refactor ethereumincomeskeeper a bit

Krigpl:
1. Moved magic constant `LOG_ID` to the payment processor - it's related to the token implementation so it belongs there, closer to the actual token logic.
2. Created `get_incomes_from_block` for `ethereumincomeskeeper` for easier communication - in particular it shouldn't know about the above mentioned `LOG_ID`
3. If I understand correctly I also fixed potential vulnerability since `address` was not passed in the `get_logs` method, meaning we were listening to all `Transfer` events, not only generated from `GNT` contract.

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -71,6 +71,9 @@ class PaymentProcessor(LoopingCallService):
     # Minimal number of confirmations before we treat transactions as done
     REQUIRED_CONFIRMATIONS = 12
 
+    # keccak256(Transfer(address,address,uint256))
+    LOG_ID = '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'  # noqa
+
     def __init__(self, client: Client, privkey, faucet=False) -> None:
         self.__client = client
         self.__privkey = privkey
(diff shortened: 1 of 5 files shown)


golem/ethereum/paymentprocessor.py
```diff
@@ -68,6 +68,9 @@ class PaymentProcessor(LoopingCallService):
     # Minimal number of confirmations before we treat transactions as done
     REQUIRED_CONFIRMATIONS = 12
 
+    # keccak(Transfer(address,address,uint256))
```
chfast: Name it `keccak256(...)` for precision.

golem/transactions/ethereum/ethereumincomeskeeper.py
```diff
@@ -73,18 +69,19 @@ def received(self,
...
             logger.debug(
                 'INCOME: from %r to %r v:%r',
                 sender,
                 receiver,
                 log_value
             )
-            # Count tokens only when we're the receiver.
-            if receiver == my_address:
-                received_tokens += log_value
+            if receiver != my_address:
```
chfast: This check can be moved to `get_incomes_from_block()`.

golem/transactions/ethereum/ethereumincomeskeeper.py
```diff
@@ -73,18 +69,19 @@ def received(self,
         received_tokens -= spent_tokens
         for income_log in incomes:
             # Should we verify sender address?
-            sender = income_log['topics'][1]
-            receiver = income_log['topics'][2]
-            log_value = int(income_log['data'], 16)
+            sender = income_log['sender']
+            receiver = income_log['receiver']
+            log_value = income_log['value']
```
chfast: How about `income_value` or just `value`?

golem/ethereum/paymentprocessor.py
```diff
@@ -390,6 +393,23 @@ def get_gnt_from_faucet(self):
             return False
         return True
 
+    def get_incomes_from_block(self, block, address):
+        logs = self.get_logs(block,
+                             block,
+                             self.TESTGNT_ADDR,
+                             [self.LOG_ID, None, address])
+        if not logs:
```
chfast: Do you mean `None` here? If yes, maybe we should be specific about this?

chfast (changes requested):

golem/ethereum/paymentprocessor.py
Krigpl: Imo an empty list is also a valid return value for the above call, so this should be fine. Do you have anything else in mind?

chfast (approved):

Elfoniok (approved):

PR #1782, opened 2017-12-19
Reduce usage of PATH on windows

### LINKED ISSUE #1612: Try to remove usage of PATH during installation
We're adding several location to the PATH when we are installing golem on Windows. We're adding golem location, hyperg and geth. Adding things to PATH is quite tricky. You can override the PATH if it's too long, if's difficult to remove parts from it or format in in ugly way by adding double ";". 

We should check if we can solve in other way. It should be fine to create other environment variable with more precise names, eg. "GOLEM", "HYPERG". The shortcut can then point to "GOLEM/golem.exe" location. 
This change will require several changes:
* GUI should try to start "GOLEM/golem.exe" and not just golem.exe
* Golem should start "HYPERG/hyperg" service instead of just hyperg
* Geth can be removed and it's installation should be optional, user should be able to connect hers own geth installation with golem.

maaktweluit:
Resolves #1612

~Blocked by golemfactory/golem-messages#74~

All `OpenSSL` and `HyperG` files are now installed in the `${PF}\Golem` folder, this reduces the need for PATH variables.

Contains extra fix for `OpenSSL`: when loading fails on windows the exe folder gets put in front of the PATH before retry.


badb (commented):
I've tested it with broken OpenSSL on PATH that stopped golem from source being used and it worked great for installers! However, keep in mind that right now we duplicate crypto code here: https://github.com/golemfactory/golem-messages/blob/master/golem_messages/cryptography.py and the changes that you proposed only apply to golem crypto file.

badb (approved):

maaktweluit:
@badb Thank you for checking out the fix!

Sounds like the crypto vs cryptography issue will break this when merged.
Would propose to:
- Use `golem_messages.cryptography` and make this the leading location for this logic
- Add `golem_cryptography` and move all related logic here.

Propose to go for option 1 now, and make a follow up issue for option 2.

maaktweluit:
Made good steps today on this issue:
- Moved the fix to golem_messages.cryptography
- Removed golem.core.crypto
- Moved tests and missing logic from crypto to cryptography
- Made PR in golem_messages, blocking this one
- Made #1880 as follow up issue to make it its own package

jiivan (approved):
I like the refactoring part. It's hard for me to verify if helps  with windows issues.

mfranciszkiewicz (approved):

PR #1791, opened 2017-12-20
Extract token logic into a separate class

Krigpl:
Abstract all the token related logic into its own class. This way `paymentprocessor` doesn't need to know anything about the details of the token implementation. 
It makes the code clearer.
It makes testing easier.
It makes migrating to GNTW easier.
The world is a better place.

This PR doesn't include any functional changes (except for making `self._awaiting` thread safe), so it's relatively safe.

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -1,8 +1,11 @@
+import abc
 import json
 import logging
 import sys
 import time
+from threading import Lock
 from time import sleep
+from typing import Any, List
 
 from ethereum import abi, utils, keys
 from ethereum.transactions import Transaction
@@ -18,10 +21,8 @@ from .node import tETH_faucet_donate
 
 log = logging.getLogger("golem.pay")
 
-gnt_contract = abi.ContractTranslator(json.loads(TestGNT.ABI))
 
-
-def _encode_payments(payments):
+def encode_payments(payments: List[Payment]):
     paymap = {}
     for p in payments:
         if p.payee in paymap:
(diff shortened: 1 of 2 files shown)


golem/ethereum/paymentprocessor.py
```diff
@@ -42,7 +44,135 @@ def _encode_payments(payments):
...
+
+class AbstractToken(object, metaclass=abc.ABCMeta):
+    """
+    This is a common interface for token transactions for the process of
+    switching between GNT and GNTW. Add awaiting payments using
+    put_awaiting_payments. Grab the next ready batch using take_next_batch and
+    make a transaction for it. This way it is possible to hold on with a batch
+    until some preprocessing has been made to make sending that batch possible.
+    """
+    def __init__(self, client : Client, privkey):
```
chfast: privkey is bytes.

golem/ethereum/paymentprocessor.py
```diff
@@ -42,7 +44,135 @@ def _encode_payments(payments):
...
+class SimpleGNTToken(AbstractToken):
+    """
+    When the main token and the batchTransfer function are in the same contract.
+    Which is the case for tGNT in testnet and eventually (after migration)
+     will be the case for the new GNT in the mainnet as well.
+    """
+    TESTGNT_ADDR = decode_hex("7295bB8709EC1C22b758A8119A4214fFEd016323")
+
+    # keccak256(Transfer(address,address,uint256))
+    LOG_ID = '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef'  # noqa
```
chfast: I think it is better to rename it to `TRANSFER_LOG_ID` or `TRANSFER_EVENT_ID`.

golem/ethereum/paymentprocessor.py
```diff
@@ -42,7 +44,135 @@ def _encode_payments(payments):
...
+    @abc.abstractmethod
+    def take_next_batch(self) -> (List[Payment], Transaction):
+        pass
+
+    @abc.abstractmethod
+    def get_incomes_from_block(self, block: int, address) -> List[Any]:
+        pass
+
+
+class SimpleGNTToken(AbstractToken):
```
chfast: I hate this `Simple` prefix everywhere. Can't it be just `GNTToken`?

Krigpl: Right. This is simpler.

Elfoniok (approved):

PR #1795, opened 2017-12-20
Issue-63: upnp implmentation using miniupnpc lib

mackode:


badb:
Hi @mackode 

Thank you for contributing, this implementation however is not really integrated in Golem. Keep in mind that even if this implementation was using ports from the config and not hardcoded ones than they may not be ports that Golem is really using. Golem is trying to find first free ports that are bigger than the defined one. 

You can also check this branch that @mfranciszkiewicz has started working on: https://github.com/golemfactory/golem/compare/upnp

mackode:
Hi @badb 

Yes, indeed ports were hardcoded, mainly for testing. The most problematic thing I can see is building miniupnpc for Windows. Did you manage to solve it?

I'm not quite sure why Golem is searching for free ports? Actually, as a user or server maintainer I would like to control explicitly which ports are open. Can I read somewhere what is the purpose of it?

mfranciszkiewicz:
@mackode Hello, yes we've built `miniupnpc` for Windows.

Golem, by default, is searching for free ports if the usual ones are already taken. You can avoid this by setting `start_port = 40102` and `end_port = 40103` in your config file (Golem binds to 2 TCP sockets).

mfranciszkiewicz:
Thank you again for contributing! I'm closing this PR in favor of https://github.com/golemfactory/golem/pull/1859

PR #1800, opened 2017-12-21
Hotfix for resource peers (0.10.1)

mfranciszkiewicz:
- this is a 0.10.0 hotfix only (no protocol version changes), for `develop` see https://github.com/golemfactory/golem/pull/1768
- resource peer addresses are always passed to hyperg
- addresses are built by the local node (unlike https://github.com/golemfactory/golem/pull/1768)


maaktweluit (approved):
LGTM

PR #1803, opened 2017-12-22
Add timestamp property to Payment

Krigpl:
It's needed to correctly calculate `closure_time` for `batchTransfer`

### DIFF
--- a/golem/model.py
+++ b/golem/model.py
@@ -29,7 +29,7 @@ db = SqliteDatabase(None, threadlocals=True,
 
 class Database:
     # Database user schema version, bump to recreate the database
-    SCHEMA_VERSION = 6
+    SCHEMA_VERSION = 7
 
     def __init__(self, datadir):
         # TODO: Global database is bad idea. Check peewee for other solutions.
@@ -216,6 +216,7 @@ class Payment(BaseModel):
     payee = RawCharField()
     value = BigIntegerField()
     details = PaymentDetailsField()
+    processed_ts = IntegerField(null=True)
 
     def __init__(self, *args, **kwargs):
         super(Payment, self).__init__(*args, **kwargs)
@@ -226,13 +227,14 @@ class Payment(BaseModel):
     def __repr__(self) -> str:
         tx = self.details.tx
         bn = self.details.block_number
-        return "<Payment sbid:{!r} v:{:.3f} s:{!r} tx:{!r} bn:{!r}>"\
+        return "<Payment sbid:{!r} v:{:.3f} s:{!r} tx:{!r} bn:{!r} ts:{!r}>"\
             .format(
                 self.subtask,
                 float(self.value) / denoms.ether,
                 self.status,
                 tx,
-                bn
+                bn,
+                self.processed_ts
             )
 
     def get_sender_node(self) -> Optional[Node]:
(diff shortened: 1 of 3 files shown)


golem/model.py
```diff
@@ -214,6 +214,7 @@ class Payment(BaseModel):
     payee = RawCharField()
     value = BigIntegerField()
     details = PaymentDetailsField()
+    timestamp = IntegerField(null=True)
```
jiivan: DateTimeField is more customary in DB context: http://docs.peewee-orm.com/en/latest/peewee/models.html#datetimefield-datefield-and-timefield

Maybe call it `processed_at`? So it will be more obvious that it's a time of processing not creation/modification time.

tests/golem/transactions/ethereum/test_paymentprocessor.py
```diff
@@ -482,6 +482,16 @@ def test_failed_transaction(self):
         self.assertEqual(p.details.fee, 55001 * self.pp.GAS_PRICE)
         self.assertEqual(self.pp._gnt_reserved(), 0)
 
+    def test_payment_timestamp(self):
```
jiivan, line 487: You can use `freezegun` to test time related cases.

tests/golem/transactions/ethereum/test_paymentprocessor.py
```diff
@@ -482,6 +482,16 @@ def test_failed_transaction(self):
...
+        self.client.get_balance.return_value = denoms.ether
+        self.client.call.return_value = hex(denoms.ether)
+        ts = 7
+
+        p = Payment.create(subtask="p1", payee=urandom(20), value=1)
+        with patch('time.time', return_value=ts):
+            self.pp.add(p)
+        self.assertEqual(ts, p.timestamp)
```
jiivan: Does this test wether mock works properly? :)

jiivan (changes requested):

golem/model.py
Krigpl: > SQLite does not have a native date type, so dates are stored in formatted text columns.

I'd rather stay with integers

tests/golem/transactions/ethereum/test_paymentprocessor.py
Krigpl: What do you mean? It tests whether `timestamp` field has been set on payment during the `PaymentProcessor::add` call.

jiivan: It seems to me that this test is too low-level. Almost like testing:
```
a = 1
assert a == 1
```

Krigpl: It isn't, it tests exactly the lines of code I've added here:
```
            ts = int(time.time())
            with Payment._meta.database.transaction():
                payment.timestamp = ts
                payment.save()
```
This test will fail if those fail to execute. Why do you think it's too low level?

tests/golem/transactions/ethereum/test_paymentprocessor.py
Krigpl: I looked at the `freezegun` package and correct me if I'm wrong but you need to specify the time in human readable format, can't just provide a unix timestamp? Which would make this test less clear about what it's doing.

jiivan: Are you suggesting that human readable time is less readable than an integer/float? ;) By default you can use datetime instance or anything that is parsable by [dateutil](https://dateutil.readthedocs.io/en/stable/parser.html).

jiivan (approved):

PR #1804, opened 2017-12-22
Implement closure_time for GNTW::batchTransfer

Krigpl:
`batchTransfer` is supposed to handle all payments that happened before some `closure_time`. To avoid race condition between calling `batchTransfer` and adding new payments we look at the payments' timestamps and choose only those older than some specified threshold (currently set to 10 seconds).

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -1,4 +1,5 @@
 import abc
+import calendar
 import json
 import logging
 import sys
@@ -45,6 +46,11 @@ def encode_payments(payments: List[Payment]):
                 "Incorrect pair length: {}. Should be 32".format(len(pair)))
         args.append(pair)
     return args
+
+
+def get_timestamp() -> int:
+    """This is platform independent timestamp, needed for payments logic"""
+    return calendar.timegm(time.gmtime())
 
 
 class AbstractToken(object, metaclass=abc.ABCMeta):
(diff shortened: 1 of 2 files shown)


golem/ethereum/paymentprocessor.py
```diff
@@ -573,10 +575,17 @@ def sendout(self):
                 log.info("Next sendout in {} s".format(self.deadline - now))
                 return False
 
-            payments = self._awaiting
-            self._awaiting = []
+            closure_time = now - self.CLOSURE_TIME_DELAY
+
+            payments = \
+                [p for p in self._awaiting if p.timestamp <= closure_time]
+            print(self._awaiting[0].timestamp)
```
jiivan: :wastebasket: `print`

tests/golem/transactions/ethereum/test_paymentprocessor.py
```diff
@@ -726,9 +730,58 @@ def test_batch_transfer(self):
...
+        self.pp.deadline = 0
+        tx = mock.Mock()
+        tx_hash = '0xdead'
+        tx.hash = decode_hex(tx_hash)
+        self.client.send.return_value = tx_hash
+        self.token.batch_transfer.return_value = tx
+
+        closure_time = 2
+        time_value = closure_time + self.pp.CLOSURE_TIME_DELAY
+        with patch('time.time', return_value=time_value):
```
jiivan: `freezegun`

golem/ethereum/paymentprocessor.py
```diff
@@ -574,10 +576,16 @@ def sendout(self):
...
-            payments = self._awaiting
-            self._awaiting = []
+            closure_time = now - self.CLOSURE_TIME_DELAY
+
+            payments = \
+                [p for p in self._awaiting if p.processed_ts <= closure_time]
+            if not payments:
+                return False
+            self._awaiting = \
+                [p for p in self._awaiting if p.processed_ts > closure_time]
```
jiivan, line 592: Is it a race condition here on `self._awaiting`?

Krigpl: This is inside a lock

jiivan (approved):

PR #1809, opened 2017-12-27
async_run: run callbacks in the main thread (0.10.1)

mfranciszkiewicz:


tests/golem/resource/test_resourceclient.py
```diff
@@ -113,7 +113,25 @@ def test_filtered(self):
...
+        assert AsyncRequest.initialized
+
+        assert request.args == []
+        assert request.kwargs == {}
+        assert reactor.suggestThreadPoolSize.call_count == 1
+
+        request = AsyncRequest(lambda x: x, "arg", kwarg="kwarg")
+        assert request.args == ("arg",)
+        assert request.kwargs == {"kwarg": "kwarg"}
+        assert reactor.suggestThreadPoolSize.call_count == 1
```
maaktweluit, line 131: looks like the mock is re-used without reset
I think the count should be 2 or the mock should be reset before the second test.

maaktweluit (changes requested):
LGTM!

One test question, it might break

tests/golem/resource/test_resourceclient.py
mfranciszkiewicz: The count should be 1 since we want to increase the pool size exactly once (see the `AsyncRequest.initialized` check in `__init__`)

maaktweluit (approved):

Issue #1680, opened 2017-11-20
Next sendout doesn't start automatically. Only after node restart


kascheri12:
Just as the title says - while watching the logs, "Next soundout in xxx s" does not show up again until I restart the node.

etam:
This is not a bug.

This message does not need to be visible all the time. If there are some payments to be made, they're done every 10 minutes. This message says that there is a payment waiting to be done.

Example:

```
INFO     [golem.pay                          ] Next sendout in 15 s 
INFO     [golem.pay                          ] Peer count: 25 
INFO     [golem.pay                          ] ETH: 0.0207827 
INFO     [golem.pay                          ] GNT: 445.8333333333333 
INFO     [golem.pay                          ] Next sendout in 2 s 
INFO     [golem.pay                          ] Peer count: 25 
INFO     [golem.pay                          ] ETH: 0.0207827 
INFO     [golem.pay                          ] GNT: 445.8333333333333 
DEBUG:eth.chain.tx      deserialized tx tx=b'8f624b46'
INFO     [golem.pay                          ] Batch payments: bec052, value: 3.333333 
INFO     [golem.pay                          ] Peer count: 25 
INFO     [golem.pay                          ] ETH: 0.0207827 
INFO     [golem.pay                          ] GNT: 442.5 
INFO     [golem.pay                          ] Checking 0xbec0 tx [4] 
INFO     [golem.pay                          ] Peer count: 25 
INFO     [golem.pay                          ] ETH: 0.02004654 
INFO     [golem.pay                          ] GNT: 442.5 
INFO     [golem.pay                          ] Checking 0xbec0 tx [4] 
INFO     [golem.pay                          ] Confirmed 0xbec0: block b7364db00ea85da79cf5e5254be3e7399f4bf96ee57673c44d9b133e4c8d6
06f (1372253), gas 36808, fee 184040000000000                                                                                      
```

After that the message is gone.

Then I created a new task, it got rendered and the message is again appearing, because there is a new payment to be done.

@kascheri12 If you think you found a bug in sending payments, please add more details and reopen this issue.

kascheri12:
I ran a very complete and successful test and have attached the log file here for review. I do believe there is a bug here and it is not sending payments as it should and I will explain below. 

This is how my test went:
* Last night I started a task with 400 subtasks. Woke up this morning and it was in Finished state. No countdown for next sendout was happening at this time.
* Shutdown golemapp
* Start golemapp -- *This is where the log file starts*
[...]

PR #1814, opened 2017-12-28
Always add payments to paymentprocessor

Krigpl:
Currently when adding payments we check for available balance and don't append the payments if we're short on funds. And this way we only re-add the payments on client restart, during the `load_from_db` phase.
Instead we could always add payments to paymentprocessor and only do `batchTransfer` for those we can afford. This way if new funds come in, we will detect that and process pending payments. Resovles #1680
Also `processed_ts` is important to be set during the `add` call, so we can pass this value in `SubtaskResultAccepted` message to the provider.

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -211,16 +211,6 @@ class PaymentProcessor(LoopingCallService):
             encode_hex(payment.payee),
             payment.value / denoms.ether))
 
-        # Check if enough ETH available to pay the gas cost.
-        if self._eth_available() < self.ETH_PER_PAYMENT:
-            log.warning("Low ETH: {} available".format(self._eth_available()))
-            return False
-
-        av_gnt = self._gnt_available()
-        if av_gnt < payment.value:
-            log.warning("Low GNT: {:.6f}".format(av_gnt / denoms.ether))
-            return False
-
         with self._awaiting_lock:
             ts = get_timestamp()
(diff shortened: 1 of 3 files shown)


badb (approved):

chfast:
This does not make sense to me. Why do you start computing tasks you cannot pay for? The payments are not going to be settled until you transfer some additional GNT?

badb:
@chfast You started computing tasks when you were able to pay for it. But in the meantime you have lost you tokens because they can be used also outside of Golem. Payment that should be done is already noted in database, but you cannot settle it until you restart Golem node. This issue allows you to settle it just by adding additional funds. @Krigpl correct me if I'm wrong here.

chfast:
I propose alternative solution:

Let's assume we control Ethereum account entirely, user can only send funds to it. So the balance can go lower only by Golem actions.

Golem will be also monitoring if anyone is performing external actions on the account. This is very easy: we just remember the expected transaction count (stating from 0, because we are creating the account on the first run). If the transaction count differs from the expected we report the problem. I think we should stop any requestor actions this moment and report problem to user.

Krigpl:
@chfast What you described is not really a solution. Yes, we can detect whether the account has been used externally, but only after the fact, so we need to be prepared to handle reality being different from what we expect.
From the other side, even we would be able to provide such guarantee then these checks are even more useless - since we wouldn't be allowing to submit a task without enough funds to pay for it.

PR #1822, opened 2017-12-29
Don't duplicate crypto logic from golem-messages

### LINKED ISSUE #1799: Messages failing to be verified
~~In Hello message there is a `rand_val: float` that is supposed to be signed by the other node. The problem is that a float cannot be trusted.~~

```
2017-12-21 14:10:15 INFO     golem.network.transport.session     Failed to verify message signature (RandVal(timestamp=1513861815.886788, encrypted=True, sig=b'...', slots=[['rand_val', 0.4781411593100001]]) from 10.30.10.95:40103)
```
~~As seen above, the float value got additional `00001` at the end and the verification failed.~~

jiivan:
fixes: #1799


golem/core/variables.py
```diff
@@ -8,7 +8,7 @@
 #       VARIABLES          #
 ############################
 APP_NAME = "Brass Golem"
-APP_VERSION = "0.10.0"
+APP_VERSION = "0.10.1"
```
Krigpl, line 11: Relevant?

golem/network/transport/tcpnetwork.py
```diff
@@ -527,12 +526,24 @@ def valid_len():
         data = self.db.read_len_prefixed_bytes()
 
         while data:
-            message = Message.deserialize(data, lambda x: x)
+            try:
+                msg = golem_messages.load(
```
Krigpl: one line?

Krigpl (changes requested):
Tests are failing, otherwise looks good.

golem/core/variables.py
jiivan: It's not relevant but I think fixing automatic updater of this const is outside of scope of this PR.

golem/network/transport/session.py
```diff
@@ -225,11 +196,5 @@ def _check_msg(self, msg):
             self.disconnect(message.Disconnect.REASON.BadProtocol)
             return False
 
-        if (type_ not in self.can_be_unsigned) and (not self.verify(msg)):
```
Krigpl, line 228: Why is it removed?

golem/core/variables.py
jiivan: Issue for that: #1824

golem/network/transport/session.py
jiivan: Verification is done in https://github.com/golemfactory/golem-messages/blob/master/golem_messages/message.py#L262

Krigpl: I can't find it - where `can_be_unsigned` is respected?

jiivan: Yes, probably `can_be_unsigned` code should also be removed from core.

In the beginning `session.key_id` is set to `0` which effectively disables signature verification i `golem_messages.load()` ( https://github.com/golemfactory/golem-messages/pull/63/files it's version mentioned in requirements )

As soon as `key_id` is set there is no reason to not verify signatures on any message.

Krigpl (approved):
Beautiful red

PR #1823, opened 2017-12-29
Concent integration

jiivan:
Refactoring related to concent integration tests.


golem/task/tasksession.py
```diff
@@ -437,17 +437,34 @@ def send_report_computed_task(
...
+                    '[CONCENT] Computing checksum (%.3fMB) of %s',
+                    p.stat().st_size / 2**20,
+                    p
+                )
+                with open(filename, 'rb') as f:
+                    while True:
+                        chunk = f.read(2**20)  # 1MB
+                        if not chunk:
+                            break
+                        result_hash.update(chunk)
```
mfranciszkiewicz: Resulting files can be big. We can't afford to calculate checksums here and block the event queue.

golem/network/concent/client.py
```diff
@@ -1,63 +1,81 @@
...
-            logger.warning('Concent RequestException %r', e)
-            response = e.response
+from golem.core import variables
+from golem.network.concent import constants
+from golem.network.concent import exceptions
+
+logger = logging.getLogger("golem.network.concent.client")
+
+
+def send_to_concent(msg: message.Message, signing_key, pubkey=b'????') \
```
maaktweluit: why is the default `????`?

maaktweluit (approved):
LGTM!

One small Q, no blocker

golem/network/concent/client.py
jiivan: Thanks for pointing this out, fixed that.

PR #1840, opened 2018-01-03
Update crossbar

Elfoniok:


maaktweluit (approved):
LGTM!

Maybe the pytest in `requirements-test.txt` can also be updated if crossbar 17.12.1 is using the new attrs

mfranciszkiewicz (approved):
Binaries built and source are starting up properly.

I second the update of `requirements-test.txt`

PR #1848, opened 2018-01-04
Disconnect when golem_messages version differs by a minor version.

jiivan:


golem/network/p2p/peersession.py
```diff
@@ -259,6 +259,7 @@ def _react_to_pong(self, msg):
         self.p2p_service.pong_received(self.key_id)
 
     def _react_to_hello(self, msg):
+        super()._react_to_hello(msg)
```
Krigpl, line 262: It'd be better if we would stop here (and in other `react_to_hello`) if messages versions do not match, so it should return boolean or something. Maybe a separate method with descriptive name (like `check_version_compability`) would be better

golem/network/transport/session.py
```diff
@@ -133,6 +136,31 @@ def _check_msg(self, msg):
             return False
         return True
 
+    def _react_to_hello(self, msg):
+        theirs_gm_version  = msg.golem_messages_version
+        try:
+            theirs_v = semantic_version.Version(theirs_gm_version)
+        except ValueError:
```
Krigpl: log it?

golem/network/transport/session.py
```diff
@@ -133,6 +136,31 @@ def _check_msg(self, msg):
...
+        spec = semantic_version.Spec(spec_str)
+        if theirs_v not in spec:
+            logger.info(
+                'Incompatible golem_messages %s (ours) != %s (theirs)',
+                ours_v,
+                theirs_v,
+            )
+            self._send_disconnect(message.Disconnect.REASON.BadProtocol)
+            self.dropped()
+            return
```
Krigpl: rm

tests/golem/network/test_session.py
```diff
@@ -0,0 +1,82 @@
...
+gm_version = semantic_version.Version(golem_messages.__version__)
+
+
+class BasicSessionTestCase(unittest.TestCase):
+    def setUp(self):
+        self.instance = BasicSession(mock.MagicMock())
+
+    @mock.patch('golem.network.transport.session.BasicSession.dropped')
+    def test_golem_messages_version_higher_minor(self, dropped_mock):
+        msg = message.Hello(
```
Krigpl: this is repeated all over, helper function would be nice

tests/golem/network/test_session.py
```diff
@@ -0,0 +1,82 @@
...
+            **dict((key, None) for key in message.Hello.__slots__),
+        )
+        msg.golem_messages_version = str(gm_version.next_minor())
+        self.instance.interpret(msg)
+        dropped_mock.assert_called_once_with()
```
Krigpl: a check whether DisconnectMessage was sent could be useful too

Krigpl (changes requested):

golem/network/transport/session.py
```diff
@@ -11,6 +13,31 @@
...
+        raise IncompatibleGolemMessages(str(e))
+    ours_v = semantic_version.Version(golem_messages.__version__)
+    spec_str = '>={major}.{minor}.0,<{next_minor}'.format(
+        major=ours_v.major,
+        minor=ours_v.minor,
+        next_minor=ours_v.next_minor(),
+    )
+    spec = semantic_version.Spec(spec_str)
+    if theirs_v not in spec:
+        raise IncompatibleGolemMessages(
```
Krigpl: Why did you decide to use exceptions here? This function is supposed to check whether the versions are compatible, it's way more natural for it to return boolean. It not really unexpected when the versions don't match so no exceptions should be used here. Looks like the only use of this exception is to return a string with the message.

jiivan: @Krigpl refactored to return boolean

Krigpl (approved):
looks good

PR #1852, opened 2018-01-04
Throttle the rate of message.SetTaskSession

mfranciszkiewicz:
Handling and forwarding excessive amounts of `message.SetTaskSession` may lock the event loop and lead to undesired delays in message processing, e.g. WebSocket ping timeouts.

- limits the number of recipients to 3 neighbors nearest to the target node
- forwarded session requests are queued and sent in batches of 12 every 2 seconds

### DIFF
--- a/golem/network/p2p/p2pservice.py
+++ b/golem/network/p2p/p2pservice.py
@@ -25,8 +25,14 @@ REFRESH_PEERS_TIMEOUT = 1200
 RECONNECT_WITH_SEED_THRESHOLD = 30
 # Should nodes that connects with us solve hashcash challenge?
 SOLVE_CHALLENGE = True
+# Number of neighbors to notify of forwarded sessions
+FORWARD_NEIGHBORS_COUNT = 3
+# Forwarded sessions batch size
+FORWARD_BATCH_SIZE = 12
+
 BASE_DIFFICULTY = 5  # What should be a challenge difficulty?
 HISTORY_LEN = 5  # How many entries from challenge history should we remember
+
 TASK_INTERVAL = 10
 PEERS_INTERVAL = 30
 FORWARD_INTERVAL = 2
@@ -782,7 +788,7 @@ class P2PService(tcpserver.PendingConnectionsServer, DiagnosticsProvider):
             if peer.key_id != node_info.key
         )
 
-        for _, peer in distances[:3]:
+        for _, peer in distances[:FORWARD_NEIGHBORS_COUNT]:
             self.task_server.task_connections_helper.forward_queue_put(
                 peer, key_id, node_info, conn_id, super_node_info
             )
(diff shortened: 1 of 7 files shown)


golem/network/p2p/p2pservice.py
```diff
@@ -224,12 +232,18 @@ def sync_network(self):
         if self.task_server:
             self.__send_message_get_tasks()
 
-        if time.time() - self.last_peers_request > PEERS_INTERVAL:
+        now = time.time()
+
+        if now - self.last_peers_request > PEERS_INTERVAL:
             self.last_peers_request = time.time()
```
maaktweluit: maybe `now` can be used here

golem/network/p2p/p2pservice.py
```diff
@@ -224,12 +232,18 @@ def sync_network(self):
...
+        now = time.time()
+
+        if now - self.last_peers_request > PEERS_INTERVAL:
             self.last_peers_request = time.time()
             self.__sync_free_peers()
             self.__sync_peer_keeper()
             self.__send_get_peers()
 
+        if now - self.last_forward_request > FORWARD_INTERVAL:
+            self.last_forward_request = time.time()
```
maaktweluit: here `now` can also be used, unless the previous `if` can take a long time.

maaktweluit (approved):
LGTM!

One small comment, would recommend at least one more review of someone more familiar with `p2pservice.py`

golem/network/p2p/p2pservice.py
```diff
@@ -766,19 +778,21 @@ def want_to_start_task_session(
                 conn_id,
                 super_node_info
             )
+            logger.debug("Starting task session with {}".format(key_id))
             return
 
         msg_snd = False
+        distances = sorted(
+            (key_distance(key_id, peer.key_id), peer)
+            for peer in list(self.peers.values())
```
etam:
```python
for peer in self.peers.values()
```

etam:
```python
distances = sorted(
    peer for peer in self.peers.values()
    if peer.key_id != node_info.key,
    key=lambda peer: key_distance(key_id, peer.key_id)
)

Comment (mfranciszkiewicz) [reply golem/network/p2p/p2pservice.py]: `list` is done on purpose, since `peers` can be changed during iteration

Comment (mfranciszkiewicz) [reply golem/network/p2p/p2pservice.py]: 2nd comment: `Generator expression must be parenthesized if not sole argument`

File: golem/task/taskconnectionshelper.py [hunk 4]
```diff
@@ -64,3 +67,17 @@ def cannot_start_task_session(self, conn_id):
         :param conn_id: id of a connection that can't be established
         """
         self.task_server.final_conn_failure(conn_id)
+
+    def forward_queue_put(self, peer, *args):
+        entry = weakref.ref(peer), args
+        self.conn_to_set_queue.appendleft(entry)
```
etam: If queue is full, this will drop entries from front. Is it OK?

mfranciszkiewicz: Those in front are the oldest. What do you think?

etam: My intuition says that dropping the oldest messages from the queue is unfair and may lead to some priority inversion.

mfranciszkiewicz: I can do this the ring buffer way - either drop the newest ones or overwrite the oldest, but the latter seems overly complicated for this use case.

Elfoniok (approved):

Issue #1137, opened 2017-06-20
[FUNC] Add identifier in version for custom builds


maaktweluit:
Hey Golem team,

A few days ago i came across your project and I think its an amazing concept. Since then i have checked out the repository, got it running and read though most of the docs.

The first thing i saw when running from source was that the version was exactly the same as the release, 85 commits ago.

I propose to add an identifier to the version so the developer can easily see what build is running. This can be done using git:
```
git describe --tags --always --dirty
```

When on a tag this picks the tag as version, when behind it appends the number of commit and a partial sha, like this:

|state|output|
|------|--------|
|tag|0.6.0|
|develop|0.6.0-85-g0eba37f|
|dirty|0.6.0-85-g0eba37f-dirty|

When [attempting to implement](https://github.com/golemfactory/golem/compare/develop...maaktweluit:build-sha-in-version?expand=1) this i found the returned value is a valid semver in all 3 cases.
Though besides this i can not for see any other impact of this change, please advise.

Please let me know what you think, if its a bad idea please toss it. But when you like it, i can finish the sample in code and make it into a merge able PR.

Thanks for reading! 
Hope it helps!

Maaktweluit

jiivan:
Fixed by #1856

PR #1856, opened 2018-01-05
Store APP_VERSION differently 

### LINKED ISSUE #1789: Make version reading a common function
Currently the version is read in 2 places: `Installer/Installer_Win/version.py` and `setup_utils/setup_commons.py`.

There should be one uniform way to generate the right version from git.

### LINKED ISSUE #1824: Store APP_VERSION differently
Right now `golem.core.variables.APP_VERSION` is updated by `setup_util.setup_commons.py`. It breaks things as it changes code that is tracked by git.

Proper solution would be to store current version in a file that is in `.gitignore` and read it from there to APP_VERSION.

Also to be compatible with PEP-396 we should set `golem.__version__` attribute instead of `APP_VERSION` and `get_golem_version()`.

It could be implemented in a similar way that is in https://github.com/golemfactory/golem-messages/blob/master/version.py and https://github.com/golemfactory/golem-messages/blob/master/golem_messages/__init__.py It will have a nice side effect of additionally marking dev nodes as i.e. `0.10.1.dev32.3jhj3`

jiivan:
fixes: #1824
fixes: #1789 

One change that we have to incorporate in our process i to merge release tags into develop.


golem/tools/version.py
```diff
@@ -0,0 +1,81 @@
...
+
+__all__ = ("get_version")
+
+import pathlib
+import subprocess
+VERSION_FILE = "RELEASE-VERSION"
+
+
+def call_git_describe(prefix='', cwd='.'):
+    cmd = 'git describe --tags --match %s[0-9]*' % prefix
```
maaktweluit, line 43: Can we use `GitPython` module for calling git like the old implementation?

If not we can delete it from requirements, i think.

maaktweluit (changes requested):
LGTM!

One q, please reply and will approve.

Also see the fix is not applied on `version.py` in installer_win. This can be picked up in #1789 to make it all use the same awesome versioning :)

golem/tools/version.py
maaktweluit: I see it is already removed, nice!

Nvm this comment, approved!

maaktweluit (approved):

maaktweluit:
Maybe you can have a look at Shippable, there are 3 errors where 2 can be removed with `  # pylint: disable=no-member` ( add `# noqa #py....` if it gets too long )

maaktweluit:
This also fixes #1137

PR #1866, opened 2018-01-08
Restart task by creating a new one

### LINKED ISSUE #1748: Change restart task behaviour
Previous analysis of restart task behaviour can be found here: https://github.com/golemfactory/golem/issues/376

There is not specified what exactly should happen with task deadline after it's restarted. 

We propose to change restart task behaviour in a way that previous task cloning worked. 
* Task state is changed to restarted and stays this way until it's deleted from task manager.
* New task is being created that is a copy of the definition of the old one. It has a new deadline and a new task id.

etam:
Fixes #1748


etam:
golem-electron part: golemfactory/golem-electron#192

apps/dummy/task/dummytask.py
```diff
@@ -166,7 +166,8 @@ def build_full_definition(cls, task_type: DummyTaskTypeInfo, dictionary):
         #     raise TypeError("Subtask data size should be int")
         sbs = int(sbs)
         # difficulty comes in hex string from GUI
-        difficulty = int(difficulty, 16)
+        if isinstance(difficulty, str):
+            difficulty = int(difficulty, 16)
```
jiivan, line 170: ~~Why?Is your intention to prevent casting to `int` anything other than `str`?~~

golem/task/taskmanager.py
```diff
@@ -541,23 +541,15 @@ def get_resources(
         return task.get_resources(resource_header, resource_type)
 
     @handle_task_key_error
-    def restart_task(self, task_id):
+    def put_task_in_restarted_state(self, task_id):
```
jiivan: Could you add a docstring that states explicitly that restarted state means that a task is stopped and duplicate is created?

tests/golem/test_client.py
```diff
@@ -412,6 +412,54 @@ def test_start_stop(self, connect_to_network, *_):
...
+        }
+
+        task_id = self.client.create_task(task_dict)
+
+        assert task_id is not None
+
+        new_task_id = self.client.restart_task(task_id)
+
+        assert task_id != new_task_id
+        assert task_manager.tasks_states[task_id].status == TaskStatus.restarted
```
jiivan, line 459: Maybe also test subtasks? Or test only `golem.client` here mocking `TaskManager.put_task_in_restarted_state()` and test `put_task_in_restarted_state` in place of removed `test_restart_task`.

jiivan (changes requested):

apps/dummy/task/dummytask.py
etam: My intention is to handle both hex string and int input.
`TaskManager.get_task_definition_dict` returns `difficulty` as int.

jiivan: Oh, now I see that casting is made with base 16. My bad.

jiivan (approved):

Elfoniok (approved):

badb:
https://github.com/golemfactory/golem-electron/pull/192 should be merged as well

PR #1869, opened 2018-01-08
Remove broken task files after they are closed.

### LINKED ISSUE #1868: Broken saved tasks files are causing Golem to crash on Windows
When Golem tries to restore file that contains broken task data it tries to remove it https://github.com/golemfactory/golem/blob/develop/golem/task/taskmanager.py#L235. This exception is still called inside `with` scope that opened file. You cannot remove opened file in windows therefore `PermissionError` occurs and Golem crashes.

badb:
Resolves #1868

### DIFF
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -211,7 +211,7 @@ class TaskManager(TaskEventListener):
 
     def restore_tasks(self) -> None:
         logger.debug('SEARCHING FOR TASKS TO RESTORE')
-        broken_paths = []
+        broken_paths = set()
         for path in self.tasks_dir.iterdir():
             if not path.suffix == '.pickle':
                 continue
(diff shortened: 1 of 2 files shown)


golem/task/taskmanager.py
```diff
@@ -211,6 +211,7 @@ def remove_dump(self, task_id: str):
 
     def restore_tasks(self) -> None:
         logger.debug('SEARCHING FOR TASKS TO RESTORE')
+        broken_paths = []
```
jiivan: Maybe a set will be enough.

It'd also be nice to see a short comment why this was implemented. For a long time POSIX user this is totally unobvious.

jiivan (changes requested):
Great, only one suggestion for a simpler type and one request for comment, as this change is unobvious for one not used to windows quirks. Will approve after that.

jiivan (approved):

mfranciszkiewicz (approved):

PR #1872, opened 2018-01-09
Rececive single set_tasks_session msg once for 20sec from each peer

Elfoniok:

### DIFF
--- a/golem/network/transport/spamprotector.py
+++ b/golem/network/transport/spamprotector.py
@@ -1,17 +1,15 @@
 import time
 import logging
 from golem_messages.message import Message
-from golem_messages.message import P2P_MESSAGE_BASE
+from golem_messages.message import SetTaskSession
 
 logger = logging.getLogger(__name__)
 
-
 class SpamProtector:
 
-    SetTaskSessionFrq = 20
-    SetTaskSessionType = P2P_MESSAGE_BASE + 16
+    SetTaskSessionInterval = 20
 
-    FeqDict = { SetTaskSessionType: SetTaskSessionFrq}
+    INTERVALS = { SetTaskSession.TYPE: SetTaskSessionInterval}
 
     def __init__(self):
 
@@ -23,21 +21,16 @@ class SpamProtector:
 
         msg_type, timestamp, is_encoded = Message.deserialize_header(
             msg[:Message.HDR_LEN])
-        return self._check_msg_type(msg_type)
 
-    def _check_msg_type(self, msg_type):
-        if msg_type == __class__.SetTaskSessionType:
-            return not self.is_spamming(msg_type)
-        else:
+        if msg_type not in self.INTERVALS:
             return True
 
-    def is_spamming(self, msg_type):
-        if msg_type in self.last_msg_map:
-            if self.last_msg_map[msg_type] - int(time.time()) < __class__.FeqDict[msg_type]:
-                logger.debug("DROPING SPAM message")
-                return True
+        now = int(time.time())
+        last_received = self.last_msg_map.get(msg_type, 0)
+        delta = now - last_received
+        if delta > self.INTERVALS[msg_type]:
+            self.last_msg_map[msg_type] = now
+            return True
 
-        self.last_msg_map[msg_type] = int(time.time())
+        logger.debug("DROPING SPAM message")
         return False
-
-
--- a/golem/network/transport/spamprotector.py
+++ b/golem/network/transport/spamprotector.py
@@ -1,0 +1,43 @@
+import time
+import logging
+from golem_messages.message import Message
+from golem_messages.message import P2P_MESSAGE_BASE
+
+logger = logging.getLogger(__name__)
+
+
+class SpamProtector:
+
+    SetTaskSessionFrq = 20
+    SetTaskSessionType = P2P_MESSAGE_BASE + 16
+
+    FeqDict = { SetTaskSessionType: SetTaskSessionFrq}
+
+    def __init__(self):
+
+        self.last_msg_map = dict()
+
+    def check_msg(self, msg):
+        if msg is None:
+            return False
+
+        msg_type, timestamp, is_encoded = Message.deserialize_header(
+            msg[:Message.HDR_LEN])
+        return self._check_msg_type(msg_type)
+
+    def _check_msg_type(self, msg_type):
+        if msg_type == __class__.SetTaskSessionType:
+            return not self.is_spamming(msg_type)
+        else:
+            return True
+
+    def is_spamming(self, msg_type):
+        if msg_type in self.last_msg_map:
+            if self.last_msg_map[msg_type] - int(time.time()) < __class__.FeqDict[msg_type]:
+                logger.debug("DROPING SPAM message")
+                return True
+
+        self.last_msg_map[msg_type] = int(time.time())
+        return False
+
+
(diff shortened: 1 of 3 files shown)


golem/network/transport/spamprotector.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def _check_msg_type(self, msg_type):
+        if msg_type == __class__.SetTaskSessionType:
+            return not self.is_spamming(msg_type)
+        else:
+            return True
+
+    def is_spamming(self, msg_type):
+        if msg_type in self.last_msg_map:
+            if self.last_msg_map[msg_type] - int(time.time()) < __class__.FeqDict[msg_type]:
```
jiivan: `./golem/network/transport/spamprotector.py:36:81: E501 line too long (92 > 80 characters)`

golem/network/transport/spamprotector.py
```diff
@@ -0,0 +1,43 @@
...
+    def check_msg(self, msg):
+        if msg is None:
+            return False
+
+        msg_type, timestamp, is_encoded = Message.deserialize_header(
+            msg[:Message.HDR_LEN])
+        return self._check_msg_type(msg_type)
+
+    def _check_msg_type(self, msg_type):
+        if msg_type == __class__.SetTaskSessionType:
```
jiivan: Why not explicitly use `golem_messages.message.SetTaskSession.TYPE`?

golem/network/transport/spamprotector.py
```diff
@@ -0,0 +1,43 @@
...
+import logging
+from golem_messages.message import Message
+from golem_messages.message import P2P_MESSAGE_BASE
+
+logger = logging.getLogger(__name__)
+
+
+class SpamProtector:
+
+    SetTaskSessionFrq = 20
```
jiivan: Unused, just move 20 to `FreqDict`?

golem/network/transport/spamprotector.py
```diff
@@ -0,0 +1,43 @@
...
+    SetTaskSessionType = P2P_MESSAGE_BASE + 16
+
+    FeqDict = { SetTaskSessionType: SetTaskSessionFrq}
+
+    def __init__(self):
+
+        self.last_msg_map = dict()
```
jiivan: If `FreqDict` then `self.last_msg_*dict*` also?

golem/network/transport/spamprotector.py
```diff
@@ -0,0 +1,43 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SpamProtector:
+
+    SetTaskSessionFrq = 20
+    SetTaskSessionType = P2P_MESSAGE_BASE + 16
+
+    FeqDict = { SetTaskSessionType: SetTaskSessionFrq}
```
jiivan: If this is meant to be constant then maybe `FREQ_DICT`?

golem/network/transport/spamprotector.py
```diff
@@ -0,0 +1,43 @@
...
+
+        self.last_msg_map = dict()
+
+    def check_msg(self, msg):
+        if msg is None:
+            return False
+
+        msg_type, timestamp, is_encoded = Message.deserialize_header(
+            msg[:Message.HDR_LEN])
+        return self._check_msg_type(msg_type)
```
jiivan: Instead of implementing helper `_check_msg_type()` just ask `FreqDict`?

```python
if msg_type not in self.FreqDict:
    return True
now = time.time()
last_received = self.last_msg_map.get(msg_type, 0)
delta = now - last_received
if delta > self.FreqDict[msg_type]:
    self.last_msg_map[msg_type] = now
    return True
return False
```

jiivan (changes requested):
Good job, but I think implementation could be simplified. Adding tests would be a good idea :) They should be quite straightforward with `freezegun`.

Also have you seen choker (pol. dławik) from original bittorrent implementation? Maybe this approach will work better for us? Not a request, just an inspiration.
https://github.com/mgp/bittorrent-dissected/blob/master/Choker.py
https://github.com/mgp/bittorrent-dissected#chokerpy

golem/network/transport/spamprotector.py
Elfoniok: No the point is to have names on such constants.

golem/network/transport/spamprotector.py
Elfoniok: Done

golem/network/transport/spamprotector.py
Elfoniok: Lets just leave this. I have changed previous name

golem/network/transport/spamprotector.py
Elfoniok: Done

golem/network/transport/spamprotector.py
Elfoniok: Done

golem/network/transport/spamprotector.py
Elfoniok: Done

Elfoniok:
Its a good idea (choker) and we might keep that in mind for future development. But bit torrent use case is for data not messages so its not easy to incorporate this, but we might get inspired and try to design something more sophisticated later in time.

golem/network/transport/spamprotector.py
```diff
@@ -0,0 +1,36 @@
...
+            return True
+
+        now = int(time.time())
+        last_received = self.last_msg_map.get(msg_type, 0)
+        delta = now - last_received
+        if delta > self.INTERVALS[msg_type]:
+            self.last_msg_map[msg_type] = now
+            return True
+
+        logger.debug("DROPING SPAM message")
```
mfranciszkiewicz, line 35: IMO this message could include the message class and remote peer's address. The latter is not available here, so this might have to be logged in `Basic/SafeProtocol`.

mfranciszkiewicz (approved):

tests/golem/network/transport/test_tcpnetwork.py
```diff
@@ -219,6 +221,27 @@ def test_dataReceived(self, load_mock):
...
+        protocol.session.interpret = mock.MagicMock()
+        msg = message.SetTaskSession(**dict((key, None) for key in message.SetTaskSession.__slots__))
+        data = msg.serialize()
+        packed_data = struct.pack("!L", len(data)) + data
+        load_mock.return_value = msg
+        with freeze_time("2012-01-14 10:30:20") as frozen_datetime:
+            for i in range (0, 100):
+                protocol.dataReceived(packed_data)
+            protocol.session.interpret.assert_called_once_with(msg)
+            frozen_datetime.move_to("2014-02-12 11:30:45")
```
jiivan, line 240:
```python
protocol.session.interpret.reset_mock()
protocol.session.interpret.assert_called_once_with(msg)
```

golem/network/transport/spamprotector.py
jiivan, line 35: `s/DROPING/DROPPING/`

golem/network/transport/spamprotector.py
```diff
@@ -0,0 +1,36 @@
+import time
+import logging
+from golem_messages.message import Message
+from golem_messages.message import SetTaskSession
+
+logger = logging.getLogger(__name__)
+
+class SpamProtector:
+
+    SetTaskSessionInterval = 20
```
jiivan, line 10: It's already named and easily referenced by INTERVALS[SetTaskSession.TYPE], why add another label?

Elfoniok: To understand what it means

jiivan:
I suggest closing this pull request. Merging #1879 to b0.10.1 and then merging b0.10.1 to develop. Otherwise we will have duplicated commits and possible merging nightmare.

PR #1879, opened 2018-01-10
Rececive single set_tasks_session msg once for 20sec from each peer

Elfoniok:

### DIFF
--- a/golem/network/transport/tcpnetwork.py
+++ b/golem/network/transport/tcpnetwork.py
@@ -1,27 +1,28 @@
-from golem_messages import message
 import logging
 import os
 import re
 import struct
 import time
 from copy import copy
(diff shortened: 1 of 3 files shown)


golem/network/transport/tcpnetwork.py
```diff
@@ -524,19 +526,21 @@ def valid_len():
...
-            message = Message.deserialize(data, lambda x: x)
+        for buf in self.db.read_len_prefixed_bytes():
+
+            if not valid_len():
+                return None
+
+            if not self.spam_protector.check_msg(buf):
+                return None
+
+            message = Message.deserialize(buf, lambda x: x)
```
jiivan: I know it's not your code but since your're touching this you may as well rename `message` to `msg` to avoid ocerwriting message namespace from imports.

jiivan (approved):

PR #1429, opened 2017-10-02
Turn off compositing only when it's needed #769

jiivan:
#769

PR #1885, opened 2018-01-11
Turn off compositing only when it's needed #769 #1429

### LINKED ISSUE #769: Turn off compositing only if it's needed
Right now compositing is turn off by default. We don't have to be as strict. We can turn it on by default, if there are more frames than subtasks.

jiivan:
resolves: #769

### DIFF
--- a/apps/blender/task/blenderrendertask.py
+++ b/apps/blender/task/blenderrendertask.py
@@ -338,8 +338,8 @@ class BlenderRenderTask(FrameRenderingTask):
             and (self.total_tasks <= len(self.frames))
         if self.compositing != task_definition.options.compositing:
             logger.warning("Task %s: Compositing not supported "
-                           "for this type of task, turning compositing off"
-                           % task_definition.task_id)
+                           "for this type of task, turning compositing off",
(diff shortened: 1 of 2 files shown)


badb (approved):

apps/blender/task/blenderrendertask.py
```diff
@@ -327,13 +327,16 @@ class BlenderRenderTask(FrameRenderingTask):
...
-        self.compositing = task_definition.options.compositing
         self.preview_updater = None
         self.preview_updaters = None
 
         FrameRenderingTask.__init__(self, task_definition=task_definition,
                                     **kwargs)
 
+        self.compositing = task_definition.options.compositing\
+            and self.use_frames\
+            and (len(self.frames) > len(self.frames_subtasks))
```
badb: `len(frames_subtasks) == len(self.frames)` is always True. 
What you tried to say probably is:
`len(self.frames) >= len(self.total_tasks)`

badb (changes requested):

badb (approved):

PR #1887, opened 2018-01-11
rinkeby.golem.network doesn't provide node list

etam:
Not trying to make empty requests may improve our situation with geth running out of file descriptors.

### DIFF
--- a/golem/ethereum/node.py
+++ b/golem/ethereum/node.py
@@ -10,7 +10,6 @@ import threading
 import time
 from distutils.version import StrictVersion
 
-import requests
 from web3 import Web3, IPCProvider, HTTPProvider
 
 from golem.core.common import is_windows, DEVNULL, SUBPROCESS_STARTUP_INFO
@@ -22,8 +21,7 @@ from golem.utils import tee_target
 log = logging.getLogger('golem.ethereum')
 
 
-NODE_LIST_URL = 'https://rinkeby.golem.network'
-FALLBACK_NODE_LIST = [
+NODE_LIST = [
     'http://188.165.227.180:55555',
     'http://94.23.17.170:55555',
     'http://94.23.57.58:55555',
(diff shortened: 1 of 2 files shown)


mfranciszkiewicz:
Few questions:
- why is the geth load balancer running on port ~80~ 443?
- is nginx configured properly on that machine? What about the fd limits?
- are we sure we want to remove the fallback list? E.g. certificate expiry might become more than an inconvenience.

maaktweluit:
Can add my 2 cents for these questions:
- The support request stated: 
  > Yes, public should be HTTPS only, private can be HTTP.
- The settings are setup by Y2 as "Round robin best practice" i did not check the config on fd-limits.
- This depends on how certain we can be this server is always up, would be nice to be able to, but safer to have a plan B.

mfranciszkiewicz:
One would expect a typical http webserver sitting behind 80 / 443. This needs to be changed.
Fallback list aside, we should make sure that nginx is ready to go before merging this PR.

maaktweluit:
You can make a support ticket at Y2 to check or update the configs.
Please let me know if you want me to pick that up.

maaktweluit:
Made a support ticket!

Requested to add port `55555` and check the configs.
Ports 443 and 80 can be removed after golem release to reduce scanners / spam.

@etam if you have any additions to check / add here please let me know!

etam:
I think that if we want to have a fallback, it should be just a second host with the same nginx reverse proxy.

mfranciszkiewicz:
@etam Does that differ from the currently implemented mechanism?

etam:
@mfranciszkiewicz Yes, because it allows to still use https (falling back to a not encrypted connection is, I think, a Very Bad Idea).

We could also go completely the other way. Setup nginx reverse proxy without load balancing on every host with geth and do load balancing in Golem, like it was before.

mfranciszkiewicz:
@etam Implementation = code you removed. Instead of HTTP addresses, you would put the HTTPS addresses of (other?) load balancers in the fallback list.

etam:
@mfranciszkiewicz That's true.

etam:
I've changed the pull request. Now it's just using rinkeby.golem.network as an another geth instance.

etam:
rebased onto current develop

maaktweluit (approved):
LGTM!

Ready for the move to 55555 :)

etam:
There is an issue with using rinkeby.golem.network as a reverse proxy with round robin multiple backends. As @Krigpl said, we make some stateful operations, so we have to connect to one geth instance and stick to it.

etam:
Meanwhile I changed the PR to just stop trying to download a node list from rinkeby.golem.network.

etam:
Let's continue the discussion at #2041

PR #1894, opened 2018-01-12
Fine tune linters

jiivan:
* Enable design/refactoring checkers
 * Use flake8 instead of bare pycodestyle to catch more problems


maaktweluit (approved):
LGTM!

Better checks FTW

maaktweluit:
After checking the test result something seems wrong:
```
./lintdiff.sh: line 95: flake8: command not found
```

PR #1901, opened 2018-01-15
[fix] cli show envs short description fix

### LINKED ISSUE #1544: Description for every environment is the same
If you choose "golemcli envs show" you can see that the description for every environment is the same. 
This is not an expected behaviour, there should be short information about every supported environment.

tworec:
fixes #1544


maaktweluit (approved):
LGTM!

Will rebuild Appveyor for you as well

tworec:
thx a lot! :)

PR #1903, opened 2018-01-16
Local database

Elfoniok:
Confine global db object. It should be used only from Client instances by everyone.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/golem-messages"]
```
etam, line 1: But why?

golem/ranking/helper/trust.py
```diff
@@ -32,14 +32,14 @@ def __init__(self, val):
         self.val = val
         self.lock = Lock()
 
-    def increase(self, node_id, database, mod=1.0):
```
etam: changes in this file in this commit could be squashed with previous commit

golem/task/taskserver.py
```diff
@@ -219,6 +222,19 @@ def __init__(self,
         dispatcher.connect(
             self.transactions_listener, signal="golem.transactions")
 
+    def get_payment_for_subtask(self, subtask_id):
```
etam, line 227: It would be better, if this change was in a separate commit.

golem/ethereum/paymentprocessor.py
```diff
@@ -13,7 +13,6 @@
 from golem.core.service import LoopingCallService
 from golem.ethereum import Client
 from golem.model import Payment, PaymentStatus
-from golem.model import Payment, PaymentStatus
```
etam: squash this with 98ce12e0bff949096c72ae2eeb45e3429f99c660

tests/golem/ranking/manager/test_database_manager.py
```diff
@@ -243,8 +245,13 @@ def test_should_update_database_records(self):
         )
 
         for case in cases:
-            case['fun_ref'](case['node_name'], case['value'])
-            self.assertAlmostEqual(getattr(dm.get_local_rank(case['node_name']), case['attribute']), case['total'],
+            if case['fun_ref'].__name__.endswith('increase') or \
+                case['fun_ref'].__name__.endswith('decrease'):
+                case['fun_ref'](case['node_name'], case['value'])
```
etam, line 250: Just pass database as an argument to Trust. You won't need to make such weird if statements and you'll avoid setting database as class property.

etam (changes requested):
There are extra golem/task/taskserver* files left.

tests/golem/test_client.py
```diff
@@ -677,8 +677,15 @@ def test_node(self, *_):
     def test_directories(self, *_):
         c = self.client
 
-        c.resource_server = ResourceServer.__new__(ResourceServer)
-        c.resource_server.dir_manager = c.task_server.task_computer.dir_manager
+        def unique_dir():
+            d = os.path.join(self.path, str(uuid.uuid4()))
+            if not os.path.exists(d):
+                os.makedirs(d)
+            return d
```
etam, line 684:
```python
d = self.new_path / str(uuid.uuid4())
d.mkdir(exist_ok=True)
return d
```

Issue #1882, opened 2018-01-10
Add user possibility to choose geth instance. 


badb:
Right no we're connecting to external geth instances.

 Advance user may still want to use their own local geth instances. The proper code for starting geth was implemented and should still work, we just need to add a command line flag `--start-geth` (or something similar) that will set `start_geth` option in EthereumTransactionSystem to `True`. 
This should of course be deeply tested if nothing was broken in the meantime. 

We need also a second option when we will not start geth on our own, but instead allow user to  configure geth instance http address. It may be set by adding `--geth-address <address>` option to command line. Please, make sure that the app behaves in an rational way if the given geth instance is not available or configured incorrectly.

tworec:
`--start-geth` is already there. I'll just add the `--get-address` option

maaktweluit:
I see this issue did not auto-close.

Next time you can use one of these keywords to link the issue and the PR: https://help.github.com/articles/closing-issues-using-keywords/

tworec:
closed via #1910

PR #1910, opened 2018-01-17
[#1182] enable user to choose geth instance

tworec:
implements #1882

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -90,7 +90,8 @@ class Client(HardwarePresetsMixin):
             use_docker_machine_manager=True,
             use_monitor=True,
             start_geth=False,
-            geth_port=None,
+            start_geth_port=None,
+            geth_address=None,
             **config_overrides):
 
         if not datadir:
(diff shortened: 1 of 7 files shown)


maaktweluit (changes requested):
LGTM!

One q on the naming: what does `start_` has to do with the `geth_port`? 
Will approve on reply :)

tworec:
I want to indicate that this port affects locally created geth instance only. This parameter has no meaning if `--start-geth` flag was not provided. It may be even more intuitive to transform `--start-geth` into option with parameter specifying port. What do you think?

maaktweluit:
Thank you for the reply.

Now the naming makes a lot of sense, will approve!

making `--start-geth` smarter sounds like a nice idea but not needed for this fix.

maaktweluit (approved):

maaktweluit:
rebuild Appveyor now it has an approval.

Please check Shippable, one fix and it will be a green build :)
```
Changed lines findings:

./golem/node.py:147:81: E501 line too long (85 > 80 characters)
```

maaktweluit:
The features in this PR can be added to the GUI in this issue: golemfactory/golem-electron#161

tworec:
closes #1882

PR #1913, opened 2018-01-18
Move sync checking to the client

Krigpl:
This is a simple code move of the `is_synchronized` and `wait_until_synchronized` methods from PaymentProcessor to `Client`, since other modules using `Client` will also be interested whether it's synchronized.

### DIFF
--- a/golem/ethereum/client.py
+++ b/golem/ethereum/client.py
@@ -1,4 +1,5 @@
 import logging
+import time
 
 import rlp
 from ethereum.utils import zpad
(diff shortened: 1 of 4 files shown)


golem/ethereum/client.py
```diff
@@ -22,6 +25,9 @@ def __init__(self, datadir, port=None, start_node=False):
         self.web3 = Client.node.web3
         # Set fake default account.
         self.web3.eth.defaultAccount = '\xff' * 20
+        self.__last_sync_check = time.time()
```
etam: For private fields there is single underscore. Double underscore is to avoid naming clashes with inheritance.

etam (approved):
LGTM, but added a comment you could resolve.

mfranciszkiewicz (approved):

PR #1917, opened 2018-01-18
Added ERROR as allowed loglevel

### LINKED ISSUE #1597: Add flag to run golem with different log severity
We can change log severity by editing logging.ini, but it will be more convenient if we are able to just change this with just adding specific flag that will change the severity. It's suggested that it will have three levels: 

* normal (no flag): log warnings and higher
* verbose (--verbose, -v): all logs (debug and higher)
* quiet (--quiet, -q): only critical and exceptions

maaktweluit:
Resolves #1597

### DIFF
--- a/golem/core/common.py
+++ b/golem/core/common.py
@@ -14,7 +14,7 @@ DEVNULL = open(os.devnull, 'wb')
 MAX_CPU_WINDOWS = 32
 MAX_CPU_MACOS = 16
 
-ALLOW_LOGLEVELS = [
+ALLOWED_LOGLEVELS = [
     'ERROR'
     'WARNING',
     'INFO',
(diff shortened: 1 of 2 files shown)


golem/core/common.py
```diff
@@ -14,6 +14,13 @@
 MAX_CPU_WINDOWS = 32
 MAX_CPU_MACOS = 16
 
+ALLOW_LOGLEVELS = [
```
jiivan: `ALLOWED_LOGLEVELS`?

maaktweluit: Good one, fixed :)

jiivan (approved):

jiivan:
@maaktweluit please remember to change help text for `--loglevel` also.

maaktweluit:
@jiivan Good catch! updated it :)

PR #1918, opened 2018-01-18
IncomesKeeper shouldn't depend on PaymentProcessor

Krigpl:
Because it just doesn't make sense to make outgoing payments through a class that is responsible for incomes.
IncomesKeeper only needs an access to the token, not the whole PaymentProcessor.

### DIFF
--- a/golem/transactions/ethereum/ethereumincomeskeeper.py
+++ b/golem/transactions/ethereum/ethereumincomeskeeper.py
@@ -1,29 +1,20 @@
 # -*- coding: utf-8 -*-
 
 import logging
-import peewee
 
-from time import sleep
 from golem.model import db
 from golem import model
 from golem.model import Income
 from golem.transactions.incomeskeeper import IncomesKeeper
-from golem.ethereum.paymentprocessor import PaymentProcessor
 
 logger = logging.getLogger('golem.transactions.ethereum.ethereumincomeskeeper')
 
 
 class EthereumIncomesKeeper(IncomesKeeper):
 
-    def __init__(self, processor: PaymentProcessor) -> None:
-        self.processor = processor
-
-    def start(self):
-        self.processor.start()
-
-    def stop(self):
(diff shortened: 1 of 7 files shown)


etam (approved):

etam (changes requested):
Tests are failing. Pylint is sad.

Krigpl:
Pylint will remain sad because I can't do much about those errors
 ``` E: tests/golem/transactions/ethereum/test_ethereumtransactionsystem.py:100:19: Instance of 'PaymentProcessor' has no '_PaymentProcessor__client' member (no-member)```
(or can I)?

Krigpl:
I guess we could update pylint https://github.com/PyCQA/pylint/issues/1643

etam:
Use `_client` instead of `__client`?

etam:
Updating pylint is also a good idea.

Krigpl:
> Use _client instead of __client?

I'm aiming to remove `Client` from `PaymentProcessor` altogether so this will resolve itself anyway.

golem/transactions/ethereum/ethereumincomeskeeper.py
```diff
@@ -1,29 +1,20 @@
...
-    def __init__(self, processor: PaymentProcessor) -> None:
-        self.processor = processor
-
-    def start(self):
-        self.processor.start()
-
-    def stop(self):
-        if self.processor.running:
-            self.processor.stop()
+    def __init__(self, eth_address: str, token) -> None:
```
mfranciszkiewicz, line 15: Just a question - why not `client` instead of `token`?

Krigpl: Because of `get_incomes_from_block` which is token dependent.

mfranciszkiewicz (approved):

etam (approved):

PR #1920, opened 2018-01-18
Pass payment timestamp in SubtaskResultAccepted message

Krigpl:
When `batchTransfer` happens it should cover all incomes where `accepted_ts <= closure_time`, thus we should store accepted_ts in ExpectedIncome.

### DIFF
--- a/golem/transactions/incomeskeeper.py
+++ b/golem/transactions/incomeskeeper.py
@@ -110,6 +110,18 @@ class IncomesKeeper(object):
             value=value
         )
 
+    def update_awaiting(self, subtask_id, accepted_ts):
+        try:
+            # FIXME: query by (sender_id, subtask_id)
+            income = ExpectedIncome.get(subtask=subtask_id)
+        except ExpectedIncome.DoesNotExist:
+            logger.error(
+                "ExpectedIncome.DoesNotExist subtask_id: %r",
+                subtask_id)
+            return
+        income.accepted_ts = accepted_ts
+        income.save()
+
     def get_list_of_all_incomes(self):
         # TODO: pagination
         union = ExpectedIncome.select(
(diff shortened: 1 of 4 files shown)


golem/transactions/incomeskeeper.py
```diff
@@ -110,6 +110,18 @@ def expect(self, sender_node_id, p2p_node, task_id, subtask_id, value):
             value=value
         )
 
+    def update_awaiting(self, subtask_id, accepted_ts):
+        try:
+            # FIXME: query by (sender_id, subtask_id)
+            income = ExpectedIncome.get(subtask=subtask_id)
```
Elfoniok, line 116: You can send hex encoded pub_key and use it in query against 'sender_node' from Expected income, otherwise, i am not sure that subtask_id is unique among all tasks.

golem/transactions/incomeskeeper.py
```diff
@@ -110,6 +110,18 @@ def expect(self, sender_node_id, p2p_node, task_id, subtask_id, value):
...
+        except ExpectedIncome.DoesNotExist:
+            logger.error(
```
Elfoniok: What is the purpose of sending accepted_ts? I am asking because i am concerning about handling this error, either because requester didn't added it to database before getting this message, or because it was already deleted after receiving payment confirmation.

Elfoniok (approved):

golem/transactions/incomeskeeper.py
Krigpl: That is the plan. This is currently replicating the existing logic where we also don't query by `sender_node` and I'm planning to change it in a separate PR. AFAICT even querying by `(sender_node, subtask_id)` is not safe at the moment, since sender doesn't guarantee `subtask_id`s will be unique. (which is again something I'll improve soon)

jiivan (approved):

PR #1921, opened 2018-01-18
Add headers to concent failed log

jiivan:


golem/network/concent/client.py
```diff
@@ -55,8 +55,15 @@ def send_to_concent(msg: message.Message, signing_key, public_key) \
         raise exceptions.ConcentUnavailableException()
 
     elif response.status_code % 500 < 100:
-        logger.warning('Concent failed with status %d and body: %r',
-                       response.status_code, response.text)
+        logger.warning(
+            'Concent failed with:\n'
```
maaktweluit: using `\n` in the logs was a bad practice at my old work.
This was so the log harvest software could not parse them natively when multiline.

Not sure if that is also the case here, maybe if we want to harvest logs later down the line.
Will approve on reply!

maaktweluit (changes requested):

golem/network/concent/client.py
jiivan: Good point @maaktweluit . Fixed

maaktweluit (approved):

PR #1922, opened 2018-01-18
ConcentService handle delay as timedelta

jiivan:
Fixes `TypeError: unorderable types: datetime.timedelta() >= int()`


Krigpl (changes requested):
golem/network/concent/client.py:189 ( I can't add an inline comment there explicitly, yey github)
Default value shouldn't be `0`, it makes type of `delay` `datetime.timedelta|int`. Although the `if` in line 194 prevents the failure, I'd change that to be consistent

jiivan:
@Krigpl good point. Fixed

Krigpl (approved):

PR #1927, opened 2018-01-19
Random crop generated based on CP & GG code

badb:
https://github.com/golemfactory/golem-verificator/blob/master/golem_verificator/blender/generate_random_crop_images.py


Elfoniok (approved):
LGTM

PR #1928, opened 2018-01-19
Show full traceback when [Task|Docker]Thread fails

jiivan:
Greatly improves debugging task thread fails.


etam (approved):
It's OK as a temporary solution. A better one would be to support passing an Exception object through TaskThread (beside error_msg).

jiivan:
@etam Agree. Even better to pass `Traceback` object.

badb (approved):

PR #1930, opened 2018-01-19
add support for allow list

### LINKED ISSUE #1836: Add support for Allowlist to Golem
There is Denylist implemented in Golem: https://github.com/golemfactory/golem/pull/1046 that allows you to block nodes with specific ids. 

For testing multi-OS integrations we would like to also have Allowlist. When on Denylist we put specific text  eg. "ALL" instead of node ID as the only content than all nodes will be blocked and only nodes with id-s that are directly place in allow.txt will be able to contect with each other. This functionality will allow to prepare us better test scenarios.

etam:
fixes #1836


tests/golem/resource/test_resourcehandshake.py
```diff
@@ -1,5 +1,6 @@
 from golem_messages import message
 import os
+from pathlib import Path
```
jiivan: `C: tests/golem/resource/test_resourcehandshake.py:3:0: standard import "from pathlib import Path" should be placed before "from golem_messages import message" (wrong-import-order)`

golem/task/acl.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def disallow(self, node_id: str) -> None:
+        self._allow_set.discard(node_id)
+
+
+def _read_set_from_file(path: Path) -> Set[str]:
+    if not path.exists():
+        return set()
+    with path.open() as f:
+        return set(line.strip() for line in f)
```
jiivan:
```python
try:
    with path.open() as f:
        #...
except OSError:
    return set()
```
Would protect against race conditions and is generally considered more pythonic.
https://docs.python.org/3/glossary.html?highlight=eafp#term-lbyl

jiivan (changes requested):
LGTM! Small remarks

Krigpl:
Having two lists for denied and allowed users and both of them being non empty feels weird.
Something like one list with the list of specified users and a mode (deny/allow) would be cleaner imo.
Not insisting though, so if it's too much to change then I'm good.

golem/task/acl.py
etam: I don't think we should care about race conditions in this case (unless you like to race with editing config files against a starting application). Also I wrote it this way, because in C++ an expected condition shouldn't be handled with exceptions. But that's python, so I think it's OK to change it as you propose.

etam:
@Krigpl good point

jiivan (approved):

tests/golem/task/test_acl.py
```diff
@@ -0,0 +1,34 @@
+from golem.task.acl import get_acl, DENY_LIST_NAME, ALL_EXCEPT_ALLOWED
+from golem.testutils import TempDirFixture
+
+
+class TestAcl(TempDirFixture):
+    def test_no_file(self):
+        acl = get_acl(self.new_path)
+        assert acl._deny_set == set()
```
Krigpl, line 8: `self.AssertEqual` and family are preferable since they give much better errors messages compared to these simple asserts

golem/task/acl.py
```diff
@@ -0,0 +1,57 @@
...
+def _read_set_from_file(path: Path) -> Set[str]:
+    try:
+        with path.open() as f:
+            return set(line.strip() for line in f)
+    except OSError:
+        return set()
+
+
+def get_acl(datadir: Path) -> Acl:
+    nodes_ids = _read_set_from_file(datadir / DENY_LIST_NAME)
```
Krigpl, line 51: I understand that `Path` is os aware and `datadir / DENY_LIST_NAME` will use `\` for Windows?

etam: Yes

tests/golem/task/test_acl.py
etam: https://docs.pytest.org/en/latest/unittest.html

> Benefits out of the box
> By running your test suite with pytest you can make use of several features, in most cases without having to modify existing code:
> - [...]
> - Use plain assert-statements instead of self.assert* functions

Krigpl: Huh, I must have been blind because I thought the messages are not descriptive. TIL

Krigpl (approved):

PR #1944, opened 2018-01-22
SmartContractInterface draft

Krigpl:
This minimal effort attempt of creating SmartContractInterface draft.
PaymentProcessor and IncomesKeeper will use this class which will hide all the communication with geth and smart contracts.
Eventually SmartContractInterface, Client and Token classes will be moved to a separate repo (https://github.com/golemfactory/golem-smart-contracts-interface) and will become a new module.

This is a preparation so make this happen. By no means SmartContractInterface API or implementation is final.

### DIFF
--- a/golem/ethereum/smartcontractinterface.py
+++ b/golem/ethereum/smartcontractinterface.py
@@ -1,0 +1,53 @@
+from typing import Any, Dict, List, Optional
+from ethereum.transactions import Transaction
+
+
+class SmartContractInterface(object):
+    def __init__(self, geth_client, token):
+        self.GAS_PRICE = token.GAS_PRICE
+        self.GAS_PER_PAYMENT = token.GAS_PER_PAYMENT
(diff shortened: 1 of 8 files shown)


golem/ethereum/smartcontractinterface.py
```diff
@@ -0,0 +1,53 @@
+from typing import Any, Dict, List, Optional
+from ethereum.transactions import Transaction
+
+
+class SmartContractInterface(object):
+    def __init__(self, geth_client, token):
+        self.GAS_PRICE = token.GAS_PRICE
```
mfranciszkiewicz, line 7: It might be a good idea to move all constants to a separate file.

Krigpl: I plan to make number of changes to refactor and clean this up so these will get sorted out.
This PR is just to move everything in one place.

Elfoniok (approved):

PR #1946, opened 2018-01-22
Ack or Reject ReportComputedTask

### LINKED ISSUE #1626: Add a response for MessageReportComputedTask, Ack or Reject
Requestor should response to MessageReportComputedTask with MessageAckReportComputedTask or MessageRejectReportComputedTask as described in
https://docs.google.com/document/d/1QMnamlNnKxichfPZvBDIcFm1q0uJHMHJPkCt24KElxc/view#heading=h.dhlbf3h55z2u

jiivan:
Reviewers: Please start from `tests/golem/task/test_concent_logic.py`. This test case should explain all other changes.

fixes: #1626


tests/golem/task/test_concent_logic.py
```diff
@@ -0,0 +1,176 @@
+"""Tests related to ConCent integration.
```
etam: Con **C** ent ?

tests/factories/messages.py
```diff
@@ -0,0 +1,68 @@
+# pylint: disable=too-few-public-methods
+import factory
+from golem_messages.message import tasks
+
+
+class TaskOwner(factory.DictFactory):
```
etam, line 6: Isn't it possible to reuse those from golem-messages?

tests/golem/task/test_concent_logic.py
```diff
@@ -0,0 +1,176 @@
...
+
+reject_reasons = concents.RejectReportComputedTask.REASON
+
+# pylint: disable=protected-access
+
+
+class ReactToReportComputedTaskTestCase(unittest.TestCase):
+    def setUp(self):
+        self.task_session = tasksession.TaskSession(mock.MagicMock())
+        self.datadir = tempfile.TemporaryDirectory('golem-test-concent')
```
etam: use `golem.testutils.TempDirFixture`?

tests/golem/task/test_concent_logic.py
```diff
@@ -0,0 +1,176 @@
...
+        self.msg.task_to_compute.sig = '31337'
+        self.task_session._react_to_report_computed_task(self.msg)
+        dropped_mock.assert_called_once_with()
+
+    @mock.patch('golem.task.tasksession.TaskSession.send')
+    def test_task_deadline(self, send_mock):
+        "Reject after task timeout"
+        after_deadline = datetime.datetime.utcnow() \
+            + datetime.timedelta(hours=1, seconds=1)
+        with freeze_time(after_deadline):
```
etam: This time freezing would be better if you'd use a single timestamp as a starting point both here and in setUp.

etam (changes requested):

tests/golem/task/test_concent_logic.py
jiivan: I'm not sure about that. Concent suggests there is a typo (consent), but it's more widely used as concent now. Changed.

tests/factories/messages.py
jiivan: It's hard because tests are not included in golem-messages library. I don't know how to resolve this reasonably now. Also they're really simple straightforward declarative definitions. I think the abuse of DRY here is not big.

tests/golem/task/test_concent_logic.py
jiivan: done

tests/golem/task/test_concent_logic.py
jiivan: done

tests/factories/messages.py
etam: You could make it as a part of golem-messages library (something like golem.testutils) and import it in both places.

tests/golem/task/test_concent_logic.py
```diff
@@ -0,0 +1,176 @@
+"""Tests related to Concent integration.
+
+Documentation:
+https://docs.google.com/document/d/1QMnamlNnKxichfPZvBDIcFm1q0uJHMHJPkCt24KElxc/
+"""
+import calendar
+import datetime
+import tempfile
```
etam: 'tempfile' imported but unused

etam (changes requested):

tests/factories/messages.py
jiivan: Could work. Created an issue for that https://github.com/golemfactory/golem-messages/issues/117

shadeofblue: @jiivan I think I'd second @etam here and - of course, acknowledging that it's _that_ much of a problem - I'd be for including the tests into the library so they can be reused, especially that we'll most likely have more and more code even there and it seems a waste indeed to have to maintain two copies...

shadeofblue:
hmm, given the above comment from @jiivan : `https://github.com/golemfactory/golem/pull/1946#discussion_r165308289` I personally don't see any other issues ... though, probably it would be better someone more versed with golem's code expressed their approval...

shadeofblue (approved):

etam (approved):

PR #1948, opened 2018-01-23
Cleanup PaymentProcessor test

Krigpl:
No changes in the production code.
Just removing unnecessary dependencies in the test, removing useless tests, fixing flaky tests by using freeze_gun and some pep8 changes.

### DIFF
--- a/golem/ethereum/node.py
+++ b/golem/ethereum/node.py
@@ -8,7 +8,6 @@ import sys
 import tempfile
 import threading
 import time
-from datetime import datetime
 from distutils.version import StrictVersion
 
 import requests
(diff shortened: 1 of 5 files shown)


Elfoniok (approved):

etam:
Do you have any tests for `SmartContractInterface`? You're removing all it's instances from tests and replacing it with mocks.

Krigpl:
@etam Not yet, as all this class does is forward calls to Token and Client which are tested. But as soon as I move it to a separate repo I start covering it with tests.

etam (approved):

PR #1949, opened 2018-01-23
Remove GNTToken

Krigpl:
TestGNT is not really compatible with GNTW so common interface doesn't make sense anymore. Maintaining TestGNT would be just annoying at this point.

### DIFF
--- a/golem/ethereum/token.py
+++ b/golem/ethereum/token.py
@@ -1,16 +1,13 @@
-import abc
 import json
 import logging
 from typing import List, Any, Optional
 
 from ethereum import abi, utils, keys
 from ethereum.transactions import Transaction
-from ethereum.utils import denoms
 
 from golem.ethereum import Client
 from golem.model import Payment
 from golem.utils import decode_hex, encode_hex
-from .contracts import TestGNT
 from .contracts import gntw
 
 logger = logging.getLogger("golem.token")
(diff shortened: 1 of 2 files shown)


golem/ethereum/token.py
```diff
@@ -40,12 +37,7 @@ def encode_payments(payments: List[Payment]):
     return args
 
 
-class AbstractToken(object, metaclass=abc.ABCMeta):
```
etam:
```python
class GNTWToken:
```

golem/ethereum/token.py
```diff
@@ -40,12 +37,7 @@ def encode_payments(payments: List[Payment]):
...
-    """
-    This is a common interface for token transactions. It hides whether we're
-    using GNT or GNTW underneath.
-    """
-
+class GNTWToken(object):
```
etam: class GNTWToken:

golem/ethereum/token.py
```diff
@@ -40,12 +37,7 @@ def encode_payments(payments: List[Payment]):
     return args
 
 
-class AbstractToken(object, metaclass=abc.ABCMeta):
-    """
-    This is a common interface for token transactions. It hides whether we're
-    using GNT or GNTW underneath.
-    """
-
+class GNTWToken():
```
etam: class GNTWToken:

Elfoniok (approved):
So after this we assume that the only thing that user can do with GNT is convert it to GNTW?

Krigpl:
> So after this we assume that the only thing that user can do with GNT is convert it to GNTW?

I wouldn't say that's gonna be the case after this PR. it was always the case with GNTW.

PR #1953, opened 2018-01-23
All requirements in .txt files

maaktweluit:
Put all requirements in `requirements.txt` files.
- No more manual requirements installed on buildbot
- Removed `-winpackager`
- Added more generic `-build`

Requires buildbot to be updated when this PR hits develop.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,8 +36,6 @@ pydispatcher
 Crypto
 scrypt
 packaging
-# GitPython < 2.1.8 has problems with latest git version
-GitPython>=2.1.8
 semantic_version
 pyOpenSSL==17.1.0
 requests==2.13.0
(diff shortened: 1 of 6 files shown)


mfranciszkiewicz (approved):
LGTM

requirements.txt
```diff
@@ -36,6 +36,8 @@ pydispatcher
 Crypto
 scrypt
 packaging
+# GitPython < 2.1.8 has problems with latest git version
+GitPython>=2.1.8
```
jiivan: Is this used anywhere outside buildbot? Maybe `requirements-build.txt` would be better place for it.

maaktweluit: Good catch! it was used for getting the version, will double check

maaktweluit: You are right, not used at all anymore, will remove it.

jiivan (approved):

PR #1976, opened 2018-01-26
Quick fix for crop generator

Elfoniok:

### DIFF
--- a/apps/blender/resources/cropgenerator.py
+++ b/apps/blender/resources/cropgenerator.py
@@ -32,7 +32,7 @@ def generate_crops(resolution, crop_scene_window, num_crops,
 
 def random_crop(min_, max_, size_):
     difference = round((max_ - size_) * 100, 2)
-    crop_min = random.randint(round(min_ * 100), difference) / 100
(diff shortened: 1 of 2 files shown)


apps/blender/resources/cropgenerator.py
```diff
@@ -32,7 +32,7 @@ def generate_crops(resolution, crop_scene_window, num_crops,
 
 def random_crop(min_, max_, size_):
     difference = round((max_ - size_) * 100, 2)
-    crop_min = random.randint(round(min_ * 100), difference) / 100
+    crop_min = random.randint(int(round(min_ * 100)), int(difference)) / 100
     crop_max = round(crop_min + size_, 2)
```
Krigpl, line 36: `crop_min` is an `int`, for consistency you should vase `crop_max` to an int as well

Krigpl (approved):

apps/blender/resources/cropgenerator.py
Elfoniok: crop_min is not int, check division after generating random int

PR #1981, opened 2018-01-26
Send UPnP ports to the monitor

### LINKED ISSUE #1978: Upnp ports should be send to the monitor
When testing Upnp on golem windows 0.11.0 binaries I get the following warning:

```
2018-01-26 13:26:28 INFO     golem.network.upnp                  b'IGD: starting discovery'
2018-01-26 13:26:29 INFO     golem.network.upnp                  b'IGD: discovery complete: http://X.X.X.X:5000/ctl/IPConn'
2018-01-26 13:26:29 INFO     golem.network.upnp                  b'IGD: network configuration: {'local_ip_address': 'X.X.X.X', 'external_ip_address': 'X.X.X.X', 'connection_type': 'IP_Routed', 'status_info': ('Connected', 1601572, 'ERROR_NONE')}'
2018-01-26 13:26:29 INFO     golem.network.upnp                  b'IGD: mapped 40102 -> 40105 (TCP)'
2018-01-26 13:26:30 INFO     golem.network.upnp                  b'IGD: mapped 40103 -> 40106 (TCP)'
2018-01-26 13:26:30 INFO     golem.network.upnp                  b'IGD: mapped 3282 -> 3283 (TCP)'
2018-01-26 13:26:30 WARNING  golem.monitor                       b'Port status: 40102: unreachable, 40103: unreachable, 3282: unreachable'
```

This is caused by my mac node having the default ports forwarded.

mfranciszkiewicz:
Resolves #1978

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -336,6 +336,8 @@ class Client(HardwarePresetsMixin):
             self.node.pub_addr)
         hyperdrive_ports = self.daemon_manager.ports()
 
+        self.node.hyperdrive_prv_port = next(iter(hyperdrive_ports))
+
         if not self.resource_server:
             resource_manager = HyperdriveResourceManager(dir_manager,
                                                          hyperdrive_addrs)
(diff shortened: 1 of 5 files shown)


golem/client.py
```diff
@@ -336,6 +336,8 @@ def start_network(self):
             self.node.pub_addr)
         hyperdrive_ports = self.daemon_manager.ports()
 
+        self.node.hyperdrive_prv_port = next(iter(hyperdrive_ports))
```
maaktweluit, line 339: what is the added value of the `iter`? 
looks like this also can be `hyperdrive_ports[0]`

maaktweluit, line 339: should we check if we received ports or will the manager handle this error?

maaktweluit (approved):
LGTM!

2 small q's, will still approve.

golem/client.py
mfranciszkiewicz: `hyperdrive_ports` (currently one) are a set, so there is no indexing, and `next` requires an iterator.

mfranciszkiewicz: A successful response from hyperdrive will contain its listening addresses / ports, unless it failed to start (and `requests` will throw a connection error). I could put a guard here but I believe it is not required.

maaktweluit: Thank you, clear :)

maaktweluit: Thank you, i agree :)

jiivan (approved):
lgtm

PR #1982, opened 2018-01-26
Explicitly check for existing UPnP mappings to prevent warnings

mfranciszkiewicz:


golem/network/upnp/igd.py
```diff
@@ -41,6 +41,12 @@ def discover(self) -> str:
         self._available = True
         return igd
 
+    def get_mapping(self,
+                    external_port: int,
+                    protocol: str = 'TCP'):
```
etam: What's the return type of this?

mfranciszkiewicz: Return type annotation added

etam (approved):

PR #1983, opened 2018-01-28
fix requirements.txt

peteut:
`pip install -r requirements.txt` fails w/ pip@9.0.1, with this patch applied it works again.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,7 +8,8 @@ peewee>=2.8.1
 Pillow==4.2.1
 psutil==5.2.2
 pluggy==0.6.0
-pyelliptic==1.5.7
+# pyelliptic==1.5.9
+git+https://github.com/mfranciszkiewicz/pyelliptic.git@openssl_1_1#egg=pyelliptic
 coincurve>=5.0.1
 asn1crypto==0.22.0
 pyasn1==0.4.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
---extra-index-url https://builds.golem.network
+# --extra-index-url https://builds.golem.network
 setuptools>=36.0.1
 appdirs>=1.4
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,4 +1,4 @@
---extra-index-url https://builds.golem.network
+# --extra-index-url https://builds.golem.network
```
jiivan, line 1: This change alone is unacceptable as there are packages missing for python3.5 which are custom build in this place.

requirements.txt
```diff
@@ -8,7 +8,8 @@ peewee>=2.8.1
 Pillow==4.2.1
 psutil==5.2.2
 pluggy==0.6.0
-pyelliptic==1.5.7
+# pyelliptic==1.5.9
+git+https://github.com/mfranciszkiewicz/pyelliptic.git@openssl_1_1#egg=pyelliptic
```
jiivan, line 12: Could be accepted only temporariliy. Finally this release should be placed on builds.golem.network.

requirements.txt
```diff
@@ -49,5 +50,5 @@ scipy==0.19.1
 pyssim
 enforce==0.3.4
 miniupnpc<2.1,>=2.0
-Golem-Messages==1.6.1
--e git://github.com/golemfactory/golem-smart-contracts-interface.git@f844eb827074d7a7278982e1c07dc167eb2bcad3#egg=golem_sci
+git+https://github.com/golemfactory/golem-messages@v1.6.1#egg=golem-messages
```
jiivan, line 53: It's a step back. All golem-messages releases are on builds.golem.network.

jiivan (changes requested):
Thank you for your contribution! Still it needs polishing to meet our qa level.

Krigpl:
Not relevant anymore

PR #1985, opened 2018-01-29
Run docker job

Elfoniok:
After generating random crop cut from subtask area, docker job will be launched in order to perform comparison. Same job will receive position of generated crop (rendered from original scene file), cut out same crop from subtask results, and come similarity metrics will be calculated and recorded in file.


apps/blender/task/verifier.py
```diff
@@ -4,26 +4,47 @@
...
 logger = logging.getLogger("apps.blender")
 
 NUM_CROPS = 3
 
 
-class BlenderVerifier(FrameRenderingVerifier):
+# pylint: disable=R0903
+class VerificationContext:
+    def __init__(self, crops_position, crop_id, crops_path):
+        self.crop_id = crop_id
```
Krigpl: I don't see there `self.crop_id` is used outside the constructor

apps/blender/task/verifier.py
```diff
@@ -4,26 +4,47 @@
...
+
+# pylint: disable=R0902
+class BlenderVerifier(FrameRenderingVerifier):
     def __init__(self, callback: Callable):
         super().__init__(callback)
         self.lock = Lock()
         self.verified_crops_counter = 0
         self.success = None
         self.failure = None
+        self.docker_image_name = 'golemfactory/image_metrics'
```
Krigpl: That could be class "constant"

Krigpl (approved):
Minor comments, I don't have any deep insights here.

apps/blender/task/verifier.py
```diff
@@ -128,16 +155,70 @@ def _generate_ctd(subtask_info, script_src):
...
+            os.mkdir(logs_dir)
+        if not os.path.exists(output_dir):
+            os.mkdir(output_dir)
+
+        resource_path = os.path.dirname(self.current_results_file)
+
+        params = dict()
+
+        params['cropped_img_path'] = os.path.join(
+            "/golem/work/tmp/output",
```
badb: Can we use constans from DockerJob here?

Elfoniok: Not really this is not '/golem/output' dir, this is path to output of blender job, where crop files are stored. This is in order to avoid copying files,

mfranciszkiewicz: I believe the constants in mind are `DockerJob.WORK_DIR`, `DockerJob.RESOURCES_DIR`, etc.

Elfoniok: Yes i know, but those cant be used here

Elfoniok:
I have tested this also on Windows, good spot with this posix path, thanks badb.

badb (approved):

apps/blender/task/verifier.py
badb: Yes, I was suggesting using DockerJob.WORK_DIR and DockerJob.RESOURCES_DIR

PR #1993, opened 2018-01-30
Add generic key value storage

Krigpl:
For cases where one needs to store some data between golem runs, like last read ethereum block.

### DIFF
--- a/golem/model.py
+++ b/golem/model.py
@@ -40,7 +40,7 @@ db = GolemSqliteDatabase(None, threadlocals=True,
 
 class Database:
     # Database user schema version, bump to recreate the database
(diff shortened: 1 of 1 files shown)


etam (approved):
LGTM, although I have a feeling it's gonna be abused somehow.

Krigpl:
@etam Maybe, maybe. If you have any other ideas how to provide functionality of storing some data between consecutive golem runs, please share. This is just what I came up with @mfranciszkiewicz yesterday which sounded good enough.

etam:
@Krigpl I have no better idea. That's why I approved this.

mfranciszkiewicz (approved):

PR #2004, opened 2018-02-01
Early RPC requests fix

mdtanrikulu:
Object attributes checked, in case of early RPC requests


golem/client.py
```diff
@@ -652,7 +652,9 @@ def get_suggested_conn_reverse(self, key_id):
         return self.p2pservice.get_suggested_conn_reverse(key_id)
 
     def get_peers(self):
-        return list(self.p2pservice.peers.values())
+        if hasattr(self.p2pservice, 'peers'):
```
mfranciszkiewicz: `self.p2pservice.peers` is always available when `self.p2pservice` is constructed; this call is actually a check whether `self.p2pservice` is equal to `None`.

golem/client.py
```diff
@@ -739,9 +741,11 @@ def get_task(self, task_id):
         return self.task_server.task_manager.get_task_dict(task_id)
 
     def get_tasks(self, task_id=None):
-        if task_id:
-            return self.task_server.task_manager.get_task_dict(task_id)
-        return self.task_server.task_manager.get_tasks_dict()
+        if hasattr(self.task_server, 'task_manager'):
```
mfranciszkiewicz: `self.task_server.task_manager` is always available when `self.task_server` is constructed. Besides `task_server` being `None`, this looks like a bug in `get_task_dict`, which assumes that every `task_id` argument is present in `task_manager.tasks`.

golem/client.py
mdtanrikulu: updated 👍

golem/client.py
mdtanrikulu: updated 👍

golem/task/taskmanager.py
```diff
@@ -661,22 +661,23 @@ def get_task_id(self, subtask_id):
...
-            # single=True retrieves one preview file. If rendering frames,
-            # it's the preview of the most recently computed frame.
-            'preview': task_type.get_preview(task, single=True)
-        }
-
-        return update_dict(dictionary,
-                           task.to_dictionary(),
-                           state.to_dictionary(),
-                           self.get_task_definition_dict(task))
+        if task:
```
badb: Can you please just write instead:
```python
 if task is None: 
    return
```
And left rest of the code with previous indendaton?

badb (approved):

mfranciszkiewicz (approved):

PR #2006, opened 2018-02-01
rebased and fixed: Requestor-oriented stats #1492 #1702

### LINKED ISSUE #1492: Send requestor-oriented stats to monitor

tworec:
rebased and fixes #1702 
resolves #1492

### DIFF
--- a/golem/task/taskrequestorstats.py
+++ b/golem/task/taskrequestorstats.py
@@ -1,0 +1,530 @@
+import logging
+import time
+from collections import defaultdict
+from typing import (  # pylint: disable=unused-import
+    NamedTuple, List, Optional, DefaultDict)
+
+from pydispatch import dispatcher
+
+from golem.task.taskstate import TaskOp, SubtaskStatus, TaskStatus, TaskState
+
+
+__all__ = ['RequestorTaskStatsManager']
+
+logger = logging.getLogger(__name__)
+
+
+TaskMsg = NamedTuple("TaskMsg", [("ts", float), ("op", TaskOp)])
+
+
+class SubtaskInfo:  # pylint: disable=too-few-public-methods
+    def __init__(self):
+        self.latest_status = SubtaskStatus.starting
+        self.messages = []
+
+
+class TaskInfo:
+    """Stores information about events related to the task.
+
+    Stores information about events that were related to a single task and
+    processes those information to get statistical information. It is probably
+    only useful for :py:class:`RequestorTaskStats` objects which fill instances
+    of this class with information.
+    """
+
+    def __init__(self):
+        self.latest_status = TaskStatus.notStarted  # type: TaskStatus
+        self._want_to_compute_count = 0
+        self.messages = []  # type: List[TaskMsg]
+        self.subtasks = defaultdict(
+            SubtaskInfo)  # type: DefaultDict[str, SubtaskInfo]
+
+    def got_want_to_compute(self):
+        """Makes note of a received work offer"""
+        self._want_to_compute_count += 1
+
+    def got_task_message(self, msg: TaskMsg, latest_status: TaskStatus):
+        """Stores information from task level message"""
+        self.messages.append(msg)
+        self.latest_status = latest_status
+
+    def got_subtask_message(self, subtask_id: str, msg: TaskMsg,
+                            latest_status: SubtaskStatus):
+        """Stores information from subtask level message"""
+        self.subtasks[subtask_id].latest_status = latest_status
+        self.subtasks[subtask_id].messages.append(msg)
+
+    def subtask_count(self) -> int:
+        """Number of subtasks of this task"""
+        return len(self.subtasks.keys())
+
+    def collected_results_count(self) -> int:
+        """Returns number of successfully received results
+
+        This is just a sum of verified and not accepted counts. That does not
+        take "unexpected" results into account, that is results received
+        which were not previously requested.
+        """
+        return (self.verified_results_count() +
+                self.not_accepted_results_count())
+
+    def verified_results_count(self) -> int:
+        """Number of verified results of the subtasks for self task
+
+        This is equal to the number of subtasks with the latest state
+        ``SubtaskStatus.finished``.
+        """
+        cnt = 0
+        for st in self.subtasks.values():
+            if st.latest_status == SubtaskStatus.finished:
+                cnt += 1
+        return cnt
+
+    def _subtasks_count_specific_task_ops(self, op: TaskOp):
+        cnt = 0
+        for st in self.subtasks.values():
+            for msg in st.messages:
+                if msg.op == op:
+                    cnt += 1
+        return cnt
+
+    def not_accepted_results_count(self) -> int:
+        """Number of times a subtask failed verification"""
+        return self._subtasks_count_specific_task_ops(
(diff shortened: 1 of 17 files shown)


tworec:
closed by mistake

tworec:
hi @maaktweluit, this is rebase + some style fixes for #1702 which you have previously reviewed

golem/task/taskrequestorstats.py
```diff
@@ -0,0 +1,521 @@
...
+        Note that if executed for a task that is still in progress this will
+        also include subtasks that are actively sending results at the moment
+        of a call.
+        """
+        cnt = 0
+        for st in self.subtasks.values():
+            download_in_progress = False
+            for msg in st.messages:
+                if msg.op == SubtaskOp.RESULT_DOWNLOADING:
+                    download_in_progress = True
```
maaktweluit, line 121: should it break the second loop when it found what it was looking for?
Or do you only want the last message result?

golem/task/taskrequestorstats.py
```diff
@@ -0,0 +1,521 @@
...
+        ``TASK_ABORTED`` and ``TASK_TIMEOUT`` messages. If the
+        task is in progress then current time is taken instead of the
+        latter. Note that the time spent paused is also included in
+        the total time.
+        """
+        start_time = 0.0
+        finish_time = 0.0
+
+        if not self.is_completed():
+            finish_time = time.time()
```
maaktweluit, line 143: @jiivan had some down-sides on using time, please double check if this is correct.

golem/task/taskrequestorstats.py
```diff
@@ -0,0 +1,521 @@
...
+        if self.is_completed():
+            return 0
+
+        cnt = 0
+        for st in self.subtasks.values():
+            if st.latest_status in [SubtaskStatus.finished,
+                                    SubtaskStatus.failure]:
+                continue
+            in_progress = False
+            for msg in st.messages:
```
maaktweluit, line 211: break loop when found?

golem/task/taskrequestorstats.py
```diff
@@ -0,0 +1,521 @@
...
+        elif op.subtask_related():
+            assert subtask_id
+            self.tasks[task_id].got_subtask_message(
+                subtask_id,
+                TaskMsg(ts=time.time(), op=op),
+                task_state.subtask_states[subtask_id].subtask_status)
+
+        else:
+            # Unknown operation, log problem
+            logger.debug("Unknown operation %r", op.name)
```
maaktweluit, line 448: would log this as an info warning, debug logs are not visible by default.
you can ignore this comment if an unknown OP has no affect for the end user

tests/apps/blender/task/test_blenderverifier.py
```diff
@@ -1,11 +1,10 @@
 import os
-import mock
 
-from golem.core.common import timeout_to_deadline
+import mock
```
maaktweluit: Please use `unittest.mock` from the stdlib, `mock` is deprecated since we upgraded to python3

tests/golem/monitor/test_monitor.py
```diff
@@ -1,23 +1,29 @@
-import mock
 import random
 from unittest import TestCase
 
+import mock
```
maaktweluit: `unittest.mock`

maaktweluit (changes requested):
LGTM!

Some small comments, will approve when replied to or fixed :)

golem/task/taskrequestorstats.py
tworec: no, we want to see the last **message**, because it is indicator of current **subtask** status.
if it is `RESULT_DOWNLOADING` then increase counter

golem/task/taskrequestorstats.py
tworec: @jiivan can you please comment on this?

golem/task/taskrequestorstats.py
tworec: same as above

golem/task/taskrequestorstats.py
tworec: should be rather hidden form user

tests/apps/blender/task/test_blenderverifier.py
tworec: I'm touching this file only to please linters - this is pure/dumb/automatic import optimization.

tests/golem/monitor/test_monitor.py
tworec: ok, I'll fix it here

golem/task/taskrequestorstats.py
maaktweluit: thank you for the explanation

in that case why do we loop?
`msg = st.messages[-1]` should also get the last message

golem/task/taskrequestorstats.py
maaktweluit: same as above

tests/apps/blender/task/test_blenderverifier.py
maaktweluit: the linter only complains on changed lines: https://github.com/golemfactory/golem/blob/develop/lintdiff.sh#L115

though i like every single linter please :)

golem/task/taskrequestorstats.py
tworec: good point! :) I was not precise.
Task is considered not_downloaded, when there was `RES_DOWNL` msg, not followed by `FINISHED` nor `NOT_ACCEPTED`. Other msg are ignored

tests/apps/blender/task/test_blenderverifier.py
tworec: ;) It was complaining here. I don't know precisely why. Maybe because of my manual rebase steps (those RDP commits were very old and had conflicts)

golem/task/taskrequestorstats.py
maaktweluit: This makes the singe item fetch a lot less feasible.
Still think the loop is not optimal, but its nice and readable.

Since we had some questions about it maybe a comment would help our future self

golem/task/taskrequestorstats.py
tworec: @maaktweluit please approve (via rocket chat)

golem/task/taskrequestorstats.py
tworec: ok. no problem. i'll put a few words

maaktweluit (approved):

PR #2008, opened 2018-02-01
Always check geth client version

Krigpl:
We weren't checking the version compatibility when connecting to remote nodes.

### DIFF
--- a/golem/ethereum/node.py
+++ b/golem/ethereum/node.py
@@ -66,11 +66,25 @@ class NodeProcess(object):
         self.web3 = None  # web3 client interface
         self.addr_list = [addr] if addr else get_public_nodes()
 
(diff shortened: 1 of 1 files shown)


golem/ethereum/node.py
```diff
@@ -105,6 +118,8 @@ def start(self, start_port=None):
         if identified_chain != self.CHAIN:
             raise OSError("Wrong '{}' Ethereum chain".format(identified_chain))
 
+        self._check_geth_version()
```
etam: I think it would be better to put this check before checking genesis block.

etam (approved):
LGTM, but added a suggestion.

mfranciszkiewicz (approved):

PR #2012, opened 2018-02-01
bump-up python version to 3.6

tworec:
I'm changing just comment as we were talking


mfranciszkiewicz (commented):
LGTM. How about adding `Programming Language :: Python :: 3.6` too?

tworec:
But do we really support it? Till today I was conviced so and was using py3.6, but my golem nodes (@ mac 10.13 & ubuntu 17.10) are not allways able to compute bigger tasks (bmw > 1000x1000) with multiple subtasks

mfranciszkiewicz:
@tworec Yes, we support 3.5+.  Did you compare 3.5 and 3.6 on the same machine?

tworec:
no, I didn't. Yet. 
But nonetheless, since we do not support 3.6 in build bots and distro, I would prefer not to add `Programming Language :: Python :: 3.6` for now.

mfranciszkiewicz:
We are using 3.6 exclusively on macOS.

tworec:
so you insist adding 3.6?

mfranciszkiewicz:
We want to keep the code backwards compatible with 3.5; this does not imply that we don't support Python 3.6.

Please note that:
- tests and releases for macOS are built with python 3.6,
- tests for Windows are done both on 3.5 (AppVeyor) and 3.6 (buildbots),
- release binaries for Windows are built with 3.6,
- our wheel repository at https://builds.golem.network/simple is populated with both 3.5 and 3.6 wheels,
- most macOS and Windows developer machines are running 3.6,
- 3.6 isn't a first choice because it's not available by default on Ubuntu LTS,
- we do fix incompatibilities caused by development done on 3.6.

Stating that we only support 3.5 is not true.

Elfoniok:
@mfranciszkiewicz 
`3.6 isn't a first choice because it's not available by default on Ubuntu LTS`
This only matter for developers, the main question here is what we can build with pyinstaller.  If you say that windows and mac'os are built with python 3.6 that's great.

tworec:
@maaktweluit I have spoken with @badb and @mfranciszkiewicz and they both support idea of switching to 3.6 as our primary and only development python version.

Can you please check if py3.6 works well on linux (ubuntu 16)?

AFAIK this is last OS platform where we supply py3.5 within golem binary installation.

maaktweluit:
@tworec py3.6 seems to work well on my test buildbot-worker.
I have followed this guide:
https://www.caseylabs.com/how-to-create-a-python-3-6-virtual-environment-on-ubuntu-16-04/
All that was missing was the `python3.6-dev` package.
Ran `pytest` and `pyinstaller` with success on a 3.6 venv.

Next step would be to add py3.6 to the ansible setup scripts so we can update all agents at once.

tworec:
great! thanks @maaktweluit !
Would you amend the setup of all agents then?

jiivan:
@tworec I've been using 3.6 since golem switching from py2 to py3. Only issues that i've experienced were related to maintaining compatibility with 3.5

The full bumping means updating all CI to use python 3.6.

tworec:
that's great @jiivan, your testimony harden basis for switching to py3.6

maaktweluit:
@tworec The pythton 3.6 pull requests are up:
https://github.com/golemfactory/ansible-role-golem-dev-linux/pull/1
internal/ansible-buildbot/merge_requests/5

Will invite reviewers and try to merge them before the buildbot release end of this week.

maaktweluit (approved):
LGTM!

PR #2016, opened 2018-02-02
propagate RemoveTask message

### LINKED ISSUE #1942: RemoveTask Message is not forwarded
Currently remove task message is not forwarded. So removed task headers are still replicated on distance node. For successfully removing task headers from the network, we must do it in one broadcast.

What is more, there is a possibility that in time of propagating this message it might happen that some node already sent a task header list with removed task header to node which already processed remove task message.

etam:
Fixes #1942


jiivan (approved):
lgtm

PR #2017, opened 2018-02-02
P2PService send get tasks cleanup

etam:


golem/network/p2p/p2pservice.py
```diff
@@ -229,11 +229,13 @@ def sync_network(self):
            Remove excess information about peers
         """
         super().sync_network(timeout=self.last_message_time_threshold)
-        if self.task_server:
-            self.__send_message_get_tasks()
 
         now = time.time()
 
+        if self.task_server and now - self.last_tasks_request > TASK_INTERVAL:
```
Krigpl: What is this `self.task_server` check for? I don't see `task_server` being used anywhere here explicitly so feels like this check either doesn't matter or should be somewhere else

etam: This is some weird state machine. `P2PService` is given access to `TaskServer` by `Client` in `start_network` method. So we don't want to send `GetTasks` messages, before we can receive them.

etam: As I see, `P2PService` makes an assumption that it won't receive any tasks messages, before it's ready to handle them. Fixing this would require bigger cleanup.

Krigpl: derp

Krigpl (approved):

golem/network/p2p/p2pservice.py
Krigpl: maybe add a comment then, because it's not obvious

PR #2015, opened 2018-02-01
Database migration: migration module

mfranciszkiewicz:

PR #2024, opened 2018-02-02
Database migration: script generator and checks with setup.py

mfranciszkiewicz:


golem/database/schemas/010_schema.py
```diff
@@ -0,0 +1,260 @@
...
+    > migrator.drop_not_null(model, *field_names)
+    > migrator.add_default(model, field_name, default)
+
+"""
+
+import datetime as dt
+import peewee as pw
+
+try:
+    import playhouse.postgres_ext as pw_pext
```
etam, line 31: unused import?

golem/database/schemas/011_schema.py
```diff
@@ -0,0 +1,72 @@
...
+except ImportError:
+    pass
+
+
+def migrate(migrator, database, fake=False, **kwargs):
+    """Write your migrations here."""
+
+    migrator.drop_index('hardwarepreset', 'name')
+
+    migrator.add_index('hardwarepreset', 'name', unique=True)
```
etam, line 40: Why is there dropping and adding index? Rollback seems to do the same.

etam:
There is a migration from 10 to 11, but what are the model changes?

golem/database/schemas/011_schema.py
mfranciszkiewicz: I see nothing strange in rebuilding indexes here, and peewee_migrate probably does this as a safeguard. In fact, migration / upgrade would a good place to vacuum the database.

golem/database/schemas/010_schema.py
mfranciszkiewicz: Also, the imports are not organized correctly. This is only one of the issues with scripts generated by peewee_migrate.

mfranciszkiewicz:
Changes are formal, since I’ve introduced 2 new field types for peewee_migrate compat. Db schema does not change.

etam:
The changes in model, I asked for, are in #2015

mfranciszkiewicz:
@etam You already have your answer. Please either approve or reject and specify the reason.

etam (approved):

PR #2037, opened 2018-02-05
Make PaymentProcessor use latest SmartContractInterface

Krigpl:
In particular SmartContractInterface is just a simple interface without any hidden logic like GNT to GNTW conversion, so this has been moved to `GNTConverter`.
Also we don't need to pass private key to PaymentProcessor anymore.
And some other minor clean ups, like consistent types for transaction hashes or simplified interfaces.

### DIFF
--- a/golem/ethereum/gntconverter.py
+++ b/golem/ethereum/gntconverter.py
@@ -1,0 +1,82 @@
+import logging
+
+from ethereum.utils import denoms
+
+log = logging.getLogger("golem.gnt_converter")
+
+
+class GNTConverter:
+    REQUIRED_CONFS = 2
+
+    def __init__(self, sci):
+        self._sci = sci
+        self._deposit_address = None
+        self._tx_hash = None
+        self._amount_to_convert = None
+
+    def convert(self, amount: int):
+        if self.is_converting():
+            # This isn't a technical restriction but rather a simplification
+            # for our use case
+            raise Exception('Can process only single conversion at once')
+
+        self._amount_to_convert = amount
+        self._update_personal_deposit_address()
+
+    def is_converting(self):
+        if self._awaiting_transaction():
+            return True
+
+        if not self._amount_to_convert:
+            return False
+
+        if self._update_personal_deposit_address():
+            return True
+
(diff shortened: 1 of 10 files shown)


golem/ethereum/gntconverter.py
```diff
@@ -0,0 +1,82 @@
+import logging
+
+from ethereum.utils import denoms
+
+log = logging.getLogger("golem.gnt_converter")
+
+
+class GNTConverter:
+    REQUIRED_CONFS = 2
```
etam, line 9: Is it possible to have `is_converting` and `is_updating_personal_deposit_address` that do not change the state of object? I understand how this works, but it's quite confusing to see that `is_converting` does some work and `update_personal_deposit_address` being used in an `if` statement.

Krigpl: `is_converting` would be split into two methods and the usage would look like:
```
if converter.is_converting():
  converter.keep_converting()
```
I dunno whether it's better. 

`update_personal_deposit_address` suggests modifying the state and imo it's fine to return bool to indicate whether something has been updated or not, what's the issue here?

etam (approved):

etam:
mypy is sad

Krigpl:
mypy is sad because it thinks it can do static analysis but it can't

etam:
> mypy is sad because it thinks it can do static analysis but it can't

What?

Here, done it for you:

> golem/ethereum/paymentprocessor.py:175: error: Invalid tuple literal type

Use `Tuple[..., ...]` instead of `(..., ...)`.

> golem/ethereum/paymentprocessor.py:183: error: Unsupported left operand type for - ("None")

Add type comment to variable initialization in constructor.

> golem/ethereum/paymentprocessor.py:71: error: Name 'Dict' is not defined

Add `from typing import Dict`.

Krigpl:
I was referring to the only one error caused by this PR which is this one:
> golem/ethereum/paymentprocessor.py:183: error: Unsupported left operand type for - ("None")

While your solution will work I don't think it's valid to just assume this variable will always be None.

PR #2051, opened 2018-02-06
Task result packaging and SHA1

### LINKED ISSUE #2021: checksum in the message ReportComputedTask
The field checksum in the message ReportComputedTask should be calculated from ZIP file without encryption.

mfranciszkiewicz:
- result packaging moved to `TaskComputer.task_computed` (executed in the computation thread)
- result package SHA1 sum calculation is a part of the packaging process
- SHA1 is a digest of the non-encrypted ZIP file
- `WaitingTaskResult` carries information on the generated package

Resolves #2021


badb (approved):

jiivan (approved):
👍

PR #2062, opened 2018-02-07
CompTaskKeeper should be informed about receiving cannot assign method

badb:


etam (approved):

tests/golem/task/test_tasksession.py
```diff
@@ -645,6 +645,20 @@ def test_subtask_to_task(self):
         assert not self.task_session._subtask_to_task('sid_1', Actor.Provider)
         assert not self.task_session._subtask_to_task('sid_2', Actor.Requestor)
 
+    def test_react_to_cannot_assign_task(self):
+        task_keeper = CompTaskKeeper(pathlib.Path(self.path))
```
jiivan: You could use `self.new_path`  instead of `pathlib.Path(self.path)`

jiivan (approved):

PR #2066, opened 2018-02-08
fix report_calls with once=True to actually call once

etam:


mfranciszkiewicz (changes requested):
`once` is for reporting, not calling, once

PR #2068, opened 2018-02-08
Merge Income and ExpectedIncome

Krigpl:
These tables are almost identical. And currently we lose `accepted_ts` when moving from `ExpectedIncome` to `Income`.
Income's lifecycle:
- created expected income
- approved by requestor - `accepted_ts` not null
- paid by requestor  - transaction not null

### DIFF
--- a/golem/transactions/incomeskeeper.py
+++ b/golem/transactions/incomeskeeper.py
@@ -1,11 +1,6 @@
 # -*- coding: utf-8 -*-
-import datetime
 import logging
-import peewee
-from pydispatch import dispatcher
 
-from golem.model import db
-from golem.model import ExpectedIncome
 from golem.model import Income
 
 logger = logging.getLogger("golem.transactions.incomeskeeper")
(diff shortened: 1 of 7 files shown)


golem/transactions/incomeskeeper.py
```diff
@@ -22,72 +17,30 @@ def stop(self):
...
-        except ExpectedIncome.DoesNotExist:
-            logger.info("ExpectedIncome.DoesNotExist "
-                        "(sender_node_id %r, "
-                        "subtask_id %r, value %r) ",
-                        sender_node_id, subtask_id, value)
-
-        try:
-            with db.transaction():
-                income = Income.create(
+        with Income._meta.database.transaction():
```
jiivan: Is it possible to receive two transactions to the same Income?

golem/transactions/incomeskeeper.py
```diff
@@ -105,31 +58,21 @@ def expect(self, sender_node_id, subtask_id, value):
     def update_awaiting(self, subtask_id, accepted_ts):
         try:
             # FIXME: query by (sender_id, subtask_id)
-            income = ExpectedIncome.get(subtask=subtask_id)
-        except ExpectedIncome.DoesNotExist:
+            income = Income.get(subtask=subtask_id)
```
jiivan, line 72: Is a check/query needed here to verify that income is awaiting (`transaction IS NULL`)?

tests/golem/transactions/test_incomeskeeper.py
```diff
@@ -42,38 +39,38 @@ def _test_expect_income(self, sender_node_id, subtask_id, value):
             value=value
         )
         with db.atomic():
-            expected_income = ExpectedIncome.get(sender_node=sender_node_id,
-                                                 subtask=subtask_id)
+            expected_income = Income.get(
+                sender_node=sender_node_id,
+                subtask=subtask_id,
```
jiivan: `transaction IS NULL`?

golem/transactions/incomeskeeper.py
Krigpl: It's not. Since we are operating on batch transfers we do the mapping which transaction covers which incomes. `received` doesn't make much sense the way it is implemented right now. I'll change that to make it more clear.

golem/transactions/incomeskeeper.py
Krigpl: Right, even `accepted_ts IS NULL`.

mfranciszkiewicz (commented):
LGTM, will approve when issues pointed out by @jiivan are resolved

jiivan (approved):

mfranciszkiewicz (approved):

PR #2077, opened 2018-02-09
Database migration: updated scripts and migration fallback

mfranciszkiewicz:
- re-create database if migration is not possible (from version < 10)
- proper migration version range support
- migration scripts updated
- additional migration tests

### DIFF
--- a/golem/database/database.py
+++ b/golem/database/database.py
@@ -4,7 +4,7 @@ from os import path
 from peewee import SqliteDatabase
 from playhouse.shortcuts import RetryOperationalError
 
-from golem.database.migration.migrate import migrate_schema
+from golem.database.migration.migrate import migrate_schema, NoMigrationScripts
 
 log = logging.getLogger('golem.db')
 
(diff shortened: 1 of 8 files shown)


golem/database/database.py
```diff
@@ -43,14 +43,23 @@ def get_user_version(self) -> int:
     def set_user_version(self, version: int) -> None:
         self.db.execute_sql('PRAGMA user_version = {}'.format(version))
 
-    def _create_database(self) -> None:
+    def _drop_tables(self):
+        log.info("Removing database")
```
maaktweluit: Should this message also be renamed to something like "removing tables"?

maaktweluit (approved):
LGTM!

One tiny comment

golem/database/database.py
mfranciszkiewicz: Oh yes, I missed that, thanks

golem/database/migration/migrate.py
```diff
@@ -1,9 +1,23 @@
+import logging
+
+from peewee import DatabaseError
 from peewee_migrate import Migrator
 
 from golem.database.migration import default_migrate_dir, patch_peewee
 from golem.database.migration.router import Router
 
 
+log = logging.getLogger('golem.db')
```
jiivan: `logger` is more popular in golem

jiivan (approved):
lgtm

PR #2082, opened 2018-02-09
Concent Ping

jiivan:
* Ping concent periodically to receive mesages
* Handle received message from any response from concent

Reviewers please look at:
 1. `golem/network/concent/handlers_library.py`
 1. `golem/network/concent/received_handler.py`


golem/network/concent/client.py
```diff
@@ -80,43 +84,15 @@ def send_to_concent(msg: message.Message, signing_key, public_key) \
     return response.content or None
 
 
-class ConcentRequestStatus(Enum):
```
shadeofblue, line 83: hmm... have those statuses proved unnecessary?

golem/task/server/resources.py
```diff
@@ -0,0 +1,123 @@
+import logging
```
shadeofblue, line 1: @jiivan it may be useful to provide a short docstring here to explain what `resources` are in this context so that someone reading the code has some initial idea of what the mixing implemented here pertains to...

shadeofblue (approved):
as far as I could tell, lgtm, with two small comments

golem/network/concent/client.py
jiivan: Unfortunately they did

tests/factories/messages.py
```diff
@@ -82,3 +84,20 @@ class Meta:
         model = tasks.SubtaskResultsRejected
 
     report_computed_task = factory.SubFactory(ReportComputedTask)
+
+
+class ServiceRefused(factory.Factory):
```
shadeofblue, line 89: I'd generally name those factories `SomethingFactory` to eliminate possible confusion between the factory classes and the message classes...

jiivan: There is some risk of confusion, but aren't namespaces the answer to that? :)

`import this`

*The Zen of Python, by Tim Peters*
[...]
*__Namespaces__ are one honking great idea -- let's do more of those!*

shadeofblue: @jiivan hmm... yes, of course, I know those classes exist in separate modules and can be referred to using the namespace prefix... maybe it's a matter of personal preference though... I'm somehow reluctant to have classes named the same when they're closely related...

PR #2090, opened 2018-02-12
Kt split ansible ci roles

ktomala:
This fixes https://github.com/golemfactory/golem-internal/issues/32


scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
```diff
@@ -2,29 +2,53 @@
     file: site_settings.yml
     name: site_settings
 
+- name: Register space required in GB for vs2017 install
```
maaktweluit, line 5: Free disk space should be stored as fact. Isnt this handled by default? then you only need to compare

Also only python fails now, i would add an assert and exit the whole role.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
```diff
@@ -2,29 +2,53 @@
...
+  set_fact:
+    required_space: 35
 
-- win_chocolatey:
-    name: "{{ item.name }}"
-    params: "{{ item.params|default('') }}"
+- name: Test if we have winrm connection
```
maaktweluit: what is the added value of this test? when collecting the facts it will fail if no connection is available.

maaktweluit (changes requested):
Good split!

Some small comments, please reply or fix and i will approve

Where are the roles installed now? I don not see them added to `requirements.yml`

We should also review the 3 new repo's are they in gitlab?

maaktweluit:
P.S. When will we move this to the other repo? I would prefer to merge it there.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
ktomala: This has to be done, because it will wait for winrm service to start (which starts up to 60 seconds after Windows launch). This is necessary so the next steps won't fail, especially when we will use on-demand workers (auto scaling).

ktomala:
I think we should merge this here, before splitting to another repo dedicated to buildbot, especially that our gitlab repo is not public.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
ktomala: This registers free disk space as a fact, so we can control how much space would be required.

maaktweluit: i mean it should be already stored as fact.
```
- debug: var=item.mount
  with_items:
  - "{{ ansible_mounts }}"
```

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
```diff
@@ -2,29 +2,53 @@
...
+- name: Register freediskspace as a fact
+  set_fact:
+    freediskspace: "{{ (getpsdrive.stdout | from_json).Free // 1073741824 }}"
+
+
+- name: Install python3
+  win_chocolatey:
+    name: python3
+    when:
+    - freediskspace | int > required_space
```
maaktweluit: Why skip python when there is not enough free disk space?
Would prefer an assert at the top that fails based on default facts.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
```diff
@@ -2,29 +2,53 @@
...
+  win_chocolatey:
+    name: visualstudio2017-workload-nativedesktop
+    params: "--includeOptional"
     timeout: 5400
-  with_items:
-    - name: visualstudio2017-workload-nativedesktop
-      params: "--includeOptional"
 
-- name: install pypiwin32
+- name: Install pypiwin32
```
maaktweluit, line 56: This change makes the `name` fields inconsistent.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
maaktweluit: AFAIK ansible will not connect without winrm.
If i am wrong and this is required i think it belongs in the playbook, it has nothing to do with buildbot-worker.

maaktweluit (changes requested):

maaktweluit:
Merging here can not be released ( deployed ) until we move to the other repo, since there are pending changes there too. I'm fine with that.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
ktomala: When I have tested on windows 10 after VM starting this was required, otherwise there was problems and ansible failed unexpectedly. It means really no harm, except for a slight delay. I'd opt for leaving this here.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
ktomala: Ansible debug command prints the fact, it doesn't set it! set_fact is for setting the fact, which is used correctly here. If you want it as variable we can move it to defaults.

maaktweluit: no need to set it, it should already been set by ansible while gathering initial facts

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
ktomala: What do you mean? All previous fields are upper case. Only the ones much below are lower case

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
maaktweluit: I'm OK with keeping it.
Though then I opt for putting it in the playbook, it has nothing to do with the role buildbot-worker.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
maaktweluit: all fields used to be consistent, not a big deal but nice to keep in one way everywhere.

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
ktomala: Ok, misunderstanding. Added FIXME comment as it will require enabling gather_facts: True which might be another can of worms. I'd suggest to do it with https://github.com/golemfactory/golem-internal/issues/49

scripts/ci/roles/buildbot-worker-windows/tasks/main.yml
ktomala: Fixed those fields to be all capitalized

maaktweluit (approved):

PR #2093, opened 2018-02-12
Local geth node cleanup

Krigpl:
`Node` class is now responsible only for creating local Geth process. It wasn't doing much for remote connections anyway.
Connectivity check is done in the SCI module.

### DIFF
--- a/golem/ethereum/node.py
+++ b/golem/ethereum/node.py
@@ -1,14 +1,11 @@
 import atexit
 import logging
 import os
-import random
 import subprocess
 import sys
 import tempfile
 import threading
 import time
-
-from web3 import Web3, IPCProvider, HTTPProvider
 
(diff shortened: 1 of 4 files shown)


mfranciszkiewicz:
I'm not really convinced that managing remote `geth` nodes should be a part of `EthereumTransactionSystem`. Why is this change required?

Krigpl:
Good point.
I wanted to move all `web3` stuff into `golem_sci`, so this project only uses high level methods like `new_sci_rpc/new_sci_ipc` and doesn't have to create providers and `Web3` objects. It's not strictly required for anything, it was meant to be a simplification. I'll go with a different approach.

etam:
@Krigpl is this PR still relevant? If not, please close it.

Krigpl:
Yeah, I meant to change it to something else but didn't have a chance yet. Will do it another time.

PR #2096, opened 2018-02-12
Use requests sessions

etam:
This allows to reuse one connection for many requests.


etam:
@jiivan has a good question: Do we want to keep connection open to monitor? This way we can quicker exhaust the file descriptors on it. We're not sending requests very often, so there's not much performance lost.

etam:
Ok, so I dropped making session with monitor. But keeping a session to hyperdrive seems reasonable, right?

mfranciszkiewicz:
@etam Well, we use the sync (requests) version of the hyperdrive client mainly during startup. This change won't affect runtime too much.

What happened to sessions for geth clients?

etam:
> What happened to sessions for geth clients?

They are managed internally by web3 library

PR #2105, opened 2018-02-13
Manual Advanced Installer

maaktweluit:
The files used for the manual installer creation at 0.11.0 including the signing patch after.


ktomala (approved):
LGTM

PR #2107, opened 2018-02-13
Fixed docker version lookup

### LINKED ISSUE #1988: Make DockerToolbox installation optional if previous version was already installed
When we update Golem version on Windows and override DockerToolbox version there is a risk of changeing VirtualBox  version that user had already installed and breaking existing VirtualMachines. Reinstallation of DockerToolbox should be option, so user should be able not to update DockerToolbox version.

maaktweluit:
Resolved #1988

Made a separate PR since this is not released yet.


ktomala (approved):

maaktweluit:
@badb should i change the target branch of this PR to b0.12.0?

mfranciszkiewicz (approved):

PR #2108, opened 2018-02-13
Moved scripts/ci to gitlab.

maaktweluit:


ktomala:
Should we also generate another github repository with removed files for general public consumption?

maaktweluit:
We can make a repo for the master python script for public to use.

Buildbot roles are quite case specific now, we can make them into galaxy roles once they get better and share them with the public this way.

Was this what you meant or am i missing a part?

ktomala:
Yes, that's ok with me. Let's make an issue for this so we can share buildbot roles as galaxy ones in the future.

ktomala (approved):

PR #2116, opened 2018-02-14
Dont't ping concent if it's disabled

jiivan:
* Don't try to ping concent if it's disabled
* Use weakrefs in library


shadeofblue (approved):
looks good as far as I could tell

PR #2117, opened 2018-02-14
Download timeout and resource size

### LINKED ISSUE #2044: Timeout for downloading resources should depend on their size
Right now we're using exactly same timeout  to wait for resource to all task. It's `WAITING_FOR_TASK_TIMEOUT` which is by default 720 seconds. We would like to have instead timeout that is connected to size of the resources. Size of the resources should be declared in task header.

### LINKED ISSUE #2045: Stop downloading resources when they exceed declared sized
Make sure that HyperG checks how much data was downloaded and does not download more than declared size of data.

mfranciszkiewicz:
- sets the value of `TaskHeader.resource_size`
- resource download timeouts are based on `resource_size`
- `resource_size` and `timeout` are included in `HyperdriveClient`'s download request

Resolves #2044
Resolves #2045

`hyperg` version required: https://github.com/mfranciszkiewicz/golem-hyperdrive/pull/18


golem/client.py
```diff
@@ -597,10 +606,11 @@ def check_test_status(self):
             return result
         return self.task_test_result
 
+    @inlineCallbacks
     def create_task(self, t_dict):
         try:
-            task = self.enqueue_new_task(t_dict)
-            return str(task.header.task_id)
+            task = yield self.enqueue_new_task(t_dict)
```
jiivan, line 601: :heart_eyes:

tests/golem/network/hyperdrive/test_hyperdrive_client.py
```diff
@@ -1,8 +1,7 @@
 import json
-import unittest
 import uuid
+from unittest import mock, TestCase, skip
```
jiivan: :fearful:

jiivan (approved):
lgtm

PR #2118, opened 2018-02-14
Set default datadir earlier in the flow

Krigpl:
It's clearer to have it set explicitly earlier in the flow instead of inside the `Client` class.
Also I'm aiming to get rid of the `config_overrides` from `Client` (we are going to need to know rpc_address and rpc_port before creating `Client`).


etam (approved):
Nice!

PR #2123, opened 2018-02-14
Last update of balance information added

### LINKED ISSUE #1855: Let the user know if the balances are stale
Using the new logic from #1851 we may show a bit outdated balances to the user. Might be more friendly to let user know (with a tooltip, warning or anything) that the balances might be outdated. Mostly for informative reasons, it should get updated eventually anyway.

mdtanrikulu:
Resolves #1855.


golem/ethereum/paymentprocessor.py
```diff
@@ -126,7 +129,8 @@ def _gnt_reserved(self):
         return self.__gntw_reserved
 
     def _gnt_available(self):
-        return self.gnt_balance() - self.__gntw_reserved
+        [gnt_balance, _] = self.gnt_balance()
```
Krigpl: I think you can do `gnt_balance, _ = ...` without the square brackets

Krigpl (changes requested):
The same applies to ETH balance, not just GNT

golem/ethereum/paymentprocessor.py
```diff
@@ -72,6 +72,8 @@ def __init__(self,
         self.__faucet = faucet
         self.deadline = sys.maxsize
         self.load_from_db()
+        self.last_gnt_update = None
+        self.last_eth_update = None
```
Krigpl: these can be private (`_` prefix)

golem/ethereum/paymentprocessor.py
```diff
@@ -86,9 +88,10 @@ def eth_balance(self, refresh=False):
             if balance is not None:
                 self.__eth_balance = balance
                 log.info("ETH: {}".format(self.__eth_balance / denoms.ether))
+                self.last_eth_update = datetime.now().timestamp()
             else:
                 log.warning("Failed to retrieve ETH balance")
-        return self.__eth_balance
+        return [self.__eth_balance, self.last_eth_update]
```
Krigpl: using a tuple instead of a list is better here

golem/ethereum/paymentprocessor.py
```diff
@@ -113,20 +116,27 @@ def gnt_balance(self, refresh=False):
                     self.__gnt_balance / denoms.ether,
                     self.__gntw_balance / denoms.ether,
                 )
-        return self.__gnt_balance + self.__gntw_balance
+                self.last_gnt_update = datetime.now().timestamp()
+
+        return [
```
Krigpl: ditto (tuple instead of a list)

golem/ethereum/paymentprocessor.py
```diff
@@ -189,7 +200,8 @@ def __get_next_batch(
 
         # we need to take either all payments with given processed_ts or none
         if ind < len(payments):
-            while ind > 0 and payments[ind-1].processed_ts == payments[ind].processed_ts:  # noqa
+            while ind > 0 and payments[
```
Krigpl: that's not really clearer imo

Krigpl (changes requested):
lgtm, just a few nits

golem/ethereum/paymentprocessor.py
mdtanrikulu: js routine 🙂  updated.

golem/ethereum/paymentprocessor.py
mdtanrikulu: 👍

golem/ethereum/paymentprocessor.py
mdtanrikulu: 👍

golem/ethereum/paymentprocessor.py
mdtanrikulu: 👍

golem/ethereum/paymentprocessor.py
mdtanrikulu: I tweaked a bit more.

Krigpl (approved):

mfranciszkiewicz (approved):

PR #2125, opened 2018-02-15
Remove config overrides

Krigpl:
These config overrides can be handled more explicitly.


mfranciszkiewicz (approved):
Tests need more love, otherwise LGTM

etam (approved):

PR #2128, opened 2018-02-15
Concent check version compatibility

jiivan:


golem/utils.py
```diff
@@ -1,10 +1,13 @@
...
 import logging
 import socket
 import sys
 
+import semantic_version
+
 from ethereum.utils import sha3
 
-import binascii
+logger = logging.getLogger("golem.network.concent.client")
```
maaktweluit: The name of the logger and the file do not match, why not `__name__`?

tests/golem/network/concent/test_concent_client.py
```diff
@@ -33,6 +35,8 @@ def setUp(self):
 
     def test_message(self, post_mock):
         response = requests.Response()
+        response.headers['Concent-Golem-Messages-Version'] = \
+            golem_messages.__version__
```
maaktweluit, line 39: this can use the new constants you added too

maaktweluit (approved):
LGTM!

2 small comments

tests/golem/network/concent/test_concent_client.py
jiivan: No, no. This is HTTP response header and requires a string.

golem/utils.py
jiivan: Good catch of Mr Paste Copy code :)

PR #2134, opened 2018-02-15
Check signature in RemoveTask message

### LINKED ISSUE #2018: Anyone can cancel any task
The `RemoveTask` message contains only `task_id`. It's possible, that a rouge node could send a `RemoveTask` message for every task header it receives.

It can be mitigated by signing the `task_id` by the requestor's key. Or signing a string `"remove {}".format(task_id)`, to make sure it can't be extracted from some other message.

badb:
Resolves #2018 

Should be merged with updated message version: https://github.com/golemfactory/golem-messages/pull/149


jiivan (commented):
It generally looks good, but... Wouldn't it be more coherent to let only owner generate `RemoveTask` and let other nodes forward it in a nested message?

```python

class RemoveTaskContainer(base.Message):
    __slots__ =  ['remove_task'] + baseslots
```

Then we would only need to validate nested message signature against owners' pubkey. We could further optimize it, so `RemoveTaskContainer` will have a list of nested messages instead of a single one.

etam:
@jiivan I think it would be more complicated. You'll have to handle `RemoveTask` and `RemoveTaskContainer` messages differently. In this PR `RemoveTask` already behaves like a container, without any bloat added by nesting messages. It can also be extended to have a list of tasks.

golem/network/p2p/peersession.py
```diff
@@ -361,9 +366,19 @@ def _react_to_tasks(self, msg):
...
 
     def _react_to_remove_task(self, msg):
+        task_owner = self.p2p_service.task_server.task_keeper.get_owner(
+            msg.task_id)
+        if task_owner is None:
+            return
+        data = remove_task_string(msg.task_id)
+        if not self.p2p_service.keys_auth.verify(msg.owner_signature, data,
+                                                 task_owner):
+            logger.warning("Someone tries to remove task header: %s without "
```
etam: I think it should be an info, not warning. It's not that important.

jiivan:
In your solution we could just change protocol and put `[[task, owner_sig], ...]` instead of task_id.

Nested messages list solution could be clarified to let owner generate `RemoveTask` message and only send it nested (possibly with other `RemoveTask` messages he got form other nodes in a container message. Thus receiving end would only have one type of message to handle.

It also has a bonus of:
 1. `RemoveTask.timestamp` acts as kind of a nonce, making it hard to run a replay attack (when task_id duplicates)
 1. Timestamp can be verified against task generation time.
 1. `RemoveTask.sig` out of the box

tests/golem/network/p2p/test_peersession.py
```diff
@@ -476,6 +480,59 @@ def test_react_to_peers(self, add_peer_mock):
...
+        task_id = "test_{}".format(uuid.uuid4())
+        owner_signature = keys_auth.sign(remove_task_string(task_id))
+        assert keys_auth.verify(owner_signature, remove_task_string(task_id),
+                                keys_auth.get_key_id())
+        msg = message.RemoveTask(task_id=task_id,
+                                 owner_signature=owner_signature)
+        with self.assertNoLogs(logger, level="WARNING"):
+            self.peer_session._react_to_remove_task(msg)
+
+        # Wrong task owner
```
etam: I think it's better to keep such test cases as a separate functions. When in one function, one test case might be affected by a state created by previous one. It's also harder for a reader to deduce the initial state of a test case. It's overall against the idea of a _unit_ test.

etam (commented):
I'm OK with this, but added some comments.

tests/golem/network/p2p/test_peersession.py
```diff
@@ -476,6 +480,59 @@ def test_react_to_peers(self, add_peer_mock):
...
+        send_mock.assert_called()
+        assert isinstance(send_mock.call_args[0][0], message.RemoveTask)
+
+    def test_react_to_remove_task(self):
+        keys_auth = EllipticalKeysAuth(self.path)
+        previous_ka = self.peer_session.p2p_service.keys_auth
+        self.peer_session.p2p_service.keys_auth = keys_auth
+
+        # Unknown task owner
+        task_server = mock.MagicMock()
```
jiivan: Maybe a factory?
https://github.com/golemfactory/golem/blob/35a553576a4d9256ae7bf85a41a9cb2b24120d2f/tests/factories/taskserver.py#L12-L18

badb:
Alternative implementation is here: https://github.com/golemfactory/golem/pull/2207 @jiivan @etam

PR #2139, opened 2018-02-16
Concent receive

jiivan:
Receive messages from Concent service. It's a prerequisite for _First use-case_

### DIFF
--- a/golem/network/concent/client.py
+++ b/golem/network/concent/client.py
@@ -226,7 +226,7 @@ class ConcentClientService(threading.Thread):
         In case of failure, service enters a grace period.
         """
         try:
-            req = self._queue.get(timeout=constants.PING_TIMEOUT)
+            req = self._queue.get_nowait()
         except queue.Empty:
             return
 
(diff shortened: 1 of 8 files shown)


golem/network/concent/client.py
```diff
@@ -22,19 +22,50 @@
...
+                response.status_code,
+                response.text
+            )
+        )
+
+    elif response.status_code % 400 < 100:
+        logger.warning('Concent request failed with status %d and '
+                       'response: %r', response.status_code, response.text)
+
+        raise exceptions.ConcentRequestError(
```
shadeofblue: @jiivan this exception seems partially duplicated ... maybe it would be more sensible to first check for 2xx and if it's okay, just break the execution flow and only it's not generate some logging messages and/or exceptions?... or provide a more DRY handler that you'd call with:

```
handle_concent_error(response, exception, reason)
```

that would construct the appropriate logging messages and/or exception messages?

shadeofblue (approved):
seems good, apart from a minor comment above

PR #2140, opened 2018-02-16
Create rpc router before creating client

Krigpl:
Lazy `Client` class creation only after RPC has been initialized.
This is needed for implementing the password feature. The idea is to get the password from RPC and only then try to create the `Client` object which may fail if the password is incorrect. In which case we can try again.

### DIFF
--- a/golem/node.py
+++ b/golem/node.py
@@ -1,10 +1,15 @@
-"""Compute Node"""
+import logging
+from twisted.internet import reactor
 
 from apps.appsmanager import AppsManager
 from golem.client import Client
 from golem.core.async import async_callback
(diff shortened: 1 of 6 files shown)


mfranciszkiewicz:
How will headless providers skip this step? Or is it going to be mandatory, e.g., via CLI?

Krigpl:
@mfranciszkiewicz I was thinking of either using CLI or passing the password via command line arguments (for developers mainly) or passing the password via env variable.

Krigpl:
The goal is to have encrypted wallet on the disk as well, so providing a password is mandatory.

mfranciszkiewicz (approved):

PR #2141, opened 2018-02-16
required key difficulty decrease until next release

tworec:
@badb up to you if we want to decrease it for a while


tworec:
actually we will go in opposite direction

PR #2146, opened 2018-02-19
Log the active hardware configuration

### LINKED ISSUE #2081: Log the active hardware configuration
Number of CPUs, amount of RAM, etc. should be visible in logs for debugging purposes.

etam:
fixes #2081

### DIFF
--- a/golemapp.py
+++ b/golemapp.py
@@ -1,8 +1,13 @@
 #!/usr/bin/env python
+import platform
 import sys
 import logging
 from multiprocessing import freeze_support
+
 import click
+import humanize
+import psutil
+from cpuinfo import get_cpu_info
 from ethereum import slogging
 
 import golem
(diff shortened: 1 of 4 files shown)


golemapp.py
```diff
@@ -14,6 +19,8 @@
 from golem.core.variables import PROTOCOL_CONST
 from golem.node import Node
 
+log = logging.getLogger('golem.version')
```
maaktweluit: Now this logger no longer only applies to the version maybe we can change it to `__name__`?

maaktweluit: `logger` is more popular in golem

maaktweluit (approved):
LGTM!

2 small comments, not blocking so approve!

mfranciszkiewicz:
Could we also log the active preset configuration (e.g. on change)? IMO that would allow us to have some insight to why some jobs are failing.

etam:
> Could we also log the active preset configuration (e.g. on change)?

done

mfranciszkiewicz (approved):

etam:
do we want this also in b0.12.0?

PR #2147, opened 2018-02-19
Fix additional test decision

Elfoniok:


badb (approved):
Some tests for that will be higly appreciated.

PR #2161, opened 2018-02-20
Set database version after each consecutive upgrade

mfranciszkiewicz:
If the schema upgrade process is somehow interrupted, the new database version will not be set. This may result in failing upgrades on next run.


maaktweluit (approved):
LGTM!

Much better if it would crash

badb (approved):

PR #2163, opened 2018-02-20
Limit the rate of new connections

mfranciszkiewicz:
Limits the rate that new peer sessions are spawned with.


golem/network/transport/limiter.py
```diff
@@ -0,0 +1,70 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class CallRateLimiter:
+
+    KEY = b'default'
+
+    def __init__(self,
+                 rate: float = 5.,
```
maaktweluit: Would it be possible to use int's here? i remember an issue before this should be preferred.

tests/golem/network/transport/test_tcpnetwork.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=no-member
```
maaktweluit, line 1: Can this be enabled again after the errors have passed? this will protect the rest of the file of errors

maaktweluit (approved):
LGTM!

Two small q's, no blockers so approved!

golem/network/transport/limiter.py
mfranciszkiewicz: Done

maaktweluit:
Thank you for the fix!

Appveyor has a time / assertion error, is this related or should we just rebuild?

golem/network/transport/limiter.py
```diff
@@ -0,0 +1,70 @@
...
+        self._delay_factor = delay_factor
+        self._limiter = Limiter(rate,
+                                capacity=int(capacity_factor * rate),
+                                storage=MemoryStorage())
+
+    @property
+    def delay_factor(self) -> float:
+        return self._delay_factor
+
+    def call(self,
```
etam, line 38: Names with double underscore at the beginning and end are reserved for Python.

This would be cleaner if `call` function was called once from outside and some other private method was passed to reactor. The problem of `*args, **kwargs` can be then solved by passing them as a tuple `(args, kwargs)` to this private method and unpacking them when required `*args[0], **args[1]`.

tests/golem/network/transport/test_limiter.py
```diff
@@ -0,0 +1,42 @@
...
+        assert limiter.delay_factor == delay_factor
+        assert limiter._limiter._capacity == int(rate * capacity_factor)
+        assert isinstance(limiter._limiter._storage, MemoryStorage)
+
+    @freeze_time("2018-01-01 00:00:00")
+    def test_call(self, reactor):
+        limiter = CallRateLimiter()
+        fn = mock.Mock(return_value=True)
+
+        for _ in range(5):
```
etam: minor: use variables from `CallRateLimiter` instead of hard-coding magic numbers.

tests/golem/network/transport/test_limiter.py
```diff
@@ -0,0 +1,42 @@
...
+            limiter.call(fn)
+        assert not reactor.callLater.called
+
+    @freeze_time("2018-01-01 00:00:00")
+    def test_delay(self, reactor):
+        limiter = CallRateLimiter()
+        fn = mock.Mock(return_value=True)
+        n = int(limiter._limiter._capacity * 1.5)
```
etam: minor: like above

etam (changes requested):

tests/golem/network/transport/test_limiter.py
mfranciszkiewicz: It's supposed to be more than current capacity.

golem/network/transport/limiter.py
mfranciszkiewicz: I agree on double underscores, but strongly disagree on calling nested / private functions and unmangling arg tuples being a cleaner solution.

etam (approved):

PR #2167, opened 2018-02-21
Fix loglevel commandline param

jiivan:


golem/core/common.py
```diff
@@ -18,10 +18,6 @@
 MAX_CPU_MACOS = 16
 
 ALLOWED_LOGLEVELS = [
```
maaktweluit: i think this variable was only used for this purpose, could be removed now too

maaktweluit (changes requested):
LGTM!

Nice fixes, much better approach!

maaktweluit (approved):

PR #2173, opened 2018-02-21
Don't request GNT from faucet unnecessarily

Krigpl:
Otherwise user ends up with 1k GNT and 1k GNTW which introduces unnecessary confusion.


maaktweluit (approved):
LGTM!

PR #2178, opened 2018-02-21
Synchronization of node time with monitor (#2038)

### LINKED ISSUE #2038: Verify that time is synchronized locally
If time on a node is not synchronized, many unpleasant things will happen (dropped messages, invalid concent timeouts etc.).

Probably golem-monitor could be used to verify that time is synchornized, possibly in the same step, that verifies open ports.

### LINKED ISSUE #2100: Display warning in logs if ports are not forwarded & upnp is not working
If user doesn't have port forwarding enabled and doesn't have router with upnp, he's able to connect to the golem network but will not be able to compute tasks (task resource handshake will fail) or be a requestor - his task will not be computed. 
We have a method in the monitor that checks whether user have open ports and return a warning if he doesn't. We should check whether this function is working correctly, fix it if it's not and put an warning in log file if user doesn't have forwarded port.
We shouldn't write this warning if monitor is offline. 
Later @jamuszyn should figure out how to display the proper warning for the user in GUI, but that's not in scope of this issue.

Wiezzel:
Local time of node is included in ping-me request. Monitor sends back difference between received timestamp and its own time. Of course this is **not** the exact difference between node's time and monitor's time because it includes network latency. If the received value exceeds defined maximum (10 seconds for now) a warning is displayed and an `unsynchronized` event is published to `golem.p2p` channel.

The whole ping-me feature was significantly refactored:
  * `listening` event has now `ports` argument instead of `port` (it was already representing multiple ports so the name was misleading)
  * requests sent to `/ping-me` endpoint have `ports` field instead of `port` as well
  * response from `/ping-me` is expected to contain `port_statuses`: an array of objects of the following format:
    ```
    {
      "port": 12345,            // port number
      "is_open": false,         // is the port ping-able
      "description": "timeout"  // description (e.g. "open", "timeout")
    }
    ```
  * `Node` contains a nice `port_statuses` dict instead of messy `port_status` all-containing string

Tests for ping-me were improved.

Because of API changes `PROTO_VERSION` was bumped to 1.

Closes #2038 
Closes #2100

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -204,17 +204,27 @@ class Client(HardwarePresetsMixin):
         self.rpc_publisher = Publisher(rpc_session)
         StatusPublisher.set_publisher(self.rpc_publisher)
 
-    def p2p_listener(self, sender, signal, event='default', **kwargs):
+    def p2p_listener(self, event='default', **kwargs):
         if event == 'unreachable':
-            self.node.port_status = kwargs.get('description', '')
-            return
-        if event == 'new_version':
-            log.warning(
-                'New version of golem available: %s',
-                kwargs['version']
-            )
-            self._publish(Network.new_version, str(kwargs['version']))
-            return
+            self.on_unreachable(**kwargs)
+        elif event == 'unsynchronized':
+            self.on_unsynchronized(**kwargs)
+        elif event == 'new_version':
+            self.on_new_version(**kwargs)
+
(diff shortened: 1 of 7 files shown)


Wiezzel:
Corresponding monitor PR: https://github.com/golemfactory/golem-monitor/pull/27

mfranciszkiewicz (approved):
LGTM

PR #2197, opened 2018-02-22
Upgrade web3 and eth libs

Krigpl:
web3 3.15.0 contains `eth_tester` which is useful for testing
ethereum-utils is deprecated in favor of eth-utils
ethereum-abi-utils is deprecated in favor of eth-abi

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,12 +16,13 @@ pytz
 six==1.10.0
 Twisted==17.9.0
 cffi==1.10.0
-web3==3.8.0
 ethereum==1.6.1
-ethereum-abi-utils==0.4.0
-#Newer version contains no code. Only deprecation warning.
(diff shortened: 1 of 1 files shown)


maaktweluit:
Circle gives an error:
```
Could not find a version that satisfies the requirement eth-abi==0.4.4 (from -r requirements.txt (line 21)) (from versions: 0.5.0, 1.0.0b0, 1.0.0b1)
No matching distribution found for eth-abi==0.4.4 (from -r requirements.txt (line 21))
```

Rest looks good, will approve when the test are running :)

Krigpl:
I just made successful batch transfer so at least we know those upgrades don't break basic functionalities :)

mfranciszkiewicz (approved):

PR #2155, opened 2018-02-19
increase default key difficulty + key generation in separate thread

tworec:
Key generation is now in separate step and can take a few minutes:
- less than 2 mins on average
- and 99% of cases less than 8 mins

Docker setup is performed in parallel with key generation before client network is started. Here are sample logs of first golem start:

```
2018-02-22 18:29:30 INFO     golemapp                            GOLEM Version: 0.11.0+dev444.g9d254350 
2018-02-22 18:29:30 INFO     golemapp                            Protocol Version: 23 
2018-02-22 18:29:30 INFO     golemapp                            golem_messages Version: 1.9.0 
2018-02-22 18:29:30 INFO     golemapp                            system: Darwin, release: 17.4.0, version: Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64, machine: x86_64 
2018-02-22 18:29:30 INFO     golemapp                            cpu: GenuineIntel Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz, 8 cores 
2018-02-22 18:29:30 INFO     golemapp                            memory: 32.0 GiB, swap: 2.0 GiB 
2018-02-22 18:29:30 INFO     golem.core.keysauth                 Generating new key pair
2018-02-22 18:29:48 WARNING  golem.docker.manager                Docker: pulling image golemfactory/blender:1.3
2018-02-22 18:30:13 INFO     golem.core.keysauth                 Keys generated in 43.10s
2018-02-22 18:31:28 WARNING  golem.docker.manager                Docker: pulling image golemfactory/luxrender:1.2
2018-02-22 18:31:53 WARNING  golem.docker.manager                Docker: pulling image golemfactory/image_metrics:1.1
2018-02-22 18:35:22 INFO     golem.client                        Client "2rec iMac", datadir: /Users/tworec/Library/Application Support/golem/default 
2018-02-22 18:35:22 INFO     golem.core.hardware                 updating config: name: custom, num_cores: 7, max_memory_size: 24.0 MiB, max_resource_size: 1000.0 KiB
2018-02-22 18:35:22 INFO     golem.pay                           Node Ethereum address: 0x39e294901d49f8ef1af0ed334f4c9f5600e8b191
2018-02-22 18:35:22 INFO     golem.ethereum                      GETH: connecting to remote RPC interface at http://94.23.57.58:55555
[...]

PR #2214, opened 2018-02-23
increase default key difficulty + key generation in separate thread (#2155)

mfranciszkiewicz:
- key difficulty changes from #2155
- key difficulty config fix from #2233


maaktweluit (approved):
LGTM!

Hard to distinguish changes from cleanups.
Maybe @tworec can have a look, since he made the original PR

Issue #2213, opened 2018-02-23
remove key-regeneration feature before main net release


tworec:
https://github.com/golemfactory/golem/blob/c66492d8a45d50f513b56f11ea5e0e106eeb87c5/golem/core/keysauth.py#L143-L147

PR #2220, opened 2018-02-26
Don't silently regenerate wallet

Krigpl:
Those cases shouldn't happen but if they do the solution isn't to quietly generate new identity imo.

### DIFF
--- a/golem/core/keysauth.py
+++ b/golem/core/keysauth.py
@@ -11,7 +11,6 @@ from golem_messages.cryptography import ECCx, mk_privkey, ecdsa_verify, \
 
 from golem.core.variables import PRIVATE_KEY
 from golem.utils import encode_hex, decode_hex
-from .simpleenv import get_local_datadir
 
 logger = logging.getLogger(__name__)
 
@@ -109,8 +108,7 @@ class KeysAuth:
 
     @staticmethod
     def _get_or_create_keys_dir(datadir: str) -> str:
-        path = datadir or get_local_datadir('default')
-        keys_dir = os.path.join(path, KeysAuth.KEYS_SUBDIR)
+        keys_dir = os.path.join(datadir, KeysAuth.KEYS_SUBDIR)
         if not os.path.isdir(keys_dir):
             os.makedirs(keys_dir)
         return keys_dir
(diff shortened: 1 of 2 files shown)


etam:
@badb decided this should be this way for now, but it should be changed before mainnet release. See #2213 

I don't know if just throwing an exception is a good enough solution.

Krigpl:
My reasoning is that this scenario either:
- will not happen in mainnet at all
- will happen in mainnet and handling it should be muuuuuch more carefully done, including informing the user about backup, why the address has changed, why they lost all theirs reputation etc

Second case is obviously much more complicated but we don't plan supporting that at the moment, so we're fine with 1. In which case exceptions are fine. For testnet we can just ask users to uninstall previous golem version before installing new one which we has done in the past iirc.

Anyway, I don't really need this change, I just thought it'd be clearer this way.

golem/core/keysauth.py
```diff
@@ -125,14 +123,13 @@ def _load_and_check_keys(priv_key_path: str,
...
-            return None
+            raise Exception(
+                "Wrong loaded private key size: {}".format(len(priv_key)))
 
         pub_key = privtopub(priv_key)
 
         if not KeysAuth.is_pubkey_difficult(pub_key, difficulty):
-            logger.warning("Loaded key is not difficult enough.")
-            return None
+            raise Exception("Loaded key is not difficult enough")
```
tworec, line 132: I think we should point out some solution, for example to reinstall golem, or delete old key file... or even some url to some FAQ page with question: how to transfer GNT(W) from old golem account...

tworec (changes requested):
@Krigpl your reasoning is ok. I have just one comment

golem/core/keysauth.py
etam: #2213

tworec: yes, but until #2213 will happened we should, at least, inform user how to deal with this issue... for example "Loaded key is not difficult enough. To regenerate please reinstall golem or remove key file `path`.

etam: I think this PR should be closed without merging. This issue should be properly and completely handled in #2213 and any partial solutions are bad in their own ways and not worth spending time.

tworec: I agree the latter part of your statement, but this PR is good foundation to final #2213 solution :)

tworec:
included in #2281

PR #2221, opened 2018-02-26
Provide private key name explicitly

Krigpl:
For password protection I will ask whether the key file exist before creating KeysAuth in `node.py`. We could keep the default in the constructor but I don't see the point.
Also tests were a bit flaky because they were reusing existing keys, changes to use random key name every time.

### DIFF
--- a/tests/golem/core/test_keysauth.py
+++ b/tests/golem/core/test_keysauth.py
@@ -17,6 +17,15 @@ from golem.utils import encode_hex
 class TestKeysAuth(testutils.PEP8MixIn, testutils.TempDirFixture):
     PEP8_FILES = ['golem/core/keysauth.py']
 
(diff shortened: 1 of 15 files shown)


tests/golem/core/test_keysauth.py
```diff
@@ -17,6 +17,15 @@
 class TestKeysAuth(testutils.PEP8MixIn, testutils.TempDirFixture):
     PEP8_FILES = ['golem/core/keysauth.py']
 
+    def _create_keysauth(self, difficulty, key_name=None) -> KeysAuth:
```
etam: Maybe add a default value `0` for `difficulty` and use keyword argument when calling with a raw value like `self._create_keysauth(difficulty=2)`. This will improve readability a bit.

etam (approved):
LGTM, but added a comment for a possible improvement.

PR #2232, opened 2018-02-26
Encrypt wallet

Krigpl:
Use password encrypted storage for ethereum wallet.
Using constant password right now just to split the changes into smaller PRs. Next PR will grab the password from RPC.

### DIFF
--- a/golem/core/keysauth.py
+++ b/golem/core/keysauth.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import math
 import os
@@ -6,11 +7,11 @@ from datetime import datetime
 from hashlib import sha256
 from typing import Optional, Tuple, Union
 
+from ethereum.keys import decode_keystore_json, make_keystore_json
+
 from golem_messages.cryptography import ECCx, mk_privkey, ecdsa_verify, \
     privtopub
-
 from golem.utils import encode_hex, decode_hex
-from .simpleenv import get_local_datadir
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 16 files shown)


golem/node.py
```diff
@@ -88,6 +88,8 @@ def _start_keys_auth(self) -> Deferred:
             KeysAuth,
             datadir=self._datadir,
             private_key_name=PRIVATE_KEY,
+            # TODO: user provided password
+            password='verystrongpasswodmuchsecurity',
```
etam, line 92: doge

Krigpl:
fwiw, this PR alone is ready for review and even merging.

tests/golem/core/test_keysauth.py
```diff
@@ -100,12 +126,14 @@ def test_get_difficulty(self):
         assert KeysAuth.is_pubkey_difficult(ek.public_key, difficulty)
         assert KeysAuth.is_pubkey_difficult(ek.key_id, difficulty)
 
+    @freeze_time("2017-11-23 11:40:27.767804")
     @patch('golem.core.keysauth.logger')
-    def test_key_recreate_on_increased_difficulty(self, logger):
+    def test_key_recreate_on_increased_difficulty__and_backup(self, logger):
```
etam: minor thing: `test_key_recreate_and_backup_on_increased_difficulty`

etam: or even better: `test_key_backup_and_recreate_on_increased_difficulty`

etam (approved):
sehr gut

golem/core/variables.py
```diff
@@ -8,7 +8,7 @@
 #       VARIABLES          #
 ############################
 APP_NAME = "Brass Golem"
-PRIVATE_KEY = "golem_private_key.peb"
+PRIVATE_KEY = "golem_wallet_const_password.json"
```
tworec, line 11: this enforce all beta-users to regenerate their keys... Can't we just read old key, and repack it into new file for first time?

golem/core/keysauth.py
```diff
@@ -68,7 +95,7 @@ class KeysAuth:
     key_id = ""  # type: str
     ecc = None  # type: ECCx
 
-    def __init__(self, datadir: str, private_key_name: str,
+    def __init__(self, datadir: str, private_key_name: str, password: str,
```
tworec, line 98: password doc is missing

etam: or rename it to `key_password` - no doc needed

golem/core/variables.py
etam: We're introducing the key difficulty thing, so most keys will be regenerated anyway.

Krigpl: Too much hassle for too little benefit.

etam: I think, that as long as we're on testnet, we don't need to care so much. On mainnet such feature would be mandatory.

tworec: ok

PR #2235, opened 2018-02-26
Concent/forced download - Connection Failure

### LINKED ISSUE #2177: [Concent] UC2 / Forced Download - Connection failure
Support for the Connection Failure reason to request Concent assistance in the "Forced Download" use case (use case 2)

shadeofblue:
closes #2177


golem/task/tasksession.py
```diff
@@ -685,18 +684,35 @@ def on_error(exc, *args, **kwargs):
...
+                )
+            else:
+                # otherwise, we're resorting to mediation through the Concent
+                # to obtain the task results
+                rct = get_task_message(
+                    'ReportComputedTask', task_id, subtask_id
+                )
+                if not rct:
+                    logger.error(
+                        "[CONCENT] Can't construct ForceGetTaskResult message")
```
mfranciszkiewicz: Maybe "retrieve" instead of "construct"? + Information on task / subtask id; sometimes it is hard to tell when log messages interleave.

golem/task/tasksession.py
```diff
@@ -675,6 +668,12 @@ def _react_to_task_result_hash(self, msg):
             client_options
         )
 
+        task_to_compute = get_task_message(
```
mfranciszkiewicz: Used only in `on_error`, we can skip a fetch from the database

mfranciszkiewicz (approved):
LGTM. Some small suggestions.

golem/task/tasksession.py
shadeofblue: @mfranciszkiewicz well, it's about retrieving `ReportComputedTask` which is needed to construct `ForceGetTaskResult` -> and the "can't find message" for the `ReportComputedTask` is anyway logged in the `get_task_message` helper

golem/task/tasksession.py
shadeofblue: @mfranciszkiewicz ah, right, good catch :+1:

golem/task/tasksession.py
mfranciszkiewicz: @shadeofblue Right, I missed that bit.

shadeofblue:
@mfranciszkiewicz addressed your remarks

PR #2245, opened 2018-02-27
Subtask deadline workaround

mfranciszkiewicz:
Removes the forward-only clock tolerance check for deadlines. Should be fixed to support backward clock desync.


maaktweluit (approved):
LGTM!

PR #2256, opened 2018-02-27
Restore startup events

mfranciszkiewicz:


golem/node.py
```diff
@@ -102,17 +117,21 @@ def start_docker():
...
-        rpc_session = Session(self.rpc_router.address,  # type: ignore
-                              methods=methods)
-        self.client.configure_rpc(rpc_session)
-        rpc_session.connect().addCallbacks(
-            async_callback(self._run), self._error('session connect'))
+        self.rpc_session.methods = methods
+        self.rpc_session.register_methods(methods)
+        self.client.configure_rpc(self.rpc_session)
+
+        async_run(AsyncRequest(self._run))
```
maaktweluit: Is it safe to remove the `self._error` here?

maaktweluit (approved):
LGTM!

One minor Q

golem/node.py
mfranciszkiewicz: Good point, I've added an error handler

PR #2257, opened 2018-02-27
Add some time tolerance for checking deadline.

### LINKED ISSUE #2246: Support backward clock desync for subtask deadline checks

badb:
Resolves #2246

### DIFF
--- a/golem/task/taskkeeper.py
+++ b/golem/task/taskkeeper.py
@@ -8,12 +8,12 @@ from typing import Optional
 import random
 from collections import Counter
 from golem_messages import message
+from golem_messages.constants import MTD
 from semantic_version import Version
 
 import golem
 from golem.core import common
 from golem.core.async import AsyncRequest, async_run
-from golem.core.variables import MAX_TIME_DIFF
 from golem.environments.environment import SupportStatus, UnsupportReason
 from .taskbase import TaskHeader
 
@@ -46,9 +46,16 @@ class CompTaskInfo:
         )
 
     def check_deadline(self, deadline):
+        """
+        Checks if subtask deadline defined in newly received ComputeTaskDef
+        is properly set, ie. it's set to future date, but not much further than
+        it was declared in subtask timeout.
+        :param float deadline: subtask deadline
+        :return bool:
+        """
         now_ = common.get_timestamp_utc()
         expected_deadline = now_ + self.header.subtask_timeout
-        if now_ < deadline < expected_deadline + MAX_TIME_DIFF:
+        if now_ < deadline < expected_deadline + MTD:
             return True
         logger.debug('check_deadline failed: (now: %r, deadline: %r, '
                      'timeout: %r)', now_, deadline,
--- a/golem/task/taskkeeper.py
+++ b/golem/task/taskkeeper.py
@@ -13,6 +13,7 @@ from semantic_version import Version
 import golem
 from golem.core import common
 from golem.core.async import AsyncRequest, async_run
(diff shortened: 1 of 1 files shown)


golem/task/taskkeeper.py
```diff
@@ -46,13 +47,13 @@ def __repr__(self):
...
         now_ = common.get_timestamp_utc()
-        # FIXME: create a proper deadline check for 0.12.1
-        if now_ > deadline:  # or deadline > now_ + self.header.subtask_timeout:
-            logger.debug('check_deadline failed: (now: %r, deadline: %r, '
-                         'timeout: %r)', now_, deadline,
-                         self.header.subtask_timeout)
-            return False
-        return True
+        expected_deadline = now_ + self.header.subtask_timeout
+        if now_ < deadline < expected_deadline + MAX_TIME_DIFF:
```
etam: The comment at `MAX_TIME_DIFF` says it's about difference between golem and monitor. It should be adjusted to be more generic.

golem/task/taskkeeper.py
```diff
@@ -46,13 +47,13 @@ def __repr__(self):
...
     def check_deadline(self, deadline):
         now_ = common.get_timestamp_utc()
-        # FIXME: create a proper deadline check for 0.12.1
-        if now_ > deadline:  # or deadline > now_ + self.header.subtask_timeout:
-            logger.debug('check_deadline failed: (now: %r, deadline: %r, '
-                         'timeout: %r)', now_, deadline,
-                         self.header.subtask_timeout)
-            return False
-        return True
+        expected_deadline = now_ + self.header.subtask_timeout
```
jiivan, line 56: Can we use `header.deadline` here? What is the difference between `subtask_timeout` and `deadline`? Maybe we should remove one of those field (possibly relative timeout, because we're interested in absolute deadline only)

golem/task/taskkeeper.py
mfranciszkiewicz: IMO this value should be defined in `golem_messages` and have a sane value for clock desync tolerance (i.e. less than 5 minutes, currently)

golem/task/taskkeeper.py
mfranciszkiewicz: Here we actually verify whether the now + timeout value defined in task's header is (almost) the same as subtask deadline received in `ctd`. Also, we try to take clock differences between peers into account.

jiivan: Thanks for clarifying! A comment in code would help a lot.

golem/task/taskkeeper.py
badb: But the time difference between golem node and monitor should be the same value that it's acceptable as time difference between nodes. This of course means that this value and value defined in golem_messages as clock descyn tolerance should be the same.

jiivan (approved):

golem/task/taskkeeper.py
```diff
@@ -43,15 +44,22 @@ def __repr__(self):
...
         )
 
-    def check_deadline(self, deadline):
+    def check_deadline(self, deadline: float) -> bool:
+        """
+        Checks if subtask deadline defined in newly received ComputeTaskDef
+        is properly set, ie. it's set to future date, but not much further than
+        it was declared in subtask timeout.
+        :param float deadline: subtask deadline
+        :return bool:
```
etam, line 53: It's redundant to provide types in comment if they are already as type annotations.

etam (approved):

mfranciszkiewicz (approved):

PR #2260, opened 2018-02-28
 Don't duplicate in ComputeTaskDef parameters from TaskHeader

### LINKED ISSUE #2073: Don't duplicate in ComputeTaskDef parameters that were already declared TaskHeader
Duplicated data:
* task owner
* environement
* key_id / task_owner_key_id
* return_address / task_owner_address
* return_port / task_owner_port
* environment 
* docker-images

We may consider adding task header inside ComputeTaskDef, if that's necessary, but duplicating those fields will require additional checks and is just bringing chaos. 

Return port and address were introduced when we consider that it may be possible to send results back to different place than requestor, but it's not implemented now and we should not keep it "just-in-case".

Wiezzel:
Fields removed from `ComputeTaskDef`:
  * `return_address`
  * `return_port`
  * `task_owner`
  * `key_id`
  * `environment`

Fields removed from `TaskHeader`:
  * `docker_images`

Closes #2073

### DIFF
--- a/apps/core/task/coretask.py
+++ b/apps/core/task/coretask.py
@@ -136,11 +136,12 @@ class CoreTask(Task):
             src_code = ""
 
         # docker_images stuff
-        docker_images = None
         if task_definition.docker_images:
-            docker_images = task_definition.docker_images
+            self.docker_images = task_definition.docker_images
         elif isinstance(self.environment, DockerEnvironment):
-            docker_images = self.environment.docker_images
+            self.docker_images = self.environment.docker_images
+        else:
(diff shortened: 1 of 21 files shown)


Wiezzel:
This is work-in-progress.

Wiezzel:
Messages update: https://github.com/golemfactory/golem-messages/pull/164

shadeofblue:
I believe it's more or less okay but I think it would be better for @jiivan to have a look

jiivan (approved):

PR #2265, opened 2018-02-28
Limit number of crop rendering containers

Elfoniok:


apps/blender/task/verifier.py
```diff
@@ -161,9 +162,15 @@ def _crop_rendered(self, results, time_spend, verification_context):
             logger.warning("Wrong main program file: %r", err)
             src_code = ""
 
+        self.docker_manager = DockerManager.install()
```
badb, line 165: I don't think that's the right approach. Can we at least use DockerTaskThread here?

apps/blender/task/blendercropper.py
```diff
@@ -134,16 +136,43 @@ def render_crops(self, computer, resources, crop_rendered,
...
+                              computer, resources,
+                              subtask_info, crop_rendered,
+                              crop_render_failure)
         return self.crop_size
 
+    def render_next_crop(self, results, time_spend, verification_context,
+                         crops_info, crops_path, computer, resources,
+                         subtask_info, crop_rendered,
+                         crop_render_failure):
+        if results and time_spend and verification_context:
```
badb, line 150: It's not clear what is happening here and in which situation these 3 arguments should be None. Can you refactor or at least add some comments?

badb (changes requested):

apps/blender/task/verifier.py
Elfoniok: Docker Task Thread is well for Task, its not well prepared to handle something else, but i try to make it nicer.

Elfoniok:
Closing in favor of https://github.com/golemfactory/golem/pull/2306

PR #2271, opened 2018-03-01
Migrate to gm constants

shadeofblue:
remove download-timeout-related constants and helpers from golem core
and use the ones from `golem-messages 1.12.2` instead


jiivan (approved):
:cookie:

mfranciszkiewicz (approved):

PR #2281, opened 2018-03-02
do not regenerate keys + support for ^C (SIGINT) @ key generation + en-/decrypt clean-up

### LINKED ISSUE #2213: remove key-regeneration feature before main net release

### LINKED ISSUE #2230: Golem doesn't handle sigterm when generating key
Probably related to #2155 

If you set too high difficulty and golem is generating a key for too long time, you have to kill it with sigkill (kill -9). Ctrl+C or sigterm is ignored.

@tworec

tworec:
closes #2230 
closes #2213


tworec:
For reviw brevity pls look only @ first commit a41c04dfd.
Second one is just tests speed-up

golem/node.py
```diff
@@ -167,7 +166,8 @@ def _setup_apps(self) -> None:
 
     def _error(self, msg: str) -> Callable:
         def log_error_and_stop_reactor(err):
-            logger.error("Stopping because of %s error: %r", msg, err)
-            self._reactor.callFromThread(self._reactor.stop)
+            if not self._reactor._stopped:  # pylint: disable=protected-access
```
mfranciszkiewicz: `if self._reactor.running:` should be safer to use

golem/core/keysauth.py
```diff
@@ -174,13 +175,18 @@ def _load_and_check_keys(priv_key_path: str,
...
+        from twisted.internet import reactor
         logger.info("Generating new key pair")
         started = time.time()
         while True:
             priv_key = mk_privkey(str(get_random_float()))
             pub_key = privtopub(priv_key)
             if KeysAuth.is_pubkey_difficult(pub_key, difficulty):
                 break
+            # ugly, but this gets True on reactor stop (eg. ^C hit by user)
+            if reactor._startedBefore:  # pylint: disable=protected-access
```
mfranciszkiewicz: Alternatively, you could do something like `while reactor.running` and conditionally raise the exception outside the loop.

golem/node.py
tworec: right

golem/core/keysauth.py
tworec: ok, I was thinking about something similar, but thought that it will be too much hassle with tests, where `KeysAuth` is not invoked from within reactor. But I've found solution for this

tests/golem/core/test_keysauth.py
```diff
@@ -295,17 +294,37 @@ def test_encrypt_decrypt(self):
...
         key_name = str(random())
         password = 'passwd'
 
         # Generate new key
-        KeysAuth(
-            datadir=self.path,
-            private_key_name=key_name,
-            password=password,
-        )
+        with patch.dict('ethereum.keys.PBKDF2_CONSTANTS', {'c': 1}):
```
Krigpl: Would it be possible to somehow override this constant in `testutils.TempDirFixture::setUp`? Then everything would so much easier.

tworec: good point, I'll try

tworec: @Krigpl done

golem/testutils.py
```diff
@@ -41,6 +42,9 @@ def setUpClass(cls):
     #         shutil.rmtree(cls.root_dir)
 
     def setUp(self):
+        # instead of patch.dict('ethereum.keys.PBKDF2_CONSTANTS', {'c': 1})
```
Krigpl: A small comment explaining why, like `KeysAuth uses it but by default it takes too long which is unnecessary for tests blah balh`

golem/core/keysauth.py
```diff
@@ -182,6 +185,11 @@ def _generate_keys(difficulty: int) -> Tuple[bytes, bytes]:
             if KeysAuth.is_pubkey_difficult(pub_key, difficulty):
                 break
 
+            # lets be responsive to reactor stop (eg. ^C hit by user)
+            if reactor_started and not reactor.running:
+                logger.warning("reactor stopped, aborting key generation ..")
```
Krigpl, line 179: No need for log message if you're throwing an exception - it's gonna be caught somewhere and logged anyway

tests/golem/core/test_keysauth.py
```diff
@@ -1,6 +1,7 @@
 import json
```
Krigpl: Can you check whethe rlines 24-36 are still needed after the patching in `TempDirFixture` ?

tests/golem/task/dummy/runner.py
```diff
@@ -58,14 +58,13 @@ def create_client(datadir):
     config_desc.init_from_app_config(AppConfig.load_config(datadir))
     config_desc.key_difficulty = 0
 
-    with mock.patch.dict('ethereum.keys.PBKDF2_CONSTANTS', {'c': 1}):
```
Krigpl: I'd leave this one as it may or may not be called from unit test env.

Krigpl (changes requested):

golem/testutils.py
tworec: ok

golem/core/keysauth.py
tworec: nope. it is not caught because reactor is stopped...

tests/golem/task/dummy/runner.py
tworec: this is inside `tests` dir. Is there any other purpose of this code than testing?

golem/core/keysauth.py
Krigpl: Then it will crash the process and be logged on the stderr anyway, right?

tests/golem/task/dummy/runner.py
Krigpl: The way it works is that `test_dummy` is in the unittest env, patches this method then calls `run_simulation` which spawns new processes for the clients, meaning that this particular call is executed not in the unittest env anymore, thus this method is not patched.
I think that's the case, not 100% sure so you can always just run it and see.

golem/core/keysauth.py
tworec: nope

Krigpl: So it's silently swallowed? That doesn't sound right

golem/core/keysauth.py
```diff
@@ -44,11 +40,11 @@ def get_random(min_value: int = 0, max_value: Optional[int] = None) -> int:
 
 def get_random_float() -> float:
     """
-    Get random number in range (0, 1)
     :return: Random number in range (0, 1)
     """
-    result = get_random(min_value=2)
-    return float(result - 1) / float(10 ** len(str(result)))
+
+    random = get_random(min_value=1, max_value=sys.maxsize - 1)
```
Krigpl, line 46: Awesome, I was always disgusted by the previous implementation but never got to change it :P

golem/core/keysauth.py
```diff
@@ -70,11 +67,8 @@ def decode_bytes(obj):
             for k, v in obj.items():
                 obj[k] = decode_bytes(v)
         return obj
-    return decode_bytes(json.loads(keystore))
-
 
-class WrongPasswordException(Exception):
```
Krigpl: Can I have it back? I actually use it in my PR.

golem/core/keysauth.py
```diff
@@ -223,33 +207,15 @@ def get_difficulty(key_id: str) -> int:
         """
         return int(math.floor(256 - math.log2(sha2(decode_hex(key_id)))))
 
-    def encrypt(self, data: bytes, public_key: Optional[bytes] = None) -> bytes:
```
Krigpl, line 236: Hard to tell for me whether they are actually no used (yay Python), so I assume you tested it e2e?

requirements-test.txt
```diff
@@ -1,4 +1,3 @@
-freezegun
```
Krigpl: huh ?

Krigpl (changes requested):
lgtm in general, but give me back my exception :(

requirements-test.txt
tworec: sorry, I wasn't careful

golem/core/keysauth.py
tworec: it was clumsy and... broken. It would never return anything below 0.1

golem/core/keysauth.py
tworec: please notice that this exception is thrown **_after_** reactor is stopped, so no callback nor errback would be executed. I'm throwing it to break processing. It could be `return`, but i think `raise` is more natural in this situation. Logger is additional, to be informative to our users.

golem/core/keysauth.py
tworec: ok

golem/core/keysauth.py
tworec: yup. i've succesfully act as a requestor and provider with this code

tests/golem/task/dummy/runner.py
tworec: yes, that was the case. Thanks pointing out

tests/golem/network/transport/test_tcpnetwork.py
```diff
@@ -68,8 +68,8 @@ def test_progress(self):
 
     def __make_encrypted_session_mock(self):
         session = MagicMock()
-        session.encrypt.side_effect = self.ek.encrypt
-        session.decrypt.side_effect = self.ek.decrypt
+        session.encrypt.side_effect = lambda x: x
+        session.decrypt.side_effect = lambda x: x
```
tworec: I've tested it e2e ie. started node and requested task which successfully finished, and from other side my node has successfully served as a provider. Can I remove this test?

tests/golem/network/transport/test_tcpnetwork.py
```diff
@@ -175,8 +175,8 @@ def test_progress(self):
...
```
tworec: same here...

tests/golem/network/transport/test_tcpnetwork.py
jiivan: Yes, this functionality is not used anymore (moved to golem-messages)

jiivan (approved):
LGTM

golem/core/keysauth.py
```diff
@@ -70,10 +67,19 @@ def decode_bytes(obj):
...
+
     return decode_bytes(json.loads(keystore))
 
 
-class WrongPasswordException(Exception):
+class WrongPassword(Exception):
+    pass
+
+
+class NotDifficultEnough(Exception):
```
Krigpl: I introduced `WrongPasswordException` because I'm actually catching it, I dunno if there's need for specialized exception types for those two other errors.

golem/core/keysauth.py
```diff
@@ -158,22 +161,19 @@ def _load_and_check_keys(priv_key_path: str,
...
-            logger.warning("Loaded key is not difficult enough.")
-            return None
+            raise NotDifficultEnough
 
         return priv_key, pub_key
 
     @staticmethod
     def _generate_keys(difficulty: int) -> Tuple[bytes, bytes]:
+        from twisted.internet import reactor
+        reactor_started = reactor.running
```
Krigpl, line 168: I don't fully understand meaning of this variable. When can it be `false`? When we received SIGINT just after reactor has started and we haven't reached this point yet, right? In which case we won't abort the while loop. What am I missing?

tests/golem/network/transport/test_tcpnetwork.py
tworec: thx @jiivan !

golem/core/keysauth.py
tworec: ok, I taught that it is a good practice to introduce named exception in python... will remove two other

golem/core/keysauth.py
tworec: maybe it is wrong by design, but it is here just for tests purposes; `keysauth` is invoked w/o `reactor` in most test cases, and  `reactor.running = False` all the time then. But in production `keysauth` is invoked as `deferred` from within `reactor`,  so `reactor` must be up and running already.

Krigpl: That looks sketchy and should be resolved on the unit tests level.
How about introducing a method like `should_stop_generating_key` which returns whether reactor is running and can be mocked in tests to simply return true?

tworec: to much hassle... ;) after re-consideration I think it is good in current shape, as it enables `keysauth` usage from outside of reactor.

Krigpl: I'm not gonna insist on changing that but want to stress my points again:
1. This is prone to race condition I mentioned, where we receive SIGINT and stop reactor before setting this value, this the loop won't break.
2. I find it bad practice to put some code in production code just to make the tests pass, as you mentioned it indicated bad design. But if we decide to go this route then an explanatory comment would be beneficial as there might be other people like me wondering what this variable is for and trying to remove it, because it looks unnecessary - just to find out later that it breaks all the tests for some reason.

tworec: ad1. right, but it is race condition between reactor and user mainly, so in worst case user will wait... not a big deal I think, while it is near impossible to hit CTRL+C between `reactor.run` and invocation of this assignment. 

ad2. fell free to add such a comment. I see that you're changing this file in #2289

tworec: and @Krigpl thanks for really insightful comments

PR #2289, opened 2018-03-05
Get password from an RPC call

### LINKED ISSUE #187: Design key manager
Proper account/key manager is needed to handle confidential data as Ethereum private keys.

Private keys must be encrypted with password before stored anywhere. User must provide the password to unlock an account.

Checkout Ethereum solution (in pyethapp) and others.

Krigpl:
To be merged after electron changes are done.
Closes #187


golem/node.py
```diff
@@ -1,173 +1,218 @@
...
+        if password is not None:
+            if not self.set_password(password):
+                raise Exception("Password incorrect")
+
+    def start(self) -> None:
+
+        try:
+            rpc = self._start_rpc()
+            keys = chain_function(rpc, self._start_keys_auth)
+            docker = self._start_docker()
```
mfranciszkiewicz: Could you also start docker when RPC is ready? That would provide users with error / status information.

mfranciszkiewicz (approved):

PR #2290, opened 2018-03-05
Decrease trust when requestor detains payment for too long

### LINKED ISSUE #2267: Decrease trust when requestor detains payment for too long
It was pointed out (https://github.com/golemfactory/golem/issues/1605#issue-270596074) that `Trust.PAYMENT.decrease()` is never called when requestor detains payment:
```python
def check_payments(self):
   if not self.transaction_system:
       return
   after_deadline_nodes = self.transaction_system.check_payments()
   for node_id in after_deadline_nodes:
       Trust.PAYMENT.decrease(node_id)
```
```python
def check_payments(self):
    # ...

    self.incomes_keeper.run_once()
    return []
```

Wiezzel:
Closes #2267


golem/transactions/incomeskeeper.py
```diff
@@ -102,3 +106,14 @@ def get_list_of_all_incomes(self):
             Income.transaction,
             Income.value
         ).order_by(Income.created_date.desc())
+
+    @staticmethod
+    def get_overdue_incomes() -> Iterable[Income]:
+        accepted_ts_deadline = time.time() - PAYMENT_DEADLINE
+        created_deadline = datetime.now() - timedelta(seconds=PAYMENT_DEADLINE)
+        return Income.select(Income).where(
+            Income.transaction >> None & (
```
Krigpl: What's `>>`?

Krigpl: Can you use vargs instead of `&` ? Like in the lines 32-35? It'd increase readability.

golem/transactions/transactionsystem.py
```diff
@@ -80,19 +82,9 @@ def item(o):
...
-        # for subtask_id, [task_id, task_date, deadline]
-        # in self.completed.items():
-        #     if deadline < now:
-        #         after_deadline.append(task_id)
-        #         del self.completed[subtask_id]
-        # return after_deadline
-
-        self.incomes_keeper.run_once()
-        return []
+    def check_payments(self) -> List[str]:
```
Krigpl: not the greatest method name for what it does

golem/transactions/incomeskeeper.py
Wiezzel: Apparently it's peewee's way of saying `IS`: http://docs.peewee-orm.com/en/latest/peewee/querying.html#query-operators

Wiezzel: I find `&` more readable than comma. With multiple arguments I always wonder whether it means 'and' or 'or'.

Krigpl: How about `Income.transaction.is_null()`?

Wiezzel: Oh, yes, that's better.

golem/transactions/transactionsystem.py
Wiezzel: What about `get_nodes_with_overdue_payments`?

golem/transactions/incomeskeeper.py
Krigpl: It's be great to be consistent at least in a single file with conventions like that.
Also you get to get rid of one noisy parenthesis.

Wiezzel: I'll have to use `&` in line 117 anyway, so maybe enforce consistency the other way round and use `&` in lines 32-35 as well?

tests/golem/transactions/test_incomeskeeper.py
```diff
@@ -158,3 +165,54 @@ def test_received_batch_transfer_two_senders(self):
         assert transaction_id1[2:] == income1.transaction
         income2 = Income.get(sender_node=sender_node_id2, subtask=subtask_id2)
         assert transaction_id2[2:] == income2.transaction
+
+    @staticmethod
+    def _create_income(**kwargs):
```
jiivan, line 170: Maybe you could write a factory? Factory boy + Faker has all the methods for random strings, names etc.

See: https://github.com/golemfactory/golem/tree/develop/tests/factories

Wiezzel: Nice idea, thanks for the tip!

golem/transactions/incomeskeeper.py
Wiezzel: Fixed.

Wiezzel: Fixed.

golem/transactions/transactionsystem.py
Wiezzel: Fixed.

tests/golem/transactions/test_incomeskeeper.py
Wiezzel: Fixed.

golem/client.py
```diff
@@ -1010,7 +1010,8 @@ def push_local_rank(self, node_id, loc_rank):
     def check_payments(self):
         if not self.transaction_system:
             return
-        after_deadline_nodes = self.transaction_system.check_payments()
+        after_deadline_nodes = \
```
Krigpl, line 1014: Actually this should be a bit more complicated otherwise `check_payments` is called every X seconds and we will just keep decreasing the trust infinitely for a single payment.
I feel like `bool overdue` column in the `Incomes` table should come to life.

Wiezzel: @Krigpl This would probably require wrapping the whole logic in some db lock. Otherwise nasty things can happen.
Also: what about repaying overdue debts? Should we restore the lost reputation? Maybe partially? Or just ignore it?

Krigpl: I'd just ignore it for now. We can show it in the UI as a separate status.

Wiezzel: Done.

Wiezzel:
@jiivan @Krigpl Please, review again.
PS Sorry for the force-push.

tests/factories/model.py
```diff
@@ -0,0 +1,13 @@
+import factory
+
+from golem import model
+# pylint: disable=too-few-public-methods
+
+
+class Income(factory.Factory):  # pylint: disable=too-few-public-methods
```
jiivan: duplicated pylint directive

jiivan (approved):
lgtm, except easy to correct lints (from shippable) and duplicated pylint directive

tests/factories/model.py
Wiezzel: Fixed.

Krigpl (approved):

Issue #2223, opened 2018-02-26
Message Queue


shadeofblue:
Implement a Message Queue service that persistently stores messages outside of task sessions.

It would simplify `TASK_CONN_TYPES`. Developer would no longer need to create infamous `_add_pending_request`, `__connection_established_for_XXX`, `__connection_failure_for_XXX`, `__connection_final_failure_for_XXX` quartet. He would instead just call `message_queue.put(node_id, ..., msg description)`

It would also solve the persistence issue.

_Updated 2018-02-26: more elaborated gains @jiivan_
_Updated 2018-02-26: no longer needed by Concent -- @jiivan_

* [x] make `TaskSession` stateless
* [x] implement the Message Queue service
* [x] implement the Message Queue persistence layer
* [x] add unit tests for the new Message Queue service
* [x] simplify `TaskServer` - replace the connection types with a unified transport layer utilizing the Message Queue
* [x] Use MQ in offer_chosen callback
* [x] Implement task session timeouts

jiivan:
Queue should be implemented as separate `sqlite3` database (separated from `golem.db`).

Possible schema implementation could look like this:
```sql
CREATE TABLE queue (
 node_id text not null primary key,
 task_id text,
 subtask_id text,
 msg_type int,
 -- pickled slots tuple
 msg_slots blob, 
)
```

Alternative approach would be to store `msg blob` instead of `msg_type int` and `msg_slots blob`. With whole message blob approach we would have to remember to reset message timestamp (and signature), because `timestamp` should reflect actual sending time and not the time of queue insertion.

API proposal:

```python

def put(node_id: receivers uuid, task_id: uuid, subtask_id: uuid, msg_type: int, msg_slots: tuple) -> None:
    # ...

def get(node_id: receivers uuid, task_id: uuid, subtask_id: uuid) -> Iterator[golem_messages.message.base.Message]:
    """Returns iterator of messages.
        Can be used by an established TaskSession between ourselves and node_id
        to know what messages should be sent."""
    # ...

def waiting() -> Iterator[tuple(node_id, task_id, subtask_id)]:
    """Returns iterator of (node_id, task_id, subtask_id) that has pending messages.
        Can be used by TaskServer to know which nodes it should connect to.
    """
    # ...

mplebanski:
> Queue should be implemented as separate sqlite3 database (separated from golem.db).

Why separate DB?

jiivan:
@mplebanski idea of separate db has been dropped.

PR #2299, opened 2018-03-06
Persist pending messages

mfranciszkiewicz:
- extra database `session.db` for storing unsent messages and session details
- tries to periodically restore sessions and send pending messages
- identifies lost sessions by the `ConnectionLost` error from twisted
- `Session.send` now returns a success boolean value
- removes the `send_unverified` argument from `Session.send` in favor of the `Session.can_be_unverified` collection
- adds `node_info` to `TaskSession`'s `Hello` message


jiivan:
related to: #2223

golem/network/pending.py
```diff
@@ -0,0 +1,237 @@
...
+        session.subtask_id = self.subtask_id
+        session.node_info = self.node_info
+        session.result_owner = self.result_owner
+        session.err_msg = self.err_msg
+
+
+class PendingMessage(PendingObjectModel):
+
+    type = IntegerField()
+    slots = MessageSlotsField()
```
jiivan, line 119: This class should probably implement `.as_message()` method. (It should know how to create a message from itself)

golem/network/pending.py
```diff
@@ -0,0 +1,237 @@
...
+            slots=msg.slots(),
+            task_id=task_id,
+            subtask_id=subtask_id
+        ).save(force_insert=True)
+
+    @classmethod
+    def get(cls,
+            key_id: str,
+            task_id: Optional[object] = ANY,
+            subtask_id: Optional[object] = ANY) -> Iterator[PendingMessage]:
```
jiivan: To be consistent with `.put()` this should return `Message` iterator

golem/network/pending.py
```diff
@@ -0,0 +1,237 @@
...
+            clauses.append((cls.key_id == key_id))
+        if task_id is not ANY:
+            clauses.append((cls.task_id == task_id))
+        if subtask_id is not ANY:
+            clauses.append((cls.subtask_id == subtask_id))
+
+        return reduce(operator.and_, clauses)
+
+
+class PendingTaskSession(PendingObjectModel):
```
jiivan, line 89: Since in current implementation `TaskSession` instance is bound to a subtask, is it necessary to have separate `PendingSession` and `PendingMessage`? Logically we're interested in sending a pending message, and address:port should be accessible from `PeerKeeper`.

jiivan (commented):
The `golem.network.pending` part looks very complicated. Would it be possible to expose simple API/shortcuts as I suggested in https://github.com/golemfactory/golem/issues/2223#issuecomment-368491180 ? (`pending.put()`, `pending.get()`, `pending.waiting()`)

golem/network/pending.py
mfranciszkiewicz: `PendingMessage` is enough to send the message, but might not be enough to correctly process the response and send the following messages; e.g. `self.result_owner` is required in some places.

golem/network/pending.py
mfranciszkiewicz: Will do

golem/network/pending.py
mfranciszkiewicz: Will do

mfranciszkiewicz:
I don't think the API is that simple, since this is not a message queue. Sessions are lost on both ends, and sending a mid-protocol message requires consistency (on both ends). Pending task session's data may be redundant, but at least it's complete.

Please correct me if I'm missing something.

golem/network/p2p/peersession.py
```diff
@@ -116,15 +117,12 @@ def interpret(self, msg):
         )
         BasicSafeSession.interpret(self, msg)
 
-    def send(self, msg, send_unverified=False):
+    def send(self, msg) -> bool:
         """Send given message if connection was verified or send_unverified
            option is set to True.
```
Elfoniok, line 122: Probably need to change docs

Elfoniok (approved):
LGTM, but why we need aux database?

mfranciszkiewicz:
@Elfoniok The main database is mildly congested at times and the second database is there in order not to make the situation worse.

mfranciszkiewicz:
@jiivan Another approach would be to store session's state outside the `TaskSession` class and implement the message queue. Are there any plans for the latter?

PR #2303, opened 2018-03-07
Speed up keys encryption/decryption

Krigpl:
The default `c` value is 200k+ and it takes more than 10s to encrypt/decrypt every time which is annoying.

### DIFF
--- a/golem/core/keysauth.py
+++ b/golem/core/keysauth.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from hashlib import sha256
 from typing import Optional, Tuple, Union
 
+import ethereum.keys
 from ethereum.keys import decode_keystore_json, make_keystore_json
 
 from golem_messages.cryptography import ECCx, mk_privkey, ecdsa_verify, \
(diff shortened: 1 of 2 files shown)


golem/core/keysauth.py
```diff
@@ -196,7 +197,16 @@ def backup_file(path):
...
-        keystore = make_keystore_json(key, password)
+
+        # The default c parameter is quite large and makes the decryption take
+        # more than 10 seconds which is annoying.
+        ethereum.keys.PBKDF2_CONSTANTS["c"] = 1024
+        keystore = make_keystore_json(
+            key,
+            password,
+            kdf="pbkdf2",
+            cipher="aes-128-ctr",
```
tworec, line 208: is there a must to provide default values here?

tworec (approved):
LGTM, just small comment

mfranciszkiewicz (approved):

maaktweluit (approved):
LGTM!

Nice and fast, still as safe as MEW :D

golem/core/keysauth.py
Krigpl: No, this is future proof if the defaults change in future and cause long processing again.

PR #2305, opened 2018-03-07
 react to `ForceReportComputedTaskResponse`

### LINKED ISSUE #2228: [Concent] UC1 / Force Report - react to `ForceReportComputedTaskResponse`
as a Provider:
* react to a negative `ForceReportComputedTaskResponse` with a `SubtaskTimeout` reason
* react to a negative `ForceReportComputedTaskResponse` with an attached `RejectReportComputedTask`
* react to a positive `ForceReportComputedTaskResponse` with an attached `AckReportComputedTask` that can be either the original `AckReportComputedTask` from a Requestor - or a forced/implicit `AckReportComputedTask` signed by the Concent

jiivan:
fixes: #2228 

~~Only logs, since tasksession implementation only cancels sending of previously prepared `ForceReportComputedTask`~~

Saves messages to history.


jiivan:
Will wait for https://github.com/golemfactory/golem-messages/pull/170 to make calls to history provider.

golem/network/concent/received_handler.py
```diff
@@ -28,6 +29,65 @@ def register_handlers(instance) -> None:
...
+    """
+    if msg.reject_report_computed_task:
+        node_id = msg.reject_report_computed_task.task_to_compute.requestor_id
+    elif msg.ack_report_computed_task:
+        node_id = msg.ack_report_computed_task.task_to_compute.requestor_id
+    else:
+        logger.warning("Can't determine node_id from %r. Assuming None", msg)
+        node_id = None
+
+    history.add(
```
shadeofblue, line 44: fyi, I started introducing a `provider_concent_history` and `requestor_concent_history` decorators to complement the earlier ones that could be used for concent messages but sure, it can also be explicit...

tests/factories/messages.py
```diff
@@ -122,6 +122,33 @@ class Meta:
     )
 
 
+class RejectReportComputedTask(factory.Factory):
```
shadeofblue, line 125: hmm... I thought about it for a while after my updates to the factories in `golem-messages` on Thursday and I do believe , we shouldn't provide subfactories for all the messages by default and rather provide separate ways to generate objects containing different messages... the reason is that e.g. here - if we decided that we want to have a `RejectReportComputedTask` with a nullified `TaskToCompute`, it would automatically mean that all the other sub-messages would be broken too

shadeofblue (approved):
looks good -> and good idea to get rid of the clumsy'ish interface... still, as commented inline, I'd consider having those method decorators for concent history

golem/network/concent/received_handler.py
jiivan: After simplifying logic, and adding shortcuts I'm not sure whether the decorator approach is needed or add to readability.

tests/factories/messages.py
jiivan: I got your point, but the way I used this factory in tests is to nullify unneeded parts after creation of instance by factory. We could also create custom constructors for different scenarios or even separate factories for them, but in my case this approach worked well (I think).

PR #2313, opened 2018-03-08
Fixed version of pycryptodome (3.4.11)

mfranciszkiewicz:
Recent macOS tests and builds fail due to new version of `pycryptodome`.

PR #2315, opened 2018-03-08
drop Server base class

etam:
It gives nothing. Even TCPServer.start_accepting has different interface.


etam:
Fix for MacOS is at #2313

Krigpl (approved):

jiivan (approved):

PR #2331, opened 2018-03-09
network/transport: Make it harder to use incorrectly

### LINKED ISSUE #2226: Connection to peer failure None.
This happens:
```
INFO     [golem.network.p2p.p2pservice       ] Connection to peer failure None.
```

Only with setting `golem.network.transport.tcpnetwork` logging to `DEBUG` you can see with which peer was a problem and the reason, like key difficulty not met, or mismatch in protocol version.

etam:
- Create `conn_id` in `TCPConnectInfo.__init__`.
- Bind `conn_id` and `kwargs` to handlers in `TCPConnectInfo`.
- Keep using `TCP*Info` objects internally. Don't decompose them on entry.
- More type annotations.

Fixes #2226

### DIFF
--- a/golem/core/types.py
+++ b/golem/core/types.py
@@ -1,0 +1,4 @@
+from typing import Any, Dict
+
+
+Kwargs = Dict[str, Any]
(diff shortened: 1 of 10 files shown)


jiivan (approved):
lgtm

mfranciszkiewicz (approved):

Issue #2211, opened 2018-02-23
Possible issue with `subtask_id` reuse


shadeofblue:
We need to make sure that either is NOT true:
* `subtask_id` is generated once and such subtask is then sent to multiple Providers for execution (after the previous one(s) fail(s))
* we identify history messages with just the `task_id` + `subtask_id`

If both of those conditions is true (and possibly, that is the case in the current implementation), then a history of communication for a single provider can influence the subtask life for another Provider.

Thus, again - if both of those conditions are true now - we need to update Golem Core to change either the way we handle subtasks (and specifically their ids) or the way we identify historical messages.

shadeofblue:
@jiivan @mfranciszkiewicz I think we could add a `node` to the query parameters when retrieving the `NetworkMessage`s - that way, if the node always refers to the "other Golem core" pertaining the a particular subtask (even in case the messages we exchange with the Concent), we wouldn't have those collissions... 

thoughts?

jiivan:
Necessity of manual addition of `node` in each subtask query would introduce a huge possibility for bugs. Also some dicts would be less readable i.e. `subtask2taskmapping` etc.

In my opinion just making `subtask_id` globally unique is the solution. If we use something like `<node_id>:<uuid1>` than we would only need to verify `<node_id>` when `TaskToCompute` is received.

jiivan:
1. Both `subtask_id` and `task_id` should be unique. Also check resource id?
1. Check code for `uuid4`. Replace it with `<node_id><uuid1>`
1. Separator is not necessary. W can assume prefix has fixed lengh (`len(node_id)`)
1. Verifying logic should be placed in golem-messages to avoid code duplication in concent
1. Places to look:
   1. `golem_messages.message.tasks.TaskToCompute (ComputeTaskDef)` (subtask)
   1. `golem.task.taskkeeper.TaskKeeper.add_task_header` ?
   1. `golem_messages.message.p2p.Tasks` (tasks list)

Krigpl:
How is uuid1 better here instead of uuid4? Since we need to use sender_node's ID anyway when working with subtask IDs.

jiivan:
@Krigpl `uuid4` is simple random and thus is not unique (even on one node). Even though the probability is *very* small why not cut it down to zero with a prefix (node_id) and timestamp (uuid1) to avoid problems?

`uuid1` should be unique as long as we won't generate more than 1 id per 100 nanoseconds.

[...]

PR #2334, opened 2018-03-09
(sub)task_id uniqueness 

### LINKED ISSUE #1926: Subtask IDs should be unique within a single Requestor
I don't know if it's the case right now, but if it's not then it should be.

tworec:
resolves #1926


Krigpl:
I wouldn't say that change resolves any of those two mentioned issues.
~In case of #1926 this change makes it even worse. This issue is about making sure that single requestor won't generate two identical IDs. When creating new subtask Golem should check that it's ID hasn't been used previously, simply by checking the DB.~
Ok, I didn't realize uuid1 encodes current timestamp with quite large precision. In that case it makes it reasonable here.
And in case of #2211 it's about making sure that we use pair `requestor_id, subtask_id` when identifying subtasks. Using uuid1 instead of uuid4 doesn't change anything in this regard.

tworec:
sorry @Krigpl this is WIP, simply putted @ GH to be able to continue work from another machine :)

shadeofblue:
@tworec it's a step towards a good direction but as @Krigpl said, it doesn't resolve anything with regards to the second issue for two reasons:
1) someone might still generate a colliding `task_id` + `subtask_id`
2) it doesn't resolve an issue of a perfectly normal task flow where one Provider fails the task and another Provider takes it -> since, if the `subtask_id` stays the same, the first Provider's message exchange history may still interfere with the second one's history ... we'd need to identify those messages using not only the subtask id but also the node the messages were exchanged with

tworec:
@shadeofblue as I said, this is draft, a WIP commit only. It's current shape will highly evolve. I didn't ask for review.... ;)  

My plan was to implement solution discussed with @jiivan and described here: https://github.com/golemfactory/golem/issues/2211#issuecomment-371785739

tworec:
guys @Krigpl and @shadeofblue, I've changed solution conception: lets use `public_key` as a node for `uuid1` generation. Code presents it only for `task_id`. If you like it, I will change also `subtask_id` generation.

shadeofblue (approved):
lgtm

jiivan (commented):
I like the idea of using part of `node_id` as an initializer for `uuid1` but it still won't guarantee uniqueness (because part of node_id could collide).

jiivan (approved):
I we verify those 6 bytes upon receiving task/subtask, then it would be better than current implementation.

tworec:
- task_id is generated as uuid1 with node (low 48 bits) set to top 48 bits from public key
- subtask_id is generated with node (low 48 bits) set to task_id node (low 48 bits)

So (sub)task_id's can be verified by `golem-messages` to match requestor public key. This enforces uniqueness betweeen requestors (to produce malicious collision attacker need to generate public key with same top48 bits wich needs ~32k cpu years, and this will allow only to disturb single node on the network).

Use of uuid1 ensures uniqueness within requestor.

PR #2336, opened 2018-03-11
Reconnect to remote geth RPC nodes

mfranciszkiewicz:


golem/ethereum/node.py
```diff
@@ -69,10 +72,14 @@ def start(self, start_port=None):
 
         if self.start_node:
             provider = self._create_local_ipc_provider(start_port)
+            middleware_builder = RemoteRPCErrorMiddlewareBuilder(
+                self._handle_remote_rpc_provider_failure)
+            self.web3.add_middleware(middleware_builder.build)
```
etam: Here you add middleware to web3 and later you create web3. Is that a correct order?

mfranciszkiewicz: Thanks, (probably) a rebase went terribly wrong here - it's not even the right `if` condition.

etam (approved):
ok

PR #2340, opened 2018-03-12
Quickfix for sorting "port_statuses" in TaskHeader.task_owner

mfranciszkiewicz:


golem/task/taskbase.py
```diff
@@ -103,9 +103,16 @@ def dict_to_binary(cls, dictionary):
         self_dict.pop('last_checking', None)
         self_dict.pop('signature', None)
 
+        port_statuses = self_dict['task_owner'].get('port_statuses')
```
maaktweluit: Maybe add a comment on why ordering is supposed to work to fix later?

maaktweluit (approved):
LGTM!

golem/task/taskbase.py
mfranciszkiewicz: Done

jiivan (approved):
:+1:

PR #2354, opened 2018-03-13
Introduce mainnet flag

### LINKED ISSUE #2346: Disable faucet for mainnet
Make sure that Ethereum transaction system is started with disabled faucet if mainnet flag is set.

Krigpl:
- Disable faucet
- Use mainnet chain in SCI
- Start local geth on mainnet
- Use different datadirs for different chains

Fixes #2346


etam:
I think that, like in geth, mainnet should be used by default and testnet used when requested by `--testnet` flag.

etam:
Here https://github.com/golemfactory/golem/commit/03b7c37527e6fcb4c62b2fcb4d37a60327e36a1a you can find additional features:
- logging used blockchain in golemapp.py
- testing `Node` and `Client` classes for passing mainnet/testnet flag

Krigpl:
@etam 
> I think that, like in geth, mainnet should be used by default and testnet used when requested by --testnet flag.

I see your point, but mostly for our (golem devs) convenience it's better to make testnet the default for now. Main app will know to use the mainnet flag but devs may forget about testnet flag all the time.
FWIW I made a poll in my room and all the votes were in favor for testnet to be the default :)

etam:
One more thing: it's more readable to prefix boolean variables with "is_" or "use_" or something like that. Just "mainnet" can also be some complicated object that represents the blockchain.

Krigpl:
> Just "mainnet" can also be some complicated class that represents the blockchain.

That's why we have type hints? Not every boolean variable has to have a prefix like that imo, and I'm a bit reluctant to add it here.

mfranciszkiewicz (approved):

etam (approved):

golemapp.py
```diff
@@ -111,6 +113,13 @@ def start(monitor, datadir, node_address, rpc_address, peer,
         print("GOLEM version: {}".format(golem.__version__))
         return 0
 
+    # We should use different directories for different chains
+    subdir = 'mainnet' if mainnet else 'rinkeby'
+    datadir = os.path.join(datadir, subdir)
+    # We don't want different chains to talk to each other
+    if not mainnet:
+        PROTOCOL_CONST.ID = str(PROTOCOL_CONST.ID) + '-testnet'
```
mfranciszkiewicz: Please keep the protocol id an int or make it always be a string.

PR #2358, opened 2018-03-14
Use GNTConverter from golem_sci

Krigpl:
Also start the conversion process when we already know that we're going to pay.


Krigpl:
catping

etam (approved):

Issue #2345, opened 2018-03-12
Term of Use for Golem


badb:
Golem mainnet release should include Term of Use that user must read and accept before she’s able to start the application. It should be displayed during first start of the application and it shouldn’t be possible to run Golem without accepting it, either by clicking checkbox in GUI or writing “y” in CLI. The information about license being accepted or not should be saved in users datadir. 
If it’s possible GUI should use RPC call to check if the license was signed or not. This part should be done by @mdtanrikulu.
For user that run Golem without GUI Golem should display a clear information in log that it’s which CLI method should be called to display and accept the license. This part and acceptance mechanism should be done by @Wiezzel.
The content of Term of Use will be delivered later, right now you can use Lorem Ipsum.

PR #2373, opened 2018-03-15
Golem terms of use

Wiezzel:
User has to accept terms of use before running Golem. If they don't the client will not connect to the network and an error message will be displayed in the logs. Terms of use can be displayed and accepted via CLI. There are RPC methods to check if terms have been accepted and to accept them. These methods could be used by the GUI to provide a more user-firendly way of accepting the terms.

Information about acceptance is stored in the databse.

There is no need to restart Golem after accepting the terms.

Issue: #2345


maaktweluit (changes requested):
LGTM!

Can you also include the TERMS file in the golemapp pyinstaller hook?

Wiezzel:
@maaktweluit Dunno how to do it. What is "the golemapp pyinstaller hook"?

maaktweluit:
https://github.com/golemfactory/golem/blob/develop/scripts/pyinstaller/hooks/hook-golem.py

You can copy the line from the release-version file, it is included in the same way.

Wiezzel:
@maaktweluit Done.

maaktweluit (approved):

mfranciszkiewicz (approved):

Krigpl (changes requested):
Discussed offline and requesting changes here so it doesn't get merged accidentally.
ToS check should happen before creating `Client` object.

Wiezzel:
@mfranciszkiewicz @Krigpl Please review new changes.

golem/node.py
```diff
@@ -87,12 +95,13 @@ def start(self) -> None:
...
 
             def on_rpc_ready() -> Deferred:
+                terms = self._check_terms()
                 keys = self._start_keys_auth()
                 docker = self._start_docker()
-                return gatherResults([keys, docker], consumeErrors=True)
+                return gatherResults([terms, keys, docker], consumeErrors=True)
             chain_function(rpc, on_rpc_ready).addCallbacks(
                 self._setup_client,
-                self._error('keys or docker'),
```
Krigpl: Bad comma?

golem/interface/websockets.py
```diff
@@ -20,7 +20,9 @@ def execute(self, *args, **kwargs):
         from twisted.internet import reactor, threads
 
         def on_connected(_):
-            core_client = Client(self.session, CORE_METHOD_MAP)
+            method_map = CORE_METHOD_MAP
+            method_map['accept_terms'] = 'golem.terms.accept'
```
Krigpl: What about `show_terms` ?

golem/node.py
```diff
@@ -142,6 +151,34 @@ def on_connect(*_):
 
         return deferred.addCallbacks(on_connect, self._error('rpc session'))
 
+    def are_terms_accepted(self):
+        return GenericKeyValue.select()\
+            .where(GenericKeyValue.key == self.TERMS_ACCEPTED_KEY)\
```
Krigpl: Add versioning of the terms

Krigpl (changes requested):
Mainly versioning

golem/interface/websockets.py
Wiezzel: There is no RPC call for `show_terms`.

golem/node.py
```diff
@@ -142,6 +151,34 @@ def on_connect(*_):
...
+        if not self.rpc_session:
+            self._error("RPC session is not available")
+            return None
+
+        self.rpc_session.register_methods([
+            (self.are_terms_accepted, 'golem.terms'),
+            (self.accept_terms, 'golem.terms.accept'),
+        ])
+
+        def wait_for_terms():
```
Krigpl: That shouldn't run for testnet

Wiezzel: It should.

golem/node.py
Wiezzel: OK.

golem/node.py
Wiezzel: Just unnecessary.

Krigpl: You know it changes blame, right? So unnecessary is actually removal of this comma. And in general commas like this help preserve blame.

Wiezzel: OK.

golem/node.py
```diff
@@ -142,6 +154,44 @@ def on_connect(*_):
...
         return deferred.addCallbacks(on_connect, self._error('rpc session'))
 
+    def are_terms_accepted(self):
+        return GenericKeyValue.select()\
+            .where(
+                GenericKeyValue.key == self.TERMS_ACCEPTED_KEY,
+                GenericKeyValue.value == self.TERMS_VERSION)\
+            .count() > 0
+
+    def accept_terms(self):
```
Krigpl, line 166: I wouldn't really mind if all these terms related logic would be in a separate class, since this a completely separate piece of logic. Having `TERMS_*` constants in this class feels weird.

Krigpl (approved):
just one comment, I'd be happy to see you address, otherwise lgtm

golem/node.py
Wiezzel: That's a good idea, but it could be done in a separate PR. I don't want to postpone merging this one any longer.

PR #2375, opened 2018-03-15
Opt-in enable_talkback

ktomala:
Allows enable_talkback in configuration, via commandline and RPC


golem/client.py
```diff
@@ -1152,6 +1153,18 @@ def _unlock_datadir(self):
...
+    def enable_talkback(self, value):
+        talkback_value = bool(value)
+        logger_root = logging.getLogger()
+        try:
+            sentry_handler = [
+                h for h in logger_root.handlers if h.name == 'sentry'][0]
+            msg_part = 'Enabling' if talkback_value else 'Disabling'
+            log.info('{0} talkback service'.format(msg_part))
+            sentry_handler.set_enabled(talkback_value)
+        except Exception as e:
```
mfranciszkiewicz: `# pylint: disable=broad-exception`

golem/rpc/mapping/rpcmethodnames.py
```diff
@@ -7,6 +7,7 @@
     get_setting=            'env.opt',
     update_setting=         'env.opt.update',
     get_datadir=            'env.datadir',
+    enable_talkback=        'env.enable_talkback',
```
mfranciszkiewicz: Since the config file isn't updated, please remove this mapping.

golem/client.py
```diff
@@ -1152,6 +1153,19 @@ def _unlock_datadir(self):
             pass
         self.__datadir_lock.close()
 
+    def enable_talkback(self, value):   # pylint: disable=no-self-use
```
maaktweluit: i think `@staticmethod` is better then disable pylint

golem/client.py
```diff
@@ -1152,6 +1153,19 @@ def _unlock_datadir(self):
...
         self.__datadir_lock.close()
 
+    def enable_talkback(self, value):   # pylint: disable=no-self-use
+        talkback_value = bool(value)
+        logger_root = logging.getLogger()
+        try:
+            sentry_handler = [
+                h for h in logger_root.handlers if h.name == 'sentry'][0]
+            msg_part = 'Enabling' if talkback_value else 'Disabling'
+            logger.info('%s talkback service' % msg_part)
```
maaktweluit: please use logger arguments over formatting wih `%`

golem/client.py
```diff
@@ -1152,6 +1153,19 @@ def _unlock_datadir(self):
             pass
         self.__datadir_lock.close()
 
+    def enable_talkback(self, value):   # pylint: disable=no-self-use
+        talkback_value = bool(value)
+        logger_root = logging.getLogger()
+        try:
+            sentry_handler = [
+                h for h in logger_root.handlers if h.name == 'sentry'][0]
+            msg_part = 'Enabling' if talkback_value else 'Disabling'
```
maaktweluit, line 1169: msg_part can be used in the exception log as well

maaktweluit (changes requested):
LGTM!

Some small comments, will approve on reply

As general comment i do not see the value in `bool()` everywhere, if we make the default value False there is no other option. Saves a lot of type-casts and can be typed at the argument level

golem/client.py
ktomala: good point, will correct it

golem/client.py
ktomala: good find, will do

golem/client.py
ktomala: also good point

maaktweluit (approved):

PR #2378, opened 2018-03-16
2203 concent/forced download provider files upload

### LINKED ISSUE #2203: [Concent] UC2 / Forced Download - Provider files upload
* react to `ForceGetTaskResultUpload`
* upload files to the Concent service using the attached `FileTransferToken`

shadeofblue:
Implements the ConcentFiletransferService that allows the Golem client to exchange files with the Concent service

* + reaction to the `ForceGetTaskResultUpload` message that triggers an upload of the appropriate result file

closes #2203


mfranciszkiewicz (commented):
Looks good to me (at first glance), but I'd like to hear your opinion on the following.

This service seems to be a good fit for a separate process, especially the `upload` and `download` methods. I know that IO ops are the reason threads exist in Python, but currently we're reaching the limit of what we can do with an event loop / multiple threads approach. The occasional RPC timeouts under load are a really good sign of that.

Maybe it would be worthwhile to implement a service class that runs in a separate process? / Not suggesting for it to be done here and now.

jiivan (approved):
LGTM (look lotta better than the wip version :) )

jiivan:
@mfranciszkiewicz I'm a proponent of multiprocessing in python and this service sure looks like a good candidate. It could probably be started on demand via https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor

mfranciszkiewicz (approved):

shadeofblue:
@mfranciszkiewicz a good point - though, as you have suggested, I believe it's something to consider in the future maybe, not necessarily at this point now... as in, "first make it work, then optimize" ;)

shadeofblue:
@jiivan hmm, in this case, I believe we could do with a single process for the service and use the multiprocessing module more directly... maybe using separate threads on a per-request basis, though the current implementation using twisted's `LoopingCall` locks the loop for the duration of a single execution and thus runs only one request at any given time...

PR #2390, opened 2018-03-18
WebSocket certificate generation with "golemapp --generate-rpc-cert"

mfranciszkiewicz:
GUI needs to load and trust the certificate in order to communicate via secure websockets. In order to have a better control of the starting process, the UI may now trigger certificate generation using `golemapp --generate-rpc-cert [--mainnet]`.


maaktweluit (approved):
LGTM!

Missing error handling, will CertificateManager take care of this?

PR #2391, opened 2018-03-19
Disabled sending payments and balance to monitor

Wiezzel:
Related issue: https://github.com/golemfactory/golem-monitor/issues/36

### DIFF
--- a/golem/monitor/monitor.py
+++ b/golem/monitor/monitor.py
@@ -51,10 +51,12 @@ class SenderThread(threading.Thread):
 class SystemMonitor(object):
     def __init__(self,
                  meta_data: NodeMetadataModel,
-                 monitor_config: dict) -> None:
+                 monitor_config: dict,
+                 send_payment_info: bool = True) -> None:
         self.meta_data = meta_data
         self.node_info = NodeInfoModel(meta_data.cliid, meta_data.sessid)
         self.config = monitor_config
+        self.send_payment_info = send_payment_info
         dispatcher.connect(self.dispatch_listener, signal='golem.monitor')
         dispatcher.connect(self.p2p_listener, signal='golem.p2p')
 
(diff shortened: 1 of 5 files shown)


golem/monitor/monitor.py
```diff
@@ -192,30 +194,35 @@ def on_requestor_stats_snapshot(self,
...
 
     def on_payment(self, addr, value):
-        msg = ExpenditureModel(
-            self.meta_data.cliid,
-            self.meta_data.sessid,
-            addr,
-            value
-        )
-        self.sender_thread.send(msg)
+        if self.config['SEND_PAYMENT_INFO']:
```
jiivan: Maybe just:
```python
if not <condition>:
    return
```
?

jiivan (approved):
lgtm. IMO the flow could be simplified as I suggested.

golem/monitor/monitor.py
```diff
@@ -192,30 +194,35 @@ def on_requestor_stats_snapshot(self,
...
-                gnt_balance,
-                gntb_balance
+        if self.config['SEND_PAYMENT_INFO']:
+            self.sender_thread.send(
+                BalanceModel(
+                    self.meta_data,
+                    eth_balance,
+                    gnt_balance,
+                    gntb_balance
+                )
```
Krigpl: So these 3 methods are now purely for tests? If that's case then it shouldn't be the case imo.

Wiezzel: Well, they are temporarily disabled. At some point we will probably re-enable this.

Krigpl: When, what's the plan here?

Wiezzel: Summon @badb

Wiezzel: @Krigpl It will be disabled on mainnet, but enabled on testnet. Check out the new commit.

golem/monitor/monitor.py
Wiezzel: Fixed.

golem/monitor/monitor.py
```diff
@@ -192,25 +196,33 @@ def on_requestor_stats_snapshot(self,
...
+            IncomeModel(
+                self.meta_data.cliid,
+                self.meta_data.sessid,
+                addr,
+                value
+            )
         )
-        self.sender_thread.send(msg)
 
     def on_balance_snapshot(self, eth_balance: int, gnt_balance: int,
```
Krigpl, line 222: This function doesn't seem to be used according to `grep`

Krigpl (approved):

golem/monitor/monitor.py
Wiezzel: @Krigpl This only shows that `grep` is not the best tool for checking if a function is called. Take a look at [this method](https://github.com/golemfactory/golem/blob/develop/golem/monitor/monitor.py#L115).

PR #2397, opened 2018-03-19
Withdraw rpc

mdtanrikulu:
The definition will do withdraw process and return transaction id(s) in array.


Krigpl:
Implementation in #2426

PR #2399, opened 2018-03-19
Close images better

maaktweluit:
Resolved these 5 errors that showed up on our windows builds out of the blue:

https://buildbot.golem.network/buildbot/#/builders/11/builds/2298/steps/6/logs/stdio

### DIFF
--- a/apps/blender/task/blenderrendertask.py
+++ b/apps/blender/task/blenderrendertask.py
@@ -645,7 +645,7 @@ class CustomCollector(RenderingTaskCollector):
             self.current_offset += new_img_res_y
             img_offset.paste(new_part, (0, offset))
             result = ImageChops.add(final_img, img_offset)
-            return result
+        return result
 
 
 def generate_expected_offsets(parts, res_x, res_y):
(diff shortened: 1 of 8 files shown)


apps/blender/task/blenderrendertask.py
```diff
@@ -645,7 +645,7 @@ def _paste_image(self, final_img, new_part, num):
             self.current_offset += new_img_res_y
             img_offset.paste(new_part, (0, offset))
             result = ImageChops.add(final_img, img_offset)
-            return result
+        return result
```
ktomala: This will fail if block under `with` is not executed (not sure because of this handle_image_error - what it does exactly)?

maaktweluit: Good one!

I'm not sure, this change did not matter in the end so reverted it, thanks!

ktomala (approved):
LGTM

mfranciszkiewicz (approved):

etam:
I'm late :(

But I'd suggest making `ImgRepr` implement context manager interface https://docs.python.org/3/library/stdtypes.html#typecontextmanager and use `with` statement as much as possible.

maaktweluit:
Sorry for merging so quickly, a PR was blocked by this.

Good idea for next steps, will add them if i get back to this code again :)

PR #2417, opened 2018-03-20
Connect to mainnet seeds and geths

Krigpl:
Those addresses don't work at the moment but they will as soon as @ktomala finish setting them up.


ktomala (approved):
LGTM

mfranciszkiewicz (approved):
lgtm for now, but IMO we should get rid of `if mainnet` wherever possible, and create runtime configurations for golem.

Krigpl:
@mfranciszkiewicz Yeah, that'd be nice
tech-debt-burden

PR #2426, opened 2018-03-20
Withdrawal RPC

Krigpl:
For withdrawing GNT/ETH.


golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -76,4 +77,58 @@ def get_balance(self):
...
+            if amount > eth:
+                raise ValueError('Not enough ETH available')
+            log.info(
+                "Withdrawing %f ETH to %s",
+                amount / denoms.ether,
+                destination,
+            )
+            return [self._sci.transfer_eth(destination, amount)]
+
+        if currency == 'GNT':
```
mfranciszkiewicz, line 99: Is it safe to assume that we have enough `ETH` for the transfer / conversion?

golem/client.py
```diff
@@ -849,6 +849,19 @@ def get_payments_list(self):
     def get_incomes_list(self):
         return self.transaction_system.get_incoming_payments()
 
+    def withdraw(
+            self,
+            amount: Union[str, int],
+            destination: str,
+            currency: str) -> List[str]:
+        if not self.is_mainnet():
```
mfranciszkiewicz: Could you change this call to `if not self.mainnet`? `is_mainnet()` has been moved to `Node` in `rpc_pass` branch

golem/transactions/ethereum/ethereumtransactionsystem.py
Krigpl: No :)

golem/client.py
Krigpl: Done

mfranciszkiewicz (approved):

PR #2434, opened 2018-03-20
Todo comments

badb:
* Add issue numbers to TODOs
* Remove some TODOs that are no longer valid 
* ... or are too complicated issues do be described as one TODO in code (first removed TODO in client)


apps/dummy/task/dummytask.py
```diff
@@ -110,10 +110,6 @@ def query_extra_data(self,
 
     # FIXME quite tricky to know that this method should be overwritten
     def accept_results(self, subtask_id, result_files):
-        # TODO maybe move it to the base method
-        if self.subtasks_given[subtask_id]["status"] == SubtaskStatus.finished:
```
maaktweluit, line 114: was this moved in this PR? dont see it added

maaktweluit (changes requested):
LGTM! nice cleanup!

One small comment, will approve on reply

apps/dummy/task/dummytask.py
badb: This line is already in the base method so this line and comments are no longer valid.

maaktweluit (approved):

PR #2435, opened 2018-03-20
is_mainnet RPC call v2

mfranciszkiewicz:
Moves `is_mainnet` call from `Client` to `Node`; `Client` instance is not yet created when the call is needed.


badb (approved):

Krigpl (approved):
verynicegreatsuccess

PR #2436, opened 2018-03-20
Use --rinkeby in NodeProcess

Krigpl:
I thought `networkid=4` is the same as `--rinkeby` but apparently not because our tests don't pass anymore.


maaktweluit (approved):
LGTM!

Builds are failing, this will rebuild so lets see :)

PR #2444, opened 2018-03-21
CLI account unlock

### LINKED ISSUE #2348: Unlock geth account with password from CLI
Propose a solution that will allow us to achieve it without just passing password with plaintext that can be easily searched with comman line history.

maaktweluit:
Resolves #2348

### DIFF
--- a/golem/interface/client/account.py
+++ b/golem/interface/client/account.py
@@ -1,4 +1,5 @@
 from typing import Dict, Any
+import getpass
 
 from ethereum.utils import denoms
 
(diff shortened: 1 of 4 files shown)


golem/interface/client/account.py
```diff
@@ -45,6 +47,33 @@ def info(self) -> Dict[str, Any]:
             )
         )
 
+    @staticmethod
+    @command(help="Unlock account, will prompt for your password")
+    def unlock() -> str:
+
```
Krigpl: rm ?

golem/interface/client/account.py
```diff
@@ -45,6 +47,33 @@ def info(self) -> Dict[str, Any]:
...
+        if not has_key:
+            confirm = getpass.getpass('Confirm password:')
+            if confirm != pswd:
+                return "Password and confirmation do not match."
+            print("Generating keys, this can take up to 5 minutes...")
+        else:
+            print("Unlocking account...")
+
+        success = sync_wait(client.set_password(pswd), timeout=300)
```
Krigpl: The timeout should be higher, IIRC generating new key averages on 2 minutes, but p95 is 8minutes. And even recently @Elfoniok had to wait ~6minutes :)
This timeout is pretty harmless so might be like 10 minutes or more as well.

Krigpl (commented):
lgtm, just a timeout thing

mfranciszkiewicz (approved):
LGTM

Wiezzel (commented):
OK, but the timeout needs to be longer.

golem/interface/client/account.py
maaktweluit: Changed it to 15 minutes timeout and the message to 10m, managing expectations here :)

Krigpl (approved):
verynicegreatsuccess

Wiezzel (approved):
OK

PR #2462, opened 2018-03-22
Dirty hack to bypass terms of use

Wiezzel:


mfranciszkiewicz (approved):

Krigpl:
tbh, test coverage of this code leaves a lot to be desired, since not a single test failed...
I made an issue for it #2475

PR #2473, opened 2018-03-22
Fix startup issues

mfranciszkiewicz:
- registered methods in Node were not re-registered after re-establishing the session
- event publishing from another thread terminated the RPC session


maaktweluit (approved):
LGTM!

PR #2485, opened 2018-03-23
Moved terms of use logic to a separate class

Wiezzel:
As suggested in https://github.com/golemfactory/golem/pull/2373#discussion_r176162100

### DIFF
--- a/golem/node.py
+++ b/golem/node.py
@@ -33,8 +33,6 @@ class Node(object):  # pylint: disable=too-few-public-methods
     """ Simple Golem Node connecting console user interface with Client
     :type client golem.client.Client:
     """
-    TERMS_ACCEPTED_KEY = 'terms_of_use_accepted'
-    TERMS_VERSION = 1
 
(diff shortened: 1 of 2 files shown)


golem/node.py
```diff
@@ -284,3 +277,27 @@ def _stop_on_error(self, msg: str, err: Any) -> None:
         if self._reactor.running:
             logger.error("Stopping because of %r error: %r", msg, err)
             self._reactor.callFromThread(self._reactor.stop)
+
+
+class TermsOfUse:
```
Krigpl: Having it in a separate file would also be fine

Krigpl (approved):
verynicegreatsuccess

golem/node.py
Wiezzel: In the same directory?

Krigpl: Yeah

Wiezzel: Voilà.

PR #2488, opened 2018-03-23
Set dummy loglevel to debug

maaktweluit:
This way when they fail we can investigate more

### DIFF
--- a/loggingconfig.py
+++ b/loggingconfig.py
@@ -32,7 +32,7 @@ LOGGING = {
         'console': {
             'class': 'logging.StreamHandler',
             'level': 'INFO',
-            'formatter': 'simple',
+            'formatter': 'date',
             'filters': [],
             'stream': 'ext://sys.stderr',
         },
--- a/loggingconfig.py
+++ b/loggingconfig.py
@@ -32,14 +32,14 @@ LOGGING = {
         'console': {
             'class': 'logging.StreamHandler',
             'level': 'INFO',
-            'formatter': 'date',
+            'formatter': 'simple',
             'filters': [],
             'stream': 'ext://sys.stderr',
         },
         'file': {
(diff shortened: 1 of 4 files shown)


mfranciszkiewicz (dismissed):

loggingconfig.py
```diff
@@ -32,7 +32,7 @@
         'console': {
             'class': 'logging.StreamHandler',
             'level': 'INFO',
-            'formatter': 'simple',
+            'formatter': 'date',
```
tworec: this will affect whole app, not only dummy task

tworec (changes requested):

loggingconfig.py
maaktweluit: Good one!

This is not meant to be merged, this was required to find what component is blocking

Changed title to WIP, will ping you when its ready for review again

maaktweluit:
Turned out UPNP was the problem this time...

Left some logs and added the dates in this test only.

Hope the next time dummy tests failures are easier to spot :)

@mfranciszkiewicz @tworec please review again!

mfranciszkiewicz (approved):

tests/golem/task/dummy/runner.py
```diff
@@ -61,6 +68,7 @@ def create_client(datadir):
     config_desc = ClientConfigDescriptor()
     config_desc.init_from_app_config(app_config)
     config_desc.key_difficulty = 0
+    config_desc.use_upnp = False
```
tworec, line 71: seems it's only line changed not regarding logger. Is this cause of our headache? :)

tworec (approved):

tests/golem/task/dummy/runner.py
maaktweluit: Yep, the one and only...

Hopefully the rest makes it easier to find next time

Issue #2475, opened 2018-03-23
Improve test coverage of ToS


Krigpl:
Change like this https://github.com/golemfactory/golem/pull/2462 should break at least one test.
And while at this it wouldn't hurt to refactor ToS as suggested here https://github.com/golemfactory/golem/pull/2373#discussion_r176162100

Wiezzel:
I will do this when #2462 is reverted.

Wiezzel:
Closed by #2489

PR #2489, opened 2018-03-23
Improved test coverage for terms of use (#2475)

Wiezzel:


tests/golem/test_terms.py
```diff
@@ -0,0 +1,56 @@
...
+
+from golem.model import GenericKeyValue
+from golem.terms import TermsOfUse
+from golem.tools.testwithdatabase import TestWithDatabase
+
+
+class TestTermsOfUse(TestWithDatabase):
+
+    def test_are_terms_accepted_no_entry(self):
+        self.assertFalse(TermsOfUse.are_terms_accepted())
```
Krigpl, line 12: This and all other `self.assert*`s can be replaced with simple `assert not bool` etc which are clearer

tests/golem/test_terms.py
```diff
@@ -0,0 +1,56 @@
...
+        self.assertTrue(TermsOfUse.are_terms_accepted())
+
+    def test_accept_terms_right_version(self):
+        GenericKeyValue.create(
+            key=TermsOfUse.TERMS_ACCEPTED_KEY,
+            value=TermsOfUse.TERMS_VERSION)
+        TermsOfUse.accept_terms()
+        self.assertTrue(TermsOfUse.are_terms_accepted())
+
+    @patch('golem.terms.get_golem_path', return_value='/tmp')
```
Krigpl: use `self.tempdir`, `/tmp` won't work on Windows for instance

Krigpl (changes requested):
lgtm, just the temporary dir thing

tests/golem/test_terms.py
Wiezzel: I am not a fan of using plain assertions in unit tests. `unittest.TestCase.failureException` could be changed to a different exception class than `AssertionError` and then we get different test outcomes. I advocate using plain `assert` only in core code.

tests/golem/test_terms.py
Wiezzel: Fixed.

tests/golem/test_terms.py
```diff
@@ -0,0 +1,56 @@
...
+    @patch('golem.terms.get_golem_path')
+    def test_show_terms(self, get_golem_path):
+        get_golem_path.return_value = self.tempdir
+        content = """
+        GOLEM TERMS OF USE
+        ==================
+        Bla bla bla bla
+        """
+        os.makedirs(self.new_path / 'golem', exist_ok=True)
+        with open(self.new_path / 'golem' / 'TERMS.html', mode='w') as terms:
```
Krigpl: Actually that path should also be a constant in the `Terms` class

Wiezzel: Good idea.

Krigpl (approved):

PR #2385, opened 2018-03-16
Cleanup socket addresses in TaskHeader, WTR/F

mfranciszkiewicz:
- `TaskHeader.task_owner` is now required
- moves the following `TaskHeader` properties to `TaskHeader.task_owner`:
    `node_name`, `task_owner_key_id`, `task_owner_addr`, `task_owner_port`
- addresses the `owner` property in `WaitingTaskResult` and `WaitingTaskFailure`

PR #2495, opened 2018-03-25
Remove PendingConnectionsServer._is_address_valid quickfix

mfranciszkiewicz:
- removes the quickfix introduced in #2352
- improves address filtering in `ip_addresses` and `ipv4_networks` (`hostaddress.py`)
- catches all exceptions in `SocketAddress.is_proper_address`

Moved from #2385

### DIFF
--- a/golem/core/hostaddress.py
+++ b/golem/core/hostaddress.py
@@ -1,10 +1,10 @@
 import ipaddress
 import logging
+import os
 import socket
 from collections import Iterable
 from typing import Union, List
 
-import os
 import netifaces
 
 from golem.network.stun import pystun as stun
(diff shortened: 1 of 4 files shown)


Elfoniok (approved):

golem/core/hostaddress.py
```diff
@@ -1,12 +1,13 @@
+import ipaddress
 import logging
-import netifaces
+import socket
+from collections import Iterable
+from typing import Union, List
 
 import os
-import socket
+import netifaces
```
etam: minor: `os` is part of standard library. should be in the first group.

etam (approved):

Issue #2419, opened 2018-03-20
Don't allow sending batchTransfer to yourself


Krigpl:
Currently provider can specify the address they want receive the payment to
https://github.com/golemfactory/golem/blob/develop/golem/transactions/transactionsystem.py#L38
if they send requestor's address then requestor will try to send the payment to itself and the batchTransfer transaction will fail.
Provider should be rejected if they specify requestor's ethereum address.

PR #2499, opened 2018-03-26
Don't use eth_account from message

Krigpl:
Mitigates #2419
We don't support custom provided Ethereum accounts anyway.


golem/task/tasksession.py
```diff
@@ -540,7 +538,7 @@ def _react_to_report_computed_task(self, msg):
             msg.key_id,
             msg.node_name,
             p2p_node.Node.from_dict(msg.node_info),
-            msg.eth_account
+            pubkeytoaddr(msg.key_id),
```
jiivan, line 541: If we want to change this logic, then it'd be better to use `msg.task_to_compute.provider_ethereum_address`

Krigpl:
In favor of #2546

PR #2502, opened 2018-03-26
Support timeouts in _DenyAcl

mfranciszkiewicz:


golem/task/acl.py
```diff
@@ -13,29 +15,43 @@ def is_allowed(self, node_id: str) -> bool:
         pass
 
     @abc.abstractmethod
-    def disallow(self, node_id: str) -> None:
+    def disallow(self, node_id: str, timeout: int) -> None:
```
etam: It's unclear in what units is this timeout. I know it's a bit verbose, but I really think it should be like `timeout_seconds`.

golem/task/acl.py
```diff
@@ -47,7 +63,7 @@ def _read_set_from_file(path: Path) -> Set[str]:
         return set()
 
 
-def get_acl(datadir: Path) -> Acl:
```
etam, line 50: Is this necessary? This function returns an object that fits in the `Acl` metaclass.

mfranciszkiewicz: It is necessary from a linter's point of view.

etam (approved):

PR #2509, opened 2018-03-28
docker logs level WARN -> INFO

tworec:
docker on macos is not so smooth. I insist to increase logs verbosity for it.


maaktweluit (approved):
yay for more clear logs and fixing those pesky bugs!

mfranciszkiewicz (approved):

PR #2512, opened 2018-03-28
Pause verification queue

### LINKED ISSUE #2472: Make sure that verification queue is stopped when Golem is stopped
When user stops Golem docker machines are stopped as well and all networks connections are disconnected. We cannot get more information about task resuts, but we need to check that results that were already sent to the verification queue will no try to be computed until Golem service is started again.

mfranciszkiewicz:
Resolves #2472


maaktweluit (commented):
LGTM!

Nice logs added everywhere!

maaktweluit (approved):

PR #2540, opened 2018-03-29
Allow users to specify the public IP address and ports

mfranciszkiewicz:


golem/client.py
```diff
@@ -363,9 +365,15 @@ def connect(ports):
...
-            if self.config_desc.use_upnp:
+            if self.config_args.use_upnp is not None:
+                use_upnp = self.config_args.use_upnp
+            else:
+                use_upnp = self.config_desc.use_upnp
+
+            if use_upnp:
                 self.start_upnp(ports + list(hyperdrive_ports))
             self.node.update_public_info()
+            self._override_node_properties(self.node)
```
maaktweluit, line 388: would be nice to call this before `use_upnp` to remove that manual check there, i guess this is not possible due to the need of upnp info

maaktweluit (approved):
LGTM!

golem/client.py
mfranciszkiewicz: I wanted to make these the true overrides. It's true that it makes little sense to force the public IP and ports with UPnP on, but now we have a `--noupnp` flag.

golem/argsparser.py
```diff
@@ -65,3 +65,14 @@ def parse_peer(ctx, param, value):
...
                 "Invalid peer address specified: {}".format(e))
     return addresses
+
+
+def parse_port(ctx, param, value):
+    del ctx, param
+    if value is not None:
+        try:
+            if not 0 < value < 65536:
+                raise Exception(value)
```
tworec: such exception w/o msg just with value would cause users/developers doubts
shouldn't we just raise `click.BadParameter` here?

golem/client.py
```diff
@@ -412,6 +420,23 @@ def terminate(*exceptions):
         self.task_server.start_accepting(listening_established=task.callback,
                                          listening_failure=task.errback)
 
+    def _override_node_properties(self, node):
+        configurations = [self.config_desc, self.config_args]  # mind the order
+        properties = {
+            'node_address': 'prv_addr',
+            'pub_node_address': 'pub_addr',
+            'pub_p2p_port': 'pub_p2p_port',
```
tworec: `pub_p2p_port` --> `p2p_pub_port` as in `golem/network/p2p/node.py:21`

golem/node.py
```diff
@@ -45,7 +45,8 @@ def __init__(self,  # noqa pylint: disable=too-many-arguments
                  start_geth: bool = False,
                  start_geth_port: Optional[int] = None,
                  geth_address: Optional[str] = None,
-                 password: Optional[str] = None) -> None:
+                 password: Optional[str] = None,
+                 config_args: Optional[ClientConfigDescriptor] = None) -> None:
```
tworec: I'd move `config_args` just after `config_desc` not being optional as it is not optional really. We always provide it within `golemapp.py`

golem/client.py
```diff
@@ -98,7 +98,8 @@ def __init__(  # noqa pylint: disable=too-many-arguments
             use_concent: bool = False,
             start_geth: bool = False,
             start_geth_port: Optional[int] = None,
-            geth_address: Optional[str] = None) -> None:
+            geth_address: Optional[str] = None,
+            config_args: Optional[ClientConfigDescriptor] = None) -> None:
```
tworec: I'd move `config_args` just after `config_desc` not being optional as it is not optional really. See my comment in `golem/node.py:49`

golemapp.py
```diff
@@ -173,6 +190,7 @@ def start(monitor, concent, datadir, node_address, rpc_address, peer, mainnet,
         node = Node(
             datadir=datadir,
             app_config=app_config,
+            config_args=config_args,
```
tworec: here you just confirm my hipotesis that this is more intuitive place for that arg

tworec (changes requested):

golem/argsparser.py
mfranciszkiewicz: We already do in `except`

golem/node.py
mfranciszkiewicz: I've moved it to the end because it affects tens of tests and is not required to build the instance.

golem/argsparser.py
tworec: I saw, but what is this raise-catch-raise for?

tworec (approved):

golem/argsparser.py
mfranciszkiewicz: If the comparison inside the `try` block raises other exception, everything is handled in a single `except`.

tworec: This is of minor importance and feel free to ignore it, but to be consistent: You are using `type=Click.INT` so `value` here has type `int`. Click lib handles non-integer input for you and prevents calling this function if input is not valid int. I can't imagine other exception in comparsion. 

I would raise `Click.BadParameter` immediately with a bit more informative msg: _'Port should be in range (1-65535)'_. If you expect other Exception, than catch it and rethrow as `Click.BadParameter`.

mfranciszkiewicz: Consistency is what I did here explicitly - this function is written in the same style as other functions.

tworec: All other functions raises `Click.BadParam` immediately and catches some specific exceptions not vanilla `Exception`

tworec: buts lets not dive deeper into it. Its ok. We have more important problems 🚀

mfranciszkiewicz: If I had a nickel for every missing exception type fix...

tworec: :wink:

PR #2543, opened 2018-03-30
Check that subtask_id was generated from key_id

badb:


mfranciszkiewicz (approved):

tests/golem/task/test_taskkeeper.py
```diff
@@ -689,12 +694,20 @@ def test_check_comp_task_def(self):
...
+               'Definition of this subtask was already received.' % (subtask_id,
+                                                                     task_id) \
                in l.output[0]
 
-        del ctk.active_tasks[task_id].subtasks['xxyyzz']
+        del ctk.active_tasks[task_id].subtasks[subtask_id]
         assert ctk.check_comp_task_def(comp_task_def)
+
+        comp_task_def['subtask_id'] = "abc"
+        with self.assertLogs(logger, level="INFO") as log_:
```
tworec, line 709: style is not consistent with beginning of the class:
- hardcoded subtask_id instead of generated one
- `log_` --> `l`

tworec (approved):
minor comment only

tests/golem/task/test_taskkeeper.py
badb: l causes linter warnings. So rather rest of the `l` should be renamed to `log_` in this class.

PR #2544, opened 2018-03-30
HyperG share timeout

mfranciszkiewicz:
- adds removal timers for every file shared by Golem

Requires HyperG v0.2.6 (master branch).


maaktweluit (approved):
LGTM!

Maybe later more can move to client_options to reduce "too-many-arguments" in some places.

Issue #2433, opened 2018-03-20
Turn off support for LuxRender if mainnet is active


badb:
Verification function and results collecting workflow for Lux is much less tested than Blender app. We don't want to add it to mainnet release yet.

PR #2549, opened 2018-04-03

Elfoniok:
https://github.com/golemfactory/golem/issues/2433


badb:
Can you remove dummy task from mainnet as well?

PR #2550, opened 2018-04-03
Payment trust

mfranciszkiewicz:
- Increase payment trust on batch transfer
- Decrease payment trust for overdue payments
- Decrease payment trust for rejected subtasks

### DIFF
--- a/golem/transactions/incomeskeeper.py
+++ b/golem/transactions/incomeskeeper.py
@@ -7,7 +7,6 @@ from typing import List
 from ethereum.utils import denoms
 from pydispatch import dispatcher
 
-from golem.core.common import get_timestamp_utc
 from golem.core.variables import PAYMENT_DEADLINE
 from golem.model import Income
 from golem.utils import encode_hex, pubkeytoaddr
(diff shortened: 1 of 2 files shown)


golem/transactions/incomeskeeper.py
```diff
@@ -80,6 +87,28 @@ def expect(self, sender_node_id, subtask_id, value):
...
+            income = Income.get(subtask=subtask_id, accepted_ts=None,
+                                overdue=False)
+        except Income.DoesNotExist:
+            logger.error(
+                "Income.DoesNotExist subtask_id: %r",
+                subtask_id)
+            return
+
+        # Mark as accepted + overdue to exclude it from other queries
+        income.accepted_ts = get_timestamp_utc()
```
Krigpl: Could we just remove it from the DB instead ? It might be better for now (until we have Concent), instead of permanently showing it in the GUI as awaiting.

golem/transactions/incomeskeeper.py
```diff
@@ -125,7 +154,15 @@ def update_overdue_incomes() -> List[Income]:
...
             )
         ))
+
         for income in incomes:
             income.overdue = True
             income.save()
+
+            dispatcher.send(
+                signal='golem.income',
+                event='rejected',
```
Krigpl: I dunno if rejected correctly describes it, maybe just `overdue` ?

mfranciszkiewicz: It sounds better; `rejected` and `overdue` do the exact same thing though

Krigpl (approved):
Some kind of unit test would be great too :)

PR #2556, opened 2018-04-04

Elfoniok:


golem/node.py
```diff
@@ -52,6 +52,9 @@ def __init__(self,  # noqa pylint: disable=too-many-arguments
         # and will prevent the IOCP / kqueue reactors from being installed.
         from twisted.internet import reactor
 
+        if mainnet:
+            DockerManager.MAINNET = True
```
Krigpl: same as `DockerManager.MAINNET = mainnet`

golem/task/taskmanager.py
```diff
@@ -63,11 +63,11 @@ class AlreadyRestartedError(Error):
     def __init__(
             self, node_name, node, keys_auth, listen_address="",
             listen_port=0, root_path="res", use_distributed_resources=True,
-            tasks_dir="tasks", task_persistence=True):
+            tasks_dir="tasks", task_persistence=True, mainnet=False):
         super().__init__()
 
-        self.apps_manager = AppsManager()
-        self.apps_manager.load_apps()
+        self.apps_manager = AppsManager(mainnet)
```
Krigpl: I'm not really familiar with this code but this seems a bit suspicious that `AppsManager` is created both in `Node` and here. Wouldn't it make sense here to receive it from `Node` ? (we wouldn't have to provide `mainnet` flag as well then)

apps/appsmanager.py
```diff
@@ -20,13 +20,19 @@ def __init__(self):
...
     """ Temporary solution for apps detection and management. """
-    def __init__(self):
+    def __init__(self, mainnet):
         self.apps = OrderedDict()
+        self._mainnet = mainnet
+
+    def load_all_apps(self):
+        self._load_apps(REGISTERED_CONFIG_FILE)
+        if not self._mainnet:
+            self._load_apps(os.path.join('apps', 'registered_test.ini'))
```
Krigpl: make a constant like `REGISTERED_CONFIG_FILE`

tests/golem/test_opt_node.py
```diff
@@ -24,6 +25,7 @@ def setUp(self):
         self.args = ['--datadir', self.path]
 
     def tearDown(self):
+        DockerManager.MAINNET = False
```
Krigpl: why is it needed? Seems weird that it's only required in `tearDown` without any changes in other places

golem/client.py
```diff
@@ -98,9 +99,11 @@ def __init__(  # noqa pylint: disable=too-many-arguments
             use_concent: bool = False,
             start_geth: bool = False,
             start_geth_port: Optional[int] = None,
-            geth_address: Optional[str] = None) -> None:
+            geth_address: Optional[str] = None,
+            apps_manager: AppsManager = None) -> None:
 
         self.mainnet = mainnet
+        self.apps_manager = apps_manager
```
Krigpl: you don't need that, do you?

golem/client.py
```diff
@@ -98,9 +99,11 @@ def __init__(  # noqa pylint: disable=too-many-arguments
             use_concent: bool = False,
             start_geth: bool = False,
             start_geth_port: Optional[int] = None,
-            geth_address: Optional[str] = None) -> None:
+            geth_address: Optional[str] = None,
+            apps_manager: AppsManager = None) -> None:
```
Krigpl: This will make a bunch of tests (`test_client`) unhappy

Krigpl (approved):
lgtm, I think

golem/client.py
Elfoniok: Thanks

golem/client.py
Elfoniok: I need it to pass in start_network function.

Krigpl: Right, github collapsed ~200 lines between these changes and I didn't the context change, derp.

apps/appsmanager.py
```diff
@@ -20,13 +20,21 @@ def __init__(self):
 
 class AppsManager(object):
     """ Temporary solution for apps detection and management. """
-    def __init__(self):
+    def __init__(self, mainnet):
         self.apps = OrderedDict()
+        self._mainnet = mainnet
+        self.isLoaded = False
```
Krigpl, line 26: `snake_case` convention
also I think it'd be better to make it private and just make `load_all_apps` do nothing if it's already loaded

golem/task/taskmanager.py
```diff
@@ -63,13 +63,14 @@ class AlreadyRestartedError(Error):
...
+            tasks_dir="tasks", task_persistence=True,
+            apps_manager=AppsManager(False)):
         super().__init__()
 
-        self.apps_manager = AppsManager()
-        self.apps_manager.load_apps()
-
-        apps = list(self.apps_manager.apps.values())
+        self.apps_manager = apps_manager
+        if not apps_manager.isLoaded:
```
Krigpl, line 71: on the other hand, do you really need that? At this point the apps have already been loaded
maybe except for the tests, but then it should be fixed in the tests, not like this

Elfoniok:
Closing in favor of which is targeted for develop:
https://github.com/golemfactory/golem/pull/2588/files

PR #2560, opened 2018-04-04
Subtask limit warning

jiivan:
related to: https://github.com/golemfactory/golem-electron/issues/253

Subtasks count calculation is refactored to separate function and exposed via RPC.


apps/rendering/task/framerenderingtask.py
```diff
@@ -26,6 +27,50 @@
 DEFAULT_PADDING = 4
 
 
+def calculate_subtasks_count_with_frames(total_subtasks: int, frames: int):
```
etam: I can haz return type annotation?

apps/rendering/task/framerenderingtask.py
```diff
@@ -26,6 +27,50 @@
...
+    est = num_frames / math.ceil(num_frames / total_subtasks)
+    est = int(math.ceil(est))
+    if est != total_subtasks:
+        logger.warning("Too many subtasks for this task. %s "
+                       "subtasks will be used.", est)
+
+    return est
+
+
+def calculate_subtasks_count(
```
etam, line 52: I can haz return type annotation?

mfranciszkiewicz (approved):

PR #2566, opened 2018-04-04
Fix closure_time in batchTransfer

Krigpl:
`closure_time` should be set to the timestamp of the last payment from the batch, not the current timestamp (it matters when batch transfer doesn't cover all the payments).


mfranciszkiewicz (approved):

maaktweluit (approved):
LGTM!

PR #2581, opened 2018-04-05
2518 concent/forced download integration unit tests

shadeofblue:


requirements.txt
```diff
@@ -54,6 +54,8 @@ token_bucket==0.2.0
 py-cpuinfo==3.3.0
 humanize==0.5.1
 raven
-Golem-Messages==1.17.2
--e git://github.com/golemfactory/golem-smart-contracts-interface.git@e1b1f4cafeaf73035effaad501d3b04770044ce6#egg=golem_sci
 html2text==2018.1.9
+factory-boy==2.9.2
+Faker==0.8.9
```
jiivan: Are those needed outside of tests?

scripts/concent_integration_tests/base.py
```diff
@@ -0,0 +1,97 @@
...
+from golem.core import variables
+
+
+# pylint:disable=no-member
+
+class ConcentBaseTest:
+
+    @staticmethod
+    def _fake_keys():
+        return cryptography.ECCx(faker.Faker().binary(length=32))
```
jiivan: I think faker is unnecessary here. Wouldn't just `cryptography.ECCx(None)` be enough?

scripts/concent_integration_tests/base.py
```diff
@@ -0,0 +1,97 @@
...
+        )
+
+    def _load_response(self, response, priv_key=None):
+        return golem_messages.load(
+            response,
+            priv_key or self.priv_key,
+            variables.CONCENT_PUBKEY,
+        )
+
+    def assertSameMessage(self, msg1, msg2):
```
jiivan: Is your intention here to not compare signature and header.encoding? If so please add a comment.

requirements.txt
shadeofblue: @jiivan ah, probably not...  just my PyCharm complained those were missing...

scripts/concent_integration_tests/base.py
jiivan: What I meant is that you could just use `.assertEqual(msg1, msg2)` because `__eq__` is implemented in `Message` class.

shadeofblue: @jiivan hmm, indeed, good catch, it was about not including `header.encrypted` in the comparison ... but on the other hand, we might consider not including it in the equality check altogether ->  https://github.com/golemfactory/golem-messages/pull/187  

would there be any reasons against such a move?

jiivan (approved):
lgtm

PR #2588, opened 2018-04-06

Elfoniok:


Krigpl (approved):
thisisgine

setup.py
```diff
@@ -65,6 +65,7 @@
         ]),
         (path.normpath('../../golem/apps'), [
             path.normpath('apps/registered.ini'),
+            path.normpath('apps/registered_test.ini'),
```
maaktweluit, line 68: I think a similar line should also be added to the pyinstaller hooks to bundle this new file in the binaries

mfranciszkiewicz: @maaktweluit We're using a wildcard for `ini` files in that directory.

maaktweluit: Thanks TIL :D

PR #2448, opened 2018-03-22
Pip freeze

maaktweluit:
Resolved #2347

- [x] Test how CI runs these new versions
- [x] Write doc on how / when to freeze

PR #2593, opened 2018-04-08
Added password strength checks as in electron

### LINKED ISSUE #2542: Account password complexity when using golemcli
I believe I remember the GUI not allowing to use a simple password. 
Should probably implement similar logic for golemcli for security.

maaktweluit:
Resolves #2542

### DIFF
--- a/golem/interface/client/account.py
+++ b/golem/interface/client/account.py
@@ -1,6 +1,6 @@
 from typing import Dict, Any
 import getpass
-from zxcvbn import zxcvbn
+import zxcvbn
 
 from decimal import Decimal
 from ethereum.utils import denoms
@@ -74,8 +74,8 @@ class Account:
 
             # Check password score, same library and settings used on electron
             account_name = getpass.getuser() or ''
-            result = zxcvbn(pswd, user_inputs=['Golem', account_name])
-            print(result['score'])
+            result = zxcvbn.zxcvbn(pswd, user_inputs=['Golem', account_name])
+            # print(result['score'])
             if result['score'] < MIN_SCORE:
                 return "Password is not strong enough. " \
                     "Please use capitals, numbers and special characters."
--- a/golem/interface/client/account.py
+++ b/golem/interface/client/account.py
@@ -1,11 +1,15 @@
 from typing import Dict, Any
 import getpass
+from zxcvbn import zxcvbn
 
 from decimal import Decimal
 from ethereum.utils import denoms
 
 from golem.core.deferred import sync_wait
 from golem.interface.command import Argument, command, group
+
+MIN_LENGTH = 5
+MIN_SCORE = 2
 
 
 @group(help="Manage account")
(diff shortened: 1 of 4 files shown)


maaktweluit:
@kascheri12 You are also invited to review, can not add you in the list

golem/interface/client/account.py
```diff
@@ -1,11 +1,15 @@
...
 import getpass
+from zxcvbn import zxcvbn
 
 from ethereum.utils import denoms
 
 from golem.core.deferred import sync_wait
 from golem.interface.command import command, group
 
+MIN_LENGTH = 5
+MIN_SCORE = 2
```
Wiezzel: I'm not sure if score 2 counts as strong password.
```python
>>> zxcvbn('golem.pass', user_inputs=['Golem', 'adam'])['score']
2
```

requirements.txt
```diff
@@ -57,3 +57,4 @@ raven
 Golem-Messages==1.17.2
 -e git://github.com/golemfactory/golem-smart-contracts-interface.git@e1b1f4cafeaf73035effaad501d3b04770044ce6#egg=golem_sci
 html2text==2018.1.9
+zxcvbn-python
```
Wiezzel: No package version specified?

golem/interface/client/account.py
maaktweluit: whow, i fully agree with you.

Wanted to match settings chosen in electron:
https://github.com/golemfactory/golem-electron/blob/039cc0a8e95bf7a3b5c1fd7676cebb842fb9ddaa/src/components/onboarding/steps/Register.js#L143

Will check if this was thought of yet and update ( both if needed ) accordingly

requirements.txt
maaktweluit: all versions will be locked in #2448 

Currently ( before freeze ) we only lock versions when it breaks.

Wiezzel: Alright, understood.

golem/interface/client/account.py
```diff
@@ -59,6 +63,19 @@ def unlock(self) -> str:  # pylint: disable=no-self-use
...
+            if len(pswd) < MIN_LENGTH:
+                return "Password is too short, minimum is 5"
+
+            # Check password score, same library and settings used on electron
+            account_name = getpass.getuser() or ''
+            result = zxcvbn(pswd, user_inputs=['Golem', account_name])
+            print(result['score'])
+            if result['score'] < MIN_SCORE:
+                return "Password is not strong enough. " \
+                    "Please use capitals, numbers and special characters."
```
jiivan: Maybe we could use `result['feedback']['suggestions']` here?

jiivan (approved):
LGTM. One suggestion

And probably some basic tests would be a good idea.

kascheri12:
@maaktweluit Thank you for the invite. LGTM; this solves my issue.

golem/interface/client/account.py
maaktweluit: Made issue with this comment, so we can round off this PR :) #2679

golem/interface/client/account.py
maaktweluit: Good idea! Does need some testing / UX love, so made an issue #2678

maaktweluit:
@kascheri12 @jiivan  @Wiezzel Thanks for the reviews!

Will add some tests for this new code and fix the merge conflict

Issue #2411, opened 2018-03-19
[Concent] UC1 Verify the reactions to `VerdictReportComputedTask` and `ForceReportComputedTask`


shadeofblue:
in the context of the missing calls to:

https://github.com/golemfactory/golem/blob/develop/golem/task/taskserver.py#L401

I believe this should have been implemented alongside one of these issues:

https://github.com/golemfactory/golem/issues/2229
https://github.com/golemfactory/golem/issues/2231

@jiivan please have a look here

PR #2599, opened 2018-04-09
Uc1 concent subtask reaction

jiivan:
* Untangle results verification from `TaskSession`
* Use it on `ForceReportComputedTask` and `VerdictReportComputedTask`

fixes: #2411


golem/task/server/resources.py
```diff
@@ -14,6 +16,92 @@
 logger = logging.getLogger(__name__)
 
 
+def noop():
```
shadeofblue: @jiivan maybe `__noop` ? or maybe just define `lambda: None` as the default for `after_success` and `after_error` below without adding this function?

golem/task/server/resources.py
```diff
@@ -14,6 +16,88 @@
...
+    # Pepare callbacks for received resources
+    def on_success(extracted_pkg, *_args, **_kwargs):
+        logger.debug("Task result extracted %r",
+                     extracted_pkg.__dict__)
+        task_server.verify_results(
+            report_computed_task=report_computed_task,
+            extracted_package=extracted_pkg,
+        )
+
+        concent_service.cancel_task_message(
```
shadeofblue: @jiivan I'm unsure here but maybe this cancellation should be performed first, before we start the verification... otherwise, theoretically we can request a forced download even after we have received the result - since we do have the extracted package by now... 

(and yes, I'm aware that this is not a new issue)

golem/task/tasksession.py
```diff
@@ -531,9 +532,30 @@ def _react_to_report_computed_task(self, msg):
...
             return
 
-        self.task_server.receive_subtask_computation_time(
-            subtask_id,
-            msg.computation_time
+        def after_success():
+            self.disconnect(message.Disconnect.REASON.NoMoreMessages)
+
+        def after_error():
+            if msg.task_to_compute.concent_enabled:
```
shadeofblue, line 539: @jiivan hmm... I don't think this is a good idea to just do nothing in case we're using concent -> it doesn't make sense to wait until the timeout if we're sure that the direct download didn't succeed and the specific action of dispatching the FGTR here should be triggered immediately...

golem/task/server/resources.py
```diff
@@ -14,6 +16,88 @@
 logger = logging.getLogger(__name__)
 
 
+def computed_task_reported(
```
shadeofblue: generally, I'm unsure if `resources` is the best place for this handler... I have a feeling this module's logic should know as little about Golem's and Concent's protocol as possible and should only be concerned with the download of the resources... 

wouldn't the taskserver be a better place for this piece of logic?

jiivan: TaskServer is already overbloated (`C: golem/task/taskserver.py:1:0: Too many lines in module (1018/1000) (too-many-lines)`) but yes, I'll move it to better place than `task.server.resources`.

shadeofblue: @jiivan ah, right, okay :+1:

golem/task/server/resources.py
jiivan: `TaskServer.verify_results()` is async, but will move for clarity

golem/task/tasksession.py
jiivan: @shadeofblue  it is in default callback (because it doesn't require `TaskSession`):
https://github.com/golemfactory/golem/blob/c0ea6b96e25548dfefc703d05559f32e4043573f/golem/task/server/resources.py#L77-L85

golem/task/server/resources.py
shadeofblue: :+1:

shadeofblue (approved):
lgtm

PR #2620, opened 2018-04-11
Sync PaymentProcessor on startup

### LINKED ISSUE #2398: Re-enable local geth sync before starting Golem
This feature was removed in: https://github.com/golemfactory/golem/commit/8502d35a4202cbbe94914eda978268e14ad808ed#diff-9f351727974c5655127c6a4dc0671995R177 
because it was causing too much problems and most users use extrenal geths.

### LINKED ISSUE #2585: Creating new task fails if transaction system doesn't know about funds yet
```
ERROR    [golem.client                       ] Cannot create task {... cut ...}                                                                                                   
  File "/home/etam/code/github/golemfactory/golem/golem/client.py", line 658, in create_task
    task = yield self.enqueue_new_task(t_dict)
  File "/home/etam/code/github/golemfactory/golem/golem/client.py", line 533, in enqueue_new_task
    self.funds_locker.lock_funds(task)
  File "/home/etam/code/github/golemfactory/golem/golem/transactions/ethereum/fundslocker.py", line 62, in lock_funds                                                                                                                             
    if tfl.gnt_lock() > gnt - lock_gnt:
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

Krigpl:
Resolves #2398
Fixes #2585

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -245,6 +245,7 @@ class Client(HardwarePresetsMixin):
         )
         self._publish(Task.evt_task_status, kwargs['task_id'])
 
+    @report_calls(Component.client, 'sync')
     def sync(self):
         self.transaction_system.sync()
 
--- a/golem/client.py
+++ b/golem/client.py
@@ -245,9 +245,8 @@ class Client(HardwarePresetsMixin):
         )
         self._publish(Task.evt_task_status, kwargs['task_id'])
 
-    # TODO: re-enable. issue #2398
     def sync(self):
-        pass
+        self.transaction_system.sync()
 
     @report_calls(Component.client, 'start', stage=Stage.pre)
     def start(self):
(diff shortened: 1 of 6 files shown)


golem/client.py
```diff
@@ -245,9 +245,8 @@ def taskmanager_listener(self, sender, signal, event='default', **kwargs):
         )
         self._publish(Task.evt_task_status, kwargs['task_id'])
 
-    # TODO: re-enable. issue #2398
     def sync(self):
-        pass
+        self.transaction_system.sync()
```
mfranciszkiewicz: Could you decorate `sync` with `@report_calls(Component.client, 'sync')`? This way we can inform the UI on what's happening and show a message to the user.

mfranciszkiewicz (approved):

etam:
Why are changes in `golem/node.py` required?

Krigpl:
@etam They aren't really. This is a leftover after I was considering a different approach where RPC calls to client would be enabled only after the client has started (which in general makes sense).

etam:
@Krigpl If so, I think it should be in a separate pull request.

Krigpl:
> If so, I think it should be in a separate pull request.

Very well.

etam (approved):
approved

PR #2628, opened 2018-04-12
Remove DB from the Client

Krigpl:
It doesn't really belong there.

### DIFF
--- a/tests/golem/test_client.py
+++ b/tests/golem/test_client.py
@@ -93,7 +93,6 @@ class TestClient(TestWithDatabase, TestWithReactor):
             app_config=Mock(),
             config_desc=ClientConfigDescriptor(),
             keys_auth=(Mock(_private_key='a' * 32)),
-            database=Mock(),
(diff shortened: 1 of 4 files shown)


tests/golem/test_client.py
```diff
@@ -388,7 +377,6 @@ def test_quit(self, *_):
             app_config=Mock(),
             config_desc=ClientConfigDescriptor(),
             keys_auth=Mock(),
-            database=Mock(),
         )
         self.client.db = None
```
Wiezzel, line 381: This line should be removed as well.

Wiezzel (approved):
OK, just one more line to delete.

mfranciszkiewicz (changes requested):
The database has been lately moved from `Client` to `Node`, but the application (reactor) shutdown trigger is still set on `Client.quit`, which closes database connections.

Krigpl:
Maybe some other day.

PR #2631, opened 2018-04-12
Explicitly express subtask price in TaskToCompute.price

jiivan:


shadeofblue (approved):
looks good as far as I could tell

PR #2638, opened 2018-04-12
DummyTask fix no 34

mfranciszkiewicz:
DummyTask computation was failing due to `subtask_id` not being generated from `task_id`, but nevertheless the test suceeded.


maaktweluit (approved):
LGTM!

only 34? though we passed 100 by now ;)

PR #2644, opened 2018-04-13
Remove TaskSession.result_owner

jiivan:
`result_owner` is now available on demand, thus leaving only `err_msg` to refactor out of `TaskSession`.


shadeofblue (approved):
seems okay

PR #2650, opened 2018-04-13
Use a shallow copy of subtasks_given while updating the preview

### LINKED ISSUE #2630: RuntimeError: dictionary changed size during iteration
#### Encountered on:
Windows
Golem source
'GOLEM Version: 0.15.0 '
'Protocol Version: 26 '
'golem_messages Version: 1.17.2

```
2018-04-12 14:37:39 ERROR    golem.client                        b'task_server.sync_network failed '
  File "C:\Users\Magdalena Szydlowska\Projects\golem\golem\client.py", line 1246, in _run
    self._client.task_server.sync_network()
  File "C:\Users\Magdalena Szydlowska\Projects\golem\golem\task\taskserver.py", line 123, in sync_network
    self.__remove_old_tasks()
  File "C:\Users\Magdalena Szydlowska\Projects\golem\golem\task\taskserver.py", line 788, in __remove_old_tasks
    nodes_with_timeouts = self.task_manager.check_timeouts()
  File "C:\Users\Magdalena Szydlowska\Projects\golem\golem\task\taskmanager.py", line 572, in check_timeouts
    t.computation_failed(s.subtask_id)
  File "C:\Users\Magdalena Szydlowska\Projects\golem\golem\core\common.py", line 148, in func_wrapper
    return func(*args, **kwargs)
  File "C:\Users\Magdalena Szydlowska\Projects\golem\apps\rendering\task\framerenderingtask.py", line 87, in computation_failed
    self._update_task_preview()
  File "C:\Users\Magdalena Szydlowska\Projects\golem\apps\rendering\task\renderingtask.py", line 154, in _update_task_preview
    for sub in self.subtasks_given.values():
RuntimeError: dictionary changed size during iteration
```

mfranciszkiewicz:
Resolves #2630

### DIFF
--- a/apps/rendering/task/renderingtask.py
+++ b/apps/rendering/task/renderingtask.py
@@ -151,7 +151,8 @@ class RenderingTask(CoreTask):
         with handle_image_error(logger), \
                 self._open_preview() as img_task:
 
-            for sub in self.subtasks_given.values():
+            subtasks_given = dict(self.subtasks_given)
+            for sub in subtasks_given.values():
                 if SubtaskStatus.is_active(sub['status']):
                     self._mark_task_area(sub, img_task, sent_color)
                 if sub['status'] in [SubtaskStatus.failure,


apps/rendering/task/renderingtask.py
```diff
@@ -151,7 +151,8 @@ def _update_task_preview(self):
         with handle_image_error(logger), \
                 self._open_preview() as img_task:
 
-            for sub in self.subtasks_given.values():
+            subtasks_given = dict(self.subtasks_given)
+            for sub in subtasks_given.values():
```
Krigpl, line 155: I don't see how the dict could be altered in this loop.
Are you sure the error you're trying to fix isn't caused by a race condition?

mfranciszkiewicz: It's not altered in this loop, but elsewhere, probably in a callback from another thread.

Krigpl: So it's a race condition then. And making a shallow copy doesn't actually fix it - initial dict can still be altered while making the copy.
Protecting this object with mutex is a way to go.

mfranciszkiewicz: It doesn't - the callback should be either guarded with a mutex too or fired within the main thread. Either way, this fix produces a preview that reflects the state at the moment of calling.

Krigpl: > Either way, this fix produces a preview that reflects the state at the moment of calling.

Making a shallow copy of a dict is not an atomic operation, so this change doesn't really fix this race condition - only slightly reduces the probability of occurring.

mfranciszkiewicz: ~Shallow copying won’t cause a context switch.~ (edited: this one is wrong; still it does not matter at this point)
We’re not iterating over a dict itself here, but over a view of it’s items (with an iterator). Copying references to those items at one point will prevent the “size changed while iterating” from occurring.

mfranciszkiewicz: A mutex would solve this issue in another way - the dictionary would not change until this loop is done or the loop won't start until the dictionary is changed. There is no guarantee that mutex will help with producing the _actual_ state of the preview. And neither does this solution.

The real race condition here is in `img_task.save`

Krigpl: My only concern is that `c = dict(d)` doesn't have to an atomic operation. It happens that under CPython it is - but this is really implementation dependent and I believe it's not enforced by Python's standard, so it's not something we should rely on.
So yes, in CPython this in fact solves the issue. The same could be achieved using a mutex which would work regardless on Python's implementation.
I assume that the probability of anyone running Golem using a Python implementation where it isn't safe is negligible, so we can use this fix, but please put an explanatory comment in the code.

Krigpl (approved):
ihavesomesortofidea

etam (approved):

PR #2651, opened 2018-04-13
Monitor updates once every minute

Elfoniok:


golem/ethereum/paymentprocessor.py
```diff
@@ -379,7 +379,6 @@ def _run(self):
 
             self.monitor_progress()
             self.sendout()
-            self._send_balance_snapshot()
```
Wiezzel, line 382: Why? Isn't it already disabled by `send_payment_info` flag in monitor?

Elfoniok: Oh it is but only for mainnet, i didn't noticed this. But imho if we are not want this updates we should not even bother dispatcher with sending them.

Wiezzel: It's not that we don't want them. We don't want them now, but we want this to reconfigurable. Removing this line basically makes the flag pointless.

Elfoniok: Because this flag is in wrong place after all

Wiezzel: OK, so maybe just move the flag to payment processor?

Elfoniok: Maybe, its not the point of this change. This is a quick fix for monitor.

Wiezzel (approved):

jiivan (approved):

mfranciszkiewicz (approved):

PR #2661, opened 2018-04-16
Allow restarting subtasks after deadline 

### LINKED ISSUE #2605: Allow to restart parts of the task after deadline
Right now we're able to restart specific subtasks but only until while task doesn't hit deadline. We would like to add users possibility to restart specific subtask after deadline. To not mess up with TaskHeader deadline we would like to make a copy of the task (similiar to cloning whole task after it hits deadline), but only with specific subtasks to be restarted. 
It may be implemented as new task with only specified subtasks or as a copy of previous subtasks with cloned finished subtasks state. Second option seems like a better choice because it will allow to merge results easily (but we need to make sure that all the data are copied and duplicated task states are not sent to Concent). 

Core method for restarting tasks should allow user to specify set of subtasks that should be restarted.
We would like to also have a method that automatically restarts all the unfinished subtasks.

Wiezzel:
Command to use: `golemcli tasks restart_subtasks <task_id> <subtask1_id> <subtask2_id> ...`
Unit tests are not ready yet.
Manual testing should be performed by @ederenn.

Closes #2605

### DIFF
--- a/golem/interface/client/tasks.py
+++ b/golem/interface/client/tasks.py
@@ -23,6 +23,7 @@ class Tasks:
 
     id_req = Argument('id', help="Task identifier")
     id_opt = Argument.extend(id_req, optional=True)
+    subtask_ids = Argument('subtask_ids', vargs=True, help="Subtask ids")
 
     sort_task = Argument(
         '--sort',
@@ -106,6 +107,12 @@ class Tasks:
     @command(argument=id_req, help="Restart a task")
     def restart(self, id):
         deferred = Tasks.client.restart_task(id)
+        return sync_wait(deferred)
+
+    @command(arguments=(id_req, subtask_ids),
+             help="Restart given subtasks from a task")
+    def restart_subtasks(self, id, subtask_ids):
+        deferred = Tasks.client.restart_subtasks_from_task(id, subtask_ids)
         return sync_wait(deferred)
 
     @command(argument=id_req, help="Abort a task")
--- a/golem/interface/client/tasks.py
+++ b/golem/interface/client/tasks.py
@@ -1,3 +1,5 @@
+# pylint: disable=no-self-use,redefined-builtin
+
 from datetime import timedelta
 import json
 import re
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -1,8 +1,13 @@
+import copy
 import logging
+import os
 import pickle
+import shutil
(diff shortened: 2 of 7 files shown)


golem/task/taskmanager.py
```diff
@@ -378,6 +383,80 @@ def check_compute_task_def():
...
+
+            if old_subtask:  # Copy results from old subtask
+                old_subtask_id = old_subtask['subtask_id']
+
+                # TODO: Refactor this using package manager (?)
+                old_result_path = old_tmp_dir / '{}.{}.zip'.format(
+                    old_task_id, old_subtask_id)
+                new_result_path = new_tmp_dir / '{}.{}.zip'.format(
+                    new_task_id, new_subtask_id)
+                shutil.copy(old_result_path, new_result_path)
```
mfranciszkiewicz, line 431: Copying and extracting should be moved to a separate thread.

golem/interface/client/tasks.py
```diff
@@ -108,6 +111,12 @@ def restart(self, id):
         deferred = Tasks.client.restart_task(id)
         return sync_wait(deferred)
 
+    @command(arguments=(id_req, subtask_ids),
+             help="Restart given subtasks from a task")
+    def restart_subtasks(self, id, subtask_ids):
```
mfranciszkiewicz, line 116: Minor: `id` is a keyword, `task_id` seems good enough.

mfranciszkiewicz (approved):
Though IMO some parts should be refactored, this is a requested feature and will be merged for this release.

golem/interface/client/tasks.py
Wiezzel: It can't be changed, because of https://github.com/golemfactory/golem/blob/b0.15.1/golem/interface/client/tasks.py#L26
I would have to change it for all methods in this module.

golem/task/taskmanager.py
Wiezzel: Why should it?

mfranciszkiewicz: You are blocking the event loop for the time required to copy and extract (potentially) gigabytes of data.

Wiezzel: But the whole `copy_results` function is running in a separate thread. Not in the main one.

Wiezzel: OK.

PR #2665, opened 2018-04-17
React to ForceSubtaskResults

### LINKED ISSUE #2188: [Concent] UC3 / Force Accept - as a Requestor, send `ForceSubtaskResultsResponse`
implement sending of `ForceSubtaskResultsResponse` to the Concent service in reaction to `ForceSubtaskResults`

jiivan:
fixes: #2188

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ venv/*
 # dummy test temp files
 apps/dummy/test_tmp
 .python-version
+luxrender-output.png
--- a/golem/network/concent/received_handler.py
+++ b/golem/network/concent/received_handler.py
@@ -104,7 +104,7 @@ def on_subtask_results_settled(msg, **_):
     """End of UC3. Nothing can be done after this point.
     I'm either a Provider or Requestor
     """
-    logger.warning("[CONCENT] End of UC3 by %r", msg)
+    logger.warning("[CONCENT] End of Force Accept scenario by %r", msg)
 
 
 class TaskServerMessageHandler():
--- a/golem/network/concent/received_handler.py
+++ b/golem/network/concent/received_handler.py
@@ -99,6 +99,14 @@ def on_force_subtask_results_rejected(msg):
     logger.warning("[CONCENT] %r", msg)
 
 
+@library.register_handler(message.concents.SubtaskResultsSettled)
+def on_subtask_results_settled(msg, **_):
+    """End of UC3. Nothing can be done after this point.
+    I'm either a Provider or Requestor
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -76,3 +76,4 @@ venv/*
 # dummy test temp files
 apps/dummy/test_tmp
 .python-version
+luxrender-output.png
```
shadeofblue, line 79: I was wondering whether we shouldn't rather fix the test not to use a file in the source tree ;p

golem/network/concent/received_handler.py
```diff
@@ -99,6 +99,14 @@ def on_force_subtask_results_rejected(msg):
     logger.warning("[CONCENT] %r", msg)
 
 
+@library.register_handler(message.concents.SubtaskResultsSettled)
+def on_subtask_results_settled(msg, **_):
+    """End of UC3. Nothing can be done after this point.
+    I'm either a Provider or Requestor
+    """
+    logger.warning("[CONCENT] End of UC3 by %r", msg)
```
shadeofblue: @jiivan I wouldn't use `UC3` in debug messages as I think it's too "niche" -> maybe rather "end of the Force Accept scenario"?

golem/network/history.py
```diff
@@ -324,6 +325,24 @@ def add(msg: message.base.Message,
     )
 
 
+def get(message_class_name: str, task_id: str, subtask_id: str)\
```
shadeofblue, line 328: :+1:

golem/network/concent/received_handler.py
```diff
@@ -202,6 +210,33 @@ def on_force_report_computed_task(self, msg, **_):
...
         )
 
+    @handler_for(message.concents.ForceSubtaskResults)
+    def on_force_subtask_results(self, msg, **_):
+        """I'm a Requestor
+
+        Concent sends his own ForceSubtaskResults with AckReportComputedTask
+        provided by a provider.
+        """
+        sra = history.get('SubtaskResultsAccepted', msg.task_id, msg.subtask_id)
```
shadeofblue, line 220: I really think we should be looking up those messages also using the node id ...

shadeofblue (approved):
generally, okay - added comments

golem/network/concent/received_handler.py
jiivan: It's doable but I think we should also update all calls to `golem.task.tasksession.get_task_message()` too.  Created an issue #2670

PR #2667, opened 2018-04-17
golemcli res show fix + user friendliness

tworec:


golem/core/fileshelper.py
```diff
@@ -189,17 +189,15 @@ def du(path):
                  if an error occurs.
     """
     try:
-        size, _ = subprocess.check_output(['du', '-sh', path]).split()
-        unit = dict(K='kB', B='B').get(size[-1], str(size[-1]) + 'B')
-        return "{} {}".format(float(size[:-1]), unit)
+        logger.warning('duuu %r', path)
```
Krigpl: `warning`? `duuu`?

golem/client.py
```diff
@@ -989,9 +989,9 @@ def add_resource_peer(self, node_name, addr, port, key_id, node_info):
         )
 
     def get_res_dirs(self):
-        return {"computing": self.get_computed_files_dir(),
-                "received": self.get_received_files_dir(),
-                "distributed": self.get_distributed_files_dir()}
+        return {"total computed files": self.get_computed_files_dir(),
```
Krigpl: As far as I understand these aren't # of files but total size of them, so these labels are not accurate

golem/core/fileshelper.py
tworec: :) upps

golem/client.py
tworec: I was considering adding *size* word, but decided to skip it for brevity. This will be obvious when you see the output
```
$ golemcli res show
total computed files    total received files    total distributed files
----------------------  ----------------------  -------------------------
0B                      4.0K                    0B
```

maaktweluit (approved):
LGTM!

Also think size would make things more clear, but the headers are long enough as is

That warning would have made a nice debug line ;)

golem/client.py
tworec: changed `files` to `data` and removed `computed` column as it's point the same as `distributed`

Issue #2439, opened 2018-03-21
Authenticate stats that are sent to the monitor


lukasz-glen:
For the moment anyone can send any stats to the monitor, it can also impersonate someone.
Provide basic security for stats. Stats that are sent to the monitor should be signed by a node, the monitor should verify the signature before processing stats.

RDProjekt:
To do this we need to use nodejs packages secp256k1 and js-sha3 in monitor, is that ok ?

RDProjekt:
Also, while working on this issue we noticed that public key used as node ids is in uncompressed form(64 bytes), but compressed form is 33 bytes, so maybe node ids might be shorter ?

RDProjekt:
We heard that release without signing stats is planned, so how should monitor react on stats sended by older clients without a signature ?

PR #2672, opened 2018-04-17
golem-monitor #21: add pinging every constant interval

RDProjekt:
Changes related to periodically requesting port status check from the monitor:
- Add option PING_INTERVAL, which specifies ping interval in seconds
- Move waiting for ping response to SenderThread in monitor.py
- Add finish action to SenderThread
- Send logout on Golem shutdown

Note that for the ping requests to work, monitor needs to be updated as well.

This now includes #2439 as well. The commits are not related, but we need to push this code and it depends on golem-monitor #21.

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -488,7 +488,7 @@ class Client(HardwarePresetsMixin):
 
     def stop_monitor(self):
         logger.debug("Stopping monitor ...")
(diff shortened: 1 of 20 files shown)


mplebanski:
Can we close this?

PR #2676, opened 2018-04-18
clean-up after PR #2667

tworec:
In #2667 I've deleted `get_computed_files_dir` but missed to remove one of its usages. Fixing this.


Wiezzel (approved):
OK, but tests need to be fixed.

PR #2677, opened 2018-04-18
Resolved double docker env init

### LINKED ISSUE #2367: Docker env called twice?
When running debug logs just before release 0.13.0 i found docker env is called twice:
```
2018-03-06 15:30:22 DEBUG    golem.docker.manager                docker_machine_command_output: b'Running\n'
2018-03-06 15:30:22 DEBUG    golem.docker.manager                docker_machine_command: ['docker-machine', 'env', '--shell', 'cmd', 'golem']


...

2018-03-06 15:30:45 DEBUG    golem.docker.manager                docker_machine_command_output: b'Running\n'
2018-03-06 15:30:45 DEBUG    golem.docker.manager                docker_machine_command: ['docker-machine', 'env', '--shell', 'cmd', 'golem']
```

Investigate why this is the case and see if we can reduce this to 1, docker on windows and mac is not the most stable when thrown the same commands in a row.

maaktweluit:
Resolves #2367 

Moved check for `_env_checked` to inside the `check_environment()` to let external calls also adhere to it.

### DIFF
--- a/golem/docker/manager.py
+++ b/golem/docker/manager.py
@@ -87,7 +87,7 @@ class DockerManager(DockerConfigManager):
     @report_calls(Component.docker, 'instance.check')
     def check_environment(self):
         if self._env_checked:
-            return True
+            return bool(self.docker_machine)
 
         if is_windows():
             import pythoncom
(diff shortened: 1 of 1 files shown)


golem/docker/manager.py
```diff
@@ -86,6 +86,8 @@ def _get_hypervisor(self):
 
     @report_calls(Component.docker, 'instance.check')
     def check_environment(self):
+        if self._env_checked:
+            return True
```
mfranciszkiewicz: `return bool(self.docker_machine)` ?

mfranciszkiewicz (approved):

PR #2687, opened 2018-04-19
Concent/uc2 fixes

shadeofblue:


golem/network/concent/filetransfers.py
```diff
@@ -100,18 +105,28 @@ def _get_upload_uri(file_transfer_token: FileTransferToken):
     def _get_download_uri(file_transfer_token: FileTransferToken):
```
jiivan, line 105: Maybe using `urljoin()` could save us some future headache with missing slashes etc?
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urljoin

jiivan (approved):
lgtm

PR #2688, opened 2018-04-19
Return if the nonce was downloaded after a handshake timeout

mfranciszkiewicz:
Gets rid of a redundant error message appearing in logs.


maaktweluit (approved):
LGTM!

PR #2691, opened 2018-04-19
Display proper task duration

Wiezzel:
The value displayed as 'duration' has been being computed as timeout - estimated time to compute. Which makes absolutely no sense and has been changed to current time - start time.


mfranciszkiewicz (approved):

mfranciszkiewicz:
BTW duration is calculated as `timeout` - `remaining_time`, which is not supposed to be a constant value like estimated time to compute.

PR #2701, opened 2018-04-23
StatusPublisher: keep last status of all components

etam:
This allows a RPC client to connect at any time and know the status of all components.

This would probably require some changes on electron side. @mdtanrikulu


tests/golem/test_client.py
```diff
@@ -917,11 +917,11 @@ def make_mock_payment_processor(sci, eth=100, gnt=100):
 
 @patch('signal.signal')
 @patch('golem.network.p2p.node.Node.collect_network_info')
-class TestClientRPCMethods(TestWithDatabase, LogTestCase):
+class TestClientRPCMethods(TestDatabaseWithReactor, LogTestCase):
     def setUp(self):
-        super(TestClientRPCMethods, self).setUp()
+        TestDatabaseWithReactor.setUp(self)
```
Wiezzel: Why not simply `super().setUp()`?

tests/golem/test_client.py
```diff
@@ -955,6 +955,7 @@ def setUp(self):
 
     def tearDown(self):
         self.client.quit()
+        TestDatabaseWithReactor.tearDown(self)
```
Wiezzel: `super().tearDown()`?

tests/golem/test_client.py
```diff
@@ -1396,20 +1397,29 @@ def test_connection_status(self, *_):
...
-        assert not self.client.get_golem_status()
+    def test_golem_status_no_publisher(self, *_):
+        component = 'component'
+        status = 'method', 'stage', 'data'
 
         # status published, no rpc publisher
-        StatusPublisher.publish(*status)
-        assert not self.client.get_golem_status()
+        StatusPublisher.publish(component, *status)
+        assert self.client.get_golem_status()[component] == status
```
Wiezzel, line 1091: I'd encourage using `assertEqual()` in tests instead of plain `assert`. It makes possible to override the failure exception class and distinguish failed test assertion from other errors. Also running `python -O` disables all `asserts`. (See https://stackoverflow.com/a/2958450)

tests/golem/test_client.py
```diff
@@ -1396,20 +1397,29 @@ def test_connection_status(self, *_):
...
+        status = 'method', 'stage', 'data'
 
         # status published, with rpc publisher
         StatusPublisher._rpc_publisher = Mock()
-        StatusPublisher.publish(*status)
-        assert self.client.get_golem_status() == status
+        StatusPublisher.publish(component, *status)
+        assert self.client.get_golem_status()[component] == status
+
+        time.sleep(0.01)
```
Wiezzel: Kinda ugly. There is no guarantee that the code will complete in the given time. Better return a `Deferred` from `publish()` and use `twisted.trial.unittest.TestCase` for testing.

tests/golem/test_client.py
etam: Which base class would that call?

tests/golem/test_client.py
etam: `pytest` encourages using raw `assert` https://docs.pytest.org/en/latest/assert.html

tests/golem/test_client.py
etam: Done (I hope correctly, please check carefully, because I'm not very familiar with `Deferred`)

tests/golem/test_client.py
Wiezzel: I have a different opinion but, whatever, that's not an issue.

tests/golem/test_client.py
Wiezzel: `TestDatabaseWithReactor` because it's the first superclass.

Wiezzel:
Looks like `tests.golem.rpc.test_router.TestRouter.test_rpc` got broken. :disappointed: 
Maybe @mfranciszkiewicz could help you with deferreds since I am not expert either.

tests/golem/test_client.py
etam: OK, I educated myself a bit on that matter, and I agree that's the proper way, and luckily base classes follows the rules to make this work properly.

tests/golem/test_client.py
shadeofblue: hmm... but doesn't it mean that we're making ourselves more dependent on a single test tool? I'd also be in favor of using the assert methods rather than raw asserts? ...

etam:
note to me: Windows tests are constantly failing on tearDown. This needs fixing.

mdtanrikulu:
It's breaking core status + behavioral mechanism in electron. If we going to implement this, there should be some more changes. 

The first thing I observed is; 
there is multiple statuses in returned object. So If I make electron adapted to the new object style, it will  try to switch both statuses in every seconds. 

for exp;
```json
[
   client{
       0: "get_password",
       1: "pre",
       2: null
   },
   docker{
       0: "instance.check",
       1: "post",
       2: null
   }
]
```

etam:
The windows tests are failing. #3031 should fix that

etam:
@mdtanrikulu The reason for this change is, if you connect to golem's rpc somewhere while it's already working, you don't know the state of all components. With this change you know them all.

If you save the results from the previous call, you can compare them and see what has changed.

mdtanrikulu:
@etam the issue is not the idea, the structure of the data I think. I have only one multi <key, value> object into an array. To get the changes from the object all the time I need to make deep object diff. I think it should not be that hard to know latest changes. Cannot we return an array with pushed states objects into? So the last element or the first can be the most recent state of app?

shadeofblue:
needs discussion between @mdtanrikulu and @etam

mdtanrikulu (approved):

mfranciszkiewicz:
What's keeping this PR away from being merged?

mdtanrikulu:
@mfranciszkiewicz Nothing, merging it.

PR #2702, opened 2018-04-23
Client.connection_status: send raw data

### LINKED ISSUE #3213: builtins.AttributeError: 'NoneType' object has no attribute 'cur_port' in golem/client.py
```
2018-08-22 21:33:47 ERROR    twisted                             AttributeError: 'NoneType' object has no attribute 'cur_port': Traceback (most recent call last):
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/autobahn/twisted/websocket.py", line 162, in _onMessage
    self.onMessage(payload, isBinary)
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/autobahn/wamp/websocket.py", line 95, in onMessage
    self._session.onMessage(msg)
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/autobahn/wamp/protocol.py", line 895, in onMessage
    on_reply = txaio.as_future(endpoint.fn, *invoke_args, **invoke_kwargs)
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/txaio/tx.py", line 417, in as_future
    return maybeDeferred(fun, *args, **kwargs)
--- <exception caught here> ---
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/home/blue/PycharmProjects/golem/golem/client.py", line 1386, in connection_status
    listen_port = self.get_p2p_port()
  File "/home/blue/PycharmProjects/golem/golem/client.py", line 947, in get_p2p_port
    return self.p2pservice.cur_port
builtins.AttributeError: 'NoneType' object has no attribute 'cur_port'
```

etam:
This allows a RPC client to understand the network status.

~~This should not have any impact on electron. @mdtanrikulu please confirm.~~

Fixes #3213

depends on https://github.com/golemfactory/golem-electron/pull/396


golem/interface/client/network.py
```diff
@@ -28,8 +28,29 @@ class Network(object):
...
+        if not status['listening']:
+            return "Application not listening, check config file."
+
+        messages = []
+
+        if status['port_statuses']:
+            port_statuses = ", ".join(
+                "{}: {}".format(port, port_status)
+                for port, port_status in self.node.port_statuses.items())
+            messages.append("Port {}.".format(port_statuses))
```
tworec: `Port` --> `Port(s):`

tests/golem/test_client.py
```diff
@@ -1385,13 +1383,11 @@ def test_connection_status(self, *_):
...
         # status with peers
-        self.assertTrue(c.connection_status().startswith("Connected"))
+        assert c.connection_status()['connected']
 
         # status without ports
         c.p2pservice.cur_port = 0
-        self.assertTrue(
-            c.connection_status().startswith("Application not listening")
-        )
+        assert not c.connection_status()['listening']
```
tworec: add missing test cases:
- unknown
- Application not listening, check config file
- Port(s): [list]
- Connected / Not Connected .. check

tworec (changes requested):

tests/golem/test_client.py
etam: Done, but in different file.

golem/client.py
```diff
@@ -1170,24 +1174,16 @@ def connection_status(self):
...
-            return "Application not listening, check config file."
-
-        messages = []
-
-        if self.node.port_statuses:
-            status = ", ".join(
-                "{}: {}".format(port, status)
-                for port, status in self.node.port_statuses.items())
-            messages.append("Port {}.".format(status))
+        status = dict()
```
tworec: You have changed return type and value for `connection_status` and corrected its usage in `golem/interface/client/network.py` but it is used also in `NetworkConnectionPublisherService` see line 1337

https://github.com/golemfactory/golem/pull/2702/files#diff-9f351727974c5655127c6a4dc0671995R1337

I don't know the purpose of this class, but have you considered and checked consequences of your change?

tests/golem/interface/test_client_commands.py
```diff
@@ -162,23 +162,85 @@ def setUpClass(cls):
...
+                'listening': False
+            }
+
+            # when
+            result = Network().status()
 
+            # then
+            assert self.client.connection_status.called
+            assert isinstance(result, str)
+            assert "not listening" in result
```
tworec: why not use exact full message which we are actually expecting and no such heuristics, where you are checking for some part is in, and two otheres are not? This is not easy readable.

tests/golem/test_client.py
tworec: ok, cool.
here you should test also `connection_status()['port_statuses']`  because you've added such a property

tworec (changes requested):

golem/client.py
etam: This class is periodically sending connection status as an event on rpc (on crossbar you can call a method over rpc, but you can also publish events).

So, yes, I know that my change affects both 'net.status' rpc method and 'evt.net.connection' event.

tests/golem/interface/test_client_commands.py
etam: To check only for important parts and not make it strictly bound to output formatting and other stuff that might appear. If someone later adds some more info to the output it won't affect the test.

tests/golem/test_client.py
etam: added

tests/golem/interface/test_client_commands.py
tworec: do not like it, but lets go on

tworec (approved):

etam:
RPC change: method `net.status` and event `evt.net.connection` (`Client.connection_status`)  is changed from string to a dictionary structure.

If golem is not listening (p2p or task port is not open)
```python
{
    'listening': False
}
```
else:
```python
{
    'listening': True,
    'connected': bool(has any connected peers),
    'port_statuses': {
        1234: "open",
        2345: "unreachable",
    }
}
```

`port_statuses` is filled on receiving data from 'ping me' service. So it might be empty shortly after starting golem.
Possible values (from https://github.com/golemfactory/golem-monitor/blob/f731eb03735f615b9a64ebcf2699874a6b113ae3/models/golemstatsmodel.js#L263):
- 'unreachable'
- 'timeout'
- 'open'

mdtanrikulu (approved):

etam:
TODO: pass the old message with the raw data

etam:
> TODO: pass the old message with the raw data

Done

PR #2716, opened 2018-04-25
client: publish Task.evt_subtask_status if subtask was changed

etam:
Currently all changes in tasks and subtasks are reported as `Task.evt_task_status` annotated with a task_id.

We have in our RPC mapping defined a `Task.evt_subtask_status`, which is unused.

This change restores reporting subtasks changes using `Task.evt_subtask_status` annotated with a task_id and subtask_id.

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -245,12 +245,17 @@ class Client(HardwarePresetsMixin):
             sender, signal, event, kwargs
         )
 
-        if 'op' in kwargs and kwargs['op'] is not None \
-                and kwargs['op'].subtask_related():
+        op = kwargs['op'] if 'op' in kwargs else None
+
+        if op is not None and op.subtask_related():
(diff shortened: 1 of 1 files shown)


etam:
And now we're also publishing an "Op" that indicate what happened.

mdtanrikulu (approved):

etam:
RPC changes: 
Before: Send only `evt.comp.task.status` with `task_id` no matter if there was change in task or subtask

After:
If there was a change in task, send `evt.comp.task.status` with:
- `task_id` 
- Op class name:
  - `TaskOp`
  - `OtherOp`
- Op value

If there was a change is subtask, send `evt.comp.subtask.status` with:
- `task_id`
- `subtask_id`
- `SubtaskOp` value

Ops definitons: https://github.com/golemfactory/golem/blob/cc069c7e346b409101fbf37f68215aa5e24f366c/golem/task/taskstate.py#L147-L198

PR #2717, opened 2018-04-25
RPC method to add nodes to block list

Wiezzel:
RPC method name: `net.peer.block`
Parameters: `node_id: str` - ID of the node to be blocked
Return value: `(success: bool, error: str)`
 * `success` - was the operation successful
 * `error` - error message (`None` if there was no error)

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -151,7 +151,7 @@ class Client(HardwarePresetsMixin):
             keys_auth=self.keys_auth,
         )
 
-        self.task_server = None
+        self.task_server: Optional[TaskServer] = None
         self.port_mapper = None
 
         self.nodes_manager_client = None
@@ -1292,6 +1292,17 @@ class Client(HardwarePresetsMixin):
     @staticmethod
     def enable_talkback(value):
         enable_sentry_logger(value)
+
+    def block_node(self, node_id: str) -> Tuple[bool, Optional[str]]:
+        if self.task_server is not None:
+            try:
+                self.task_server.acl.disallow(node_id, persist=True)
+                return True, None
+
+            except Exception as e:  # pylint: disable=broad-except
+                return False, str(e)
+
+        return False, 'Client is not ready'
 
 
 class DoWorkService(LoopingCallService):
--- a/golem/task/acl.py
+++ b/golem/task/acl.py
@@ -15,14 +15,16 @@ class Acl(abc.ABC):
         pass
 
     @abc.abstractmethod
-    def disallow(self, node_id: str, timeout_seconds: int) -> None:
+    def disallow(self, node_id: str, timeout_seconds: int, persist: bool) \
+            -> None:
         pass
 
 
 class _DenyAcl(Acl):
-    def __init__(self, deny_coll: Iterable[str]) -> None:
+    def __init__(self, deny_coll: Iterable[str], list_path: Path) -> None:
         key_sequence = list(deny_coll)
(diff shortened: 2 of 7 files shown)


golem/client.py
```diff
@@ -1245,6 +1245,10 @@ def _unlock_datadir(self):
     def enable_talkback(value):
         enable_sentry_logger(value)
 
+    def block_node(self, node_id):
+        assert isinstance(self.task_server, TaskServer)
```
etam: This is too bold. Change it to return `True` on success and add printing an error message in golemcli on error.

golem/task/acl.py
```diff
@@ -64,11 +80,18 @@ def _read_set_from_file(path: Path) -> Set[str]:
         return set()
 
 
+def _write_set_to_file(path: Path, node_set: Set[str]):
+    with path.open('w') as f:
+        for node_id in node_set:
```
etam: Suggestion: `in sorted(node_set)`. Otherwise it'll mix the contents every time.

etam (changes requested):
so close

golem/client.py
Wiezzel: You mean something like this?
```python
try:
    # here goes all code
    return True, None
except Exception e:
    return False, e
```

etam: `return False, str(e)` in last line.

But also instead of `isinstance`, it would be more straightforward to use `if self.task_server is not None:` (it's `None` until `start_network` is called for the first time).

Wiezzel: `str(e)` for sure.
But I'll keep the type assertion. It enables type-checking. Of course it would be better to declare this field as of `TaskServer` type, but that would require a bigger refactoring.

etam: The type of `task_server` is `Optional[TaskServer]`. No big refactoring needed, no type checking needed.

Wiezzel: Is it? I don't see any type declaration here: https://github.com/golemfactory/golem/blob/develop/golem/client.py#L156

Wiezzel: I get nice hints from IDE when using type assertion:
image
But not with not-none check:
image

etam: > Is it? I don't see any type declaration here

You can add it.

Lots of our code was originally wrote for python 2. Don't expect to see type annotations everywhere. If there are none, then it doesn't mean "it can be type T, but it can be also completely different type U". In most cases it means "it's type T, or Optional[T]".

Wiezzel: OK, fixed.

golem/task/acl.py
Wiezzel: Fixed.

tests/golem/interface/test_client_commands.py
```diff
@@ -162,6 +162,9 @@ def setUpClass(cls):
         cls.n_clients = len(peer_info)
         cls.client = client
 
+    def tearDown(self):
+        self.client.reset_mock()
```
etam, line 166: Wouldn't that reset everything that was prepared in `setUpClass`? If this is required, then it seems reasonable to create a proper `setUp` method.

Wiezzel: Nope. It just resets calls to mocked methods. ([sauce](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.reset_mock))

etam (approved):
:+1:

mfranciszkiewicz (approved):

PR #2718, opened 2018-04-25
Enable concent by default on testnet

### LINKED ISSUE #2666: Enable concent by default on testnet
we need to set concent urls in Golem client cmdline params
 1. `None`
 1. `dev.`
 1. `staging.`
 1. `test.`
 1. `main.`

jiivan:
fixes: #2666


shadeofblue (approved):
seems okay

shadeofblue (changes requested):
actually, I think it would be better, instead of passing the of the variant name around, pass the resultant configuration dictionary ... it seems somewhat artificial that each component needs to know that it has to retrieve the config from the choices dictionary...

golem/client.py
```diff
@@ -97,7 +97,8 @@ def __init__(  # noqa pylint: disable=too-many-arguments
             connect_to_known_hosts: bool = True,
             use_docker_manager: bool = True,
             use_monitor: bool = True,
-            use_concent: bool = False,
+            # SEE: golem.core.variables.CONCENT_CHOICES
+            concent_variant_name: str = 'disabled',
```
shadeofblue: so, I believe it should be just `concent_variant` here, the default value of which, should be `golem.core.variables.CONCENT_CHOICES.get('disabled')`

golem/client.py
```diff
@@ -138,13 +139,13 @@ def __init__(  # noqa pylint: disable=too-many-arguments
 
         self.p2pservice = None
         self.diag_service = None
-        self.use_concent = use_concent
         self.concent_service = ConcentClientService(
-            enabled=self.use_concent,
+            variant_name=concent_variant_name,
```
shadeofblue: just `variant`

golem/client.py
```diff
@@ -138,13 +139,13 @@ def __init__(  # noqa pylint: disable=too-many-arguments
...
         self.diag_service = None
-        self.use_concent = use_concent
         self.concent_service = ConcentClientService(
-            enabled=self.use_concent,
+            variant_name=concent_variant_name,
             keys_auth=self.keys_auth,
         )
         self.concent_filetransfers = ConcentFiletransferService(
             keys_auth=self.keys_auth,
+            variant_name=concent_variant_name,
```
shadeofblue: just `variant`

golem/network/concent/client.py
```diff
@@ -64,7 +64,8 @@ def verify_response(response: requests.Response) -> None:
 
 def send_to_concent(
         msg: message.Message,
-        signing_key) -> typing.Optional[bytes]:
+        signing_key,
+        variant: dict) -> typing.Optional[bytes]:
```
shadeofblue: and I think it would be good to call this parameter `concent_variant` so that it's perfectly clear the variant of _what_ we're referring to...

golem/network/concent/client.py
```diff
@@ -170,13 +174,12 @@ class ConcentClientService(threading.Thread):
     MAX_GRACE_TIME = 5 * 60  # s
     GRACE_FACTOR = 2  # n times on each failure
 
-    def __init__(self, keys_auth: keysauth.KeysAuth, enabled=True):
+    def __init__(self, keys_auth: keysauth.KeysAuth, variant_name='staging'):
```
shadeofblue: btw, why does it default to `staging` here all of a sudden?

golem/network/concent/client.py
```diff
@@ -185,6 +188,10 @@ def __init__(self, keys_auth: keysauth.KeysAuth, enabled=True):
         self._delayed = dict()
         self.received_messages: queue.Queue = queue.Queue(maxsize=100)
 
+    @property
+    def enabled(self):
+        return None not in self.variant.values()
```
shadeofblue, line 192: wouldn't it be more future-proof to check just e.g. the URL? I imagine we might add some setting later on that wouldn't be crucial but could be `None` as a default and it would fail this check...

golem/network/concent/filetransfers.py
```diff
@@ -43,7 +43,9 @@ class ConcentFiletransferService(LoopingCallService):
 
     def __init__(self,
                  keys_auth: keysauth.KeysAuth,
-                 interval_seconds: int = 1) -> None:
+                 interval_seconds: int = 1,
+                 variant_name: str = 'dev') -> None:
```
shadeofblue: and again, why a different default yet here?

golem/network/concent/client.py
jiivan: because before this change `enabled=True` meant `staging`. :)

shadeofblue (approved):

PR #2727, opened 2018-04-25
Respect block gas limit

### LINKED ISSUE #2699: Respect block gas limit when sending batch transfer
We have to limit # of recipients in batchTransfer so the total gas cost won't exceed gas block limit.

Krigpl:
So we don't try to send a transaction with higher gas limit then the block gas limit - it wouldn't get accepted to the pool at all.

Fixes #2699

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -52,6 +52,8 @@ class PaymentProcessor(LoopingCallService):
     REQUIRED_CONFIRMATIONS = 12
 
     CLOSURE_TIME_DELAY = 2
+    # Don't try to use more than 75% of block gas limit
+    BLOCK_GAS_LIMIT_RATIO = 0.75
 
     def __init__(self,
                  sci,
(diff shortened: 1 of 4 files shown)


golem/ethereum/paymentprocessor.py
```diff
@@ -194,13 +196,24 @@ def __get_next_batch(
         eth_balance = eth_balance - self.ETH_BATCH_PAYMENT_BASE
         ind = 0
         eth_per_payment = self.get_gas_cost_per_payment()
+        gas_limit = \
+            self._sci.get_latest_block().gas_limit * self.BLOCK_GAS_LIMIT_RATIO
+        payees = set()
```
mfranciszkiewicz, line 201: Are payments per payee summed?

Krigpl: Yes. Inside `sci.batch_transfer`

mfranciszkiewicz (approved):

golem/ethereum/paymentprocessor.py
```diff
@@ -194,13 +196,24 @@ def __get_next_batch(
...
         for p in payments:
             if p.processed_ts > closure_time:
                 break
             gntb_balance -= p.value
-            eth_balance -= eth_per_payment
```
maaktweluit, line 201: no more need for updating the ETH balance?

maaktweluit (approved):
LGTM!

One small q, reply can be enough

golem/ethereum/paymentprocessor.py
Krigpl: Yeah, previous estimation wasn't accounting for multiple payments to a single provider, current version does so the check is performed in a different way.

PR #2734, opened 2018-04-26
requestor to check performance index before work offer acceptance

### LINKED ISSUE #2771: Add minimal performance option in GUI
The aim of Golem is to speed up computation therefore requestors may only want to choose providers with performance equal or above their machine’s performance. 
We are able to set performance by hand from CLI but it would be good to add it nicely presented in GUI. 

What is required:
* Defining if we want to set this option per task or as a general config - all
* Design of convenient way for users to set this option. Users probably should be able to check their own performance and edit required performance accordingly to their own benchmark result. @jamuszyn
* Implementation of the designed GUI @mdtanrikulu
* Implementation of proper RPC calls (if one that are right now are not adequate to chosen method). @tworec

tworec:
initially this change was in favour of @viggith 
on requestor side it adds a check if
```
provider performance > minimal acceptd requestor performance
````
If not, such provider will rejected, and:

```
minimal accepted performance = min accepted step * performance multiplier
```
where `min accepted step` value is stored per environment (use case) and by default is `300`.
`performance multiplier` is global, `0` by default and can be manipulated via RPC. Accepted values are within [0, 100]. `golemcli` options for getting/setting it were added:

```
$ golemcli envs
...
    perf_mult           Gets accepted performance multiplier
    perf_mult_set       Sets accepted performance multiplier
```

fixes #2771 

btw. I also cleared code where providers are checked by requestor

### CODE REVIEW

File: golem/task/taskserver.py [hunk 1]
```diff
@@ -578,7 +578,40 @@ def add_forwarded_session_request(self, key_id, conn_id):
...
+            max_resource_size,
+            max_memory_size,
+            num_cores):  # noqa pylint: disable=unused-argument
+
+        task = self.task_manager.tasks[task_id]
+
+        env = self.get_environment_by_id(task.header.environment)
+        performance = env.get_performance() / 2
+        if performance > int(perf_index):
+            logger.info('insufficient provider performance index: '
```
mfranciszkiewicz: Could you also add `node_name`, `node_id` and `task_id` to these messages? That would provide clear information as of who and why.

golem/task/taskmanager.py
```diff
@@ -321,26 +321,15 @@ def get_next_subtask(
         if task.header.max_price < price:
             return None, False, False
 
-        def has_subtasks():
+        def needs_computation():
             if self.tasks_states[task_id].status not in self.activeStatus:
-                logger.debug('state no in activestatus')
+                logger.info('task is not active')
```
mfranciszkiewicz: Please see my other comment regarding logs

golem/task/taskserver.py
```diff
@@ -578,7 +578,40 @@ def add_forwarded_session_request(self, key_id, conn_id):
...
+                        f'{max_resource_size} KiB')
+            return False
+
+        if task.header.estimated_memory > (int(max_memory_size) * 1024):
+            logger.info('insufficient provider memory size: '
+                        f'{max_memory_size} KiB')
+            return False
+
+        # TODO: where to find num cores needed by task
+        # if task.header.num_cores > num_cores:
```
mfranciszkiewicz: Unfortunately, currently there is no such parameter.

golem/task/taskserver.py
tworec: good point

mfranciszkiewicz (dismissed):

mfranciszkiewicz:
It looks like more tests need updating.

golem/task/taskmanager.py
```diff
@@ -321,26 +321,16 @@ def get_next_subtask(
...
+            ids = f'provider: {node_name} - {node_id}, task_id: {task_id}'
             if self.tasks_states[task_id].status not in self.activeStatus:
-                logger.debug('state no in activestatus')
+                logger.info(f'task is not active; {ids}')
                 return False
             if not task.needs_computation():
-                logger.debug('not task.needs_computation')
+                logger.info(f'no more computation needed; {ids}')
                 return False
-            if task.header.resource_size > (int(max_resource_size) * 1024):
```
Krigpl: Why do you remove these checks? They are still useful so the provider doesn't even bother requestor unnecessarily.

mfranciszkiewicz: @Krigpl These checks have been moved to `should_accept_provider`

Krigpl: I know, I'm saying that they are still useful on the provider side.

Krigpl: Oh, this is still requestor side. So my point is that these checks could be beneficial on the provider side as well.

Krigpl: Also, `max_resource_size` and `max_memory_size` are no longer needed in this method.
I'd also look into why `estimated_performance` and `num_cores` is needed

tworec: @Krigpl thanks for feedback

tworec:
@mfranciszkiewicz @badb if we want this to be in next release please merge it.

badb:
@tworec Can you make PR to develop with this change?

golem/task/taskserver.py
```diff
@@ -578,12 +578,51 @@ def add_forwarded_session_request(self, key_id, conn_id):
...
+
+        if task_id not in self.task_manager.tasks:
+            logger.info(f'Cannot find task in my tasks: {ids()}')
+            return False
+
+        task = self.task_manager.tasks[task_id]
+        env = self.get_environment_by_id(task.header.environment)
+        requestor_perf = env.get_performance()
+
+        if requestor_perf > int(provider_perf) * int(num_cores):
```
badb: Why are you multiplying it by number of cores? It's not a performance per core, but general performance for the assigned amount of cores. This multiplication creates extremely uneven advantage for people with many cores assigned.

badb (changes requested):
Check the comment

golem/task/taskserver.py
tworec: it was intentional, as we discussed. But will change, as Blender benchmark uses multi-core

golem/database/schemas/015_schema.py
```diff
@@ -0,0 +1,36 @@
...
+
+    migrator.add_index('performance', 'environment_id', unique=True)
+
+
+def rollback(migrator, *_, **__):
+    """Write your rollback migrations here."""
+
+    migrator.remove_fields('performance', 'min_accepted')
+
+    migrator.drop_index('performance', 'environment_id')
```
mfranciszkiewicz: `drop` + `add` (rebuilding) indexes isn't harmful and may even be desirable in the current state. We should implement some other machinery for that purpose, though. Could you remove these lines ?

golem/rpc/mapping/rpcmethodnames.py
```diff
@@ -83,6 +83,8 @@
     disable_environment=    'comp.environment.disable',
     run_benchmark=          'comp.environment.benchmark',
     get_performance_values= 'comp.environment.performance',
+    get_min_performance=    'comp.environment.performance.min.get',
```
mfranciszkiewicz: Can i haz (in a former fashion) `comp.environment.performance.min` as `get` and `comp.environment.performance.min.update` for updating? Or does it cascade changes in GUI?

golem/database/schemas/015_schema.py
```diff
@@ -0,0 +1,36 @@
+import peewee as pw
+
+SCHEMA_VERSION = 15
+
+
+def migrate(migrator, *_, **__):
```
mfranciszkiewicz, line 8: We've talked about this (semi) offline; could you create an issue for `setup.py develop migration` + data directory changes?

golem/client.py
```diff
@@ -1200,6 +1204,22 @@ def get_estimated_cost(self, task_type, options):
     def get_performance_values(self):
         return self.environments_manager.get_performance_values()
 
+    # TODO: hardcoded Blender just for 0.16.0
+    @inlineCallbacks
+    def get_min_performance(self, env_id: str = BlenderEnvironment.get_id()) \
+            -> Generator[Deferred, float, float]:
+        env = self.environments_manager.get_environment_by_id(env_id)
+        result = yield defer.execute(env.get_min_accepted_performance)
```
mfranciszkiewicz: `yield` (+ `inlineCallbacks`) isn't required here, since `get_min_accepted_performance` is 100% synchronous. A straightforward call should suffice.

golem/client.py
```diff
@@ -1200,6 +1204,22 @@ def get_estimated_cost(self, task_type, options):
...
+        env = self.environments_manager.get_environment_by_id(env_id)
+        result = yield defer.execute(env.get_min_accepted_performance)
+        return result
+
+    # TODO: hardcoded Blender just for 0.16.0
+    @inlineCallbacks
+    def set_min_performance(self, min_accepted_performance: float,
+                            env_id: str = BlenderEnvironment.get_id()):
+        env = self.environments_manager.get_environment_by_id(env_id)
+        yield defer.execute(env.set_min_accepted_performance,
```
mfranciszkiewicz: Same as above

mfranciszkiewicz (commented):
Some details; almost approved

golem/client.py
```diff
@@ -1200,6 +1204,22 @@ def get_estimated_cost(self, task_type, options):
     def get_performance_values(self):
         return self.environments_manager.get_performance_values()
 
+    # TODO: hardcoded Blender just for 0.16.0
```
badb: Can you create an issue for this and attach its nubmer to this TODO comment?

golem/database/schemas/015_schema.py
tworec: ok

golem/rpc/mapping/rpcmethodnames.py
tworec: we need to align it with @mdtanrikulu No problem from my side

golem/database/schemas/015_schema.py
tworec: will do

golem/client.py
tworec: As we discussed, I've added it just for catching exceptions, because without `yield` and `inlineCallbacks` it was'nt working for me. But now seems to work ;) It was too late yesterday...

golem/client.py
tworec: since we've decided to do it more generic (see https://github.com/golemfactory/golem/issues/2771#issuecomment-391344430) this hardcode will be removed

golem/rpc/mapping/rpcmethodnames.py
tworec: @mdtanrikulu agreed, so I've changed. Please check

mfranciszkiewicz (approved):

PR #2738, opened 2018-04-27
Remove unused ethereum.password module

Krigpl:
I believe this is unused, let's see if CI confirms.


etam (approved):
Searching for `get_saved_password` gives no results (other than the test you're also removing).

I'll give it a yes.

Krigpl:
>I'll give it a yes.

I'd rather have some gif, but fine.

PR #2743, opened 2018-04-27
Return locks in fundslocker

badb:


Krigpl (approved):
thumbupglitch

maaktweluit (approved):
LGTM!

Is this now also available in a GUI call?

badb:
Do not merge before @mdtanrikulu accepts!

golem/client.py
```diff
@@ -924,10 +924,11 @@ def get_balance(self):
...
             last_gnt_update, \
             last_eth_update = yield self.transaction_system.get_balance()
+        gnt_lock, eth_lock = self.funds_locker.sum_locks()
         if gnt is not None:
-            return str(gnt), str(av_gnt), str(eth), str(
-                last_gnt_update), str(last_eth_update)
-        return None, None, None, None, None
+            return str(gnt), str(av_gnt), str(eth), str(gnt_lock), \
+                   str(eth_lock), str(last_gnt_update), str(last_eth_update)
+        return None, None, None, None, None, None, None
```
mfranciszkiewicz: Uhm, since we now have a 7-element tuple, maybe it's a good idea to format the return value as (e.g). a `dict`? And get rid of all `"None"` strings, simply return `None`s.

badb:
@mdtanrikulu Call result changed to dict, check if you are ok with this change.

golem/client.py
```diff
@@ -921,13 +922,22 @@ def get_task_computer_stat(self, name):
 
     @inlineCallbacks
     def get_balance(self):
+        if not self.transaction_system:
```
Krigpl: This isn't needed.

golem/client.py
```diff
@@ -511,7 +511,8 @@ def quit(self):
         logger.info('Shutting down ...')
         self.stop()
 
-        self.transaction_system.stop()
+        if self.transaction_system:
```
Krigpl: Not needed.

badb: It was neede to pass a test that was previously broken. Previous test for this function was written in a way that always returned True (even though it should fail). So when I fixed it I actually find out that part of the test was failing. I can remove this tests if we're sure that we don't want to accept situations when transaction system is disabled / not working properly.

Krigpl: Which test is it ?

badb: Test get_balance in test_client lines 1144-1146

Krigpl: Yeah, just remove these lines, let's adjust test code to the production code, not the other way around and currently there's no production scenario where `transaction_system` is not set.

mdtanrikulu (approved):

PR #2750, opened 2018-05-03
Allow HTTPS scheme for --geth-address

gcarq:
This PR modifies `argparser.parse_http_addr` to accept https urls, that allows users to pass a custom geth node which is behind a https reverse proxy.

### DIFF
--- a/golem/argsparser.py
+++ b/golem/argsparser.py
@@ -1,4 +1,5 @@
 import ipaddress
+from urllib.parse import urlparse
 
 import click
 
@@ -18,14 +19,16 @@ def parse_http_addr(ctx, param, value):
     del ctx, param
     if value:
         try:
-            http_prefix = 'http://'
-            if not value.startswith(http_prefix):
+            parsed_url = urlparse(value)
+            if parsed_url.scheme not in ['http', 'https']:
                 raise click.BadParameter(
-                    "Address without http:// prefix"
+                    "Address without http(s):// prefix "
                     "specified: {}".format(value))
-            SocketAddress.parse(value[len(http_prefix):])
+            SocketAddress.parse(
+                value.replace(parsed_url.scheme + '://', '')
+            )
             return value
-        except ipaddress.AddressValueError as e:
+        except (ValueError, ipaddress.AddressValueError) as e:
             raise click.BadParameter(
                 "Invalid network address specified: {}".format(e))
     return None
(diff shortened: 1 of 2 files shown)


golem/argsparser.py
```diff
@@ -18,14 +19,16 @@ def parse_http_addr(ctx, param, value):
...
-            if not value.startswith(http_prefix):
+            parsed_url = urlparse(value)
+            if parsed_url.scheme not in ['http', 'https']:
                 raise click.BadParameter(
-                    "Address without http:// prefix"
+                    "Address without http(s):// prefix "
                     "specified: {}".format(value))
-            SocketAddress.parse(value[len(http_prefix):])
+            SocketAddress.parse(
+                value.replace(parsed_url.scheme + '://', '')
```
maaktweluit, line 28: I liked the `val[len(x):]` more then a string replace, not sure what is more efficient / readable

maaktweluit (approved):
LGTM!

One small comment, can be replied to only so i approve.

tests/golem/test_opt_node.py
```diff
@@ -120,7 +120,7 @@ def test_missing_node_address_should_fail(self, *_):
 
     @patch('twisted.internet.reactor', create=True)
     @patch('golemapp.Node')
-    def test_geth_address_should_be_passed_to_node(self, mock_node, *_):
```
Krigpl, line 123: Let's not change that. This test checks whether the `geth_address` argument is correctly passed down the stack, not whether the argument itself is correct.
So instead of this and adding the other test `test_geth_https_address_should_be_passed_to_node` it should remain as it is and you may add a test for the `parse_http_addr` itself. I see that the `tests/test_argsparser.py` file is missing so feel free to add it and write a test checking whether it correctly parses both http and https.

tests/golem/test_opt_node.py
```diff
@@ -142,8 +142,33 @@ def test_geth_address_should_be_passed_to_node(self, mock_node, *_):
...
+                                     start_geth=False,
+                                     start_geth_port=None,
+                                     concent_variant=variables.CONCENT_CHOICES[
+                                         'test'
+                                     ],
+                                     use_monitor=True,
+                                     password=None)
+
     @patch('golem.node.Client')
-    def test_geth_address_should_be_passed_to_client(self, mock_client, *_):
```
Krigpl, line 146: Same as above, do not change it.

Krigpl (changes requested):
Thank you for your contribution. The change in the parser code looks good but let's improve the tests, see inline comments.

gcarq:
Thanks for the feedback, I will change that and solve the file conflicts

Krigpl:
@gcarq Please feel free to reopen if you decide to pick this up again.

PR #2753, opened 2018-05-08
Verifier module

Elfoniok:


etam (changes requested):
- task_computer.data is added but seems not needed
- imports from `golem_verificator` should be grouped separately from `golem` and `apps`.

etam (approved):
ok

jiivan (approved):

PR #2759, opened 2018-05-09
Concent negotiation

### LINKED ISSUE #2752: Concent negotiation
Concent should be enabled in the TaskToCompute if and only if both parties have Concent service enabled.

If Provider has Concent enabled and Requestor has Concent disabled, then Requestor should not send TaskToCompute at all - there is no agreement, negotiation fails.

If Provider has Concent enabled and Requestor has Concent disabled and Requestor sent TaskToCompute with Concent distabled, they should reject any TTC that has concent disabled (CannotComputeTask).

jiivan:
fixes: #2752

### DIFF
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -525,7 +525,7 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin):
             self.dropped()
             return
 
-        def _cannot(reason):
+        def _cannot_compute(reason):
             logger.debug("Cannot %r", reason)
             self.send(
                 message.tasks.CannotComputeTask(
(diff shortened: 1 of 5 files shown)


golem/task/tasksession.py
```diff
@@ -524,6 +524,29 @@ def _react_to_task_to_compute(self, msg):
...
+                    reason=reason,
+                ),
+            )
+            self.task_computer.session_closed()
+            self.dropped()
+
+        reasons = message.CannotComputeTask.REASON
+
+        logger.debug("cnc enabled: %r", self.concent_service.enabled)
+        logger.debug("msg enabled: %r", msg.concent_enabled)
```
shadeofblue: maybe the debug message should say `msg concent enabled:` instead of just `msg enabled`?

shadeofblue: or even `ttc concent enabled:`

golem/task/tasksession.py
```diff
@@ -524,6 +524,29 @@ def _react_to_task_to_compute(self, msg):
             self.task_computer.session_closed()
             self.dropped()
             return
+
+        def _cannot(reason):
```
shadeofblue: and I'd change it to either `_cannot_compute` or maybe `_cct` as the lone "cannot" sounds awkward-ish to me ...

shadeofblue (approved):

golem/task/tasksession.py
jiivan: Those are just development artefacts. Thanks for pointing them out.

Issue #2761, opened 2018-05-10
[Concent] Automatic Concent desposits


jiivan:
| Role | When |-|
|-|-|-|
| Requestor | immediately after creating `TaskHeader` and putting it in `P2P service`| Done |
| Provider | When requesting additional verification | In progress |

| Role | Min amount | Optimal amount |-|
|-|-|-|-|
| Requestor | `2tp deposit + 1tp gntw` | `4tp deposit + 1tp gntw` | Done |
| Provider | `subtask_price` | `4*subtask_price`| In progress |

lukasz-glen:
Please refer to [G 5.4](https://docs.google.com/document/d/1fY2E8uRXVJjdmWjl1l5qz02TterB_wIlspHGbBw_LkM/edit#heading=h.ifd047f71yjw) and [G 6.1 and G 6.2](https://docs.google.com/document/d/1fY2E8uRXVJjdmWjl1l5qz02TterB_wIlspHGbBw_LkM/edit#heading=h.h2xnlx3ide7r)

jiivan:
Already done in:
https://github.com/golemfactory/golem/blob/2aa5d432594227aeba1c6302eba9a76f0096e431/golem/task/tasksession.py#L702-L719

Issue #2766, opened 2018-05-14
[Concent] Provider checks Requestor's GNTB balance and deposits


lukasz-glen:
When Provider gets new subtask to compute and Concent is enabled for this subtask, he checks if Requestor has required GNTB balance and deposit. Please refer to [documentation (G 6.3, G 6.3.1 and G 6.3.2)](https://docs.google.com/document/d/1fY2E8uRXVJjdmWjl1l5qz02TterB_wIlspHGbBw_LkM/edit#heading=h.h2xnlx3ide7r)

lukasz-glen:
@jiivan you have done some work with Concent negotiation, please check it up if it is partially done

jiivan:
> G 6.3
> Provider gets new subtask to compute
> This existing use case. This description regards only its interaction with contracts.
> After Provider receives TaskToCompute but before he starts to compute, Provider checks if Concent flag in TaskToCompute is set properly and then he checks Requestor’s wallet and deposit. If Concent flag in TaskToCompute is not set properly, then Provider cancels computation by sending CannotComputeTask. Provider checks if Requestor’s GNTW balance is at least (total task price). If Concent is enabled, then Provider checks if Requestor’s deposit is at least 2*(total task price). If it is not then Provider cancels computation by sending CannotComputeTask.
> Relations: G 6.3.1, G 6.3.2
> 
> G 6.3.1
> Provider checks Requestor’s deposit
> Input: required value, Requestor’s Etherum address
> This simple function checks if deposit of the given Ethereum address is not less than the required value. It also checks locked_until field, if it is set and will elapse within 48 hours, then deposit should be considered as invalid.
> TODO confirm 48 hours time interval
> 
> G 6.3.2
> Provider checks Requestor’s GNT balance
> Input: required GNTW value, Requestor’s Etherum address
> This simple function checks if GNTW balance of the given Ethereum address is not less than the required value.
>

PR #2774, opened 2018-05-16
Automatic requestor deposit

jiivan:
Solves requestor part of #2761 
Provider part requires additional verification.

### DIFF
--- a/apps/core/task/coretask.py
+++ b/apps/core/task/coretask.py
@@ -170,6 +170,10 @@ class CoreTask(Task):
         self.tmp_dir = None
         self.max_pending_client_results = max_pending_client_results
 
+    @property
+    def price(self) -> int:
+        return self.subtask_price * self.total_tasks
+
     @staticmethod
     def create_task_id(public_key: bytes) -> str:
         return generate_id(public_key)
--- a/golem/client.py
+++ b/golem/client.py
@@ -546,7 +546,7 @@ class Client(HardwarePresetsMixin):
             self.transaction_system.concent_deposit(
                 required=min_amount,
                 expected=opt_amount,
-                reserved=task.price,
+                reserved=self.funds_locker.sum_locks()[0],
             )
 
         task_id = task.header.task_id
--- a/golem/client.py
+++ b/golem/client.py
@@ -537,15 +537,16 @@ class Client(HardwarePresetsMixin):
 
         if self.transaction_system:
             self.funds_locker.lock_funds(task)
+
+        if self.transaction_system and self.concent_service.enabled:
             min_amount, opt_amount = msg_helpers.requestor_deposit_amount(
                 task.price,
             )
-
-        if self.transaction_system and self.concent_service.enabled:
             # Could raise golem.transactions.ethereum.exceptions.NotEnoughFunds
             transaction_hash = self.transaction_system.concent_deposit(
                 required=min_amount,
                 expected=opt_amount,
+                reserved=task.price,
             )
         else:
             transaction_hash = None
--- a/golem/client.py
+++ b/golem/client.py
@@ -535,10 +535,9 @@ class Client(HardwarePresetsMixin):
         else:
             task = task_dict
 
-        if self.transaction_system:
-            self.funds_locker.lock_funds(task)
+        self.funds_locker.lock_funds(task)
 
-        if self.transaction_system and self.concent_service.enabled:
+        if self.concent_service.enabled:
             min_amount, opt_amount = msg_helpers.requestor_deposit_amount(
                 task.price,
             )
--- a/golem/client.py
+++ b/golem/client.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from threading import Lock, Thread
 from typing import Dict, Hashable, Optional, Union, List, Iterable, Tuple
 
+from golem_messages import helpers as msg_helpers
 from pydispatch import dispatcher
 from twisted.internet.defer import (
     inlineCallbacks,
@@ -536,6 +537,14 @@ class Client(HardwarePresetsMixin):
 
         if self.transaction_system:
             self.funds_locker.lock_funds(task)
+            min_amount, opt_amount = msg_helpers.requestor_deposit_amount(
+                task.price,
+            )
+            # Could raise golem.transactions.ethereum.exceptions.NotEnoughFunds
+            self.transaction_system.concent_deposit(
+                required=min_amount,
+                expected=opt_amount,
+            )
 
         task_id = task.header.task_id
         logger.info('Enqueue new task "%r"', task_id)
--- a/golem/client.py
+++ b/golem/client.py
@@ -540,11 +540,15 @@ class Client(HardwarePresetsMixin):
             min_amount, opt_amount = msg_helpers.requestor_deposit_amount(
                 task.price,
             )
+
+        if self.transaction_system and self.concent_service.enabled:
             # Could raise golem.transactions.ethereum.exceptions.NotEnoughFunds
-            self.transaction_system.concent_deposit(
+            transaction_hash = self.transaction_system.concent_deposit(
                 required=min_amount,
                 expected=opt_amount,
             )
+        else:
+            transaction_hash = None
 
         task_id = task.header.task_id
         logger.info('Enqueue new task "%r"', task_id)
@@ -577,9 +581,25 @@ class Client(HardwarePresetsMixin):
                 task_state.resource_hash = resource_manager_result[0]
             except Exception as exc:  # pylint: disable=broad-except
                 error(exc)
-            else:
-                request = AsyncRequest(task_manager.start_task, task_id)
+                return
+
+            request = AsyncRequest(task_manager.start_task, task_id)
+
+            if transaction_hash is None:
                 async_run(request, lambda _: _result.callback(task), error)
+                return
+
+            def on_receipt(receipt):
+                if not receipt.status:
+                    error("Blockchain transaction failed")
+                    return
+                async_run(request, lambda _: _result.callback(task), error)
+
+            self.transaction_system._sci.on_transaction_confirmed(
+                tx_hash=transaction_hash,
+                required_confs=6,
+                cb=on_receipt,
+            )
 
         def error(exception):
             logger.error("Task '%s' creation failed: %r", task_id, exception)
--- a/golem/client.py
+++ b/golem/client.py
@@ -543,13 +543,11 @@ class Client(HardwarePresetsMixin):
                 task.price,
             )
             # Could raise golem.transactions.ethereum.exceptions.NotEnoughFunds
-            transaction_hash = self.transaction_system.concent_deposit(
+            self.transaction_system.concent_deposit(
                 required=min_amount,
                 expected=opt_amount,
                 reserved=task.price,
             )
-        else:
-            transaction_hash = None
 
         task_id = task.header.task_id
         logger.info('Enqueue new task "%r"', task_id)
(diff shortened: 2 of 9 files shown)


golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -156,3 +156,18 @@ def withdraw(
             return res
 
         raise ValueError('Unknown currency {}'.format(currency))
+
+    def concent_balance(self) -> int:
+        return self._sci.get_deposit_value(account_address=self._sci._address)  # noqa pylint: disable=protected-access
```
Krigpl: `sci.get_eth_address` returns the address so use that instead of `sci._address`

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -156,3 +156,18 @@ def withdraw(
             return res
 
         raise ValueError('Unknown currency {}'.format(currency))
+
+    def concent_balance(self) -> int:
```
Krigpl: I don't see it being used anywhere else so either make it private or just inline below since it's quite simple anyway

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -156,3 +156,18 @@ def withdraw(
...
+        return self._sci.get_deposit_value(account_address=self._sci._address)  # noqa pylint: disable=protected-access
+
+    def concent_deposit(self, required: int, expected: int) -> None:
+        current = self.concent_balance()
+        if current >= required:
+            return
+        required -= current
+        expected -= current
+        gntb_balance = self.payment_processor._gnt_available()  # noqa pylint: disable=protected-access
```
Krigpl: `_gnt_available` gives you GNT+GNTB so later when you try to call `deposit_payment` it may fail because there's not enough GNTB

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -156,3 +156,18 @@ def withdraw(
...
+        if gntb_balance < required:
+            raise NotEnoughFunds(required, gntb_balance, 'GNTB')
+        max_possible_amount = min(expected, gntb_balance)
+        self._sci.deposit_payment(max_possible_amount)
```
Krigpl: This just sends the transaction to the network and it may take some time before it gets mined - do you care about that?
This may also throw if there isn't ETH for gas

golem/client.py
```diff
@@ -536,6 +537,14 @@ def enqueue_new_task(self, task_dict):
 
         if self.transaction_system:
             self.funds_locker.lock_funds(task)
+            min_amount, opt_amount = msg_helpers.requestor_deposit_amount(
+                task.price,
+            )
+            # Could raise golem.transactions.ethereum.exceptions.NotEnoughFunds
+            self.transaction_system.concent_deposit(
```
Krigpl, line 546: Shouldn't it happen only when Concent is enabled ?

Krigpl (changes requested):

golem/transactions/ethereum/ethereumtransactionsystem.py
Krigpl: In general any time you feel like you need to access protected member in SCI just tell me - I'll either point you to the right direction or change the API appropriately if needed.

golem/transactions/ethereum/ethereumtransactionsystem.py
jiivan: It'll be used in provider deposits and to verify whether other side has sufficient deposit ( #2766 ). And there no such thing as private method in python. :)

golem/client.py
jiivan: Good point.

golem/client.py
```diff
@@ -536,6 +537,14 @@ def enqueue_new_task(self, task_dict):
 
         if self.transaction_system:
             self.funds_locker.lock_funds(task)
+            min_amount, opt_amount = msg_helpers.requestor_deposit_amount(
```
Krigpl: Nit picking but I find it confusing that `requestor_deposit_amount` is coming from the `golem_messages` module

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -156,3 +156,18 @@ def withdraw(
...
+        return self._sci.get_deposit_value(account_address=self._sci._address)  # noqa pylint: disable=protected-access
+
+    def concent_deposit(self, required: int, expected: int) -> None:
+        current = self.concent_balance()
+        if current >= required:
+            return
+        required -= current
+        expected -= current
+        gntb_balance = self.payment_processor._gnt_available()  # noqa pylint: disable=protected-access
+        if gntb_balance < required:
```
Krigpl: Another thing: we shouldn't end up putting too much to the deposit, not to end up in a situation when we can't cover the task from the main balance. AFAICT right now you could transfer whole balance to the deposit and basically be forced to pay from the deposit instead of the usual batchTransfer

jiivan: @Krigpl isn't it covered by fundslocker?

golem/client.py
```diff
@@ -537,6 +538,19 @@ def enqueue_new_task(self, task_dict):
         if self.transaction_system:
             self.funds_locker.lock_funds(task)
 
+        if self.transaction_system and self.concent_service.enabled:
```
Krigpl: `self.transaction_system` cannot be `None` anymore (same applies to the `if` above)

golem/client.py
```diff
@@ -537,6 +538,19 @@ def enqueue_new_task(self, task_dict):
         if self.transaction_system:
             self.funds_locker.lock_funds(task)
 
+        if self.transaction_system and self.concent_service.enabled:
+            min_amount, opt_amount = msg_helpers.requestor_deposit_amount(
+                task.price,
+            )
+            # Could raise golem.transactions.ethereum.exceptions.NotEnoughFunds
+            transaction_hash = self.transaction_system.concent_deposit(
```
shadeofblue: is this `transaction_hash` used anywhere?

shadeofblue (approved):
seems okay

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -156,3 +156,24 @@ def withdraw(
...
+
+    def concent_deposit(
+            self, required: int, expected: int, reserved: int) -> Optional[str]:
+        current = self.concent_balance()
+        if current >= required:
+            return None
+        required -= current
+        expected -= current
+        # TODO migrate funds from gnt to gntb
+        gntb_balance = self._sci.get_gntb_balance(self._sci.get_eth_address())
```
Krigpl: That doesn't account for GNTB that is locked in PaymentProcessor

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -156,3 +156,24 @@ def withdraw(
...
+        gntb_balance -= reserved
+        if gntb_balance < required:
+            raise NotEnoughFunds(required, gntb_balance, 'GNTB')
+        max_possible_amount = min(expected, gntb_balance)
+        transaction_id = self._sci.deposit_payment(max_possible_amount)
```
Krigpl: just put `return` here

Krigpl (changes requested):
Please wait with landing this until new GNT-GNTB conversion is done.

golem/client.py
shadeofblue: @Krigpl I believe it's part of the protocol so indeed, it should be included in `golem_messages`...

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -190,6 +190,26 @@ def withdraw(
 
         raise ValueError('Unknown currency {}'.format(currency))
 
+    def concent_balance(self) -> int:
+        return self._sci.get_deposit_value(
+            account_address=self._sci.get_eth_address(),
+        )
+
+    def concent_deposit(
```
Krigpl, line 198: A test for this method would be beneficial

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -190,6 +190,26 @@ def withdraw(
...
+            return None
+        required -= current
+        expected -= current
+        gntb_balance = self._sci.get_gntb_balance(self._sci.get_eth_address())
+        gntb_balance -= reserved
+        gntb_balance -= self.payment_processor.reserved_gntb
+        if gntb_balance < required:
+            raise NotEnoughFunds(required, gntb_balance, 'GNTB')
+        max_possible_amount = min(expected, gntb_balance)
+        return self._sci.deposit_payment(max_possible_amount)  # tx_hash
```
Krigpl: Please spawn an info log message about the amount and including the tx hash

apps/core/task/coretask.py
```diff
@@ -170,6 +170,10 @@ def __init__(self,
         self.tmp_dir = None
         self.max_pending_client_results = max_pending_client_results
 
+    @property
+    def price(self) -> int:
```
Krigpl, line 174: Since you introduce this property, can you make `FundsLocker` use it in the `gnt_lock` method?

golem/client.py
```diff
@@ -535,8 +536,18 @@ def enqueue_new_task(self, task_dict):
...
+
+        if self.concent_service.enabled:
+            min_amount, opt_amount = msg_helpers.requestor_deposit_amount(
+                task.price,
+            )
+            # Could raise golem.transactions.ethereum.exceptions.NotEnoughFunds
+            self.transaction_system.concent_deposit(
+                required=min_amount,
+                expected=opt_amount,
+                reserved=task.price,
```
Krigpl: This has to be total amount of locked funds from `FundsLocker`
Let's say you have 100 GNT
Then create a task for 5 GNT
You lock 5 GNT in FundsLocker
You make deposit for 20 GNT
Then you create a task for 25 GNT
You lock 25 GNT in FundsLocker (with total of 30 GNT)
You want to make deposit for 80 (100 - 20), but have only 70 (100 - 5 - 25) available, so you make deposit for 70 GNT
Total deposit is 90 GNT, your balance is 10 GNT but you have tasks for 30 GNT

Please add tests for these cases.

Krigpl (changes requested):

golem/client.py
mfranciszkiewicz: I'm sorry to interrupt, but `golem_messages` shouldn't try to define the protocol.

Krigpl: Maybe we should try to clarify what `golem_messages` actually is. My thinking so far was that it just provides all the messages (so simple structs basically) + some implementation helpers (like serialization), but it doesn't provide any constants/mechanisms defined by the protocol.
But looks like it's the former. I wonder whether that was intentional or just convenient, since it's a shared module between Golem and Concent.
The thing is that if these constants get changed it requires to bump the golem-messages version, since it's not backwards compatible. And that affects non Concent users even though they shouldn't be affected because the change has nothing to do with them.

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -190,6 +190,31 @@ def withdraw(
...
+        )
+
+    def concent_deposit(
+            self, required: int, expected: int, reserved: int) -> Optional[str]:
+        current = self.concent_balance()
+        if current >= required:
+            return None
+        required -= current
+        expected -= current
+        gntb_balance = self._sci.get_gntb_balance(self._sci.get_eth_address())
```
Krigpl: This is the same as `gntb_balance = self._gntb_balance` but without an extra request to Geth

tests/golem/test_client.py
```diff
@@ -12,10 +13,10 @@
 from pydispatch import dispatcher
 from twisted.internet.defer import Deferred
 
-import golem
 from apps.appsmanager import AppsManager
 from apps.dummy.task.dummytask import DummyTask
 from apps.dummy.task.dummytaskstate import DummyTaskDefinition
+import golem
```
Krigpl, line 19: Why is it needed?

Krigpl (approved):
lgtm, just minor comments

tests/golem/test_client.py
jiivan: Ungrouped imports lint

PR #2784, opened 2018-05-17
Graceful shutdown

maaktweluit:
- Added new setting `in_shutdown`
- On `init()` from `Client` this setting gets set to `False`
- Added new rpc call `graceful_shutdown` to be called from cli / gui and control the `in_shutdown` setting
- Added `golemcli account shutdown` to call rpc
- As provider `in_shutdown` disables new tasks like the setting `accept_tasks`
- As requestor, `in_shutdown` raises an error like "out of eth"
- Added handler `node.check_shutdown` to be called from `taskcomputer` and `taskmanager` when a task is done.
- Check `subtask` timeouts before `task` timeouts in `taskmanager`

### DIFF
--- a/golem/appconfig.py
+++ b/golem/appconfig.py
@@ -35,6 +35,7 @@ SEND_PEERS_NUM = 10
 USE_IP6 = 0
 USE_UPNP = 1
 ACCEPT_TASKS = 1
+IN_SHUTDOWN = 0
 SEND_PINGS = 1
 ENABLE_MONITOR = 1
 DEBUG_THIRD_PARTY = 0
@@ -125,6 +126,7 @@ class AppConfig:
             key_difficulty=KEY_DIFFICULTY,
             # flags
             accept_tasks=ACCEPT_TASKS,
+            in_shutdown=IN_SHUTDOWN,
             send_pings=SEND_PINGS,
             enable_talkback=ENABLE_TALKBACK,
             enable_monitor=ENABLE_MONITOR,
--- a/golem/appconfig.py
+++ b/golem/appconfig.py
@@ -35,7 +35,6 @@ SEND_PEERS_NUM = 10
 USE_IP6 = 0
 USE_UPNP = 1
 ACCEPT_TASKS = 1
-IN_SHUTDOWN = 0
 SEND_PINGS = 1
 ENABLE_MONITOR = 1
 DEBUG_THIRD_PARTY = 0
@@ -125,8 +124,8 @@ class AppConfig:
             opt_peer_num=OPTIMAL_PEER_NUM,
             key_difficulty=KEY_DIFFICULTY,
             # flags
+            in_shutdown=0,
             accept_tasks=ACCEPT_TASKS,
-            in_shutdown=IN_SHUTDOWN,
             send_pings=SEND_PINGS,
             enable_talkback=ENABLE_TALKBACK,
             enable_monitor=ENABLE_MONITOR,
--- a/golem/client.py
+++ b/golem/client.py
@@ -110,7 +110,8 @@ class Client(HardwarePresetsMixin):
             start_geth: bool = False,
             start_geth_port: Optional[int] = None,
             geth_address: Optional[str] = None,
-            apps_manager: AppsManager = AppsManager()) -> None:
+            apps_manager: AppsManager = AppsManager(),
+            task_finished_cb=None) -> None:
 
         self.apps_manager = apps_manager
         self.datadir = datadir
@@ -124,6 +125,9 @@ class Client(HardwarePresetsMixin):
         self.app_config = app_config
         self.config_desc = config_desc
         self.config_approver = ConfigApprover(self.config_desc)
+
+        if self.config_desc.in_shutdown:
+            self.update_setting('in_shutdown', False)
 
         logger.info(
             'Client "%s", datadir: %s',
@@ -207,6 +211,7 @@ class Client(HardwarePresetsMixin):
         self.use_monitor = use_monitor
         self.monitor = None
         self.session_id = str(uuid.uuid4())
+        self._task_finished_cb = task_finished_cb
 
         dispatcher.connect(
             self.p2p_listener,
@@ -317,7 +322,8 @@ class Client(HardwarePresetsMixin):
             use_ipv6=self.config_desc.use_ipv6,
             use_docker_manager=self.use_docker_manager,
             task_archiver=self.task_archiver,
-            apps_manager=self.apps_manager
+            apps_manager=self.apps_manager,
+            task_finished_cb=self._task_finished_cb,
         )
 
         monitoring_publisher_service = MonitoringPublisherService(
--- a/golem/client.py
+++ b/golem/client.py
@@ -529,6 +529,9 @@ class Client(HardwarePresetsMixin):
         self._unlock_datadir()
 
     def enqueue_new_task(self, task_dict):
+        if self.config_desc.in_shutdown:
+            raise Exception('Can not enqueue tasks when shutdown is in progress, '
+                            'toggle shutdown mode off to create a new tasks.')
         task_manager = self.task_server.task_manager
         _result = Deferred()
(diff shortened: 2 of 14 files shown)


golem/appconfig.py
```diff
@@ -34,6 +34,7 @@
 USE_IP6 = 0
 USE_UPNP = 1
 ACCEPT_TASKS = 1
+IN_SHUTDOWN = 0
```
Krigpl: I didn't look closely at the whole PR but this looks fishy. Do you expect someone to set it to 1 and then run Golem? If not then it doesn't belong in the config file.

maaktweluit: I agree this looks fishy, but from all options this was picked as the best approach to this problem. I'm open to alternatives to this setting

No, it wont even work as expected when the users sets it before running golem, this setting is only set after golem is running. ( by gui or rpc )

Please read the description of the PR at least, i tried to make all steps / changes clear to get to a graceful shutdown.

The reason this was chosen is that config_desc is available in all places where the in_shutdown should work, and update_config handles updated from all sides properly.

Krigpl: How about just removing this line and changing line 127 to `in_shutdown=0` ? afaict it should be fine.

maaktweluit: this would make the file inconsistent, all settings are set in this way
What is the added value / clarity you see on removing only this line?

Krigpl: `node_name`, `node_address`, `eth_account`, `seed_host`, `seeds` are all set to empty string literals so it's already inconsistent in this sense. 
I understand that all the values in this file are configuration properties of Golem and can be tweaked by the user if the need arises. While `IS_SHUTDOWN` is not something that is supposed to be changed by the user because it not really a configuration property, it's just a runtime boolean flag for proper execution.
That's why I find it fishy to put this property here but if it's the easiest then it shouldn't be exposed among other real configuration properties.

maaktweluit: clear! thanks for the explanation!

Will update this :)

Elfoniok:
I have also doubts about this.
1. This started from request to close golem VM after shutdown, but this not solve this, instead we wait for golem to finish task, strange.

2. Have you considered using `addSystemEventTrigger("before", "shutdown", ...)`, this should be used if one want to do anything before twisted will start shutting down process.

maaktweluit:
@Elfoniok Thanks for the feedback!

1. I'm not sure what you mean with this, the original issue title was: "Finish important tasks in the background after closing Golem". This PR does that, it disables using golem and closes automatically when all important tasks are finished. The GUI can be closed after the shutdown bit is set.

2. Yes, this has been considered and is possible to hook in there eventually. For the first implementation we decided to not hook here to reduce impact and first see how this works.. POC -> MVP -> enforce on each user

golem/node.py
```diff
@@ -189,6 +190,51 @@ def accept_terms():
     def show_terms():
         return TermsOfUse.show_terms()
 
+    def graceful_shutdown(self):
+        # is in shutdown? turn off as toggle
+        if self._config_desc.in_shutdown:
+            self.client.update_setting('in_shutdown', False)
```
jiivan: I'm not sure about toggle here. Explicit shutdown cancelling would be nice.

jiivan (approved):
Looks like a step in good direction. I have doubts about toggle though.

golem/node.py
maaktweluit: Good idea!

Will check this when we add this feature to the GUI, then either the settings can be adjusted directly from there or we have to make an explicit call like you propose

maaktweluit:
All tests have been added!

@Krigpl @Elfoniok @jiivan Please check again: removed WIP, fixed some comments and added all tests.

golem/node.py
```diff
@@ -205,6 +206,51 @@ def accept_terms(self,
...
+            self.quit()
+            return 'quit'
+
+        # configure in_shutdown
+        logger.info('Enabling shutdown mode, no more tasks can be started')
+        self.client.update_setting('in_shutdown', True)
+
+        # subscribe to events
+
+        return 'on'
```
Krigpl: How about not using magic string literals? :)

golem/node.py
```diff
@@ -205,6 +206,51 @@ def accept_terms(self,
     def show_terms():
         return TermsOfUse.show_terms()
 
+    def graceful_shutdown(self):
```
Krigpl: Please add return types to all newly introduced methods

golem/node.py
```diff
@@ -205,6 +206,51 @@ def accept_terms(self,
...
+
+        # configure in_shutdown
+        logger.info('Enabling shutdown mode, no more tasks can be started')
+        self.client.update_setting('in_shutdown', True)
+
+        # subscribe to events
+
+        return 'on'
+
+    def check_shutdown(self):
```
Krigpl: I think this could be a protected method.
Also I don't find this name descriptive (`check` in the name suggests boolean return imo) but can't come up with a better one. Maybe `try_quit`?

tests/golem/task/test_taskcomputer.py
```diff
@@ -149,6 +151,8 @@ def test_computation(self):
         assert tc.counting_thread is not None
         self.assertGreater(tc.counting_thread.time_to_compute, 9)
         self.assertLessEqual(tc.counting_thread.time_to_compute, 10)
+        mock_finished.assert_called_once()
```
Krigpl: `assert_called_once_with()` even

golem/node.py
jiivan: @Krigpl there is no such thing as a "protected method" in python. Just sayin' :)

Krigpl: I've heard you first time you said that, but there are conventions and linters and we use them.

jiivan: I think using the right vocabulary for the tool helps a lot. There is a *single underscore convention* in python slightly enforced by exclusion of single underscore functions and variables from `__all__` (can be overridden). It's also supported by some linters. But there is not such thing as protected method.

OOP concept of access control is neither supported nor encouraged in python.

Krigpl: Considering that you have to write `disable-protected-access` in our codebase when accessing things starting with a single underscore I think calling a method 'protected' is the right vocabulary.

golem/node.py
maaktweluit: good one! fixed :)

golem/node.py
maaktweluit: added

golem/node.py
maaktweluit: changed it to `_try_shutdown()` does that make it protected enough? ;)

tests/golem/task/test_taskcomputer.py
maaktweluit: good catch! find and replaced them all

golem/node.py
```diff
@@ -205,6 +213,68 @@ def accept_terms(self,
...
+        if self.client is None:
+            logger.debug('_is_task_in_progress? False: client=None')
+            return False
+
+        task_server = self.client.task_server
+        if task_server is None or task_server.task_manager is None:
+            logger.debug('_is_task_in_progress? False: task_manager=None')
+            return False
+
+        task_requester_progress = task_server.task_manager.get_progresses()
```
Krigpl: `s/requester/requestor/`

golem/client.py
```diff
@@ -523,6 +529,9 @@ def quit(self):
         self._unlock_datadir()
 
     def enqueue_new_task(self, task_dict):
+        if self.config_desc.in_shutdown:
+            raise Exception('Can not enqueue task: shutdown is in progress,'
```
Krigpl: nit: space after comma

Krigpl (approved):
thisisgine

golem/node.py
maaktweluit: Fixed!

golem/client.py
maaktweluit: Fixed!

Issue #2768, opened 2018-05-15
Tasks states are preserved in a pickle instead of the database


etam:
In `TaskManager` we put `TaskState` into a pickle. In #2709 I've modified `TaskState` (`subtask_status` is now an enum, not string). Now, when loading old objects from pickle, we get random exceptions (because string is not an enum, ofc).

Quick fix: convert `subtask_status` from string to enum, when loading from pickle.

Proper fix: Using pickle for storing data in golem should be banned. We have a database for that. And it supports migrations now!

etam:
Quick fix (migration for pickled data) is merged. The issue stays open for development of migration to database.

PR #2709, opened 2018-04-24
Improve taskstate

etam:
And fix a bug, where we initialize `subtask_status` with a `TaskStatus`.

PR #2785, opened 2018-05-17
TaskManager: convert str to enum when unpickling

etam:
Workarounds a regression introduced by #2709 

This is a quickfix for #2768

### DIFF
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -219,6 +219,21 @@ class TaskManager(TaskEventListener):
         except (FileNotFoundError, OSError) as e:
             logger.warning("Couldn't remove dump file: %s - %s", filepath, e)
 
+    @staticmethod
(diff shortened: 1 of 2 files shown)


golem/task/taskmanager.py
```diff
@@ -230,7 +230,19 @@ def restore_tasks(self) -> None:
             task_id = None
             with path.open('rb') as f:
                 try:
+                    task: Task
+                    state: TaskState
                     task, state = pickle.load(f)
+
+                    # workaround #2768
```
jiivan: Since this is only one time migration for old pickles, I would be really pleased with a comment that it's a migration.

Also maybe it's possible to enclose this migration logic in a function with proper docstring? Something like `_migrate_status_to_enum()`

jiivan (approved):
LGTM with one but

tworec (approved):
LGTM

PR #2793, opened 2018-05-18
Samplespp

Elfoniok:
Blender task have now option to define samples per pixel parameters:
It should be pass from UI or JSON as 
`['options']['samples']`

If option is not available, blender will use default value from .blend file.

### DIFF
--- a/apps/blender/resources/blenderloganalyser.py
+++ b/apps/blender/resources/blenderloganalyser.py
@@ -24,7 +24,10 @@ def make_log_analyses(log_content, return_data):
     engine_type = find_engine_type(log_content)
     if engine_type:
         return_data['engine_type'] = to_unicode(engine_type)
-        return_data['samples'] = find_samples_for_scenes(log_content)
+        if engine_type == "CYCLES":
+            samples_pp = find_samples_for_scenes(log_content)
+            if samples_pp:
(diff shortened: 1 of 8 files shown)


apps/blender/resources/blenderloganalyser.py
```diff
@@ -21,6 +21,20 @@ def make_log_analyses(log_content, return_data):
...
+        return_data['engine_type'] = to_unicode(engine_type)
+        if engine_type == "CYCLES":
+            samples_pp = find_samples_for_scenes(log_content)
+            if samples_pp:
+                return_data['samples'] = samples_pp
+
+
+def find_samples_for_scenes(log_content):
+    samples = re.search("^Info: Samples: (.*)", log_content,
+                        re.MULTILINE | re.IGNORECASE)
```
jiivan: I know it's c&p from other regexes, but is `re.MULTILINE | re.IGNORECASE` really needed here? Especially `re.IGNORECASE` could slow parsing down significantly.

jiivan (approved):
LGTM, one comment.

apps/blender/resources/blenderloganalyser.py
Elfoniok: Good point thanks.

PR #2795, opened 2018-05-21
Convert GNT to GNTB asap

Krigpl:
The idea is to convert GNT to GNTB as soon as it is possible and make the rest of the system not aware of GNT at all and only operate on GNTB which makes a lot of stuff easier (eg. deposits to Concent).

I removed a bunch of stuff from PaymentProcessor and moved some stuff to EthereumTransactionSystem. This removes some dependencies and makes PaymentProcessor closer to have only one responsibility.

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -2,13 +2,12 @@ import calendar
 import logging
 import time
 from collections import defaultdict
-from typing import Dict, List, Optional
+from typing import Dict, List
 from threading import Lock
 
 from sortedcontainers import SortedListWithKey
-from ethereum.utils import normalize_address, denoms
+from ethereum.utils import denoms
 from pydispatch import dispatcher
-import requests
 
 import golem_sci
 from golem.core.variables import PAYMENT_DEADLINE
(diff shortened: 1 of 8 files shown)


golem/ethereum/paymentprocessor.py
```diff
@@ -59,91 +37,30 @@ def _make_batch_payments(payments: List[Payment]) -> List[golem_sci.Payment]:
...
-
-    def _gnt_available(self):
-        gnt_balance, _ = self.gnt_balance()
-        return gnt_balance - self.__gntb_reserved
+
+    def get_reserved_eth(self) -> int:
+        return self._eth_reserved + self.ETH_BATCH_PAYMENT_BASE
+
+    def get_reserved_gntb(self) -> int:
+        return self._gntb_reserved
```
jiivan: This kind of getter/setter encapsulation is suboptimal in python. You can just rename `_gntb_reserved` attribute to `gntb_reserved`. If a need arises you could always make it a **property** in the future.

jiivan (approved):
LGTM, with one remark about getters in python.

golem/ethereum/paymentprocessor.py
Krigpl: These are readonly, so I will use `property` indeed (not a personal fan of this solution, but I guess this is Pythonic way, so I'm fine).

mfranciszkiewicz (approved):

PR #2798, opened 2018-05-21
Activate preset async

### LINKED ISSUE #346: performance not recalculating after resources (CPU, RAM) reallocation
change to CPU cores allocation impacts node's performance, but now change CPU cores or RAM will not trigger performance recalculation. This leads to wrong perception of node as a provider.

needs to be fixed before golemfactory/golem#2771

tworec:
`activate_hw_preset` is changed to be async, and return performance values
fixes golemfactory/golem-electron#346
works only with golemfactory/golem-electron#347

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -1126,7 +1126,6 @@ class Client(HardwarePresetsMixin):
 
     @inlineCallbacks
     def run_benchmark(self, env_id):
-        logger.info('Running benchmark for %s', env_id)
         deferred = Deferred()
 
         if env_id != DefaultEnvironment.get_id():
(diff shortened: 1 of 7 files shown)


prekucki (approved):

golem/client.py
```diff
@@ -1126,7 +1126,7 @@ def get_environments(self):
 
     @inlineCallbacks
     def run_benchmark(self, env_id):
-        logger.info('Running benchmarks ...')
+        logger.info('Running benchmark for %s', env_id)
```
maaktweluit: This log line is now duplicated ( benchmarkmanager.py ), maybe "requesting benchmark" followed by "running benchmark"?

maaktweluit (commented):
LGTM! Small comment, will approve on reply

golem/client.py
tworec: ok, I'll remove one of them

tworec:
related to #2771

PR #2799, opened 2018-05-22
Withdraw GNTB only

Krigpl:
Now that we convert GNT to GNTB asap and operate on GNTB only withdrawals logic is a bit easier.

### DIFF
--- a/golem/transactions/ethereum/ethereumtransactionsystem.py
+++ b/golem/transactions/ethereum/ethereumtransactionsystem.py
@@ -150,6 +150,8 @@ class EthereumTransactionSystem(TransactionSystem):
                 self._gnt_balance + self._gntb_balance - pp.reserved_gntb
             if amount > total_gnt - lock:
                 raise NotEnoughFunds(amount, total_gnt - lock, currency)
+            # This can happen during unfinished GNT-GNTB conversion,
+            # so we should wait until it finishes
             if amount > self._gntb_balance:
                 raise Exception('Cannot withdraw right now, '
                                 'background operations in progress')
--- a/golem/transactions/ethereum/ethereumtransactionsystem.py
+++ b/golem/transactions/ethereum/ethereumtransactionsystem.py
@@ -121,15 +121,7 @@ class EthereumTransactionSystem(TransactionSystem):
         if currency == 'ETH':
             return 21000 * gas_price
         if currency == 'GNT':
-            total_gnt = self._gnt_balance + self._gntb_balance
-            gnt = self._sci.get_gnt_balance(self._sci.get_eth_address())
-            gntb = total_gnt - gnt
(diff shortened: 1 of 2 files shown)


golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -154,39 +146,19 @@ def withdraw(
...
-
-            if gntb >= amount:
-                log.info(
-                    "Withdrawing %f GNTB to %s",
-                    amount / denoms.ether,
-                    destination,
-                )
-                return [self._sci.convert_gntb_to_gnt(destination, amount)]
-
+            if amount > self._gntb_balance:
```
maaktweluit: Should one ( or both ) of these checks be `>=` to allow withdraw of the full balance?

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -154,39 +146,19 @@ def withdraw(
...
+                raise Exception('Cannot withdraw right now, '
+                                'background operations in progress')
```
maaktweluit: what operations is still in progress?

maaktweluit (commented):
LGTM!

Few small questions then i'll approve :)

golem/transactions/ethereum/ethereumtransactionsystem.py
Krigpl: Not sure I get it. The way it is now you are able to withdraw full balance

golem/transactions/ethereum/ethereumtransactionsystem.py
Krigpl: GNT-GNTB conversion, I imagine this message could be shown to the user, so it's better to use vague `background operations` rather than GNTB conversion.

mfranciszkiewicz (approved):

golem/transactions/ethereum/ethereumtransactionsystem.py
maaktweluit: Then it should be fine, i see the old code was `>=` and in the new version it changed to `>`

maaktweluit: looking at it more closely it is indeed the same:
old = `gnt >= amt` = success
new = `amt > gnt` = failure

great this has been simplified now ;)

golem/transactions/ethereum/ethereumtransactionsystem.py
maaktweluit: Agree with the end-user message, but not sure if all of our future selves will find this easily

Maybe a comment for the unfortunate dev that has to debug this later would be nice.

maaktweluit (approved):

PR #2817, opened 2018-05-25
New command: rpc provider state

### LINKED ISSUE #2704: Add RPC call to see provider state
Providers may be interested if their nodes is currently computing task or not. It will be display in GUI in footer but we need a proper RPC call that displays current state of provider, ie. 
* computing subtask (deadline: xxx)
* trying to get subtask from task
* waiting for tasks
We may want to add additional states if that's possible. 

We want to also add proper CLI method that displays result of this method.

etam:
resolves #2704


badb:
@mdtanrikulu Please check this PR, how this changes will affect footer implementation?

golem/client.py
```diff
@@ -1257,21 +1258,41 @@ def connection_status(self):
...
-            msg = "Computing {} subtask(s):".format(len(progress))
-            for k, v in list(progress.items()):
-                msg = "{} \n {} ({}%)\n".format(msg, k, v.get_progress() * 100)
-        elif self.config_desc.accept_tasks:
-            msg = "Waiting for tasks...\n"
-        else:
-            msg = "Not accepting tasks\n"
-
-        peers = self.p2pservice.get_peers()
+    def get_provider_status(self):
```
tworec: nice to have return type

golem/client.py
```diff
@@ -1257,21 +1258,41 @@ def connection_status(self):
...
+
+        task_computer = self.task_server.task_computer
+
+        # computing
+        subtask_progress: ComputingSubtaskStateSnapshot = \
+            task_computer.get_progress()
+        if subtask_progress is not None:
+            return {
+                'status': 'computing',
+                'subtask': subtask_progress.__dict__,
```
tworec, line 1406: do we need those details?

golem/manager/nodestatesnapshot.py
```diff
@@ -1,28 +1,39 @@
...
-        self.cpu_power = cpu_power
-        self.est_time_left = est_time_left
+from copy import copy
+from pathlib import Path
+from typing import List
+
+
+# pylint: disable=too-few-public-methods
+# pylint: disable=too-many-instance-attributes
+class ComputingSubtaskStateSnapshot:
```
tworec, line 8: why do we need/want to supply all these details with state?

golem/task/taskkeeper.py
```diff
@@ -609,8 +609,8 @@ def get_owner(self, task_id) -> typing.Optional[str]:
 
     def get_task(self) -> typing.Optional[TaskHeader]:
         """ Returns random task from supported tasks that may be computed
-        :return TaskHeader|None: returns either None if there are no tasks
-                                 that this node may want to compute
+        :return: either None if there are no tasks
```
tworec: this doc is malformed

golem/task/taskcomputer.py
```diff
@@ -41,16 +44,11 @@ class TaskComputer(object):
...
     dir_lock = Lock()
 
-    def __init__(self, node_name, task_server, use_docker_manager=True) -> None:
-        """ Create new task computer instance
-        :param node_name:
-        :param task_server:
-        :return:
-        """
-        self.node_name = node_name
+    def __init__(self, task_server: 'TaskServer', use_docker_manager=True) \
```
tworec: 👍 for removing unnecessary `node_name` :)

golemcli.py
```diff
@@ -29,7 +30,7 @@
 # prevent 'unused' warnings
 _ = {
     Terms, Tasks, Subtasks, Network, Environments, Resources, Settings,
-    Account, incomes, payments, debug
+    Account, incomes, payments, debug, provider_status
```
tworec: what is it for?

tworec (changes requested):
I have some questions...

golem/client.py
etam: I don't know what exactly do we need. I just put there everything that seems reasonable. Keep in mind, that our GUI might not be the only RPC client interested in this data.

golemcli.py
etam: As the comment says, it was supposed to prevent 'unused' warnings. But since we also have proper comments for linters and stuff, probably this could be removed. If it is, I think it should go in a separate PR.

golem/client.py
tworec: this is not best practice. I'd rather provide minimal set of data, not maximal one. If anyone in future will need something (s)he would simply enrich this.

golemcli.py
tworec: i think you can remove this _preventive_ code, but need to add `, unused-import` at line 14

tworec: this is not a blocker for me, feel free to ignore this comment

golem/client.py
etam: What's the minimal set of data, then? @badb @mdtanrikulu ?

golem/manager/nodestatesnapshot.py
etam: I don't understand your question.

golem/task/taskkeeper.py
etam: How about `None if there are no tasks that this node may want to compute` (just removing 'either' from the beginning)?

tworec: ok

golemcli.py
tworec: I've done it here https://github.com/golemfactory/golem/pull/2734/commits/57c36cb8410f84581757408352f1376e6d20255e

golem/manager/nodestatesnapshot.py
tworec: it is same Q as above regarding details

golem/client.py
etam: fixed

golem/task/taskkeeper.py
etam: fixed

golem/client.py
tworec: 👍

golem/client.py
```diff
@@ -1264,21 +1265,41 @@ def connection_status(self):
...
+            }
+
+        # trying to get subtask from task
+        waiting_for_task: Optional[str] = task_computer.waiting_for_task
+        if waiting_for_task is not None:
+            return {
+                'status': 'waiting for task',
+                'task_id': waiting_for_task,
```
tworec: the `waiting_for_task ` var holds `None` or `task_id`, hence name you've proposed for it is misleading. I know `task_computer` uses this name, but it is also misleading. Please clarify that by changing to eg. `task_id_waited_for`

golem/interface/client/provider.py
```diff
@@ -0,0 +1,7 @@
+from golem.core.deferred import sync_wait
+from golem.interface.command import command
+
+
+@command(help="Show provider status", root=True)
+def provider_status():
```
tworec: I would do a `provider` class/group with the `status` method/command. Than cli invocation would have form:
```
golemcli provider status
```

golem/client.py
```diff
@@ -1264,21 +1265,41 @@ def connection_status(self):
...
-            for k, v in list(progress.items()):
-                msg = "{} \n {} ({}%)\n".format(msg, k, v.get_progress() * 100)
-        elif self.config_desc.accept_tasks:
-            msg = "Waiting for tasks...\n"
-        else:
-            msg = "Not accepting tasks\n"
-
-        peers = self.p2pservice.get_peers()
+    def get_provider_status(self) -> Dict[str, Any]:
+        # golem is starting
```
tworec: those comments are not necessary, as you provide string result saying the same

tworec (approved):
LGTM (stylistics comments only)

tworec:
One general comment: please do not use force push after review was made, because it hinder next review phase. Rather add new commits and push them

golem/interface/client/provider.py
etam: As long as there are no other `golemcli provider ...` commands, I see no reason to do this.

golem/client.py
etam: Yes, but they make sections of the code more visually separated.

golem/client.py
etam: done

golem/interface/client/provider.py
tworec: to be consistent with other commands

etam:
RPC change: ~~new method `provider.status`, takes no arguments~~ dict returned by method 'comp.tasks.stats' (get_task_stats) is extended with a new key 'provider_state' with possible values:

```python
{
    'status': 'golem is starting'
}
```
```python
{
    'status': 'computing',
    'subtask': ComputingSubtaskStateSnapshot as dict
}
```
```python
{
    'status': 'waiting for task',
    'task_id_waited_for': str(task_id),
}
```
```python
{
    'status': 'not accepting tasks',
}
```
```python
{
    'status': 'idle',
}
```

mdtanrikulu:
@badb I'm unable to test it. 

Whenever I try to run task from any other instance I have following error on provider node;
```
INFO     [golem.task.taskcomputer            ] Starting computation of subtask '3089a422-86a5-11e8-a5cb-ec276b3f4294' (task: '2803ddfe-86a5-11e8-a44b-ec276b3f4294', deadline: 1531526453.387266, docker images: [{'image_id': None, 'repository': 'golemfactory/blender', 'tag': '1.4'}])
WARNING  [golem.docker.task_thread           ] Task stderr:
/usr/bin/python: can't open file '/golem/work/job.py': [Errno 2] No such file or directory
```

golem/client.py
mdtanrikulu: None of them are necessary in `subtask_progress.__dict__` for now, but good to see what we have as information about computed task. We can consider to use at least some of them to show to providers for the future.

mdtanrikulu:
Some time ago, I implemented the same logic for `get_host_state` function here at #2788 instead of creating another RPC, maybe we can consider to move this function into that, so we can ask for the provider stats and status at the same time?

etam:
> Some time ago, I implemented the same logic for get_host_state function here at #2788 instead of creating another RPC, maybe we can consider to move this function into that, so we can ask for the provider stats and status at the same time?

Done

mdtanrikulu (approved):

PR #2818, opened 2018-05-25
Concent integration deposit

jiivan:


golem/task/tasksession.py
```diff
@@ -820,22 +821,15 @@ def _react_to_ack_report_computed_task(self, msg):
             ack_report_computed_task=msg,
         )
         logger.debug('[CONCENT] ForceResults: %s', delayed_forcing_msg)
-        report_computed_task = get_task_message(
-            'ReportComputedTask',
-            msg.task_id,
-            msg.subtask_id,
+        ttc_deadline = datetime.datetime.utcfromtimestamp(
```
shadeofblue, line 824: hmm... @jiivan wasn't the `report_computed_task` validation/logging needed here?

shadeofblue (approved):
looks good as far as I can tell

golem/task/tasksession.py
jiivan: RCT was fetched from history only because it was historically absent in `AckReportComputedTask` back in the days.

shadeofblue: ah, gotcha

PR #2821, opened 2018-05-29
Stop geth sync on client.quit

### LINKED ISSUE #2777: Stop geth synchronization on exit
Calling `stop` on Client doesn't stop the Client when synchronization with Geth is in process.

maaktweluit:
Resolves #2777


Krigpl (approved):
verynicegreatsuccess

mfranciszkiewicz (approved):

PR #2822, opened 2018-05-29
2193 react to subtask results settled

### LINKED ISSUE #2193: [Concent] UC4 / Additional Verification - react to `SubtaskResultsSettled`
implement the reaction to the `SubtaskResultsSettled` message as both the Provider, and as a Requestor, effectively ending the subtask's life.

* status for subtask that `SRS` was registered and the payment will be settled by the Concent

shadeofblue:
closes #2193

### DIFF
--- a/golem/transactions/incomeskeeper.py
+++ b/golem/transactions/incomeskeeper.py
@@ -8,7 +8,7 @@ from ethereum.utils import denoms
 from pydispatch import dispatcher
 
 from golem.core.variables import PAYMENT_DEADLINE
-from golem.model import Income, IncomeOrigin
+from golem.model import Income
 from golem.utils import encode_hex, pubkeytoaddr
 
 logger = logging.getLogger("golem.transactions.incomeskeeper")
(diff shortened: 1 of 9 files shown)


jiivan (approved):
lgtm

golem/transactions/incomeskeeper.py
```diff
@@ -104,6 +104,18 @@ def reject(subtask_id):
             subtask_id=subtask_id
         )
 
+    @staticmethod
+    def settled(subtask_id, settled_ts):
+        try:
+            income = Income.get(subtask=subtask_id)
```
Krigpl: I dunno what's the current state of the uniqueness of the subtask ids, but bulletproof way of grabbing the correct income is using pair `(sender, subtask_id)`, like in `update_awaiting` method.

Krigpl (approved):
obama_approves

golem/transactions/incomeskeeper.py
shadeofblue: @Krigpl good point indeed, sir, good point

PR #2824, opened 2018-05-29
Provider deposit for additional verification in Concent

jiivan:


golem/task/tasksession.py
```diff
@@ -713,15 +713,42 @@ def _react_to_subtask_results_rejected(
...
+                self.concent_service.submit_task_message(
+                    subtask_id=msg.subtask_id,
+                    msg=srv,
+                )
+
+            # pylint: disable=protected-access
+            self.task_server.client.transaction_system._sci \
+                .on_transaction_confirmed(
+                    tx_hash=thash,
+                    required_confs=3,
```
shadeofblue: @jiivan shouldn't the `3` be some kind of a `golem.core.variables` setting? ... or something?

shadeofblue (approved):
dla mnie bomba, ale niech jeszcze @Krigpl spojrzy

golem/task/tasksession.py
```diff
@@ -713,15 +713,42 @@ def _react_to_subtask_results_rejected(
...
+                srv = message.concents.SubtaskResultsVerify(
+                    subtask_results_rejected=msg
+                )
+
+                self.concent_service.submit_task_message(
+                    subtask_id=msg.subtask_id,
+                    msg=srv,
+                )
+
+            # pylint: disable=protected-access
```
Krigpl: Please don't do that. There's a reason for it.
Make `concent_deposit` take a callback instead.

Krigpl (changes requested):

golem/task/tasksession.py
```diff
@@ -713,14 +713,39 @@ def _react_to_subtask_results_rejected(
...
+                srv = message.concents.SubtaskResultsVerify(
+                    subtask_results_rejected=msg
+                )
+
+                self.concent_service.submit_task_message(
+                    subtask_id=msg.subtask_id,
+                    msg=srv,
+                )
+
+            def transaction_receipt(receipt):
```
Krigpl: Since the goal is just to call `ask_for_verification` after there's enough GNT deposited I'd say you could move this one inside `concent_deposit` and hide the receipt check there.
In other words just invoke callback immediately if there's enough GNT otherwise use `on_transaction_confirmed`. And then you don't need to return `tx_hash` from `concent_deposit` as well.

The idea is that `tasksession` shouldn't really care about any blockchain implementation details, there belong in `EthereumTransactionSystem` better.

Krigpl (changes requested):

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -188,6 +192,12 @@ def concent_deposit(
             max_possible_amount,
```
Krigpl: You need to divide this amount by `denoms.ether`

jiivan: Are you sure? Looks like it's an int value in golem_sci:
https://github.com/golemfactory/golem-smart-contracts-interface/blob/e53673630f107c3f03c185dc729a971eff71a40e/golem_sci/implementation.py#L536-L537

Krigpl: Yeah, this is the log message and it easier to read 1 GNT instead of 1000000000000000000

Krigpl: But yeah. you don't `need to`, but it's way better

golem/task/tasksession.py
jiivan: @Krigpl done.

golem/transactions/ethereum/ethereumtransactionsystem.py
jiivan: gotcha

golem/task/tasksession.py
```diff
@@ -714,14 +714,29 @@ def _react_to_subtask_results_rejected(
...
+                    msg=srv,
+                )
+
+            thash = self.task_server.client.transaction_system.concent_deposit(
+                required=amount,
+                expected=expected,
+                reserved=self.task_server.client.funds_locker.sum_locks()[0],
+                cb=ask_for_verification,
             )
+            if thash is None:
```
Krigpl: You don't need that anymore, right?

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -174,24 +178,45 @@ def concent_balance(self) -> int:
         )
 
     def concent_deposit(
-            self, required: int, expected: int, reserved: int) -> Optional[str]:
+            self,
+            required: int,
+            expected: int,
+            reserved: int,
+            cb=None) -> Optional[str]:
```
Krigpl: Fix return type

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -174,24 +178,45 @@ def concent_balance(self) -> int:
...
         )
-        return tx_hash
+
+        def transaction_receipt(receipt):
+            if not receipt.status:
+                log.warning(
+                    "Deposit failed. Receipt: %r",
+                    receipt,
+                )
+                raise exceptions.TransactionFailed(receipt)
```
Krigpl: `on_transaction_confirmed` callbacks should not throw - there's no error handling for them. Just log the message (even with `critical` severity).

Krigpl (changes requested):
lgtm, just a bit of clean up

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -174,24 +178,47 @@ def concent_balance(self) -> int:
...
-        return tx_hash
+
+        def transaction_receipt(receipt):
+            if not receipt.status:
+                log.critical(
+                    "Deposit failed. Receipt: %r",
+                    receipt,
+                )
+                return
+            cb(receipt=receipt)
```
Krigpl: Why do you pass the receipt if it's not needed for anything ?

jiivan: It's needed to know whether transaction succeeded or not. See  scripts/concent_integration_tests/force_accept/test_requestor_doesnt_send.py

Krigpl: It's a test. I don't see production code using the receipt which is a code smell. Depending on what the test checks you can either assume the transaction was successful or check the deposit amount whether it increased this way knowing if it was successful.

jiivan: Done

Krigpl (approved):
obama_approves

PR #2825, opened 2018-05-29
run default benchmark once

tworec:
This PR adds running `DEFAULT` benchmark initially and only once, since it always uses single CPU core, and resource re-allocation (cores, RAM, disk) does not influence its value.  

This PR also includes benchmark code refactor and some tests clean-up.

Fixes golemfactory/golem-electron#144


golem/client.py
```diff
@@ -1150,23 +1148,11 @@ def get_environments(self):
...
-            kwargs = {'func': DefaultEnvironment.run_default_benchmark,
-                      'callback': deferred.callback,
-                      'errback': deferred.errback,
-                      'num_cores': self.config_desc.num_cores,
-                      'save': True}
-            Thread(target=callback_wrapper, kwargs=kwargs).start()
-            result = yield deferred
-            returnValue(result)
+        result = yield deferred
+        returnValue(result)
```
etam: Minor: according to `inlineCallbacks` documentation, since python 3.3 you can use regular `return` instead of `returnValue`.

golem/environments/environment.py
```diff
@@ -154,10 +155,13 @@ def get_source_code(self):
                 return f.read()
 
     @classmethod
-    def run_default_benchmark(cls, num_cores=1, save=False):
+    def run_default_benchmark(cls, save=False):
```
etam: Is this still used? The last usage of it was in `Client.run_benchmark`, which you removed, so it seems it's just a dead code.

tests/golem/task/test_benchmarkmanager.py
```diff
@@ -10,17 +11,57 @@
...
 
-        assert not b.benchmarks_needed()
+        Performance.update_or_create(Environment.get_id(), 3)
+
+        # then
+        assert not self.b.benchmarks_needed()
+
+    @patch("golem.task.benchmarkmanager.BenchmarkManager.run_default_benchmark")
+    @patch("golem.task.benchmarkmanager.BenchmarkRunner")
+    def test_run_all_benchmarks(self, br_mock, rdb_mock, *_):
```
etam: Is this `, *_` needed?

mfranciszkiewicz (approved):

etam (approved):

PR #2831, opened 2018-05-30
Disable sentry logger early (b0.16.0)

mfranciszkiewicz:
Sentry logger is currently enabled between calling `config_logging` and starting the client.


maaktweluit (approved):
LGTM!

PR #2848, opened 2018-06-05
PoC firejail

RDProjekt:


apps/registered_envs.ini
```diff
@@ -0,0 +1,4 @@
+[Blender]
+envs=apps.blender.dockerenvironment.blenderenvironment.BlenderEnvironment();
```
Elfoniok, line 2: So provider need to modify this manually?

apps/blender/firejailenvironment/task_thread.py
```diff
@@ -0,0 +1,181 @@
...
+        ]
+        return cmd
+
+    def _format_firejail_cmd(self):
+        command = [
+            FIREJAIL_COMMAND,
+            f'--profile={self.profile_path}'
+        ]
+        if self.memory_limit:
+            memory_limit_in_bytes = self.memory_limit * 1024
```
Elfoniok: this can be double and firejail doesn't  support it

golem/environments/environmentsmanager.py
```diff
@@ -1,83 +1,158 @@
...
+            -> Optional[Environment]:
+        """ Return environment suitable for given task_type and requirements.
+        Takes into account environments 'accept_tasks' and support
+        status, returns first Environment for given task_type that is enabled
+        and satisfies all the requirements
+        :param task_type: type of task
+        :param requirements: list of requirements which must be satisfied
+        :return: environment or None
+        """
+        envs = (e for e in self.environments_for_tasks.get(task_type, [])
```
Elfoniok, line 134: why no requirements here? What are purpose of it then? Could you provide example?

Elfoniok (changes requested):

golem/environments/environmentsmanager.py
Elfoniok: Shouldn't requirements from task definition be passed here instead of empty list?

golem/task/localcomputer.py
```diff
@@ -168,19 +179,31 @@ def _prepare_tmp_dir(self):
...
         os.makedirs(self.tmp_dir)
 
-    def _get_task_thread(self, ctd: ComputeTaskDef) -> DockerTaskThread:
-        return DockerTaskThread(
+    def _get_task_thread(self, ctd: ComputeTaskDef) -> TaskThread:
+        environment = self.environment
+        if self.environments_manager:
+            # TODO No requirements checks for local computer?
+            environment = self.environments_manager\
+                .get_environment_for_task(ctd['task_type'], [])
```
Elfoniok, line 187: empty requirements list?

golem/environments/environmentsmanager.py
RDProjekt: this is get() on a map, so empty list is the default value if no environments were reqistered for this type. Requirements are checked 3 lines below.

apps/registered_envs.ini
RDProjekt: Yes, if the Provider wishes to prefer other environments, that list has to be modified manually. It does not affect tasks which have specific requirements not satisfied by this environment though.

apps/core/task/coretask.py
```diff
@@ -118,34 +117,24 @@ def __init__(self,
...
+                    src_code = src_file.read()
+            except OSError as err:
+                logger.warning("Wrong main program file: %s", err)
 
         th = TaskHeader(
+            task_type=task_definition.task_type,
             task_id=task_definition.task_id,
-            environment=self.environment.get_id(),
             task_owner=owner,
+            requirements=task_definition.requirements,
```
Elfoniok, line 132: So requirements are not passed in message?

golem/task/localcomputer.py
```diff
@@ -36,7 +40,14 @@ def __init__(self,
...
         self.success = False
         self.lock = Lock()
-        self.tt: Optional[DockerTaskThread] = None
+        if (environments_manager and environment)\
+                or (not environments_manager and not environment):
+            raise AttributeError("Either environments_manager or a single"
+                                 " environment is required for LocalComputer"
+                                 " to run")
+        self.environments_manager = environments_manager
+        self.environment = environment
```
Elfoniok, line 49: Whats the point of passing this? Shouldn't  environments manager be used for getting environment for specific ctd?

golem/task/localcomputer.py
```diff
@@ -168,19 +179,31 @@ def _prepare_tmp_dir(self):
             shutil.rmtree(self.tmp_dir, True)
         os.makedirs(self.tmp_dir)
 
-    def _get_task_thread(self, ctd: ComputeTaskDef) -> DockerTaskThread:
-        return DockerTaskThread(
+    def _get_task_thread(self, ctd: ComputeTaskDef) -> TaskThread:
+        environment = self.environment
```
Elfoniok, line 183: Once again why this if we are using manager?

golem/task/localcomputer.py
```diff
@@ -190,10 +213,12 @@ def __init__(self):
         self.computer = None
 
     def start_computation(self, root_path, success_callback, error_callback,
-                          compute_task_def, resources, additional_resources):
+                          environment, compute_task_def, resources,
+                          additional_resources):
         self.computer = LocalComputer(root_path=root_path,
                                       success_callback=success_callback,
                                       error_callback=error_callback,
+                                      environment=environment,
```
Elfoniok, line 221: why not environments manager?

golem/task/taskserver.py
```diff
@@ -75,14 +75,13 @@ def __init__(self,
...
-        benchmarks = self.task_manager.apps_manager.get_benchmarks()
+        benchmarks = client.environments_manager.get_benchmarks()
         self.benchmark_manager = BenchmarkManager(config_desc.node_name, self,
                                                   client.datadir, benchmarks)
-        udmm = use_docker_manager
         self.task_computer = TaskComputer(
             config_desc.node_name,
             task_server=self,
-            use_docker_manager=udmm)
+            use_docker_manager=use_docker_manager)
```
Elfoniok, line 84: What is the point of using docker manager, looks like Task Computer should be agnostic if it can use different environments, and some of them will not use docker at all.

golem/task/localcomputer.py
RDProjekt: As in the TODO - it's actually not clear if requirements should be passed here.

The problem is that LocalComputer is used for local computation, including verification and creating reference data. For that reason it should be able to compute given tasks.

So the question is - should it be possible to create a task that requestor cannot compute itself, e.g. require CUDA rendering engine when the requestor does not have it? We've chosen 'yes' as the answer because it makes golem way more powerful for end-users. But that means that LocalComputer should ignore requirements and select any environment for the task - at least one such environment must be present.

apps/core/task/coretask.py
RDProjekt: "message" as in "ctd"? No - requirements are added to TaskHeader to that providers have them at hand from the beginning and so can decide if they support the task before possibly asking for it (see TaskHeaderKeeper.checkEnvironment(header))

golem/task/taskserver.py
RDProjekt: In general DockerManager is still required because some (most) environments require working docker setup. But I do see your point that 1. docker manager might not be required if all configured environments are not docker-based, 2. maybe TaskComputer is not the best place to have it.

Moving it out of TaskComputer would require decision on "move to where". There is no better suited place right now, so we would have to create a whole new mechanism of setting up executables based on golem's configuration.

golem/task/localcomputer.py
RDProjekt: When computation is requested straight from a task (e.g. in luxrender __generate_final_file ) there is no environments_manager reference at hand. Passing explicit environment is a workaround, as more robust solution would require quite a lot of debate and refactoring:

* environments_manager must be globally referencable, manually passing it from one entity to another doesn't work in that case.
* discussion what exactly are tasks and what should they carry - e.g. in blender verifier calculacting verification scores is done by explicitly running a docker job. But in luxrender merging flm results is done by running an artifical (created on spot, not received from external requestor) ctd with different script_src - that way it runs in the same docker as standard luxrender task.
* should tasks locally be computed the same way as global ones (i.e. including requirements) or do we allow local computation to differ. If they can differ - should local computation environment be hardcoded, configured by a separate config, or reuse environments config?

golem/task/localcomputer.py
RDProjekt: see reply above

golem/task/localcomputer.py
RDProjekt: see reply above

golem/task/localcomputer.py
Elfoniok: Ok, i suspected that but this should be noted clearly.

golem/environments/environmentsmanager.py
Elfoniok: Right, thanks!

apps/registered_envs.ini
Elfoniok: Ok

apps/core/task/coretask.py
Elfoniok: No i was asking about Tasks Message. But its all right.

golem/task/localcomputer.py
Elfoniok: Ok this is also important to note.

golem/task/taskserver.py
Elfoniok: Ok

Elfoniok:
I will accept for slow test to run.

Elfoniok (approved):

Elfoniok (changes requested):
First of all there are some lint errors to be fixed.

Second, this changes shouldn't break other platforms. Ignoring errors on removing files, is not a good solution. Files are not deleted and removing them raise errors because they are probably open. This wreak havoc on Mac OS with very low file descriptors.

Third there are 2 failing tests on Windows.

Elfoniok (approved):

Elfoniok:
Two offending test are due to docker images not being available on MacOs and Windows buildbots. Since DummyTaskEnvironment inherits from DockerEnvironment it is reasonable to not support those if docker images are not available, but this also force us to ci_skip this tests.

RDProjekt:
Actually we want to try mocking part of DockerEnvironment instead of using ci_skip. Failing tests do not really need Docker to run, and both of them are testing fairly important pieces of functionality, so it would be good to leave them running.

Elfoniok:
borat

PR #2852, opened 2018-06-06
Fix shutting down golem monitor

Elfoniok:


golem/monitor/monitor.py
```diff
@@ -37,10 +37,16 @@ def send(self, o):
...
         while not self.stop_request.isSet():
             try:
-                msg = self.queue.get(True, self.monitor_sender_thread_timeout)
-                self.sender.send(msg)
+                if not msg:
+                    msg = self.queue.get(True, 1)
+                if time.time() - last_send > self.monitor_sender_thread_timeout:
+                    last_send = time.time()
+                    self.sender.send(msg)
+                    msg = None
```
Krigpl: I think it makes sense to put `else time.sleep(1)` or something similar here, otherwise this loop may become wasteful active waiting loop (when `msg` is not `None` but `timeout` is not reached yet

Elfoniok: Yeas actually i was thinking the same, but somehow hesitated

tests/golem/monitor/test_monitor.py
```diff
@@ -79,7 +79,7 @@ def check(f, msg_type):
                 + 'DefaultHttpSender.post_json') \
                     as mock_send:
                 f()
-                time.sleep(0.005)
+                time.sleep(0.05)
```
tworec: why do we need increasing this?

tworec (changes requested):

golem/monitor/monitor.py
jiivan: I think it's proper to use `.get_nowait()`. My proposition:
```python
last_sent = 0
while not event.isSet():
    if time.time() - last_sent < self.monitor_sender_thread_timeout:
        time.sleep(veryshort)
        continue
    try:
        msg = self.queue.get_nowait()
    except queue.Empty:
        msg = self.node_info
    last_sent = time.time()
    self.sender.send(msg)
```

jiivan (commented):
I propose a different algorithm.

golem/monitor/monitor.py
Krigpl: @jiivan Your way changes the logic to send at most 1 message per time interval which wasn't the case before. I dunno what the requirements are for the monitor but I imagine there might be cases where we end up sending a bunch of messages at the same moment.

Another way I see is to just put `None` in the queue in the `join` method and break the above loop if we retrieve it from the queue (so 3 lines change).

jiivan: @Krigpl AFAIK it's the intention.

Elfoniok:
What about ping messages?

tests/golem/monitor/test_monitor.py
Elfoniok: i think because of time needed for additional logic in send

golem/monitor/monitor.py
```diff
@@ -40,13 +40,16 @@ def run(self):
         while not self.stop_request.isSet():
             try:
                 msg = self.queue.get(True, self.monitor_sender_thread_timeout)
+                if not msg:
+                    continue
```
Krigpl: If you use `break` here you can remove `self.stop_request` altogether

Krigpl (approved):
verynicegreatsuccess

golem/monitor/monitor.py
Elfoniok: I though about it but, i cant be sure that none is set only on exit, who nows what gets in here. It might be that we don't want to break when getting none because of some possible bug or feature.

golem/monitor/monitor.py
```diff
@@ -40,13 +40,16 @@ def run(self):
         while not self.stop_request.isSet():
             try:
                 msg = self.queue.get(True, self.monitor_sender_thread_timeout)
+                if not msg:
```
jiivan: Better use `if msg is None:`, because many objects can have False boolean value. As you said it can be a future bug or feature :)

jiivan (approved):

tworec (approved):

mfranciszkiewicz (approved):

PR #2853, opened 2018-06-06
Remove unnecessary mutex in PaymentProcessor

Krigpl:
Everything's happening from twisted thread.


mfranciszkiewicz (approved):
`PaymentProcessor` isn't thread-safe anyway.

PR #2854, opened 2018-06-07
Change default ethereum logging level

Krigpl:
To get rid of annoying `DEBUG:eth.chain.tx	deserialized tx tx=b'103006db'` messages,


maaktweluit (approved):
silence

PR #2856, opened 2018-06-07
Upgrade to better SCI

### LINKED ISSUE #2414: Consider monitoring the latest block in Payment Processor
Right now we observed the pending block. Consider unlocking funds only after the payment is in the latest block. 

https://github.com/golemfactory/golem/blob/develop/golem/ethereum/paymentprocessor.py#L263

Krigpl:
See https://github.com/golemfactory/golem-smart-contracts-interface/pull/30

Resolves #2414

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -8,6 +8,7 @@ from threading import Lock
 from sortedcontainers import SortedListWithKey
 from ethereum.utils import denoms
 from pydispatch import dispatcher
+from twisted.internet import threads
 
 import golem_sci
 from golem.core.variables import PAYMENT_DEADLINE
@@ -35,11 +36,7 @@ def _make_batch_payments(payments: List[Payment]) -> List[golem_sci.Payment]:
     return res
 
 
-# pylint: disable=too-many-instance-attributes
 class PaymentProcessor:
-    # Minimal number of confirmations before we treat transactions as done
-    REQUIRED_CONFIRMATIONS = 6
-
     CLOSURE_TIME_DELAY = 2
     # Don't try to use more than 75% of block gas limit
     BLOCK_GAS_LIMIT_RATIO = 0.75
@@ -48,15 +45,16 @@ class PaymentProcessor:
         self.ETH_BATCH_PAYMENT_BASE = \
             sci.GAS_PRICE * sci.GAS_BATCH_PAYMENT_BASE
         self._sci = sci
-        self._eth_reserved = 0
         self._gntb_reserved = 0
         self._awaiting = SortedListWithKey(key=lambda p: p.processed_ts)
-        self._inprogress: Dict[str, List[Payment]] = {}  # Sent transactions.
         self.load_from_db()
 
     @property
     def reserved_eth(self) -> int:
-        return self._eth_reserved + self.ETH_BATCH_PAYMENT_BASE
+        if not self._awaiting:
+            return 0
+        return self.ETH_BATCH_PAYMENT_BASE + \
+            len(self._awaiting) * self.get_gas_cost_per_payment()
 
     @property
     def reserved_gntb(self) -> int:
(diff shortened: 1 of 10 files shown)


golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -232,9 +227,13 @@ def _get_gnt_from_faucet(self) -> None:
             return
         if self._eth_balance < 0.001 * denoms.ether:
             return
-        if self._gnt_balance + self._gntb_balance < 100 * denoms.ether:
+        if self._gnt_balance + self._gntb_balance < 100 * denoms.ether and\
```
mfranciszkiewicz: Was a single donation per application run intended?

golem/ethereum/paymentprocessor.py
```diff
@@ -69,18 +67,64 @@ def get_gas_cost_per_payment(self) -> int:
...
+            receipt,
+            fee / denoms.ether,
+        )
+        for p in payments:
+            p.status = PaymentStatus.confirmed  # type: ignore
+            p.details.block_number = receipt.block_number
+            p.details.block_hash = receipt.block_hash[2:]
+            p.details.fee = fee
+            p.save()
+            self._gntb_reserved -= p.value
```
mfranciszkiewicz, line 115: Unguarded updates of `_gntb_reserved` are fine for now, but are prone to future errors.

golem/transactions/ethereum/ethereumtransactionsystem.py
Krigpl: This allows multiple donations per single run when you run out of GNT. Although it's not exactly the logic I intended, lemme change that.

golem/ethereum/paymentprocessor.py
Krigpl: Not sure what you meant but if we ever change the threading model in Golem not to run everything from Twisted when many more places in Golem will have the same problem.

golem/ethereum/paymentprocessor.py
```diff
@@ -69,18 +67,64 @@ def get_gas_cost_per_payment(self) -> int:
...
             for awaiting_payment in Payment \
                     .select() \
                     .where(Payment.status == PaymentStatus.awaiting):
                 self.add(awaiting_payment)
 
+    def _on_batch_confirmed(self, payments: List[Payment], receipt) -> None:
+        if not receipt.status:
+            log.critical("Failed batch transfer: %s", receipt)
+            for p in payments:
+                p.status = PaymentStatus.awaiting  # type: ignore
```
maaktweluit, line 95: Why do we need the `# type: ignore`?

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -243,34 +237,17 @@ def _balance_known(self) -> bool:
...
         addr = self._sci.get_eth_address()
 
-        eth_balance = self._sci.get_eth_balance(addr)
-        if eth_balance is not None:
-            self._eth_balance = eth_balance
-            self._last_eth_update = time.mktime(datetime.today().timetuple())
-        else:
-            log.warning("Failed to retrieve ETH balance")
+        self._eth_balance = self._sci.get_eth_balance(addr)
+        self._last_eth_update = time.mktime(datetime.today().timetuple())
```
maaktweluit, line 241: this could be calculated once as f.i. `refresh_time` and used for both last_update values

maaktweluit (commented):
LGTM!

Some small comments, will approve on reply!

golem/ethereum/paymentprocessor.py
Krigpl: `mypy` complains because `p.status` is of type `PaymentStatusField` so the assignment is not compatible

golem/transactions/ethereum/ethereumtransactionsystem.py
Krigpl: Yeah, but it doesn't really matter, does it? I aiming to get rid of these values anyway and replace them with more useful one (geth's last block number/timestamp)

mfranciszkiewicz (approved):

golem/ethereum/paymentprocessor.py
maaktweluit: clear, thanks :D

golem/transactions/ethereum/ethereumtransactionsystem.py
maaktweluit: haha, not very much.. but every computation saved helps
to use blocks for this sounds better for sure, will approve!

PR #2860, opened 2018-06-08
Simplify FundsLocker a bit

Krigpl:
`TaskFundsLock` doesn't need to know about `transaction_system`.
Also fixed corresponding test.

### DIFF
--- a/tests/golem/transactions/ethereum/test_fundslocker.py
+++ b/tests/golem/transactions/ethereum/test_fundslocker.py
@@ -3,7 +3,6 @@ from unittest import mock
 
 from golem.core.common import timeout_to_deadline
 from golem.core.variables import PAYMENT_DEADLINE
-from golem.task.taskkeeper import compute_subtask_value
 from golem.testutils import TempDirFixture
 from golem.transactions.ethereum.fundslocker import (logger, FundsLocker,
                                                      TaskFundsLock)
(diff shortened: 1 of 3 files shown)


tests/golem/transactions/ethereum/test_fundslocker.py
```diff
@@ -13,9 +12,9 @@ class TestFundsLocker(TempDirFixture):
     def setUp(self):
         super().setUp()
         self.ts = mock.MagicMock()
-        self.ts.eth_for_batch_payment.return_value = 13000
+        self.ts.eth_for_batch_payment.side_effect = lambda n: n * 13000
```
maaktweluit, line 15: These values could be stored in a shared variable for easy updates later on.

They could even link to the constants in SCI, any reason you did not?

maaktweluit (approved):
LGTM!

One small q, not blocking...

tests/golem/transactions/ethereum/test_fundslocker.py
Krigpl: Yeah, that wasn't me who introduced this code :)

Krigpl: But yeah, these could be non magic numbers - someday...

PR #2862, opened 2018-06-08
Show precise balance in CLI

### LINKED ISSUE #2735: Rounding of GNT balances
When I try to do "golemcli account withdraw BAL ADDRESS GNT" with BAL being my available GNT balance and replacing ADDRESS with my valid Ethereum address, I get an unknown exception. This is due to rounding. The available balance is displayed as 4.916667 GNT, but I really have 4.916666999999999998 GNTB, according to Etherscan. To fix this, you should probably round down GNT balances. I am on the latest version.

Krigpl:
Fixes #2735
Sample output:
```
  available_balance: 938.303333333333333332 GNT
  eth_address: '0xA84CA4E4119F677431a51DB85F3567f2B59B4C2A'
  eth_balance: 0.019257347 ETH
  eth_locked: 0.00054 ETH
  gnt_locked: 0 GNT
  reserved_balance: 0 GNT
  total_balance: 938.303333333333333332 GNT
```

### DIFF
--- a/golem/interface/client/account.py
+++ b/golem/interface/client/account.py
@@ -44,12 +44,12 @@ class Account:
         gnt_balance = balance['gnt']
         gnt_available = balance['av_gnt']
         eth_balance = balance['eth']
-        gnt_balance = float(gnt_balance)
-        gnt_available = float(gnt_available)
-        eth_balance = float(eth_balance)
+        gnt_balance = int(gnt_balance)
+        gnt_available = int(gnt_available)
+        eth_balance = int(eth_balance)
         gnt_reserved = gnt_balance - gnt_available
-        gnt_locked = float(balance['gnt_lock'])
-        eth_locked = float(balance['eth_lock'])
+        gnt_locked = int(balance['gnt_lock'])
+        eth_locked = int(balance['eth_lock'])
 
         return dict(
             node_name=node['node_name'],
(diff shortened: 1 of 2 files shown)


etam (changes requested):
TestAccount.test_show fails

golem/interface/client/account.py
```diff
@@ -44,12 +44,12 @@ def info(self) -> Dict[str, Any]:  # pylint: disable=no-self-use
         gnt_balance = balance['gnt']
```
etam: what is the type we originally get here? string? int?

golem/interface/client/account.py
```diff
@@ -117,4 +117,8 @@ def withdraw(  # pylint: disable=no-self-use
 
 
 def _fmt(value: float, unit: str = "GNT") -> str:
```
etam: `value` is not a float anymore.

golem/interface/client/account.py
Krigpl: string

etam: Then maybe there's no need to convert it to int and back to string?

Krigpl: Calculating `gnt_reserved` requires `ints`

Krigpl: Also `_fmt` expects `ints` for division

etam: > Calculating `gnt_reserved` requires `ints`

That's fair

> Also `_fmt` expects `ints` for division

That could be done by inserting a dot a a specific place in string. (I'm not saying it must, but could)

Krigpl: > That could be done by inserting a dot a a specific place in string. (I'm not saying it must, but could)

With value less than 10^18 it would be awkward.

etam (approved):
ok

PR #2866, opened 2018-06-08
Network masking for tasks

### LINKED ISSUE #2706: Masking for tasks
Right now network is very asymmetrical - there are many more providers than requestors. As a result requestors may be flooded with job offers from providers when they create a new task.
To mitigate this we may want to add masks that will allow providers from only one subset of the network to show their job offers for given task. Masking should be based on nodes’ ids. We should start from generally small set of providers and increase the size of subset by changing the mask if we don't get number of offers from providers that is enough in given tour. 
The specific parameters for timeouts for a given tour that must be set down as a result of experiments and take into consideration task deadline.

Wiezzel:
Closes #2706


golem/client.py
```diff
@@ -556,6 +566,18 @@ def enqueue_new_task(self, task_dict):
...
             package_path, package_sha1 = packager_result
             task.header.resource_size = path.getsize(package_path)
+
+            # Minimum num_workers is 4 to avoid delayed start in case of
+            # task with very few subtasks (for small number of subtasks it is
+            # likable that initial mask would rule out all the nodes)
+            num_workers = max(
+                task.get_total_tasks() *
+                self.config_desc.initial_mask_size_factor,
+                4)
```
mfranciszkiewicz: Please replace the magic value with a constant

golem/client.py
```diff
@@ -1481,3 +1503,29 @@ def __init__(self,
...
+            update_num_bits: int
+    ) -> None:
+        self._task_manager: TaskManager = task_manager
+        self._update_num_bits = update_num_bits
+        self._interval = interval_seconds
+        super().__init__(interval_seconds)
+
+    def _run(self) -> None:
+        logger.info('Updating masks')
+        for task_id, task in self._task_manager.tasks.items():
```
mfranciszkiewicz: `tasks` may change during iteration (each `_run` call is delegated to a thread)

golem/task/taskbase.py
```diff
@@ -70,53 +73,159 @@ def __init__(self,  # pylint: disable=too-many-arguments
...
-        self_dict.pop('signature', None)
+        self_dict.pop('checksum', None)
 
         # "port_statuses" is a nested dict and needs to be sorted;
         # Python < 3.7 does not guarantee the same dict iteration ordering
-        port_statuses = self_dict['task_owner'].get('port_statuses')
-        if isinstance(port_statuses, dict):
-            self_dict['task_owner']['port_statuses'] = \
-                cls._ordered(port_statuses)
+        if 'task_owner' in self_dict:
```
mfranciszkiewicz: Maybe the `task_owner` entry should be mandatory and validated?

mfranciszkiewicz: It seems to already be the case in `validate`, where error messages expect a valid `task_owner`

golem/task/taskmanager.py
```diff
@@ -176,6 +176,28 @@ def add_new_task(self, task):
...
                                  persist=False)
 
+    @handle_task_key_error
+    def increase_task_mask(self, task_id: str, num_bits: int = 1) -> None:
+        """ Increase mask for given task i.e. make it more restrictive """
+        task = self.tasks[task_id]
+        try:
+            task.header.mask.increase(num_bits)
+        except AssertionError:
+            return  # Mask cannot be further increased
```
mfranciszkiewicz: Could you add some logs here? The same goes for `decrease_task_mask`. Now they both fail silently

golem/task/taskmanager.py
```diff
@@ -532,6 +554,10 @@ def verify_subtask(self, subtask_id):
         else:
             return False
 
+    def is_this_my_task(self, header: TaskHeader) -> bool:
```
mfranciszkiewicz, line 559: 👍

mfranciszkiewicz (commented):
Not an in-depth review yet

golem/client.py
Wiezzel: Good catch, thanks.

golem/task/taskmanager.py
Wiezzel: Yes, but that's not really an error. Just reaching max/min size. I thought it would be easier this way but I can as well add some check before increase/decrease and not throw any AssertionError at all.

golem/client.py
Wiezzel: Fixed.

golem/client.py
Wiezzel: Fixed.

golem/task/taskbase.py
Wiezzel: Fixed.

golem/task/taskmanager.py
Wiezzel: Fixed.

golem/client.py
```diff
@@ -1481,3 +1504,30 @@ def __init__(self,
...
+        self._task_manager: TaskManager = task_manager
+        self._update_num_bits = update_num_bits
+        self._interval = interval_seconds
+        super().__init__(interval_seconds)
+
+    def _run(self) -> None:
+        logger.info('Updating masks')
+        # Using list() because tasks could be changed by another thread
+        for task_id, task in list(self._task_manager.tasks.items()):
+            if task.needs_computation():
```
mfranciszkiewicz: You can decrease the level of nesting with
```
if not ...:
    continue
```

for both conditions.

golem/network/p2p/peerkeeper.py
```diff
@@ -189,6 +190,43 @@ def buckets_by_id_distance(self, key_num):
...
+        """
+        Get estimated network size
+        Based on https://gnunet.org/bartmsthesis p. 55
+        """
+        def depth(peer):
+            """ Get peer 'depth' i.e. number of common leading digits in binary
+            representations of peer's key and own key which is equivalent to the
+            position of the first '1' in (peer_key XOR own_key)"""
+            return format(
+                node_id_distance(peer, self.key_num),
```
mfranciszkiewicz: `self.k_size - int(math.log2( node_id_distance(peer, self.key_num) ))`

Wiezzel: It's `self.k_size - int(math.log2(node_id_distance(peer, self.key_num))) - 1` actually.

golem/task/taskmanager.py
```diff
@@ -176,6 +176,28 @@ def add_new_task(self, task):
                                  op=TaskOp.CREATED,
                                  persist=False)
 
+    @handle_task_key_error
+    def increase_task_mask(self, task_id: str, num_bits: int = 1) -> None:
```
badb, line 182: These two functions should be moved in the future to some external marketplace module. But that's just a comment for future refactor, it's probably not worth it to separate them now (in this PR especially).

golem/task/masking.py
```diff
@@ -0,0 +1,82 @@
...
+    ALL_BITS: Set[int] = set(range(MASK_LEN))
+
+    def __init__(self, byte_repr: bytes = b'\x00' * MASK_BYTES) -> None:
+        self.byte_repr = byte_repr
+
+    def increase(self, num_bits: int = 1) -> None:
+        bits = self.to_bits()
+        if len(bits) == self.MASK_LEN:
+            return  # Cannot increase any further
+        assert 0 <= num_bits <= self.MASK_LEN - len(bits)
```
badb: Can you replace asserts with if & Exceptions?

golem/client.py
Wiezzel: Fixed.

golem/network/p2p/peerkeeper.py
Wiezzel: Fixed.

golem/task/masking.py
Wiezzel: Sure. Actually I was thinking about replacing this with:
```python
num_bits = min(num_buts, self.MASK_LEN - len(bits))
if num_bits < 0:
    raise ValueError
```

golem/client.py
```diff
@@ -1481,3 +1504,33 @@ def __init__(self,
...
+        super().__init__(interval_seconds)
+
+    def _run(self) -> None:
+        logger.info('Updating masks')
+        # Using list() because tasks could be changed by another thread
+        for task_id, task in list(self._task_manager.tasks.items()):
+            if not task.needs_computation():
+                continue
+            task_state = self._task_manager.query_task_state(task_id)
+            if task_state.elapsed_time < self._interval:
```
badb, line 1533: Just to be clear - the purpose of this checking is just to make sure, that we don't decrease the mask in the first 30 seconds of task computation. Later it shold be regulated by looping call interval, right?

golem/task/masking.py
Wiezzel: Fixed and updated tests to cover it.

mfranciszkiewicz (approved):

golem/task/masking.py
```diff
@@ -0,0 +1,82 @@
...
+    def from_dict(cls, dict_repr: dict) -> 'Mask':
+        return Mask(**dict_repr)
+
+    @classmethod
+    def get_mask_for_task(
+            cls, desired_num_workers: int, network_size: int) -> 'Mask':
+        if network_size < 1:
+            return Mask()
+        num_bits = max(
+            -math.floor(math.log2(desired_num_workers / network_size)), 0)
```
badb, line 90: Are you sure that you want to use floor and not ceil here?

golem/task/masking.py
```diff
@@ -0,0 +1,88 @@
...
+        return (digest & self.to_int()) == 0
+
+    @classmethod
+    def _bits_to_bytes(cls, bits: Set[int]) -> bytes:
+        int_repr = reduce(ior, (1 << x for x in bits), 0)
+        return int_repr.to_bytes(cls.MASK_BYTES, 'big', signed=False)
+
+    @classmethod
+    def generate(cls, num_bits: int = 0) -> 'Mask':
+        assert 0 <= num_bits <= cls.MASK_LEN
```
badb: One more assert left

badb (approved):

golem/client.py
Wiezzel: Yea, exactly.

golem/task/masking.py
Wiezzel: Floor gives masks that are possibly more restrictive i.e. yield less work offers (that's because the logarithm value is negative). This what we want because we could extend the mask later if no sufficient number of workers appear.

golem/task/masking.py
Wiezzel: Ooops, will fix.

Wiezzel: Done.

PR #2873, opened 2018-06-11
Concent fullnode tests for UC1-UC4

shadeofblue:


jiivan (approved):
lgtm

scripts/concent_node_tests/nodes/provider/impatient_frct.py
```diff
@@ -0,0 +1,32 @@
...
+import datetime
+import mock
+import sys
+
+from golem_messages.message.concents import ForceReportComputedTask
+from scripts.concent_node_tests import params
+
+sys.path.insert(0, 'golem')
+
+from golemapp import start  # noqa: E402
```
maaktweluit: The name of the `#noqa` rule (instead of number ) would be more readable here, is that possible like with `pylint: disable`?

scripts/concent_node_tests/nodes/requestor/no_sra.py
```diff
@@ -0,0 +1,33 @@
...
+sys.path.insert(0, 'golem')
+
+from golemapp import start  # noqa: E402
+
+sys.argv.extend(params.REQUESTOR_ARGS)
+
+original_send = TaskSession.send
+
+
+def send(self, msg, send_unverified=False):
```
maaktweluit: wildcard args would make this more future proof, only not so nice when using `msg`.. up to you, nit comment..

scripts/concent_node_tests/params.py
```diff
@@ -0,0 +1,45 @@
+import os
+import tempfile
+
+
+def get_datadir(role: str):
```
maaktweluit, line 5: This could use `golem.testutils.TestDirFixture` ( maybe this needs the tests to be ran as `pytest` )

maaktweluit (approved):
LGTM!

Nice set of tests!

Left some small comments, all minor

PR #2891, opened 2018-06-18
Fix the --geth-address argument

mfranciszkiewicz:
- allow only `https://` node addresses
- fix `NodeProcess` argument ordering


Krigpl:
good-catch-hamster


> allow only https:// node addresses

FWIW there's https://github.com/golemfactory/golem/pull/2750 to allow both http and https (there is no sensitive data here so http is fine too).

Krigpl (approved):

PR #2897, opened 2018-06-18
Fix concent uc 2 for 16.1

shadeofblue:
Fixes issues with `ReportComputedTask`'s payload and concent file processing - required to fulfill Concent's "Forced Download" use case (UC2)

### DIFF
--- a/golem/task/result/resultmanager.py
+++ b/golem/task/result/resultmanager.py
@@ -93,12 +93,13 @@ class EncryptedResultPackageManager(TaskResultPackageManager):
         for resource in self.resource_manager.get_resources(
                 task_result.task_id):
             if file_name in resource.files:
-                return (resource.hash,
-                        file_path,
-                        sha1,
-                        package_size,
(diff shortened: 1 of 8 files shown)


badb (approved):

golem/task/result/resultmanager.py
```diff
@@ -84,13 +86,20 @@ def create(self, node, task_result, key_or_secret=None):
...
+        package_size = os.path.getsize(package_path)
 
         self.resource_manager.add_file(path, task_result.task_id)
         for resource in self.resource_manager.get_resources(
                 task_result.task_id):
             if file_name in resource.files:
-                return resource.hash, file_path, sha1, size
+                return (
+                    resource.hash,
+                    file_path,
```
mfranciszkiewicz: Could you rename this to something like `encrypted_package_path`?

mfranciszkiewicz (approved):
Thanks blue

PR #2901, opened 2018-06-20
Replace golem.utils.encode_hex

Krigpl:
No need for our own implementation is there's already one existing.
NOTE: `eth_utils.encode_hex` prepends `0x` while `golem.utils` doesn't.

### DIFF
--- a/golem/model.py
+++ b/golem/model.py
@@ -59,7 +59,7 @@ class RawCharField(CharField):
     """ Char field without auto utf-8 encoding."""
 
     def db_value(self, value):
-        return str(encode_hex(value)[2:])
+        return encode_hex(value)[2:]
 
     def python_value(self, value):
         return decode_hex(value)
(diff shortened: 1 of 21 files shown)


golem/model.py
```diff
@@ -59,7 +59,7 @@ class RawCharField(CharField):
     """ Char field without auto utf-8 encoding."""
 
     def db_value(self, value):
-        return str(encode_hex(value))
+        return str(encode_hex(value)[2:])
```
maaktweluit: does it matter here if the `str(` cast is before or after the slice `[2:]`?

maaktweluit (approved):
LGTM!

Nice cleanup!

Small q, probably not trivial

golem/model.py
Krigpl: Hm, the `str` cast in unnecessary even

etam:
Seems to be good, but tests are failing

Krigpl:
> Seems to be good, but tests are failing

aka "the usual"

golem/model.py
maaktweluit: awesome! even better :D

etam (approved):
ok

PR #2905, opened 2018-06-20
FundsLocker doesnt need to clean old locks

Krigpl:
Old tasks are periodically removed in `Client` which results in removing the locks as well. Unless I'm missing something `FundsLocker` shouldn't take care of old locks on its own.


maaktweluit (approved):
So many removals, how did this work before :P

Krigpl:
This is not exactly precise, will not merge it just yet.

maaktweluit:
Ok, please let me know if you want another round of review :)

mplebanski:
Can we close this?

PR #2899, opened 2018-06-19
Benchmarks run on linux hw cfg change

tworec:
fixes #2864 
tests for this are in #2906

PR #2906, opened 2018-06-20
hardware preset tests added

tworec:
tests for #2899


golem/config/presets.py
```diff
@@ -1,18 +1,13 @@
...
-from golem.core.hardware import HardwarePresets
 from golem.model import HardwarePreset
 
 log = logging.getLogger("golem.config")
 
 
 class HardwarePresetsMixin(object):
 
-    @staticmethod
-    def get_hw_caps():
```
mfranciszkiewicz: `HardwarePresetsMixin` is a mixin class for `golem.client.Client`; this method handles the `env.hw.caps` RPC request

tworec: ups, I did it again ... thx

tworec: those two classes `HardwarePresets` and `HardwarePresetsMixin` should be merged I think... cause their responsibility are pretty the same, furthermore this would shorten `Client` code

tworec:
those two classes `HardwarePresets` and `HardwarePresetsMixin` should be merged I think... cause their responsibility are pretty the same, furthermore this would shorten `Client` code

mfranciszkiewicz (approved):

PR #2911, opened 2018-06-21
Fix GNT lock in FundsLocker

Krigpl:
Current GNT lock value depends on the number of the unfinished tasks which is being updated but not taken into calculations after #2774. This PR fixes that.

### DIFF
--- a/golem/transactions/ethereum/fundslocker.py
+++ b/golem/transactions/ethereum/fundslocker.py
@@ -11,9 +11,13 @@ logger = logging.getLogger(__name__)
 
 class TaskFundsLock:  # pylint: disable=too-few-public-methods
     def __init__(self, task):
-        self.gnt_lock = task.price
+        self.price = task.subtask_price
         self.num_tasks = task.total_tasks
         self.task_deadline = task.header.deadline
+
+    @property
+    def gnt_lock(self):
+        return self.price * self.num_tasks
 
 
 class FundsLocker(LoopingCallService):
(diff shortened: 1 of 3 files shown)


golem/transactions/ethereum/fundslocker.py
```diff
@@ -11,10 +11,14 @@
 
 class TaskFundsLock:  # pylint: disable=too-few-public-methods
     def __init__(self, task):
-        self.gnt_lock = task.price
+        self.price = task.subtask_price
```
maaktweluit, line 14: is the name `price` here clear enough or would `subtask_price` be better?

tests/golem/transactions/ethereum/test_fundslocker.py
```diff
@@ -28,15 +28,15 @@ def test_lock_funds(self):
...
+        task.subtask_price = 320
         task.total_tasks = 10
         task.header.deadline = task_deadline
         fl.lock_funds(task)
         tfl = fl.task_lock['abc']
 
         def test_params(tfl):
             assert isinstance(tfl, TaskFundsLock)
-            assert tfl.gnt_lock == 320
+            assert tfl.gnt_lock == 320 * 10
```
maaktweluit, line 39: maybe the magic numbers can be replaced with `task.*`?

maaktweluit (commented):
LGTM!

2 small comments, will approve on reply

golem/transactions/ethereum/fundslocker.py
Krigpl: I dunno, it was `price` before, either is fine. But also I'm aiming to change the way FundsLocker works altogether (as you might have noticed already) so this should't matter in the long run

tests/golem/transactions/ethereum/test_fundslocker.py
Krigpl: Yeah, but as I mentioned when I'm finished with FundsLocker most of this stuff won't exist anymore so I didn't bother to fix it now :)

maaktweluit (approved):
Thank you for the replies, looks good to me!

jiivan (approved):

PR #2921, opened 2018-06-22
General fixes

Elfoniok:


apps/core/task/coretask.py
```diff
@@ -145,7 +145,7 @@ def __init__(self,
             subtask_timeout=task_definition.subtask_timeout,
             resource_size=self.resource_size,
             estimated_memory=task_definition.estimated_memory,
-            max_price=task_definition.max_price,
+            max_price=task_definition.max_price
```
jiivan: :-1:

golem/environments/environmentsmanager.py
```diff
@@ -6,7 +6,8 @@
 
 
 class EnvironmentsManager(object):
-    """ Manage known environments. Allow user to choose accepted environment, keep track of supported environments """
+    """ Manage known environments. Allow user to choose accepted environment,
+    keep track of supported environments """
```
jiivan: First line should be a summary without a continuation

jiivan (approved):

apps/core/task/coretask.py
Elfoniok: Ok

golem/environments/environmentsmanager.py
Elfoniok: Ok

PR #2922, opened 2018-06-22
Prevent provider locking in TaskComputer

mfranciszkiewicz:
If a `KeyError` is raised in `TaskComputer.task_computed`, the provider may never set their `counting_task` property to `None` (a.k.a. providers stuck in `Computing` state)

### DIFF
--- a/golem/task/taskcomputer.py
+++ b/golem/task/taskcomputer.py
@@ -168,6 +168,8 @@ class TaskComputer(object):
         self.reset()
 
     def task_computed(self, task_thread):
+        self.reset()
+
         if task_thread.end_time is None:
             task_thread.end_time = time.time()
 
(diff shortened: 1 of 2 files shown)


golem/task/taskcomputer.py
```diff
@@ -168,6 +168,8 @@ def resource_request_rejected(self, subtask_id, reason):
         self.reset()
 
     def task_computed(self, task_thread):
+        self.reset()
+
```
badb, line 172: If you call reset here, you can also remove self.counting_thread checks below (lines 176-178).

badb (approved):

golem/task/taskcomputer.py
mfranciszkiewicz: Spoken too soon - these lines cannot be simply removed. `reset` does not modify the `counting_thread` property.

PR #2925, opened 2018-06-25
TaskManager.task_needs_computation

Wiezzel:
Refactored nested function into a public method. Used it in MaskUpdateService instead of Task.needs_computation(). This will prevent updating masks for timed-out tasks.

### DIFF
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -375,6 +375,8 @@ class TaskManager(TaskEventListener):
             return None, False, False
 
         if not self.task_needs_computation(task_id):
+            logger.info(f'Task does not need computation; '
+                        f'provider: {node_name} - {node_id}')
             return None, False, False
 
         extra_data = task.query_extra_data(
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -327,6 +327,16 @@ class TaskManager(TaskEventListener):
                                      op=TaskOp.WORK_OFFER_RECEIVED,
                                      persist=False)
 
+    def task_needs_computation(self, task_id: str) -> bool:
+        if self.tasks_states[task_id].status not in self.activeStatus:
+            logger.info(f'task is not active: {task_id}')
+            return False
+        task = self.tasks[task_id]
+        if not task.needs_computation():
+            logger.info(f'no more computation needed: {task_id}')
+            return False
+        return True
+
     def get_next_subtask(
             self, node_id, node_name, task_id, estimated_performance, price,
             max_resource_size, max_memory_size, num_cores=0, address=""):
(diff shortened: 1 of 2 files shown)


golem/task/taskmanager.py
```diff
@@ -327,6 +327,16 @@ def got_wants_to_compute(self,
                                      op=TaskOp.WORK_OFFER_RECEIVED,
                                      persist=False)
 
+    def task_needs_computation(self, task_id: str) -> bool:
+        if self.tasks_states[task_id].status not in self.activeStatus:
+            logger.info(f'task is not active: {task_id}')
```
tworec, line 332: @mfranciszkiewicz asked to put node info into logs. It helps debugging

tworec (changes requested):

golem/task/taskmanager.py
Wiezzel: I can add another logging command in `get_next_subtask` but not here.

Wiezzel: Ok, there you go.

tworec: ok, but now task_id is redundantly twice...

tworec (changes requested):

golem/task/taskmanager.py
Wiezzel: Better now?

tworec:
selfie-1

tworec (approved):

golem/task/taskmanager.py
jiivan: Please use logging function parameters for string format arguments.

https://gist.github.com/jiivan/eedb505574ccb2503af48fa2e3043036

Wiezzel: @jiivan Actually, you didn't include the f-string method in your logtest. I did and the difference for string argument is negligible:
```
#'message'
               correct_logging: 1069.55ms
             incorrect_logging: 1213.96ms
    newstyle_incorrect_logging: 1250.48ms
         format_string_logging: 1089.07ms
```
And using f-strings is more readable than using `%s`.

tworec: interesting comparison, I think you can post it on #developers channel on our rocket. I also like f'string readability but thought that it is much slower than _'correct logging'_

tworec: @Wiezzel create PR for @jiivan code with f'string ;)

jiivan: @Wiezzel That's interesting, I didn't know about that. Maybe that's why linter didn't complain about f string method. Readability is subjective. I'll leave it up to you then.

tworec: we can do a poll inside our team for readablity of f'strings and other string formatting methods !

Wiezzel: I will leave it as it is. But we surely need some explicit, team-wide coding standards.

PR #2926, opened 2018-06-25
Verificator refactor

Elfoniok:


apps/blender/blender_reference_generator.py
```diff
@@ -0,0 +1,254 @@
+import math
```
etam, line 1: Not a single type annotation in this file :(

apps/blender/blender_reference_generator.py
```diff
@@ -0,0 +1,254 @@
...
+import random
+import os
+import logging
+from copy import deepcopy
+from functools import partial
+import numpy
```
etam: Second point of https://www.python.org/dev/peps/pep-0008/#imports

apps/blender/verification_queue.py
```diff
@@ -0,0 +1,111 @@
...
+import threading
+import queue
+from types import FunctionType
+from typing import Optional, Type, Dict
+from collections import namedtuple
+from twisted.internet.defer import Deferred, gatherResults
+from apps.blender.blender_reference_generator import BlenderReferenceGenerator
+from golem_verificator.verifier import Verifier
+from golem.core.common import deadline_to_timeout
+from golem.task.localcomputer import ComputerAdapter
```
etam: Second point of https://www.python.org/dev/peps/pep-0008/#imports

etam:
It would be easier if you wrote in advance that this is related to https://github.com/golemfactory/golem-verificator/pull/34

apps/blender/verification_queue.py
```diff
@@ -0,0 +1,111 @@
...
+                raise Exception("Task deadline passed")
+
+        except Exception as exc:  # pylint: disable=broad-except
+            with self._lock:
+                deferred_job.errback(exc)
+                self._jobs.pop(subtask_id, None)
+
+            logger.error("Failed to start verification of subtask %r: %r",
+                         subtask_id, exc)
+            self._process_queue()
```
etam, line 109: That's ugly. It would be better if `_run` would return `bool` and `_process_queue` would loop based on that.

apps/dummy/task/verifier.py
```diff
@@ -7,19 +7,26 @@
 class DummyTaskVerifier(CoreVerifier):
     # subtask_info is what sits in the task.subtasks_given["subtask_id"]
     # it is set in the query_extra_data
-    def _verify_result(self, subtask_info: dict, result: str,
-                       reference_data: list, resources: list):
+    def __init__(self, callback, verification_data=None):
```
etam: Type annotations? They were here before.

apps/dummy/task/verifier.py
```diff
@@ -7,19 +7,26 @@
...
-        if len(result_data) != subtask_info["result_size"]:
-            return False
+        for result in results:
+            _, ext = os.path.splitext(result)
+            ext = ext.lower()
+            if ext != subtask_info["result_extension"]:
+                return False
+
+            with open(result, "r") as f:
+                result_data = f.read()
```
etam: If we only want the size of file contents, there's no need to waste memory for that. Use `os.stat('path').st_size`

etam: Oh, `result_data` is used below outside of this loop. So it uses the value of the last result. Is that correct?

etam (changes requested):

apps/blender/verification_queue.py
mfranciszkiewicz: `_process_queue` should be called only if we cannot start the verification process or the verification process has finished (via `callback`), not when `_run` returns.

apps/blender/blender_reference_generator.py
Elfoniok: Done

apps/blender/blender_reference_generator.py
Elfoniok: Done

apps/blender/verification_queue.py
Elfoniok: Done

apps/dummy/task/verifier.py
Elfoniok: Done

apps/dummy/task/verifier.py
Elfoniok: Done

etam (changes requested):
Tuple of what? List of what? Dict of what? Please! That's **the most** important part.
And don't forget about return types.

apps/blender/blender_reference_generator.py
```diff
@@ -0,0 +1,269 @@
...
+    CROP_STEP = 0.01
+    CROPS_NO_FIRST = 3
+    CROPS_NO_SECOND = 6
+
+    def __init__(self):
+        self.crop_counter: int = 0
+        self.crop_size: Tuple[float, float] = ()
+        self.split_values: List[Tuple[float, float, float, float]] = []
+        self.split_pixels: List[Tuple[int, int]] = []
+        self.rendered_crops_results: Dict[int, [[str], float, CropContext]] = {}
```
etam: `Dict[int, Union[List[str], float, CropContext]]`? Or maybe just `Dict[int, Any]`?

apps/blender/blender_reference_generator.py
```diff
@@ -0,0 +1,269 @@
...
+        # survive in edge cases
+        max_ -= 1
+        min_ += 1
+        difference = (max_ - size_)
+        if difference < 0:
+            raise Exception("Subtask is to small to reliable verifcation")
+        split_min = random.randint(min_, difference)
+        split_max = split_min + size_
+        logger.info("split_min=%r, split_max=%r", split_min, split_max)
+        return [split_min, split_max]
```
etam, line 252: minor: it would make more sense to return as a tuple

Elfoniok: No, list is good for now.

apps/blender/blender_reference_generator.py
```diff
@@ -0,0 +1,269 @@
...
+        crops_path = os.path.join(subtask_info['tmp_dir'],
+                                  subtask_info['subtask_id'])
+
+        crops_info = self.generate_split_data((subtask_info['res_x'],
+                                               subtask_info['res_y']),
+                                              subtask_info['crop_window'],
+                                              num_crops,
+                                              crop_size)
+
+        verify_ctx = CropContext({'paths': crops_path, 'position': crops_info},
```
etam, line 164: That's the weirdest way to pass arguments I've ever seen. Why not just drop those dicts and use `path=crop_path, position=crops_info, ...`?

etam (approved):
Would be prettier if CropContext was a data class, but well, we're not there yet.

Elfoniok:
I agree hopefully we will be there soon.

PR #2931, opened 2018-06-26
Fix trust calculation

### LINKED ISSUE #2874: Incorrect trust calculation in taskserver
That should say `min_trust`
https://github.com/golemfactory/golem/blob/26642de9ede50b79acd2a14f1f414b20536c4ff6/golem/task/taskserver.py#L489

Krigpl:
Fixes #2874

### DIFF
--- a/golem/task/taskserver.py
+++ b/golem/task/taskserver.py
@@ -489,7 +489,7 @@ class TaskServer(
     def decrease_trust_payment(self, task_id):
         node_id = self.task_manager.comp_task_keeper.get_node_for_task_id(
             task_id)
-        Trust.PAYMENT.decrease(node_id, self.max_trust)
+        Trust.PAYMENT.decrease(node_id, self.min_trust)
 
     def reject_result(self, subtask_id, account_info):
         mod = min(


golem/task/taskserver.py
```diff
@@ -489,7 +489,7 @@ def increase_trust_payment(self, task_id):
     def decrease_trust_payment(self, task_id):
         node_id = self.task_manager.comp_task_keeper.get_node_for_task_id(
             task_id)
-        Trust.PAYMENT.decrease(node_id, self.max_trust)
+        Trust.PAYMENT.decrease(node_id, self.min_trust)
```
maaktweluit, line 492: Are you sure the `mod` should be `0.0`?
https://github.com/golemfactory/golem/blob/daf6d0223db7270835d55b196ccf40d84140afee/golem/ranking/helper/trust.py#L43

I think its better to remove the second param, but i only..
ihavesomesortofidea

maaktweluit (changes requested):

golem/task/taskserver.py
Krigpl: Oh yea, I dunno actually. I also don't have an idea what I'm doing. I assumed this is wrong based on the api and nomenclature and high likeliness of copy-paste bug from symmetrical `increase_trust_payment`. But I guess you're right, 0 doesn't make sense here.

Krigpl:
Not a bug.

PR #2937, opened 2018-06-28
hardware config change detection fix

tworec:
Changes `HardwarePresets.update_config` to return true only when config is not initial (all zeros) and was changed.


etam:
Are you sure you want to merge it to b0.16.1?

golem/core/hardware.py
```diff
@@ -1,24 +1,25 @@
...
+from golem.core.common import get_cpu_count, is_osx, is_windows, \
     MAX_CPU_MACOS, MAX_CPU_WINDOWS
 from golem.core.fileshelper import free_partition_space
 from golem.model import HardwarePreset
 
 logger = logging.getLogger(__name__)
 
 
-def cpu_cores_available():
+def cpu_cores_available() -> List:
```
etam: List of what?

golem/core/hardware.py
```diff
@@ -28,48 +29,59 @@ def cpu_cores_available():
...
         return list(range(0, num_cores - 1)) or [0]
 
 
-def memory_available():
+def memory_available() -> int:
     """
     :return int: 3/4 of total available memory
     """
-    return max(int(virtual_memory().total * 0.75) / 1024, MIN_MEMORY_SIZE)
+    return int(max(int(virtual_memory().total * 0.75) / 1024, MIN_MEMORY_SIZE))
```
etam: Maybe instead of adding `int` around `max` change
`int(virtual_memory().total * 0.75) / 1024`
to
`int(virtual_memory().total * 0.75 / 1024)`
?

golem/core/hardware.py
etam: List of what?

golem/core/hardware.py
```diff
@@ -28,48 +29,59 @@ def cpu_cores_available():
...
-        logger.debug("Couldn't read CPU affinity: {}".format(e))
+        logger.debug("Couldn't read CPU affinity: %r", e)
         num_cores = get_cpu_count()
         return list(range(0, num_cores - 1)) or [0]
 
 
-def memory_available():
+def memory_available() -> int:
     """
     :return int: 3/4 of total available memory
```
etam: While at it, you could add that it's in kilobytes.

tests/golem/core/test_hardware_presets.py
```diff
@@ -76,20 +76,23 @@ def test_update_config_upper_bounds(self):
...
-            DEFAULT_HARDWARE_PRESET_NAME, self.config)
-        assert HardwarePresets.update_config(
-            CUSTOM_HARDWARE_PRESET_NAME, self.config)
-        assert not HardwarePresets.update_config(
-            CUSTOM_HARDWARE_PRESET_NAME, self.config)
+        # given
+        HardwarePreset.create(name='foo', cpu_cores=1, memory=1200000,
+                              disk=2000000)
+
+        # than
```
etam: then

golem/core/hardware.py
```diff
@@ -97,7 +109,7 @@ def from_config(cls, config):
         )
 
     @classmethod
-    def caps(cls):
+    def caps(cls) -> dict:
```
etam: `-> Dict[str, int]`?

golem/core/hardware.py
```diff
@@ -106,8 +118,9 @@ def caps(cls):
         }
 
     @classmethod
-    def values(cls, preset_or_name):
-        preset_or_name = preset_or_name or DEFAULT_HARDWARE_PRESET_NAME
+    def values(cls, preset_or_name: Union[str, HardwarePreset]) \
+            -> Tuple[str, dict]:
```
etam: `-> Tuple[str, Dict[str, int]]`?

etam (changes requested):
I like type annotations :)

badb:
@tworec Can you please make this PR to develop?

golem/core/hardware.py
tworec: 👍

golem/core/hardware.py
tworec: whaa? we need to assure that it is not below MIN, how do your code achieve this?

golem/core/hardware.py
tworec: 👍

golem/core/hardware.py
tworec: 👍

tests/golem/core/test_hardware_presets.py
tworec: 😄

golem/core/hardware.py
tworec: 👍

golem/core/hardware.py
etam: whaa? I don't understand what are you talking about.

Lets split this one-liner a bit:

Before your changes:
```python
proposed_memory_in_kb = int(virtual_memory().total * 0.75) / 1024
result = max(proposed_memory_in_kb, MIN_MEMORY_SIZE)
return result
```
MIN_MEMORY_SIZE is int.

The problem here is that `proposed_memory_in_kb` is a float, not int. So `result` also might be float.

What you're proposing is changing `return result` to `return int(result)`.

I propose changing calculation of `proposed_memory_in_kb` so it's int instead of float.

tworec: @etam, I'm totally sorry - it was my misunderstanding as of y-day. I didn't notice double `int(` constructor

tworec: done

etam (approved):
ok

PR #2940, opened 2018-06-29
Switch back provider and requestor in clean res CLI command

### LINKED ISSUE #2889: Requesting node runs out of disk space
Affects: 0.16.0

When running a node that has `accept tasks = 0` and is requesting tasks, it quickly fills the disk space (100GB in 2-3 days). The space can be reclaimed by running `./golemcli.py res clear --provider`.

badb:
Resolves #2889

### DIFF
--- a/golem/interface/client/resources.py
+++ b/golem/interface/client/resources.py
@@ -35,7 +35,7 @@ class Resources(object):
 
         clear = Resources.client.clear_dir
 
-        if provider:
+        if requestor:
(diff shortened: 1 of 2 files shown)


mfranciszkiewicz (approved):

etam (approved):
That's a good change, but I don't think it resolves #2889.

badb:
@etam Why?

etam:
Because the issue is about running out of disk space. CLI confusion was something that was also found, but wasn't the main problem.

PR #2952, opened 2018-07-03
Reverted unicode fixes for windows logs

maaktweluit:
Since they seem fixed in python3.

Original PR: #910

This removes the `b''` surrounding each log line on windows.


jiivan (approved):
LGTM

PR #2955, opened 2018-07-03
display message about waiting for password repeatedly

### LINKED ISSUE #2879: Display information that golem is waiting for password more clearly in console
When Golem cannot be started because terms of use are not accepted than message with this information is displayed many times in regular interval. When it's waiting for password to unlock the account, however, it's only displayed once and it's not the last message that is displayed usually (some additional information about docker state can be displayed after it). Therefore many users may not notice that golem is waiting for some form of interaction. 
We would like this message to be displayed many times in regular intervals like term of use or at least to make it more visible

etam:
Fixes #2879

### DIFF
--- a/golem/node.py
+++ b/golem/node.py
@@ -306,14 +306,15 @@ class Node(object):  # pylint: disable=too-few-public-methods
 
             if self.key_exists():
                 event = 'get_password'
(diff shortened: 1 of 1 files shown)


Krigpl (approved):
coolkid

PR #2958, opened 2018-07-04
update cli 'account withdraw' interface

### LINKED ISSUE #2880: Fix help message to "account withdraw" command
> account withdraw
the following arguments are required: amount, destination, currency

Should be: the following arguments are required: destination amount currency. For eg. account withdraw 0x74446B4FB2Cca2fd8A475Bc40606b77d8d7659C4 3 GNT

Also, information about locked withdraws on the testnet should be shown earlier for the user, not after he types all necessary commands. Maybe just add comment in account “withdrawals are not available for the testnet”

etam:
Fixes #2880

### DIFF
--- a/golem/interface/client/account.py
+++ b/golem/interface/client/account.py
@@ -102,12 +102,13 @@ class Account:
         return "Account unlock success"
 
     @command(
-        arguments=(amount_arg, address_arg, currency_arg),
-        help="Withdraw GNT/ETH")
+        arguments=(address_arg, amount_arg, currency_arg),
+        help=("Withdraw GNT/ETH\n"
+              "(withdrawals are not available for the testnet)"))
     def withdraw(  # pylint: disable=no-self-use
             self,
-            amount,
             destination,
+            amount,
             currency) -> str:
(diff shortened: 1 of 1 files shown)


jiivan (approved):
lgtm

golem/interface/client/account.py
```diff
@@ -102,12 +102,13 @@ def unlock(self) -> str:  # pylint: disable=no-self-use
...
-        arguments=(amount_arg, address_arg, currency_arg),
-        help="Withdraw GNT/ETH")
+        arguments=(address_arg, amount_arg, currency_arg),
+        help=("Withdraw GNT/ETH\n"
+              "(withdrawals are not available for the testnet)"))
     def withdraw(  # pylint: disable=no-self-use
             self,
-            amount,
             destination,
+            amount,
```
Krigpl, line 111: Why the reorder? All other `withdraw` functions have `amount` first

Krigpl: I found the answer so nevermind

etam: hat's what #2880 is about. And this is only a CLI command, that's why I'm not reordering further.

PR #2960, opened 2018-07-04
Cli 'account unlock' can be called only if account is locked

### LINKED ISSUE #2882: Make sure that "account unlock" can be called only if account is locked
Right now user can call account unlock with proper password many times. It shouldn't cause any problems but may be nonintuitive for many users.

etam:
Fixes #2882

### DIFF
--- a/golem/node.py
+++ b/golem/node.py
@@ -158,6 +158,9 @@ class Node(object):  # pylint: disable=too-few-public-methods
     def key_exists(self) -> bool:
         return KeysAuth.key_exists(self._datadir, PRIVATE_KEY)
 
+    def is_account_unlocked(self) -> bool:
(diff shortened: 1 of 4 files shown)


golem/node.py
```diff
@@ -158,6 +158,9 @@ def set_password(self, password: str) -> bool:
     def key_exists(self) -> bool:
         return KeysAuth.key_exists(self._datadir, PRIVATE_KEY)
 
+    def is_account_unlocked(self) -> bool:
```
Krigpl, line 159: You can use it in lines 305 and 319 for consistency as well.

etam: done

tworec (approved):
lgtm

PR #2964, opened 2018-07-04
 improve EnvironmentsManager and cli envs command

### LINKED ISSUE #2883: Make "envs disable <ENVIRONMENT>" behaviour more intuitive.
If you type **envs disable blender** than you get _Completed in 0.006350994110107422_s but it’s not really working, as command should be **envs disable BLENDER**. 
We should make sure that proper error method is displayed if command fails. 

After disabling BLENDER and enabling it again in the table active it’s showing ‘True’ instead of ‘1’ as in other cases.
We need to check if this behaviour is a result of bug that may cause wrong Golem behaviour or is it just a problem with displaying values.

etam:
Fixes #2883

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -1186,13 +1186,13 @@ class Client(HardwarePresetsMixin):
     def get_environments(self):
         envs = copy(self.environments_manager.get_environments())
         return [{
-            'id': str(env.get_id()),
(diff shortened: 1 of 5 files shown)


maaktweluit (approved):
LGTM!

Nice cleanup!

PR #2967, opened 2018-07-04
[Concent] UC5 / Force Payment - react to  ForcePaymentCommitted

### LINKED ISSUE #2195: [Concent] UC5 / Force Payment - react to `ForcePaymentCommitted`
implement reaction to the `ForcePaymentCommitted` message (essentially meaning that the payment has been effected and the subtask's life has ended)
* as a Requestor
* as a Provider

### LINKED ISSUE #2196: [Concent] UC5 / Force Payment - react to `ForcePaymentRejected`
implement Provider's reaction to `ForcePaymentRejected` message from the Concent - effectively exhausting the set of available actions and ending the subtask's life from the perspective of the Provider

jiivan:
fixes: #2195 
fixes: #2196


golem/network/concent/received_handler.py
```diff
@@ -527,3 +539,39 @@ def on_subtask_results_settled(self, msg, **_):
...
+
+    @handler_for(message.concents.ForcePaymentCommitted)
+    def on_force_payment_committed(self, msg, **_):
+        handler_name = 'on_force_payment_committed_for_{}'.format(
+            msg.recipient_type.value,
+        )
+        getattr(self, handler_name)(msg)
+
+    def on_force_payment_committed_for_requestor(self, msg):  # noqa pylint: disable=no-self-use
+        # There is no mechanism to cancel started payment.
```
shadeofblue: @jiivan you mean, there's no way to abort an already-started transaction, right? or do you mean that we have no way to tell the requestor node not to send a payment that's in the waiting period (for which the blockchain transaction has not yet already been started)?

golem/network/concent/received_handler.py
```diff
@@ -527,3 +539,39 @@ def on_subtask_results_settled(self, msg, **_):
...
+        SEE: IncomesKeeper.received_forced_subtask_payment
+        """
+        incomes_keeper = self.task_server.client \
+            .transaction_system.incomes_keeper
+        upi = incomes_keeper.get_list_of_unpaid_incomes(
+            sender=utils.pubkeytoaddr(msg.task_owner_key),
+            closure_time=msg.payment_ts,
+        )
+        for income in upi:
+            incomes_keeper.update_forced(
```
shadeofblue: @jiivan hmm... shouldn't we somehow take the `amount_paid` and `amount_pending` into consideration here? ... and, e.g., maybe only update those incomes for which the sum of expected incomes is lower than `amount_paid`? ...

golem/transactions/ethereum/ethereumincomeskeeper.py
```diff
@@ -36,6 +37,11 @@ def __init__(self, sci) -> None:
                 from_block,
                 self._on_forced_subtask_payment,
             )
+            self.__sci.subscribe_to_forced_payments(
+                requestor_address=None,
+                provider_address=self.__sci.get_eth_address(),
+                from_block=from_block,
+                cb=self._on_forced_payment,
```
Krigpl: Missing closing bracket

golem/transactions/incomeskeeper.py
```diff
@@ -79,6 +75,13 @@ def received_batch_transfer(
             value=amount,
         )
 
+    def received_forced_payment(self, *args, **kwargs):
```
Krigpl: Don't use `args` and `kwargs` here, they are not required and greatly reduce legibility.

golem/transactions/incomeskeeper.py
```diff
@@ -59,11 +56,10 @@ def received_batch_transfer(
         amount_left = amount
 
         for e in expected:
-            value = min(amount_left, e.value)
-            amount_left -= value
+            if amount_left < e.value:
```
Krigpl: So, if provider is expecting 3 payments 1 GNT each and will receive 1.5 GNT from the requestor and then 1.5 GNT from Concent, one of the payments will still not be marked as done.

golem/network/concent/received_handler.py
```diff
@@ -527,3 +538,38 @@ def on_subtask_results_settled(self, msg, **_):
...
+        from golem.model import Payment
+        from golem.model import PaymentStatus
+        logger.warning(
+            "[CONCENT] Our deposit was used to cover payment of %.6f GNT"
+            " for eth address: %s",
+            msg.amount_paid / denoms.ether,
+            msg.provider_eth_account,
+        )
+        # Try to stop awaiting payments
+        payments = Payment.select().where(
```
Krigpl: Remove it, it will be taken care of when blockchain event is detected (also this `select` is quite incorrect).

tests/golem/transactions/test_incomeskeeper.py
```diff
@@ -54,6 +55,17 @@ def _test_expect_income(self, sender_node_id, subtask_id, value):
         assert expected_income.accepted_ts is None
         assert expected_income.transaction is None
 
+    @mock.patch("golem.transactions.incomeskeeper.IncomesKeeper"
+                ".received_batch_transfer")
+    def test_received_forced_payment(self, batch_mock):
+        args = range(10)
+        kwargs = dict((str(x), x**2) for x in range(10))
```
Krigpl: What is this supposed to be ?

Krigpl (changes requested):

golem/transactions/incomeskeeper.py
jiivan: IMO it's perfect place for `*args, **kwargs`. It clearly states that this method is just pass-through for `received_batch_transfer` and it doesn't modify arguments. Stating arguments explicitly here would reduce legibility as a reader would have to check whether all arguments are passed unchanged.

Krigpl: You're doing `kwargs.get('sender', None)` below which is completely unnecessary since `sender` will always be present in the arguments and that would be obvious if the arguments would be listed explicitly. Otherwise it raises questions like "when `sender` is not in the args" with the answer "never". This is the lack of legibility I'm talking about.

golem/transactions/incomeskeeper.py
```diff
@@ -59,18 +56,18 @@ def received_batch_transfer(
...
-            # TODO don't change the value, wait for Concent. issue #2255
-            e.value = value
             e.save()
 
-            dispatcher.send(
-                signal='golem.income',
-                event='confirmed',
-                subtask_id=e.subtask,
-            )
+            if e.value_expected <= 0:
```
Krigpl: There's no way it's gonna be negative.

Krigpl (changes requested):

golem/database/schemas/017_schema.py
```diff
@@ -0,0 +1,17 @@
...
+# pylint: disable=too-few-public-methods
+from golem import model
+
+SCHEMA_VERSION = 17
+
+
+def migrate(migrator, database, fake=False, **kwargs):
+    migrator.add_fields(
+        'income',
+        value_received=model.HexIntegerField(default=0),
```
Krigpl: `value_received` should be equal to `value` for confirmed incomes (those with transaction hash set). Is it possible to make it happen in this migration?

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -128,6 +128,17 @@ def _subscribe_to_events(self) -> None:
                     event.amount,
                 )
             )
+            self._sci.subscribe_to_forced_payments(
+                requestor_address=None,
+                provider_address=self._sci.get_eth_address(),
+                from_block=from_block,
+                cb=lambda event: ik.received_forced_payment(
+                    tx_hash=event.tx_hash,
+                    sender=event.sender,
```
Krigpl: The field is `requestor`, not `sender`

Krigpl (commented):
The migration thing would be great to have, because we may start showing it in the UI and for old tasks it would show 0.

golem/ethereum/incomeskeeper.py
```diff
@@ -56,18 +53,42 @@ def received_batch_transfer(
...
-                subtask_id=e.subtask,
-            )
+            if e.value_expected == 0:
+                dispatcher.send(
+                    signal='golem.income',
+                    event='confirmed',
+                    subtask_id=e.subtask,
+                )
+
+        dispatcher.send(
```
Krigpl: This has been removed in #3165

golem/network/concent/received_handler.py
```diff
@@ -527,3 +538,30 @@ def on_subtask_results_settled(self, msg, **_):
                 subtask_id=msg.subtask_id,
                 settled_ts=msg.timestamp,
             )
+
+    @handler_for(message.concents.ForcePaymentCommitted)
+    def on_force_payment_committed(self, msg, **_):
+        handler_name = 'on_force_payment_committed_for_{}'.format(
+            msg.recipient_type.value,
+        )
+        getattr(self, handler_name)(msg)
```
Krigpl: This magic isn't needed here, the following is imo more readable and more importantly greppable.
```
if msg.recipient_type == requestor:
  self.on_force_payment_committed_for_requestor
else:
  on_force_payment_committed_for_provider
```

tests/golem/network/concent/test_received_handler.py
```diff
@@ -817,3 +817,31 @@ def test_settled(self):
 
         self.assertEqual(
             Income.get(subtask=srs.subtask_id).settled_ts, srs.timestamp)
+
+
+class ForcePaymentTest(TaskServerMessageHandlerTestBase):
+    def setUp(self):
+        super().setUp()
+        self.client.transaction_system = mock.Mock()
+        self.client.transaction_system.incomes_keeper = IncomesKeeper()
```
Krigpl: Why is it needed?

golem/ethereum/incomeskeeper.py
jiivan: Good catch! Thanks.

golem/network/concent/received_handler.py
jiivan: No magic here, but greppable argument sounds valid to me.

tests/golem/network/concent/test_received_handler.py
jiivan: Good catch, it was an artefact from when handler interacted with IncomesKeeper.

Krigpl (approved):
rickandmorty_yes

PR #2971, opened 2018-07-05
Fix loading payments from DB

Krigpl:
Lambda was binding to the wrong reference

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -77,8 +77,8 @@ class PaymentProcessor:
         for tx_hash, payments in sent.items():
             self._sci.on_transaction_confirmed(
                 tx_hash,
-                lambda r: threads.deferToThread(
-                    self._on_batch_confirmed, payments.copy(), r),
+                lambda r, p=payments: threads.deferToThread(
+                    self._on_batch_confirmed, p, r),
             )
 
         for awaiting_payment in Payment \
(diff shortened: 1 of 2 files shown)


golem/ethereum/paymentprocessor.py
```diff
@@ -77,8 +77,8 @@ def load_from_db(self):
         for tx_hash, payments in sent.items():
             self._sci.on_transaction_confirmed(
                 tx_hash,
-                lambda r: threads.deferToThread(
-                    self._on_batch_confirmed, payments.copy(), r),
+                lambda r, p=payments: threads.deferToThread(
+                    self._on_batch_confirmed, p, r),
```
etam, line 81: How that works? What is the difference?

Krigpl: Now: copy `payments` by value in the loop, at the time callback is executed `p` has the expected value
Before: `payments` is just a pointer to values from `sent` map, after the loop is over it points to the last element and at the time callback is executed `copy` copies the last value from `sent` instead of the expected one.

mfranciszkiewicz: @Krigpl are you sure? `p` should hold the reference to `payments` at the time of definition, different for each loop.

Krigpl: I don't fully understand your question. I am rather sure my description is correct.
What I don't know is why `p=payments` is enough instead of `p=payments.copy()`, but it is.
Feel free to play with different approaches and use the corresponding test.

mfranciszkiewicz (approved):

golem/ethereum/paymentprocessor.py
etam: https://docs.python.org/3/faq/programming.html#why-do-lambdas-defined-in-a-loop-with-different-values-all-return-the-same-result

etam (approved):
Can we move to rust already?

PR #2986, opened 2018-07-06
Better management of locked funds

Krigpl:
Transaction system manages locked funds, not `FundsLocker`. `FundsLocker` is now just a simple class managing number of locked subtasks of a task.
This way everything funds related belongs in a single place - `TransactionSystem`.
No need to provide `locked` arguments in methods like `withdraw` or `concent_deposit`.
No need to remember about `FundsLocker` when querying for balance.

### DIFF
--- a/golem/transactions/ethereum/ethereumtransactionsystem.py
+++ b/golem/transactions/ethereum/ethereumtransactionsystem.py
@@ -146,7 +146,11 @@ class EthereumTransactionSystem(TransactionSystem):
         if eth > eth_available:
             raise NotEnoughFunds(eth, eth_available, 'ETH')
 
-        log.info("Locking %f GNT and ETH for %d payments", gnt, num)
+        log.info(
+            "Locking %f GNT and ETH for %d payments",
+            gnt / denoms.ether,
+            num,
+        )
         self._gntb_locked += gnt
         self._payments_locked += num
 
@@ -163,7 +167,11 @@ class EthereumTransactionSystem(TransactionSystem):
                 self._payments_locked,
 
             ))
-        log.info("Unlocking %f GNT and ETH for %d payments", gnt, num)
+        log.info(
+            "Unlocking %f GNT and ETH for %d payments",
+            gnt / denoms.ether,
+            num,
+        )
         self._gntb_locked -= gnt
         self._payments_locked -= num
 
--- a/golem/transactions/ethereum/ethereumtransactionsystem.py
+++ b/golem/transactions/ethereum/ethereumtransactionsystem.py
@@ -3,7 +3,7 @@ import time
 from enum import Enum
 from datetime import datetime
 from pathlib import Path
-from typing import List
+from typing import Any, Dict, List
 
 from ethereum.utils import privtoaddr, denoms
 from eth_utils import encode_hex, is_address, to_checksum_address
@@ -15,7 +15,7 @@ from golem.ethereum.node import NodeProcess
 from golem.ethereum.paymentprocessor import PaymentProcessor
 from golem.transactions.ethereum.ethereumincomeskeeper \
     import EthereumIncomesKeeper
-from golem.transactions.ethereum import exceptions
+from golem.transactions.ethereum.exceptions import NotEnoughFunds
 from golem.transactions.transactionsystem import TransactionSystem
 
 log = logging.getLogger('golem.pay')
@@ -73,6 +73,8 @@ class EthereumTransactionSystem(TransactionSystem):
         self._gntb_balance: int = 0
         self._last_eth_update = None
         self._last_gnt_update = None
+        self._payments_locked: int = 0
+        self._gntb_locked: int = 0
         self._is_stopped = False
 
     def stop(self):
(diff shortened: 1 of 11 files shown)


golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -102,18 +105,72 @@ def get_payment_address(self):
...
+                gnt / denoms.ether,
+                self._gntb_locked / denoms.ether,
+            ))
+        if num > self._payments_locked:
+            raise Exception("Can't unlock {} payments, locked: {}".format(
+                num,
+                self._payments_locked,
+
+            ))
+        log.info("Unlocking %f GNT and ETH for %d payments", gnt, num)
```
maaktweluit: Where is the ETH unlocked in this function?

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -102,18 +105,72 @@ def get_payment_address(self):
...
+
+        eth = self._eth_for_batch_payment(num)
+        if self._payments_locked == 0 and \
+           self.payment_processor.reserved_eth == 0:
+            eth += self._eth_base_for_batch_payment()
+        eth_available = self.get_available_eth()
+        if eth > eth_available:
+            raise NotEnoughFunds(eth, eth_available, 'ETH')
+
+        log.info("Locking %f GNT and ETH for %d payments", gnt, num)
```
maaktweluit: Where is the ETH locked in this function?

tests/golem/transactions/ethereum/test_ethereumtransactionsystem.py
```diff
@@ -25,9 +25,11 @@ class TestEthereumTransactionSystem(TestWithDatabase, LogTestCase,
     def setUp(self):
         super().setUp()
         self.sci = Mock()
-        self.sci.GAS_PRICE = 0
-        self.sci.GAS_BATCH_PAYMENT_BASE = 0
+        self.sci.GAS_PRICE = 10 ** 9
```
maaktweluit, line 29: Repeated magic numbers, maybe they can have a common source?

tests/golem/transactions/ethereum/test_ethereumtransactionsystem.py
```diff
@@ -179,9 +186,44 @@ def test_withdraw(self):
 
         # Not enough GNT with lock
         with self.assertRaises(NotEnoughFunds):
-            self.ets.withdraw(gnt_balance + gntb_balance - 1, dest, 'GNT', 2)
+            self.ets.withdraw(gntb_balance + locked_gnt + 1, dest, 'GNT')
         self.sci.reset_mock()
 
+    def test_locking_funds(self):
+        eth_balance = 10 * 10 ** 18
```
maaktweluit: same here, isnt it better to use denoms.ether?

maaktweluit (commented):
LGTM!

Some q's, will approve on reply or fix :)

golem/transactions/ethereum/ethereumtransactionsystem.py
Krigpl: Locked ETH is calculated based on the `_payments_locked` variable.

golem/transactions/ethereum/ethereumtransactionsystem.py
Krigpl: Ditto

tests/golem/transactions/ethereum/test_ethereumtransactionsystem.py
Krigpl: The value doesn't really matter and is not used anywhere else (as in changing this value doesn't trigger changing any other `10**9`), that's why imo it's fine like this.

tests/golem/transactions/ethereum/test_ethereumtransactionsystem.py
Krigpl: This is somewhat arbitrary as well but will use `denoms.ether` for clarity

golem/transactions/ethereum/ethereumtransactionsystem.py
maaktweluit: ahhh, now i see, its all there :D

golem/transactions/ethereum/ethereumtransactionsystem.py
maaktweluit: thanks! ditto

maaktweluit (approved):

golem/transactions/ethereum/ethereumtransactionsystem.py
```diff
@@ -102,18 +105,80 @@ def get_payment_address(self):
...
-        gnt_total = self._gnt_balance + self._gntb_balance
-        gnt_av = gnt_total - self.payment_processor.reserved_gntb
-        return gnt_total, gnt_av, self._eth_balance, \
-            self._last_gnt_update, self._last_eth_update
+    def get_available_eth(self) -> int:
+        return max(self._eth_balance - self.get_locked_eth(), 0)
+
+    def get_locked_eth(self) -> int:
+        eth = self.payment_processor.reserved_eth + \
+            self._eth_for_batch_payment(self._payments_locked)
```
mfranciszkiewicz: Is eth for batch payment always a part of locked eth? What about providers?

Krigpl: ETH for batch payments has two parts: base and per payment. This only adds the per payment part which for providers will be 0. The logic below checks whether the base should be added.

PR #2988, opened 2018-07-06
pubkeytoaddr should return checksum'd address

Krigpl:
`IncomesKeeper` doesn't work well without it.
Changing `BLOCK_NUMBER_DB_KEY` will trigger listening to old events which will correctly mark awaiting incomes as confirmed.

### DIFF
--- a/golem/transactions/ethereum/ethereumincomeskeeper.py
+++ b/golem/transactions/ethereum/ethereumincomeskeeper.py
@@ -11,7 +11,7 @@ logger = logging.getLogger(__name__)
 
 
 class EthereumIncomesKeeper(IncomesKeeper):
-    BLOCK_NUMBER_DB_KEY = 'eth_incomes_keeper_block_number'
+    BLOCK_NUMBER_DB_KEY = 'eth_incomes_keeper_block_number_v2'
     BLOCK_NUMBER_BUFFER = 50
 
     def __init__(self, sci) -> None:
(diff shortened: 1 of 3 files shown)


golem/transactions/ethereum/ethereumincomeskeeper.py
```diff
@@ -11,7 +11,7 @@
 
 
 class EthereumIncomesKeeper(IncomesKeeper):
-    BLOCK_NUMBER_DB_KEY = 'eth_incomes_keeper_block_number'
+    BLOCK_NUMBER_DB_KEY = 'eth_incomes_keeper_block_number_v2'
```
maaktweluit, line 14: How is this related to the checksum in the public address?

maaktweluit (approved):
LGTM!

One question but that is more that i do not know :)

golem/transactions/ethereum/ethereumincomeskeeper.py
Krigpl: Because `pubkeytoaddr` didn't return checksum'd address IncomesKeeper didn't match any incomes correctly from blockchain events (which contain checksum'd addresses). Thus many events have been processed without correctly marking incomes as confirmed. Changing this value is one way to enforce listening to all income events from the beginning of the blockchain which in turn will correctly mark the incomes as confirmed.

mfranciszkiewicz (approved):

golem/transactions/ethereum/ethereumincomeskeeper.py
maaktweluit: Super clear! thank you for the explanation!

PR #3017, opened 2018-07-11
Remove local Geth support

Krigpl:
Light clients are not reliable enough and this wasn't supported for a while now anyway, so this is just dead code causing problems with maintenance.
We'll consider bringing it back when light clients are more stable.


maaktweluit (approved):
2dsnoj

PR #3031, opened 2018-07-12
Client.{create,restart}_task: restore returning task id

### LINKED ISSUE #2953: TestClient::test_restart_task is not testing
This was changed: 1b45aa10ff5ae22ed8afab1886526135060b7ecb 

This needs to be fixed:
```patch
modified   tests/golem/test_client.py
@@ -508,7 +508,7 @@ class TestClient(TestDatabaseWithReactor):
 
         def add_task(*_args, **_kwargs):
             resource_manager_result = 'res_hash', ['res_file_1']
-            result = resource_manager_result, 'res_file_1', 'package_hash'
+            result = resource_manager_result, 'res_file_1', 'package_hash', 0
             return done_deferred(result)
 
         self.client.resource_server = Mock(
```

But before applying this patch, fix the test, to make it execute this part of code and fail on it.

etam:
Also:
- test_restart_task should no longer cause failures on windows of #2701
- fixes #2953

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -558,6 +558,9 @@ class Client(HardwarePresetsMixin):
         if self.config_desc.in_shutdown:
             raise Exception('Can not enqueue task: shutdown is in progress, '
                             'toggle shutdown mode off to create a new tasks.')
+        if self.task_server is None:
+            raise Exception("Golem is not ready")
+
         task_manager = self.task_server.task_manager
         _result = Deferred()
 
@@ -743,8 +746,11 @@ class Client(HardwarePresetsMixin):
             task = yield self.enqueue_new_task(t_dict)
             return task.header.task_id, None
         except CreatingTaskFailed as ex:
-            logger.exception("Cannot create task %r", t_dict)
+            logger.error("Cannot create task %r: %s", t_dict, str(ex.__cause__))
             return ex.task_id, str(ex.__cause__)
+        except Exception as ex:  # pylint: disable=broad-except
+            logger.error("Cannot create task %r: %s", t_dict, str(ex))
+            return None, str(ex)
 
     def abort_task(self, task_id):
         logger.debug('Aborting task "%r" ...', task_id)
--- a/golem/client.py
+++ b/golem/client.py
@@ -100,11 +100,6 @@ class ClientTaskComputerEventListener(object):
         self.client.config_changed()
 
 
-class CreatingTaskFailed(Exception):
-    def __init__(self, task_id: str) -> None:
-        self.task_id = task_id
-
-
 class Client(HardwarePresetsMixin):
     _services = []  # type: List[IService]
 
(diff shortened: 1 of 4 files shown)


etam:
This should also fix tests failing on windows at #2701

golem/client.py
```diff
@@ -725,21 +732,31 @@ def check_test_status(self):
...
 
     def abort_task(self, task_id):
         logger.debug('Aborting task "%r" ...', task_id)
         self.task_server.task_manager.abort_task(task_id)
 
-    def restart_task(self, task_id: str) -> Tuple[bool, Optional[str]]:
+    @inlineCallbacks
+    def restart_task(self, task_id: str):
+        """
+        -> Tuple[Optional[str], Optional[str]]
```
etam: The return type annotation is in comment, because mypy hates `inlineCallbacks` decorator.

etam:
Changes in RPC:
Before:
- `comp.task.create` and `comp.task.restart` returns `Tuple[bool, str]`:
  - `True, ''` on success
  - `False, 'error message'` on failure

After:
- `comp.task.create` and `comp.task.restart` returns `Tuple[Optional[str], Optional[str]]`:
  - `task_id, None` on success
  - `task_id or None, 'error message'` on failure

golem/client.py
```diff
@@ -638,7 +643,9 @@ def task_created(resource_server_result):
 
         def error(exception):
             logger.error("Task '%s' creation failed: %r", task_id, exception)
-            _result.errback(exception)
+            e = CreatingTaskFailed(task_id)
+            e.__cause__ = exception
```
Wiezzel: Maybe move this assignment to `CreatingTaskFailed.__init__()`?

golem/client.py
Wiezzel: :(

golem/interface/client/tasks.py
```diff
@@ -112,8 +112,8 @@ def subtasks(self, id, sort):
     @command(argument=id_req, help="Restart a task")
     def restart(self, id):
         deferred = Tasks.client.restart_task(id)
-        ok, error = sync_wait(deferred)
-        if not ok:
+        new_task_id, error = sync_wait(deferred)
+        if not new_task_id:
             return error
         return None
```
Wiezzel: Actually it might be good idea to return `new_task_id` here as well.

Wiezzel (commented):
Could you check that these changes actually fix test errors in #2701 ?

etam:
> Could you check that these changes actually fix test errors in #2701 ?

I've created a temporary PR with those changes merged (#3035) and the tests on buildbot passed (https://buildbot.golem.network/buildbot/#/builders/3/builds/5263)

golem/client.py
etam: Maybe...? I'm violating Python anyway, because `__cause__` is supposed to be set by calling `raise CreatingTaskFailed(task_id) from exception`. So I don't think it's legitimate to assign this in constructor.

golem/interface/client/tasks.py
etam: :+1:

Wiezzel (approved):
:+1:

mdtanrikulu:
RPC does not return any exception reason while task creation. I have all conditions to have "Not enough GNT" for task creation, in terminal I have the exception but I don't have it on RPC error callback.

etam:
> RPC does not return any exception reason while task creation. I have all conditions to have "Not enough GNT" for task creation, in terminal I have the exception but I don't have it on RPC error callback.

Fixed

mdtanrikulu:
The only problem is, it takes too much time to create task, or at least "response" from the "create task" definition,  if the task file is heavy/detailed or etc. The return/response shouldn't wait for the task to be created. It was a known problem some time ago and solved by the team, but this branch cause it again.

etam:
Done

mdtanrikulu (approved):

PR #3032, opened 2018-07-12
Modified task cost & fee information

Wiezzel:
* Changed `get_total_payment_for_subtasks`:
  - return `(None, None)` instead of `(0, 0)` as empty result
  - include 'sent' payments, not only 'confirmed'
  - return Nones if not all tasks are sent (no partial sums)
* Added `estimated_cost` & `estimated_fee` to `TaskStatus`

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -937,6 +937,12 @@ class Client(HardwarePresetsMixin):
         subtask_ids = list(task_state.subtask_states.keys())
         task_dict['cost'], task_dict['fee'] = \
             self.transaction_system.get_total_payment_for_subtasks(subtask_ids)
+
+        # Convert to string because RPC serializer fails on big numbers
+        for k in ('cost', 'fee', 'estimated_cost', 'estimated_fee'):
+            if task_dict[k] is not None:
+                task_dict[k] = str(task_dict[k])
+
         return task_dict
 
     def get_tasks(self, task_id: Optional[str] = None) \
--- a/golem/client.py
+++ b/golem/client.py
@@ -615,7 +615,9 @@ class Client(HardwarePresetsMixin):
             else:
                 task.header.mask = Mask()
 
-            task_manager.add_new_task(task)
+            estimated_fee = self.transaction_system.eth_for_batch_payment(
+                task.total_tasks)
+            task_manager.add_new_task(task, estimated_fee=estimated_fee)
 
             client_options = self.task_server.get_share_options(task_id, None)
             client_options.timeout = deadline_to_timeout(task.header.deadline)
--- a/golem/transactions/paymentskeeper.py
+++ b/golem/transactions/paymentskeeper.py
@@ -1,6 +1,6 @@
 import logging
 from datetime import datetime
-from typing import Iterable, Tuple
+from typing import Iterable, Tuple, Optional
 
 from eth_utils import encode_hex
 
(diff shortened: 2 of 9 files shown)


Wiezzel:
The functionality has been discussed with @jamuszyn.
I am aware that some tests are failing therefore [WIP].

mdtanrikulu (changes requested):
If you can return fee & cost as string or divided by eth denom, that might be better to avoid application from big number issues.

Wiezzel:
@mdtanrikulu Done.
Test are fixed as well.
@mfranciszkiewicz @Krigpl Please, review.

golem/client.py
```diff
@@ -935,6 +937,12 @@ def get_task(self, task_id: str) -> Optional[dict]:
         subtask_ids = list(task_state.subtask_states.keys())
         task_dict['cost'], task_dict['fee'] = \
             self.transaction_system.get_total_payment_for_subtasks(subtask_ids)
+
+        # Convert wei -> eth because front-end has troubles with large numbers
```
mfranciszkiewicz: Our front-end uses a bignum library, that's an issue with the RPC serializer.

golem/client.py
```diff
@@ -935,6 +937,12 @@ def get_task(self, task_id: str) -> Optional[dict]:
         subtask_ids = list(task_state.subtask_states.keys())
         task_dict['cost'], task_dict['fee'] = \
             self.transaction_system.get_total_payment_for_subtasks(subtask_ids)
+
+        # Convert wei -> eth because front-end has troubles with large numbers
+        for k in ('cost', 'fee', 'estimated_cost', 'estimated_fee'):
+            if task_dict[k]:
+                task_dict[k] = task_dict[k] / denoms.ether
```
mfranciszkiewicz: IMO `str`s would be a better option

Wiezzel: @mdtanrikulu do you agree?

Wiezzel: OK, done.

mdtanrikulu (approved):

mfranciszkiewicz (approved):

golem/transactions/paymentskeeper.py
```diff
@@ -34,23 +34,30 @@ def get_payment_for_subtask(subtask_id):
 
     @staticmethod
     def get_total_payment_for_subtasks(subtask_ids: Iterable[str])\
-            -> Tuple[int, int]:
+            -> Tuple[Optional[int], Optional[int]]:
```
Krigpl, line 37: afaict there's no way one value will be `None` while other won't, so `Optional[Tuple]` looks like a better choice

Wiezzel: Returning a single `None` would make this assignment much uglier: https://github.com/golemfactory/golem/blob/6f614648d4b4d4c5be4068b22040c8cb0662cacb/golem/client.py#L936-L937
I can go with `Union[Tuple[int, int], Tuple[None, None]]` if you don't like `Tuple[Optional[int], Optional[int]]`

Krigpl:
```
task_dict['cost'], task_dict['fee'], *_ = \ 
self.transaction_system.get_total_payment_for_subtasks(subtask_ids), None
```
would work then, but lets not do that :p
I'm fine either way, was just nitpicking.

PR #3043, opened 2018-07-16
Remove TaskComputer dependency from TaskThread classes

mfranciszkiewicz:


tworec (approved):
LGTM.

I just do not like two letter var names eg. `tt`, but it's just a metter of style, and it's applied though whole golem code unfortunately.

Issue #2670, opened 2018-04-17
Use `node_id` when fetching messages from history


jiivan:
Update `golem.network.history.get()` and `golem.task.tasksession.get_task_message()` to use mandatory `node_id` parameter.

PR #3046, opened 2018-07-16
Send ForcePayment

### LINKED ISSUE #2194: [Concent] UC5 / Force Payment - send `ForcePayment`
implement sending `ForcePayment` as a Provider to the Concent service in a situation where some time has passed since the results have been accepted and the payment is not to be found for a subtask or multiple subtasks.

jiivan:
Send force payment when there are overdue incomes in database.
fixes: #2194

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -1235,6 +1235,7 @@ class Client(HardwarePresetsMixin):
         self.p2pservice.push_local_rank(node_id, loc_rank)
 
     def check_payments(self):
+        self.transaction_system.incomes_keeper.update_overdue_incomes()
         after_deadline_nodes = \
             self.transaction_system.get_nodes_with_overdue_payments()
         for node_id in after_deadline_nodes:
--- a/golem/client.py
+++ b/golem/client.py
@@ -1235,7 +1235,6 @@ class Client(HardwarePresetsMixin):
         self.p2pservice.push_local_rank(node_id, loc_rank)
 
     def check_payments(self):
-        self.transaction_system.incomes_keeper.update_overdue_incomes()
         after_deadline_nodes = \
             self.transaction_system.get_nodes_with_overdue_payments()
         for node_id in after_deadline_nodes:
--- a/golem/network/history.py
+++ b/golem/network/history.py
@@ -328,8 +328,8 @@ def add(msg: message.base.Message,
 def get(
         message_class_name: str,
         subtask_id: str,
-        task_id: Optional[str] = None,  # deprecated
-        node_id: Optional[str] = None)\
+        node_id: Optional[str] = None,
+        task_id: Optional[str] = None) \
         -> Optional[message.Message]:
     #  FIXME: Use node_id in queries
     #         https://github.com/golemfactory/golem/issues/2670
--- a/golem/network/history.py
+++ b/golem/network/history.py
@@ -328,8 +328,8 @@ def add(msg: message.base.Message,
 def get(
         message_class_name: str,
         subtask_id: str,
-        task_id: str = None,  # deprecated
-        node_id: str = None)\
+        task_id: Optional[str] = None,  # deprecated
+        node_id: Optional[str] = None)\
         -> Optional[message.Message]:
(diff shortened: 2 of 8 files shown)


golem/network/history.py
```diff
@@ -325,21 +325,36 @@ def add(msg: message.base.Message,
     )
 
 
-def get(message_class_name: str, task_id: str, subtask_id: str)\
+def get(
+        message_class_name: str,
+        subtask_id: str,
+        task_id: Optional[str] = None,  # deprecated
```
maaktweluit: since the order has already changed maybe its better to put the deprecated argument last?

golem/network/history.py
```diff
@@ -325,21 +325,36 @@ def add(msg: message.base.Message,
...
+        node_id: Optional[str] = None)\
         -> Optional[message.Message]:
     #  FIXME: Use node_id in queries
     #         https://github.com/golemfactory/golem/issues/2670
```
maaktweluit: nice first step on this issue :)

so until #2670 is fully fixed `node_id` should be optional?

golem/network/history.py
```diff
@@ -325,21 +325,36 @@ def add(msg: message.base.Message,
...
-            msg_cls=message_class_name,
+    if not (node_id or task_id):
+        raise RuntimeError("Please use node_id, subtask_id pair")
+
+    query = {}
+    query['subtask'] = subtask_id
+    query['msg_cls'] = message_class_name
+    if task_id:
+        query['task'] = task_id
+        warnings.warn(
```
maaktweluit, line 344: what is `warnings.warn` over `logger.warning`?

golem/network/history.py
```diff
@@ -325,21 +325,36 @@ def add(msg: message.base.Message,
     )
 
 
-def get(message_class_name: str, task_id: str, subtask_id: str)\
+def get(
+        message_class_name: str,
+        subtask_id: str,
+        task_id: Optional[str] = None,  # deprecated
+        node_id: Optional[str] = None)\
```
maaktweluit: i think its neater to add an extra space at the end: `) \`
Linter does not enforce this so can leave it :)

maaktweluit (approved):
LGTM!

Some questions, none are breaking so approved!

golem/client.py
```diff
@@ -1235,6 +1235,7 @@ def push_local_rank(self, node_id, loc_rank):
         self.p2pservice.push_local_rank(node_id, loc_rank)
 
     def check_payments(self):
+        self.transaction_system.incomes_keeper.update_overdue_incomes()
```
Krigpl: This isn't necessary and it's even harmful.
It's unnecessary because the line below calls `update_overdue_incomes` in the process.
It's harmful because with two consecutive `update_overdue_incomes` calls the second will always return an empty list, so the `TRUST` below won't get updated.

Krigpl (changes requested):

golem/client.py
jiivan: Thanks @Krigpl , missed that call. Although it seems clanky that `update_...()` is called as a side effect of getter method.

Krigpl: True, summoning @Wiezzel here

golem/network/history.py
jiivan: IMO they're better for everything that fits into [warning categories](https://docs.python.org/3/library/warnings.html#warning-categories). They can be nicely controlled by command line arg (`-W`) and programmatically. It makes it easier to find all occurences of a deprecated call by changing severity to error.

golem/client.py
Wiezzel: OK, that's not a simple "getter method". You're welcome to rename it so that no one else gets confused.

golem/network/history.py
maaktweluit: Thank you for the explanation! super clear and handy

golem/client.py
jiivan: It's outside of scope for this PR. I've created #3055

mfranciszkiewicz (approved):

PR #3061, opened 2018-07-19
Remove sync from EthereumTransactionSystem

Krigpl:
That fixes an issue where `FundsLocker` tries to restore locks at startup but the balances are not known yet so it fails.
But also I think this `sync` approach was not great in general.


maaktweluit (approved):
Woot, old syncs die hard ;)

mfranciszkiewicz (approved):

PR #3062, opened 2018-07-19
expose testing tasks via cli

etam:
Also:
- move task json validation from cli to golemapp
- rpc changes:
  - comp.task.create (create_task):
    - task definition is more rigorously checked and errors in more cases are returned.
  - comp.task.test.status (check_test_status):
    - instead of sending serialized structure as json string, send the
      structure itself. RPC transport takes care of serialization
      already.
    - don't cleanup results.
    - if there's no error, value of "error" is None, instead of True

depends on https://github.com/golemfactory/golem-electron/pull/418


golem/client.py
```diff
@@ -735,6 +740,42 @@ def create_task(self, t_dict):
             logger.exception("Cannot create task %r", t_dict)
             return False, str(ex)
 
+    def _validate_task_dict(self, t_dict) -> None:
+        task_name = ""
+        if 'name' in t_dict.keys():
```
jiivan: `.keys()` is unnecessary

jiivan (approved):
LGTM

mdtanrikulu (changes requested):
Probably while creating test task, there is no task name and it causes error from `_validate_task_dict`. So test result is always `{error: "Length of task name cannot be less than 4 or more than 24 characters." , status: "Error"}` in GUI. If I pass task name for test reason it's checking task name and everything is okay on this side. Not sure maybe it's a good idea to check task name length here.

Apart from that, test result is always from previous test. Is not waiting for the new test results most of the time.

etam:
I don't fully understand your comment

> Probably while creating test task, there is no task name and it causes error from _validate_task_dict. So test result is always {error: "Length of task name cannot be less than 4 or more than 24 characters." , status: "Error"} in GUI. If I pass task name for test reason it's checking task name and everything is okay on this side.

I moved validating of task description from golemcli into core and also added it to check test task description. What you're saying is correct. If you don't set a name for a test task, it will fail at validation. Setting any proper name should solve the problem, and I think it's a better solution than making special cases at validation.

> Not sure maybe it's a good idea to check task name length here.

Where?

> Apart from that, test result is always from previous test. Is not waiting for the new test results most of the time.

Could you elaborate more on that? Steps to reproduce or something?

mdtanrikulu:
> Not sure maybe it's a good idea to check task name length here.

Ignore this, I solved in electron side as you mentioned. 

> Could you elaborate more on that? Steps to reproduce or something?

This local task "Atv Buggy" started after "Pavillion missing" task and results are belong to "Pavillion missing" when I have the results.
screen shot 2018-08-20 at 12 00 55

Same for the other way around for "Pavillion missing" task, results belong to "Atv Buggy" one.

screen shot 2018-08-20 at 12 01 18

You can reproduce the same thing if you have one blend project folder with missing assets and one without.

mdtanrikulu (approved):

PR #3070, opened 2018-07-19
Hardware presets: do not assume exact values in tests

mfranciszkiewicz:
`test_update_config` and `test_update_config_not_changed` should now be deterministic.

### DIFF
--- a/tests/golem/core/test_hardware_presets.py
+++ b/tests/golem/core/test_hardware_presets.py
@@ -16,7 +16,9 @@ class TestHardwarePresets(testutils.DatabaseFixture):
 
     def setUp(self):
         super().setUp()
-        HardwarePresets.initialize(self.tempdir)
+        with patch('golem.core.hardware.free_partition_space',
+                   return_value=7e9):
+            HardwarePresets.initialize(self.tempdir)
         self.config = ClientConfigDescriptor()
 
     def test_initialize(self, *_):
@@ -40,7 +42,7 @@ class TestHardwarePresets(testutils.DatabaseFixture):
     def test_memory(self, *_):
         assert HardwarePresets.memory(-1) == MIN_MEMORY_SIZE
         assert HardwarePresets.memory(1e6) == MIN_MEMORY_SIZE
-        assert HardwarePresets.memory(2**20) == 2 ** 20
+        assert HardwarePresets.memory(2 ** 20) == 2 ** 20
         assert HardwarePresets.memory(1e7) == 1e7
         assert HardwarePresets.memory(7e7) == 7e7
         assert HardwarePresets.memory(9e9) == 7e7
@@ -48,7 +50,7 @@ class TestHardwarePresets(testutils.DatabaseFixture):
     def test_disk(self, *_):
         assert HardwarePresets.disk(-1) == MIN_DISK_SPACE
         assert HardwarePresets.disk(1e6) == MIN_DISK_SPACE
-        assert HardwarePresets.disk(2**20) == 2 ** 20
+        assert HardwarePresets.disk(2 ** 20) == 2 ** 20
         assert HardwarePresets.disk(1e7) == 1e7
         assert HardwarePresets.disk(7e9) == 7e9
         assert HardwarePresets.disk(9e19) == 7e9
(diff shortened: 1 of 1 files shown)


tests/golem/core/test_hardware_presets.py
```diff
@@ -84,15 +90,29 @@ def test_update_config_upper_bounds(self):
...
         # then
-        assert not HardwarePresets.update_config(DEFAULT, self.config)
-        assert not HardwarePresets.update_config(DEFAULT, self.config)
-        assert not HardwarePresets.update_config(DEFAULT, self.config)
-        assert HardwarePresets.update_config(CUSTOM, self.config)
-        assert not HardwarePresets.update_config(CUSTOM, self.config)
-        assert HardwarePresets.update_config(DEFAULT, self.config)
-        assert HardwarePresets.update_config('foo', self.config)
-        assert not HardwarePresets.update_config('foo', self.config)
+        with mock.patch(
```
tworec: with those mocks, config can differ only on `name` field. This test is not testing real/prod code any more... 
I'd rather mock `cpu_cores_available()`, `memory_available()` and `free_partition_space()` methods here.

tests/golem/core/test_hardware_presets.py
```diff
@@ -84,15 +90,29 @@ def test_update_config_upper_bounds(self):
 
     def test_update_config_not_changed(self):
         # given
-        HardwarePreset.create(name='foo', cpu_cores=1, memory=1200000,
-                              disk=2000000)
+        cpu_cores = HardwarePresets.cpu_cores(1)
```
tworec: those are specially crafted values for this test to pass. 
I guess you're trying to bypass test environment limitations/characteristics here, am I right?
But in this shape you are stripping lots of logic from tested class.

tests/golem/core/test_hardware_presets.py
```diff
@@ -54,18 +56,22 @@ def test_update_config_lower_bounds(self):
...
         # when
         config_changed = HardwarePresets.update_config('foo', self.config)
 
         # then
         assert not config_changed
         assert self.config.hardware_preset_name == 'foo'
-        assert self.config.num_cores == 1
-        assert self.config.max_memory_size == 1200000
-        assert self.config.max_resource_size == 2000000
+        assert self.config.num_cores == HardwarePresets.cpu_cores(cpu_cores)
```
tworec: It is even better now, since we are testing `update_config` method here.
I suggest creating three additional tests for those three methods used here (mocking methods checking system resources ie. `cpu_cores_available()` for `cpu_cores()`)

tworec (changes requested):

tests/golem/core/test_hardware_presets.py
mfranciszkiewicz: Those values are not constant and had to be changed. That additional logic needs to be tested elsewhere

tests/golem/core/test_hardware_presets.py
mfranciszkiewicz: But why are we then testing the same deterministic outcome 3 times? This test is assuming a bit too much about the environment.

tworec: three times is abundant, 2x is enough, your're right. You can remove one invocation if you want.

This test is assuming that environment has more than one cpu core, more than 1200000KiB RAM and more than 2000000KiB / 3 * 4 disk space. When you mock methods I've enumerated above, it will be independent from actual env limitations.

mfranciszkiewicz: The problem is that sometimes this test fails, because each `update_config` call reads hardware caps independently and that can change (i.e. free disk space). I made it always return the same value, which _is_ retrieved from the current hw configuration.

tests/golem/core/test_hardware_presets.py
tworec: I think mocking same methods as I've mentioned below/above is sufficient also here. 

What additional logic are you mentioning?

tests/golem/core/test_hardware_presets.py
mfranciszkiewicz: I sincerely thought that this test was supposed to verify that applying the same hw configuration more than once won't trigger a "hardware changed" event. In production, we care only about the real environment and do not care about the values in fashion that this test is written in.

tests/golem/core/test_hardware_presets.py
mfranciszkiewicz: The logic that tests invariance of hardware parameters.

tworec:
here is my proposition. It was easier and shorter to code it than write about it ;)

mfranciszkiewicz:
That’s another way to solve this issue :)

tworec (approved):

tworec:
So, I'm proudly approving my own solution, seeing you said it solves our problem ;)

tworec:
You've correctly interpreted my intention about `test_update_config_not_changed`.  Mocking methods that are building blocks of `update_config` reduces this test coverage, so I've proposed solution with mocking lower level functions which are responsible for reading current env limits. But reading your comments allowed me to realize that we should also add a test with changing this mocks between `update_config` calls, because this happens on prod all the time..

tworec:
Main part of this PR was merged into develop with  #3071

PR #3083, opened 2018-07-24
Remove ethereumpaymentskeeper

Krigpl:
There's no such thing as `EthereumPaymentsKeeper`.
`EthAccountInfo` contained a lot of unnecessary data.
`PaymentDetails` in the DB do not require this data.


golem/task/tasksession.py
```diff
@@ -204,9 +203,15 @@ def verification_finished():
             task_to_compute = get_task_message(
                 'TaskToCompute', task_id, subtask_id)
 
+            eth_address = get_task_message(
+                'ReportComputedTask',
+                task_id,
+                subtask_id,
+            ).eth_account
```
mfranciszkiewicz, line 210: Could you handle the case when `get_task_message` returns `None`?

Krigpl: It wasn't handled before. Also this is response on receiving `ReportComputedTask` so `None` here would indicate some major bug imo.

mfranciszkiewicz: > `None` here would indicate some major bug imo

And we don't care what happens next?

Krigpl: I dunno, we didn't until now apparently  :)
But this is also applies to the `get_task_message` above which is also not handled if it returns None.
Might be worth changing that but it's outside of the scope of this PR.

mfranciszkiewicz (approved):

PR #3086, opened 2018-07-25
B0.17.0 merge 0.16.2

shadeofblue:


shadeofblue:
ok, after a heated discussion we have concluded this does not make any sense

PR #3090, opened 2018-07-26
Replace callback with deferred

Krigpl:
This is cleaner as it doesn't require passing arbitrary callbacks.

### DIFF
--- a/tests/golem/transactions/ethereum/test_ethereumtransactionsystem.py
+++ b/tests/golem/transactions/ethereum/test_ethereumtransactionsystem.py
@@ -304,26 +304,23 @@ class TestEthereumTransactionSystem(TestWithDatabase, LogTestCase,
 
     def test_concent_deposit_enough(self):
         self.sci.get_deposit_value.return_value = 10
-        cb = Mock()
-        self.ets.concent_deposit(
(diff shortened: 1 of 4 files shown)


tests/golem/transactions/ethereum/test_ethereumtransactionsystem.py
```diff
@@ -304,26 +304,23 @@ def test_subscriptions(self):
...
-        cb = Mock()
-        self.ets.concent_deposit(
+        deferred = self.ets.concent_deposit(
             required=10,
             expected=40,
-            cb=cb,
         )
-        cb.assert_called_once_with()
+        called = []
+        deferred.addCallback(lambda _: called.append(True))
```
mfranciszkiewicz:
```
deferred.addErrback(lambda _: self.fail('reasons'))
assert deferred.called
```

just saying

mfranciszkiewicz (approved):

tests/golem/transactions/ethereum/test_ethereumtransactionsystem.py
Krigpl: noice

PR #3096, opened 2018-07-26
Refactoring docker paths

inexxt:
This PR refactors usage of docker paths (`/golem`, `/golem/output`, etc.)

Now, all paths are kept in `DockerJob`, which removed high code duplication - every docker script can read them from params file.


Krigpl (commented):
lgtm, but you need to fix buildbots

inexxt:
Fixed

Krigpl (approved):
lgtm.
btw, in order for the lint buildbot to pass you only need to fix the lines that you touched, not everything that it reported. When there are too many lint fixes it introduces unnecessary noise to the PR which makes it harder to review.

badb (approved):
Ok, just some conflict resolving needed.

PR #3112, opened 2018-08-01
Sharing performance information

Wiezzel:
Estimated performance (benchmark results) is shared in the `metadata` field of `Hello` message. In order to achieve this 'metadata providers' were added to P2PService.

Received metadata from other nodes are stored in the `KnownHosts` table in the database. Performance information obtained this way are used to computed the potential number of providers for netmasking (instead of relying on hard-coded list).


Wiezzel:
One problem with this feature is that `Hello` message could be sent *before* benchmarks finish so that 0.0 would be sent instead of true performance. A way to solve this would be to wait with starting network until benchmarks are computed, but I don't know if this solution wouldn't cause some further trouble. On the other hand the problem is not very serious since benchmarks are computed quite fast and this affects only a couple of first `Hello` messages (most likely sent to seeds).

@mfranciszkiewicz @Krigpl What do you think?

mfranciszkiewicz (approved):
LGTM

mfranciszkiewicz:
@Wiezzel Benchmarks can take time on slower machines (looking at you, Blender), but since the first run is always the longest, IMO we can wait for the results. Either that or we can ignore `performance <= 0` for now.

Wiezzel:
@mfranciszkiewicz I will add waiting for benchmark computation in a separate PR.

golem/network/p2p/p2pservice.py
```diff
@@ -42,6 +42,9 @@
 RANDOM_DISCONNECT_INTERVAL = 5 * 60
 RANDOM_DISCONNECT_FRACTION = 0.1
 
+# Indicates how many KnownHosts can be stored in the DB
+MAX_STORED_HOSTS = 100
```
Krigpl, line 48: `connect_to_network` iterates over all KnownHosts and tries to connect to all of them. Can you confirm that trying to connect to 100 random nodes sequentially won't increase the startup time?

Wiezzel: I modified `connect_to_network` to use only `opt_peer_num` hosts instead of all 100: https://github.com/golemfactory/golem/pull/3112/commits/c90991a9b7e2c325066526092da6ad02cf7f9e54

PR #3115, opened 2018-08-01
Concent integration tests retry until timeout

shadeofblue:
improve DRY-ness


scripts/concent_acceptance_tests/base.py
```diff
@@ -279,25 +303,20 @@ def put_deposit(self, sci: SmartContractsInterface, amount: int):
...
+            lambda: sci.get_eth_balance(sci.get_eth_address()) == 0,
+            "Acquiring tETH timed out",
+        )
 
         self.wait_for_gntb(sci)
 
-        start2 = datetime.datetime.now()
         deposit = False
 
         def deposit_confirmed(_):
```
Krigpl, line 322: You could get rid of this indirect way by explicitly checking whether deposit is already greater than 0.

Krigpl (approved):
thumbupglitch

PR #3119, opened 2018-08-02
Bump cryptography version

mfranciszkiewicz:


mfranciszkiewicz:
Newer versions of `crossbar` are broken on macOS - they use `psutil.Process.cpu_affinity` which is not available on that system.

Elfoniok:
Can we change this label to deferred?

PR #3120, opened 2018-08-02
Use Ethereum address for incomes

Krigpl:
Instead of `sender_node_id`.
Had to use two migration steps because peewee_migrate doesn't handle adding `not_null` constraint very well.


Krigpl:
I need to rethink that.

PR #3121, opened 2018-08-02
TransactionSystem to have its own Ethereum key

Krigpl:
But use old key for backward compatibility for now.

### DIFF
--- a/golem/node.py
+++ b/golem/node.py
@@ -4,6 +4,7 @@ import logging
 import time
 from typing import List, Optional, Callable, Any
 
+from pathlib import Path
 from twisted.internet import threads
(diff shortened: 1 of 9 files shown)


golem/node.py
```diff
@@ -144,6 +153,14 @@ def set_password(self, password: str) -> bool:
                 password=password,
                 difficulty=self._config_desc.key_difficulty,
             )
+            # When Golem is ready to use different Ethereum accounta for
```
mfranciszkiewicz: Typo-h in `accounta`

mfranciszkiewicz (approved):
One test to go.

PR #3128, opened 2018-08-06
Structured local test result

mdtanrikulu:
To apply new test result component on GUI, warning of local task result structured as json


apps/blender/resources/blenderloganalyser.py
```diff
@@ -38,21 +38,23 @@ def find_samples_for_scenes(log_content):
...
 
     wrong_engine = find_wrong_renderer_warning(log_content)
     if wrong_engine:
-        warnings.append("\n{}\n".format(wrong_engine))
+        warnings['wrong_engine'] = wrong_engine
 
     if warnings:
         if return_data.get("warnings"):
-            return_data["warnings"] += "".join(warnings)
+            return_warnings = return_data.get("warnings")
```
Elfoniok, line 53: This may return None

apps/blender/resources/blenderloganalyser.py
```diff
@@ -38,21 +38,23 @@ def find_samples_for_scenes(log_content):
...
+            if return_warnings.get("missing_files"):
+                return_data["warnings"]["missing_files"] = return_data.get(
```
Elfoniok: Its not needed here to access list once again, use += operator or extend metohd return_data["warnings"]["missing_files"].extend(warnings['missing_files'])

apps/blender/resources/blenderloganalyser.py
```diff
@@ -64,14 +66,19 @@ def find_wrong_renderer_warning(log_content):
...
             # extract filename from warning message
-            warnings.add(os.path.basename(missing_file.group(1)))
-    return warnings
+            missing_path = missing_file.group(1)
+            fileInfo = {
+                'baseName': os.path.basename(missing_path),
+                'dirName': os.path.dirname(missing_path)
+            }
+            warnings.append(fileInfo)
+    return list(map(dict, set(tuple(sorted(f.items())) for f in warnings)))
```
Elfoniok: Cant you leave this as it was for missing files? Just store whole missing_path. Otherwise you are unnecessary doubling data storage, where it could be up to the consumer of this class to either get basename or dir name.

Elfoniok (changes requested):

apps/blender/resources/blenderloganalyser.py
Elfoniok: Alternatively you can leave set, as warnings type, and apply sort when adding element to it. This way you don't need to iterate over it so many times,  and you know that you are actually sorting by keys, so this will not change anything ( baseName will always be before dirName) :)

apps/blender/resources/blenderloganalyser.py
mdtanrikulu: While having an if condition  `if return_data.get("warnings"):` how it can return None ?

apps/blender/resources/blenderloganalyser.py
mdtanrikulu: Done, thanks!

apps/blender/resources/blenderloganalyser.py
mdtanrikulu: Oh seems like I forgot to add attribute for sorted, like 
```
sorted(f.items, key=lambda f: f['baseName'])
```
which is fixed now. 

For your first comment, this is the goal of this PR. To make data structured. We are using both information (dirName, baseName) in new local task test result as below to give user more information about the place of the data, so they can place the missing files easily; 

screen shot 2018-08-18 at 18 47 33

Elfoniok: Ok but why don't split path on basename and dir name, right before displaying it?

mdtanrikulu: While having at least 2 different file system to check (Windows, Mac/Linux) and having built in path method already for it, I couldn't find any reason to write another function.

apps/blender/resources/blenderloganalyser.py
```diff
@@ -64,13 +66,22 @@ def find_wrong_renderer_warning(log_content):
...
-    warnings = set()
+    warnings = list()
     for l in log_content.splitlines():
         missing_file = re.search("^Warning: Path '(.*)' not found", l,
                                  re.IGNORECASE)
         if missing_file:
             # extract filename from warning message
-            warnings.add(os.path.basename(missing_file.group(1)))
+            missing_path = missing_file.group(1)
+            fileInfo = (
```
Elfoniok, line 77: Please explain, because i see this unjustified. Why not store whole `missing_path` and extract basename and filename where is necessary?

mdtanrikulu: I already explained it in [here](https://github.com/golemfactory/golem/pull/3128#discussion_r211081287). 
Instead of having one full path of the file, I would rather to create an english syntax for UX purposes. To do it, I needed to have file name and the directory separately. And the lines you pointed, was the best place to split path correctly for all related OS file systems. Thus, rather than writing another function to check and split different file system paths, I used built-in one without any additional effort.

apps/blender/resources/blenderloganalyser.py
```diff
@@ -64,13 +66,22 @@ def find_wrong_renderer_warning(log_content):
...
+                ('baseName', os.path.basename(missing_path)),
+                ('dirName', os.path.dirname(missing_path))
+            )
+            warnings.append(fileInfo)
+
+    if bool(warnings):
```
Elfoniok: conversion to bool is not required

apps/blender/resources/blenderloganalyser.py
```diff
@@ -64,13 +66,22 @@ def find_wrong_renderer_warning(log_content):
...
+        warnings = sorted(warnings, key=lambda t: (t[0][1]))
```
Elfoniok: That is better but you could sort items before inserting them so save some iterations.

Elfoniok (changes requested):

apps/blender/resources/blenderloganalyser.py
jiivan: I also feel that it's a client/frontend responsibility to split & format paths. Backend should just return unique fullpaths.

apps/blender/resources/blenderloganalyser.py
mdtanrikulu: Done, thanks good to know.

apps/blender/resources/blenderloganalyser.py
mdtanrikulu: @jiivan I will keep the same argument, If I was doing any other string manipulations I would definitely go for it. But for the cross platform file paths, I see it more suitable.

apps/blender/resources/blenderloganalyser.py
mdtanrikulu: Done.

Elfoniok (approved):

apps/blender/resources/blenderloganalyser.py
jiivan:
```python
if warnings:
    try:
        return_data["warnings"]["missing_files"].extend(warnings["missing_files"})
    except KeyError:
        return_data["warnings"] = warnings
```

tests/apps/blender/task/test_blenderrendertask.py
```diff
@@ -298,12 +305,17 @@ def test_after_test(self):
...
         after_test_data = self.bt.after_test(results, None)
         warnings = after_test_data["warnings"]
-
-        self.assertTrue("f1.png" in warnings)
-        self.assertTrue("file2.png" in warnings)
-        self.assertTrue("file3.png" in warnings)
-        self.assertEqual(warnings.count("file2.png"), 1)
-        self.assertFalse("file4.png" in warnings)
+        self.assertTrue([f for f in warnings['missing_files']
+                         if f['baseName'] == "f1.png"])
```
jiivan:
```python

self.assertIn({'baseName': 'f1.png', 'dirName': 'XXX'}, warnings['missing_files'])
```

tests/apps/blender/task/test_blenderrendertask.py
```diff
@@ -298,12 +305,17 @@ def test_after_test(self):
...
+        self.assertTrue([f for f in warnings['missing_files']
+                         if f['baseName'] == "file2.png"])
+        self.assertTrue([f for f in warnings['missing_files']
+                         if f['baseName'] == "file3.png"])
+        self.assertEqual(
+            sum(f['baseName'] ==
+                "file2.png" for f in warnings['missing_files']), 1)
```
jiivan: Would this test be necessary if you left `warnings` as set?

jiivan (approved):
Some suggestions

tests/apps/blender/task/test_blenderrendertask.py
mdtanrikulu: Test was there, only thing I did is update to the new form. Who wrote the test probably wanted to be sure if there's any duplicate still. If needless, which it is with set usage, I removed it.

tests/apps/blender/task/test_blenderrendertask.py
mdtanrikulu: Thanks good one.

PR #3162, opened 2018-08-07
Move files around

Krigpl:
Ethereum files were scattered around in two directories without clear reason. One directory is just as fine.
Basically moving all files from `golem/transactions` to `golem/ethereum`.


Wiezzel (commented):
How about renaming `golem.ethereum.ethereumtransactionsystem` to `golem.ethereum.transactionsystem`?

Wiezzel (approved):
👍

PR #3167, opened 2018-08-08
Cleanup overdue income events

Krigpl:
This approach seems to be cleaner. We already (sort of) were decreasing trust in taskserver so no need for that in Client.
Concent does something weird for `overdue` events, they expect a list and I'm pretty sure their logic is not correct - to be resolved later.


mfranciszkiewicz (approved):

Wiezzel (approved):
OK

PR #3171, opened 2018-08-10
RPC client wampcra auth

inexxt:
Implementation of *wampcra*-based client authentication for crossbar router. ([auth based on challenge-response mechanism](https://crossbar.io/docs/Challenge-Response-Authentication/))

PR also adds additional `datadir` parameter to `golemcli`.

Needs to be merged together with related electron PR - https://github.com/golemfactory/golem-electron/pull/413


.gitignore
```diff
@@ -77,3 +77,4 @@ venv/*
 apps/dummy/test_tmp
 .python-version
 luxrender-output.png
+.testmondata
```
badb, line 80: What's that for?

golem/rpc/mapping/rpcmethodnames.py
```diff
@@ -1,6 +1,8 @@
 # pylint: disable=bad-whitespace
 # flake8: noqa
 
+DOCKER_URI = "docker"
```
badb, line 4: Is this necessary?

golemcli.py
```diff
@@ -35,6 +39,7 @@ def start():
         address=('-a', '--address'),
         port=('-p', '--port'),
         trust=('-t', '--verify-trust'),
+        datadir=("-d", "--datadir")
```
badb, line 42: So right now, if there are two Golem instances on the same machine, to connect with the second one with golemcli we must give both  - proper rpc port and datadir, right?

golem/rpc/cert.py
```diff
@@ -56,6 +69,28 @@ def generate_if_needed(self):
         import gc
         gc.collect()
 
+        self.generate_secrets()
```
badb, line 74: Starting golem with just golemapp (without electron) should generate secrets as well.

.gitignore
inexxt: There is a tool that automatically, continually executes tests for code that changed- https://github.com/tarpas/pytest-testmon, and `.testmondata` are temporary files used by this tool to mark what was changed.

golemcli.py
inexxt: Unfortunately, yes.

golem/rpc/mapping/rpcmethodnames.py
inexxt: It was used in two places - and since crossbar errors are sometimes quite difficult to debug, IMO it is better to keep additional constant then to spend time trying to fix potential issues stemming from code duplication later on.

golem/rpc/cert.py
inexxt: Fixed.

golem/rpc/cert.py
```diff
@@ -23,12 +25,24 @@ class CertificateManager:
     PRIVATE_KEY_FILE_NAME = "rpc_key.pem"
     CERTIFICATE_FILE_NAME = "rpc_cert.pem"
 
+    @enum.unique
+    class Crossbar_users(enum.Enum):
```
mfranciszkiewicz: Why not CamelCase?

golem/rpc/router.py
```diff
@@ -114,6 +117,23 @@ def _build_config(address: WebSocketAddress,
                 "dhparam": cert_manager.dh_path,
             }
 
+        # configuration for crsb_users with admin priviliges
+        crsb_users = {
```
mfranciszkiewicz: Could you extract the creation of `crsb_users`?

mfranciszkiewicz (commented):
Looks OK, still have to do some testing though

golem/rpc/cert.py
inexxt: Fixed

golem/rpc/router.py
inexxt: Fixed

mfranciszkiewicz (approved):

mdtanrikulu (approved):

PR #3178, opened 2018-08-14
3125 bump messages to 2.11.0 3

shadeofblue:


golem/network/transport/session.py
```diff
@@ -51,15 +52,15 @@ def __init__(self, conn):
         self.last_message_time = time.time()
         self._disconnect_sent = False
         self._interpretation = {
-            message.Disconnect.TYPE: self._react_to_disconnect,
-            message.Hello.TYPE: self._react_to_hello,
+            library.get_type(message.Disconnect): self._react_to_disconnect,
+            library.get_type(message.Hello): self._react_to_hello,  # fixme  # noqa pylint:disable=no-member
```
etam: Is there a reason for registering `_react_to_hello` here? It's defined in derived classes (that's why `no-member` is required). Derived classes also register their `_react_to_hello` in `__set_msg_interpretations` methods. I think this line should be dropped.

etam (changes requested):

golem/network/transport/session.py
shadeofblue: @etam honestly, I have no idea - I'll check if the tests pass without that line but generally, as you can see, this pull request doesn't add that reaction callback - it only updates the code to use the newer golem-messages API

etam (approved):

jiivan (changes requested):
How about taking this opportunity to refactor out this strange int "`TYPE`" comparison and using message class for it?

```python

# using isinstance here can fail because some messages inherit from other ones
self._interpretation.get(msg.__class__)
```

I mean changing:
https://github.com/golemfactory/golem/blob/39b877ab0c37db438412716d296d9173999fb977/golem/network/transport/session.py#L74

and

https://github.com/golemfactory/golem/blob/39b877ab0c37db438412716d296d9173999fb977/golem/network/transport/session.py#L204-L221

golem/task/tasksession.py
```diff
@@ -339,7 +338,7 @@ def send_report_computed_task(
         # the `ForceReportComputedTask` message to the Concent will be
         # cancelled and thus, never sent to the Concent.
 
-        delayed_forcing_msg = message.ForceReportComputedTask(
+        delayed_forcing_msg = message.concents.ForceReportComputedTask(
```
jiivan, line 341: Yes, using proper path looks like a good route to take :)

jiivan (commented):
Proper paths would look so nice here.
`message.concent.ForceReportComputedTask` instead of `message.ForceReportCOmputedTask`

jiivan (approved):

PR #3180, opened 2018-08-14
Change return value of add_payment_info

Krigpl:
The whole `Payment` object is simply not needed.


maaktweluit (approved):
LGTM!

No tests changes needed?

Krigpl:
@maaktweluit Looks like it

PR #3182, opened 2018-08-14
Remove random test

Krigpl:
This really is an integration test for third party library, I doubt it's valuable.


maaktweluit (approved):
I think this could be valuable when:
- We do not trust the ethereum package tests
- We use it in exactly this way

Guess this is not applicable so LGTM!

Wiezzel (approved):

PR #3183, opened 2018-08-14
Ability to set custom gas price during withdrawal

Krigpl:
One can pass a custom gas price during RPC withdrawal call. The price in is wei, not gwei. And is expected to be an int since, it'll be lower than 10^13.
Also `get_withdraw_gas_cost` now returns cost in gas, not in eth, so it should be multiplied by gas price to get the total ETH cost.


shadeofblue (approved):
looks good as far as I can tell - please communicate with @mdtanrikulu to make sure that both this and the Electron's changes make it into the nearest release before merging

mdtanrikulu (approved):

golem/ethereum/transactionsystem.py
```diff
@@ -1,797 +1,806 @@
-import functools
```
shadeofblue: @mdtanrikulu line endings? ...

PR #3189, opened 2018-08-16
Check next subtask in "react to wants to compute"

maaktweluit:
before starting handshake

This is required for the task "buffer provider requests" and is an improvement that could be merged before the next release

### DIFF
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -421,8 +421,7 @@ class TaskManager(TaskEventListener):
                                  op=SubtaskOp.ASSIGNED)
         return ctd, False, False
 
-
-    def check_next_subtask(
+    def check_next_subtask(  # noqa pylint: disable=too-many-arguments
             self, node_id, node_name, task_id, estimated_performance, price,
             max_resource_size, max_memory_size, num_cores=0):
         """ Check next subtask from task <task_id> to give to node with
@@ -450,7 +449,7 @@ class TaskManager(TaskEventListener):
             max_resource_size, max_memory_size, num_cores,
         )
         if task_id not in self.tasks:
-            logger.info("Cannot find task {} in my tasks".format(task_id))
+            logger.info("Cannot find task in my tasks. task_id=%r", task_id)
             return False, True, False
 
         task = self.tasks[task_id]
@@ -459,8 +458,8 @@ class TaskManager(TaskEventListener):
             return False, False, False
 
         if not self.task_needs_computation(task_id):
-            logger.info(f'Task does not need computation; '
-                        f'provider: {node_name} - {node_id}')
+            logger.info('Task does not need computation. provider=%r - %r',
+                        node_name, node_id)
             return False, False, False
 
         verdict = task.should_accept_client(node_id)
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -361,13 +361,10 @@ class TaskManager(TaskEventListener):
         :param max_memory_size:
         :param num_cores:
         :param address:
-        :return (ComputeTaskDef|None, bool, bool): Function returns a triplet.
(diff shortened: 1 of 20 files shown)


mfranciszkiewicz (approved):

golem/task/taskmanager.py
```diff
@@ -361,42 +361,38 @@ def get_next_subtask(
...
-        if not self.task_needs_computation(task_id):
-            logger.info(f'Task does not need computation; '
-                        f'provider: {node_name} - {node_id}')
-            return None, False, False
+        if self.should_wait_for_node(task_id, node_id):
+            return None
+
+        task = self.tasks[task_id]
+
+        if task.get_progress == 1.0:
```
mfranciszkiewicz: Should be `task.get_progress()`

golem/task/taskmanager.py
```diff
@@ -361,42 +361,38 @@ def get_next_subtask(
...
+            logger.error("Task already computed")
```
mfranciszkiewicz: Could you add more information to this message, e.g. `task_id`, `node_id`, `node_name`?

golem/task/taskmanager.py
```diff
@@ -426,7 +424,62 @@ def check_compute_task_def():
...
+
+        task = self.tasks[task_id]
+
+        verdict = task.should_accept_client(node_id)
+        if verdict == AcceptClientVerdict.ACCEPTED:
+            return False
+        elif verdict == AcceptClientVerdict.SHOULD_WAIT:
+            logger.warning("Waiting for results from %s on %s", node_id,
+                           task_id)
+        else:
```
mfranciszkiewicz: I think we should return `False` here - we're sending `WaitingForResults` instead of `CannotAssignTask` to banned nodes.

golem/task/tasksession.py
```diff
@@ -441,39 +441,50 @@ def _react_to_want_to_compute_task(self, msg):
...
-            self.dropped()
-            return
+                # TODO: Queue requests here
+
+                ctd = self.task_manager.get_next_subtask(
+                    self.key_id, msg.node_name, msg.task_id, msg.perf_index,
+                    msg.price, msg.max_resource_size, msg.max_memory_size,
+                    msg.num_cores, self.address)
+            else:
+                ctd = None
```
mfranciszkiewicz: Minor: could be simplified with `ctd` set initially to `None`

tests/golem/task/dummy/runner.py
```diff
@@ -184,6 +180,10 @@ def shutdown():
         reactor.running and reactor.callFromThread(reactor.stop)
         logging.shutdown()
         if os.path.exists(datadir):
+            logfile = path.join(datadir, "logs", "golem.log")
```
mfranciszkiewicz: Minor: code is copy-pasted, maybe a function instead?

mfranciszkiewicz (changes requested):
2 important issues

golem/task/taskmanager.py
maaktweluit: Good catch, done!

golem/task/taskmanager.py
maaktweluit: added

golem/task/taskmanager.py
maaktweluit: Kept that as it was, but was thinking the same

Should we fix that?

golem/task/tasksession.py
maaktweluit: Nice one! will clean it up

tests/golem/task/dummy/runner.py
maaktweluit: like it! moved to a private function

golem/task/taskmanager.py
maaktweluit: fixed

mfranciszkiewicz (approved):

PR #3191, opened 2018-08-16
Verify requestors balance and deposit

### LINKED ISSUE #2766: [Concent] Provider checks Requestor's GNTB balance and deposits

jiivan:
fixes: #2766


golem/ethereum/transactionsystem.py
```diff
@@ -507,6 +521,17 @@ def _refresh_balances(self) -> None:
         except Exception as e:  # pylint: disable=broad-except
             log.warning('Failed to update balances: %r', e)
 
+    def balance(self, account_address) -> int:
```
Krigpl: How about making `get_available_gntb` take an address argument and return `sci.get_gntb_balance` when address is provided and not equal to our address?

golem/task/tasksession.py
```diff
@@ -551,6 +551,31 @@ def _cannot_compute(reason):
         if not self.concent_service.enabled and msg.concent_enabled:
             # We can't provide what requestors wants
             _cannot_compute(reasons.ConcentDisabled)
+            return
+
+        # TODO number_of_subtasks
+        total_task_price = msg.price * number_of_subtasks
+        transaction_system = self.task_server.client.transaction_system
+        requestors_gntb_balance = transaction_system.balance(
+            account_address=msg.requestor_ethereum_public_key,
```
Krigpl: Address is not the same as public key

golem/task/tasksession.py
```diff
@@ -551,6 +551,31 @@ def _cannot_compute(reason):
...
+            requestors_deposit_value = transaction_system.concent_balance(
+                account_address=msg.requestor_ethereum_public_key,
+            )
+            if requestors_deposit_value < (total_task_price * 2):
+                _cannot_compute(reasons.InsufficientDeposit)
+                return
+            requestors_deposit_timelock = transaction_system.concent_timelock(
+                account_address=msg.requestor_ethereum_public_key,
+            )
+            if requestors_deposit_timelock < 48 * 3600:
```
Krigpl: magic const ?

Krigpl: also `timelock` is a timestamp in the future or 0, so this check is invalid.

Krigpl (changes requested):

golem/ethereum/transactionsystem.py
jiivan: Thanks, missed that method.

golem/task/tasksession.py
```diff
@@ -551,6 +550,34 @@ def _cannot_compute(reason):
...
+            _cannot_compute(reasons.InsufficientBalance)
+            return
+        if msg.concent_enabled:
+            requestors_deposit_value = transaction_system.concent_balance(
+                account_address=msg.requestor_ethereum_address,
+            )
+            if requestors_deposit_value < (total_task_price * 2):
+                _cannot_compute(reasons.InsufficientDeposit)
+                return
+            requestors_deposit_timelock = transaction_system.concent_timelock(
```
Krigpl, line 573: This can also return 0 which means that the deposit is not timelocked - so safe to use.

golem/core/variables.py
```diff
@@ -53,6 +53,8 @@
 }
 
 CONCENT_PULL_INTERVAL = (MAX_CONCENT_PING_INTERVAL / 2).total_seconds()
+# 48 hours to accept requestors deposit
+CONCENT_MIN_DEPOSIT_TIMELOCK = 48 * 3600
```
Krigpl: iirc the deposit timelock will be set to 48h, so using the same value as minimum is pointless.
shouldn't this value depend on the task timeout for example?

jiivan: As far as I understand the documentation, it states that provider should reject subtask if requestor has timelock shorter than 48h. @lukasz-glen ?

jiivan: Now I understand @Krigpl . Removed it completely. :)

Krigpl (approved):
thumbupglitch

shadeofblue (approved):
as far as I could tell, looks more or less good ;)

PR #3202, opened 2018-08-20
Marketplace implementation stub in Rust

Krigpl:
This is my proposal for integrating Rust into Golem and it's supposed to be an RFC, I'm not married to any approach used here.

Please disregard the `pick_provider` implementation as it's not relevant at all and only demonstrates the idea.

And the idea is that, there's gonna be one rust library which all the public API exported in `rust/golem/src/lib.rs`. This is the only file that will take care of converting to/from Python/Rust.
The Rust code itself will be modularized and able to share common code. It will have extensive coverage of unit tests as usual for libraries.

Then in certain places in Golem core we will just import relevant pieces of the Rust library. And these will be covered with very simple integration test(s) (I'd expect usually just a single one) to check whether the Python/Rust conversions are correct.


.gitignore
```diff
@@ -75,5 +79,5 @@ venv/*
 
 # dummy test temp files
 apps/dummy/test_tmp
-.python-version
-luxrender-output.png
+.python-version
```
maaktweluit: What do these files have to do with marketplace / rust?

rust/golem/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+#[macro_use]
+extern crate cpython;
+
+use cpython::{Python, PyResult, PyList};
+
+mod marketplace;
+
+#[allow(non_snake_case)]
+fn marketplace__pick_provider(py: Python, pyoffers: PyList) -> PyResult<u32> {
```
maaktweluit: Why the double underscore? 

i think we should stick to rust naming conventions an not disable them.

rust/golem/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+    let mut offers = vec![];
```
maaktweluit: Since the capacity of the vec is already known you could make it:

```
let mut offers = Vec::with_capacity(pyoffers.len(py));

maaktweluit (commented):
LGTM!

Some small questions, will approve on reply :)

For the tests, should we also add some breaking cases? like passing it integers or string?

.gitignore
Krigpl: Looks like my editor changed new line characters or something

rust/golem/src/lib.rs
Krigpl: It aims to use `module__function_name` style, it's of course arbitrary but since we're gonna have more modules in Rust I tried to make it legible (since module can also have underscores, so to avoid `module_name_function_name`). But we can use a different style.

rust/golem/src/lib.rs
Krigpl: Quite indeed.

mfranciszkiewicz:
I'm OK with this solution, but I have a question to all reviewers: is it better to build the extension as a part of Golem sources or to only have the pyext library as a dependency, but keep the sources in the same repo?

rust/golem/src/lib.rs
maaktweluit: ok, sounds good for the `py-ext` part of the lib.

lets keep the plain rust libs clean and 100% rust standards :)

maaktweluit:
@igor replied :) did you apply the change?

@mfranciszkiewicz i think the `py-ext` rust lib can be part of the golem code-base, and the plain rust libraries should be published as crate.

Krigpl:
> i think the py-ext rust lib can be part of the golem code-base, and the plain rust libraries should be published as crate.

What's your reasoning for that?
I don't see Rust stuff as an external generic library that Golem will use, but part of Golem which just happens to be written in Rust.

maaktweluit:
This depends, i had in mind they would also be used in for example Unlimited.

When those libs are only used in golem i fully agree with you

Wiezzel:
If it's in the same repo then it should be built as a part of Golem I believe, not as an external crate.
But I also believe we should consider splitting the Golem codebase into separate repositories and crates/libraries as we move towards a more decoupled architecture.

Krigpl:
> But I also believe we should consider splitting the Golem codebase into separate repositories and crates/libraries as we move towards a more decoupled architecture.

Separate repos only make sense when there's absolute need for them, otherwise they only introduce annoying dependency management issues.
We're building an application not a library so moving a chunk of code to a separate repo just because it can be moved is not justified. 

> This depends, i had in mind they would also be used in for example Unlimited.

I don't know how Unlimited is supposed to be related to Golem Core but if we end up sharing significant amount of code I think it could live in this repo as well. FWIW I also think Concent should live in this main repo as well - this would eliminate the need for golem-messages (which currently includes much more than just the messages, e.g. cryptography and protocol constants, so it's abused already), golem-smart-contracts-interface and golem-verificator (and most likely more in the future). Even golem-electon could be moved here - solving the RPC interface synchronization issues.
Not saying we should do all of that right now, but that's my reasoning why Rust should stay in this repo.

A nice read on this topic: https://cacm.acm.org/magazines/2016/7/204032-why-google-stores-billions-of-lines-of-code-in-a-single-repository/fulltext

maaktweluit:
> Separate repos only make sense when there's absolute need for them

Agreed, i thought there was a need, but if concent and unlimited join the repo there is none indeed.

Lets keep them together for now and split when we see a good reason to

mfranciszkiewicz (approved):

Issue #3037, opened 2018-07-13
Setting CPU affinity for Docker containers is broken


mfranciszkiewicz:
Short description of `golem.core.hardware.cpu_cores_available()`:
- reads CPU affinity of Golem's process or the number of system cores as fallback,
- returns a list of available cores, e.g. `[1, 3, 4]`
- `len(cpu_cores_available())` determines the number of cores assigned to the VM (Windows, macOS)
- the returned list is used for specifying cores to be used by Docker containers

`cpu_cores_available()` is broken because:
- _fixed_ ~it no longer works as described in its doc and returns the first (`0`) core~,
- is used both for containers running inside a VM (Windows, macOS) and on the host OS (Linux)
- core numbering inside a VM is not related to Golem's process affinity on host

Issues that follow:
- the first (`0`) core should be skipped on the host OS due to performance reasons
    - Docker VM execution should not be scheduled on the first core (we need to set CPU affinity accordingly) on Windows and macOS
    - Docker containers should not be running on the `0` core on a Linux host
- containers running inside the VM should not fully utilize the `0` core (as seen from inside the VM); a cap is needed for the guest OS to function properly.

Update:
28 V 2019 - "no longer works as described in doc" marked as fixed

maaktweluit:
@mfranciszkiewicz was this solved by the 2 linked PR's?

mfranciszkiewicz:
@maaktweluit The issue is still valid. I've updated the comment with fixes introduced.

PR #3203, opened 2018-08-20
Docker for Mac

mfranciszkiewicz:
Initial support for Docker for Mac.

- computations are performed on the default VM
- `docker-machine` is no longer used on macOS
- manages VM configuration via a JSON configuration file
- installer PR: https://github.com/golemfactory/homebrew-golem/pull/37

TODO (in follow up PRs):
- ~decouple DockerManager and Hypervisor classes~ #3219
- ~move *Hypervisor classes to their own modules~ #3223
- ~stop the VM on golem shutdown~


golem/docker/config.py
```diff
@@ -37,14 +37,18 @@ def build_config(self, config_desc):
...
                 cpu_set = [str(c) for c in cpu_cores[:max_cpus]]
                 host_config['cpuset'] = ','.join(cpu_set)
+            except Exception as exc:  # pylint: disable=broad-except
+                logger.warning('Cannot set the CPU set: %r', exc)
 
-            with self._try():
+            try:
                 host_config['mem_limit'] = int(max_memory_size) * 1024
+            except Exception as exc:  # pylint: disable=broad-except
+                logger.warning('Cannot set the CPU set: %r', exc)
```
maaktweluit: `Memory limit` instead of `CPU set`?

requirements-win_to-freeze.txt
```diff
@@ -1,5 +1,5 @@
 --extra-index-url https://builds.golem.network
-pywin32=221
+pywin32=223
```
maaktweluit, line 2: this update should also be in the non frozen version

maaktweluit (commented):
LGTM!

2 small comments

Will have one more look tomorrow since its quite a big PR

requirements-win_to-freeze.txt
mfranciszkiewicz: It's there
https://github.com/golemfactory/golem/blob/48921417e18aa054e76fcfcbed8d6a1a3ab7eee3/requirements-win.txt#L1-L5

golem/docker/config.py
mfranciszkiewicz: Whoops, fixing

requirements-win_to-freeze.txt
maaktweluit: i see, sorry then, did not see it change

golem/docker/commands.py
```diff
@@ -0,0 +1,96 @@
...
+            shell: bool = False) -> Optional[str]:
+
+        command = cls.commands.get(command_name)
+        if not command:
+            logger.warning('Unknown command: %s', command_name)
+        elif isinstance(command, list):
+            return cls._command(command[:], vm_name, args, shell)
+        elif callable(command):
+            return command(vm_name, args, shell)
+        return str()
```
Wiezzel: ??

golem/docker/commands.py
```diff
@@ -0,0 +1,96 @@
...
+
+        if args:
+            command += args
+        if vm_name:
+            command += [vm_name]
+
+        logger.debug('Docker command: %s', command)
+
+        try:
+            output = subprocess.check_output(
```
Wiezzel: Maybe add a `timeout`?

golem/docker/config.py
```diff
@@ -37,14 +37,18 @@ def build_config(self, config_desc):
             num_cores = config_desc.num_cores
             max_memory_size = config_desc.max_memory_size
 
-            with self._try():
+            try:
                 cpu_cores = cpu_cores_available()
                 max_cpus = min(len(cpu_cores), max(int(num_cores), 1))
                 cpu_set = [str(c) for c in cpu_cores[:max_cpus]]
                 host_config['cpuset'] = ','.join(cpu_set)
+            except Exception as exc:  # pylint: disable=broad-except
```
Wiezzel, line 73: Why do we need such broad except clause while possible errors are very limited?

golem/docker/config.py
```diff
@@ -37,14 +37,18 @@ def build_config(self, config_desc):
...
+                logger.warning('Cannot set the CPU set: %r', exc)
 
-            with self._try():
+            try:
                 host_config['mem_limit'] = int(max_memory_size) * 1024
+            except Exception as exc:  # pylint: disable=broad-except
```
Wiezzel: See comment for line 45.

golem/docker/manager.py
```diff
@@ -41,64 +43,60 @@ class DockerManager(DockerConfigManager):
...
 
-    def __init__(self, config_desc=None):
 
-        super(DockerManager, self).__init__()
+class DockerManager(DockerConfigManager):
 
-        self.hypervisor = None
-        self.docker_machine = FALLBACK_DOCKER_MACHINE_NAME
+    def __init__(self, config_desc=None):
+        super(DockerManager, self).__init__()
```
Wiezzel: In Python 3 you can write
```python
super().__init__()
```
😮

golem/docker/manager.py
```diff
@@ -41,64 +43,60 @@ class DockerManager(DockerConfigManager):
...
-            return VirtualBoxHypervisor.instance(self)
-        elif is_osx():
-            return XhyveHypervisor.instance(self)
-        return None
-
     @report_calls(Component.docker, 'instance.check')
     def check_environment(self):
         if self._env_checked:
-            return bool(self.docker_machine)
+            return bool(self.hypervisor)
```
Wiezzel, line 35: What if `self._select_hypervisor()` succeeds but `self.command('help')` fails? Wouldn't it return false positive in such situation?

golem/docker/manager.py
```diff
@@ -167,25 +145,27 @@ def build_config(self, config_desc):
         cpu_count = self.min_constraints['cpu_count']
         memory_size = self.min_constraints['memory_size']
 
-        with self._try():
+        try:
             cpu_count = max(int(config_desc.num_cores), cpu_count)
+        except Exception as exc:  # pylint: disable=broad-except
```
Wiezzel: Why do we need such broad except clause while possible errors are very limited?

golem/docker/manager.py
```diff
@@ -167,25 +145,27 @@ def build_config(self, config_desc):
...
-        with self._try():
+        try:
             memory_size = max(int(config_desc.max_memory_size) // 1024,
                               memory_size)
-
-        with self._try():
-            if config_desc.docker_machine_name:
-                self.docker_machine = config_desc.docker_machine_name
-                self._env_checked = False
+        except Exception as exc:  # pylint: disable=broad-except
```
Wiezzel: See comment for line 150.

golem/docker/manager.py
```diff
@@ -401,25 +300,157 @@ def _wait_for_tasks(self, sb, cb):
...
+class Hypervisor(object):
+
+    POWER_UP_DOWN_TIMEOUT = 30 * 1000  # milliseconds
+    SAVE_STATE_TIMEOUT = 120 * 1000  # milliseconds
+    COMMAND_HANDLER = DockerCommandHandler
+
+    _instance = None
+
+    def __init__(self, docker_manager: DockerManager,
+                 docker_vm: str = DOCKER_VM_NAME) -> None:
```
Wiezzel: `docker_vm_name` or simply `vm_name` would make a better field/variable name.

golem/docker/manager.py
```diff
@@ -401,25 +300,157 @@ def _wait_for_tasks(self, sb, cb):
...
+    @classmethod
+    def instance(cls, docker_manager):
+        if not cls._instance:
+            cls._instance = cls._new_instance(docker_manager)
+        return cls._instance
+
+    def create(self, name: Optional[str] = None, **params):
+        raise NotImplementedError
+
+    def remove(self, name: Optional[str] = None) -> bool:
```
Wiezzel: If `name` is an optional parameter we should probably add this as well:
```python
name = name or self._docker_vm
```

golem/docker/manager.py
```diff
@@ -401,25 +300,157 @@ def _wait_for_tasks(self, sb, cb):
...
+    def restart_ctx(self, name: Optional[str] = None):
+        raise NotImplementedError
+
+    @contextmanager
+    def recover_ctx(self, name: Optional[str] = None):
+        raise NotImplementedError
+
+    @classmethod
+    def _new_instance(cls,
+                      docker_manager: DockerManager) -> Optional['Hypervisor']:
```
Wiezzel: Why `Optional`?

golem/docker/manager.py
```diff
@@ -401,25 +300,157 @@ def _wait_for_tasks(self, sb, cb):
...
+            cls._instance = cls._new_instance(docker_manager)
+        return cls._instance
+
+    def create(self, name: Optional[str] = None, **params):
+        raise NotImplementedError
+
+    def remove(self, name: Optional[str] = None) -> bool:
+        logger.info("Hypervisor: removing VM '%s'", name)
+        try:
+            self._docker_manager.command('rm', name)
```
Wiezzel: Is it really necessary to create a two-way dependency like this? `DockerManager.command()` calls the hypervisor's `COMANND_HANDLER` so why cannot hypervisor call it by itself?

golem/docker/manager.py
```diff
@@ -858,5 +871,182 @@ def _config(self, name):
...
+        'app': f'{APP}/Contents/MacOS/Docker',
+        'driver': 'com.docker.driver',
+        'hyperkit': 'com.docker.hyperkit',
+        'vpnkit': 'com.docker.vpnkit',
+    }
+
+    @classmethod
+    def start(cls, *_args, **_kwargs) -> None:
+        try:
+            subprocess.check_call(['open', '-g', '-a', cls.PROCESSES['app']])
```
Wiezzel: `timeout`?

golem/docker/manager.py
Wiezzel: OK, I see it is done in the next PR.

golem/docker/manager.py
```diff
@@ -858,5 +871,182 @@ def _config(self, name):
...
+        started = time.time()
+
+        while any(map(cls._pid, cls.PROCESSES)):
+            if time.time() - started >= cls.TIMEOUT:
+                logger.error('Docker for Mac: VM start timeout')
+                return
+            time.sleep(0.5)
+
+    @classmethod
+    def _pid(cls, key: str = 'app') -> Optional[int]:
```
Wiezzel: This method (and `_pipe()` as well) looks like a general macOS/linux utility function rather than something docker-for-mac-specific.

golem/docker/manager.py
```diff
@@ -858,5 +871,182 @@ def _config(self, name):
...
+        except (FileNotFoundError, subprocess.CalledProcessError):
+            return None
+
+        try:
+            return int(line.split()[1])
+        except (IndexError, TypeError, ValueError):
+            return None
+
+    @staticmethod
+    def _pipe(cmd: List[str], pipe: List[str]):
```
Wiezzel: I'd rename args to `source_cmd`, `sink_cmd`.
Also see comment for `_pid()`.

golem/docker/manager.py
```diff
@@ -858,5 +871,182 @@ def _config(self, name):
...
+        update = dict(cpus=cpu, memoryMiB=mem)
+
+        try:
+            self._update_config(update)
+        except Exception as e:  # pylint: disable=broad-except
+            logger.error("Docker for Mac: unable to update config: %r", e)
+
+    def constraints(self, name: Optional[str] = None) -> Dict:
+        if not os.path.exists(self.CONFIG_FILE):
+            self.start_vm()
```
Wiezzel: Isn't `wait_until_started()` necessary here?

Wiezzel (commented):
I have some change suggestions. Probably will need to re-review after all further refactoring is merged.

golem/docker/commands.py
mfranciszkiewicz: `== ''`

golem/docker/config.py
mfranciszkiewicz: `cpu_cores_available` is severely broken at the moment (https://github.com/golemfactory/golem/issues/3037) and can throw a wide range of exceptions.

golem/docker/config.py
mfranciszkiewicz: Here - agreed

golem/docker/manager.py
mfranciszkiewicz: Yes, this was probably auto-generated by the IDE and overlooked by me.

golem/docker/commands.py
Wiezzel: Ok, so why not `None` (return type is `Optional`) and if for some reason empty string is better then why call `str()` instead of using literal `''`?

golem/docker/manager.py
mfranciszkiewicz: If `help` fails, an `EnvironmentError` is raised and Golem will refuse to start.

golem/docker/commands.py
mfranciszkiewicz: Timeouts will vary between pulling images and checking vm's status. Should every command have one?

Wiezzel: Pulling images may take long time, so probably not every one should have a timeout. But if checking vm's status does not finish in 1 minute then something is definitely wrong and we don't want to hang forever. This probably means we need some command -> timeout mapping instead of single constant.

mfranciszkiewicz: `docker` and `docker-machine` commands already have timeouts built in. I wouldn't add another layer of complexity that will bring new problems to the table.

Wiezzel: OK, built-in timeouts will do.

golem/docker/manager.py
mfranciszkiewicz: Ok, will do.

golem/docker/manager.py
mfranciszkiewicz: It's included in `start`

golem/docker/manager.py
mfranciszkiewicz: `open` terminates instantly

golem/docker/manager.py
mfranciszkiewicz: Fixed in subsequent PRs

golem/docker/commands.py
mfranciszkiewicz: Resolved in https://github.com/golemfactory/golem/pull/3223

golem/docker/config.py
mfranciszkiewicz: Resolved in https://github.com/golemfactory/golem/pull/3223

golem/docker/manager.py
mfranciszkiewicz: Resolved in https://github.com/golemfactory/golem/pull/3223

golem/docker/manager.py
mfranciszkiewicz: Resolved in https://github.com/golemfactory/golem/pull/3223

golem/docker/manager.py
mfranciszkiewicz: Resolved in #3223

golem/docker/manager.py
mfranciszkiewicz: Resolved in #3223

golem/docker/manager.py
mfranciszkiewicz: Resolved in #3223

golem/docker/manager.py
mfranciszkiewicz: Resolved in #3223

golem/docker/commands/docker.py
```diff
@@ -0,0 +1,96 @@
...
+        if not command:
+            logger.error('Unknown command: %s', command_name)
+        elif isinstance(command, list):
+            return cls._command(command[:], vm_name, args, shell)
+        elif callable(command):
+            return command(vm_name, args, shell)
+        return None
+
+    @classmethod
+    def wait_until_started(cls):
```
Wiezzel: `-> None`

golem/docker/commands/docker_for_mac.py
```diff
@@ -0,0 +1,70 @@
...
+        try:
+            subprocess.check_call(['kill', str(pid)])
+        except subprocess.CalledProcessError:
+            logger.error('Docker for Mac: unable to stop the app')
+            return
+
+        cls.wait_until_stopped()
+
+    @classmethod
+    def pid(cls):
```
Wiezzel: `-> Optional[int]`

golem/docker/commands/docker_for_mac.py
```diff
@@ -0,0 +1,70 @@
...
+        return unix_pid(cls.PROCESSES['app'])
+
+    @classmethod
+    def status(cls) -> str:
+        return 'Running' if cls.pid() else ''
+
+    @classmethod
+    def wait_until_stopped(cls):
```
Wiezzel: `-> None`

golem/docker/commands/docker_for_mac.py
```diff
@@ -0,0 +1,70 @@
...
+
+        cls.wait_until_stopped()
+
+    @classmethod
+    def pid(cls):
+        return unix_pid(cls.PROCESSES['app'])
+
+    @classmethod
+    def status(cls) -> str:
+        return 'Running' if cls.pid() else ''
```
Wiezzel: This `'Running'` is a kind of magic constant, isn't it?

golem/docker/config.py
```diff
@@ -24,11 +31,31 @@
...
+
+
+GetConfigFunction = Callable[[], Dict[str, Any]]
+
 
 class DockerConfigManager(object):
 
     def __init__(self):
         self.container_host_config = dict(DEFAULT_HOST_CONFIG)
+        self.hypervisor = None
```
Wiezzel: `: Optional[Hypervisor]`

golem/docker/config.py
```diff
@@ -55,9 +86,6 @@ def install(cls, *args, **kwargs):
             DockerTaskThread.docker_manager = docker_manager
         return DockerTaskThread.docker_manager
 
-    @contextmanager
-    def _try(self):
-        try:
-            yield
-        except Exception:
-            pass
+    def quit(self):
```
Wiezzel: `-> None`

golem/docker/hypervisor/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def __init__(self,
+                 get_config: GetConfigFunction,
+                 vm_name: str = DOCKER_VM_NAME) -> None:
+
+        self._get_config = get_config
+        self._vm_name = vm_name
+
+    @classmethod
+    def is_available(cls):
```
Wiezzel: `-> bool`

golem/docker/hypervisor/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+        logger.info("Docker: stopping %s", name)
+
+        try:
+            self.command('stop', name)
+            return True
+        except subprocess.CalledProcessError as e:
+            logger.warning("Docker: failed to stop the VM: %r", e)
+        return False
+
+    def create(self, name: Optional[str] = None, **params):
```
Wiezzel: `-> bool`

golem/docker/hypervisor/docker_for_mac.py
```diff
@@ -0,0 +1,96 @@
...
+        return False
+
+    def remove(self, name: Optional[str] = None) -> bool:
+        # We do not control VM removal
+        return False
+
+    def constrain(self, name: Optional[str] = None, **params) -> None:
+        cpu = params.get(CONSTRAINT_KEYS['cpu'])
+        mem = params.get(CONSTRAINT_KEYS['mem'])
+        update = dict(cpus=cpu, memoryMiB=mem)
```
Wiezzel: These should be constants: `cpus`, `memoryMiB`.

golem/docker/hypervisor/docker_for_mac.py
```diff
@@ -0,0 +1,96 @@
...
+            logger.error("Docker for Mac: unable to update config: %r", e)
+
+    def constraints(self, name: Optional[str] = None) -> Dict:
+        if not os.path.exists(self.CONFIG_FILE):
+            self.start_vm()
+        if not os.path.exists(self.CONFIG_FILE):
+            raise RuntimeError('Docker for Mac: unable to read VM config')
+
+        with open(self.CONFIG_FILE) as config_file:
+            config = json.load(config_file)
```
Wiezzel: Maybe extract lines 54-60 to a separate method `_get_config()` and re-use it in `_update_config()`?

Wiezzel (approved):
I think it's fine. Just add a few type annotations, extract constants, and fix the linter error.

golem/docker/commands/docker_for_mac.py
mfranciszkiewicz: Yes, should probably be less magic.

mfranciszkiewicz:
@Wiezzel @maaktweluit Please review the changes in 55d9ad7.

Few things have changed:
- usage of the `psutil` package instead of a custom `ps` function
- ~reading the running configuration from the process command line
    Docker writes the configuration to a file on exit, so this was bad from the UX point.~
Edit: this was a bug in current implementation.

maaktweluit (approved):
LGTM!

PR #3205, opened 2018-08-21
3067 fill ethsig

### LINKED ISSUE #3067: fill in `TaskToCompute.ethereum_address_proof`
depends on: https://github.com/golemfactory/golem-messages/issues/241

shadeofblue:
closes https://github.com/golemfactory/golem/issues/3067

### DIFF
--- a/golem/network/transport/tcpnetwork.py
+++ b/golem/network/transport/tcpnetwork.py
@@ -465,8 +465,7 @@ class SafeProtocol(ServerProtocol):
             return None
 
         logger.info(
-            'Sending %r, using session %r: %r',
-            msg.__class__, self.session.__class__, self.session)
+            'Sending: %r, using session: %r', msg.__class__, self.session)
         serialized = golem_messages.dump(
             msg,
             self.session.my_private_key,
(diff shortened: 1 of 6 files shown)


golem/network/transport/tcpnetwork.py
```diff
@@ -464,6 +464,8 @@ def _prepare_msg_to_send(self, msg):
             logger.error("Wrong session, not sending message")
             return None
 
+        logger.info(
```
Krigpl: rm? or reduce severity

golem/task/tasksession.py
```diff
@@ -497,6 +497,7 @@ def _react_to_want_to_compute_task(self, msg):
                 price=price,
                 size=task_state.package_size
             )
+            ttc.generate_ethsig(self.my_private_key)
```
Krigpl, line 500: 1. that should use private key from the transaction system
2. and because of that the API should be different, transaction system won't be exposing its private key, instead it can provide a method to sign whatever you need

shadeofblue: @Krigpl okay, added issues that will change that:

* https://github.com/golemfactory/golem/issues/3206
* https://github.com/golemfactory/golem-messages/issues/252

please add one that will add the appropriate call to the ETS API and link it as a dependency, thank you :)

Krigpl (approved):
solid_eng

PR #3214, opened 2018-08-22
+ add an even more baseline integration test with concent disabled for both nodes

shadeofblue:
[ please merge it after approval ]


scripts/concent_integration_tests/params.py
```diff
@@ -43,3 +43,9 @@ def get_datadir(role: str):
 REQUESTOR_ARGS_DEBUG = REQUESTOR_ARGS + [
     '--log-level', 'DEBUG',
 ]
+
+REQUESTOR_ARGS_NO_CONCENT = list(REQUESTOR_ARGS_DEBUG)
+REQUESTOR_ARGS_NO_CONCENT[1] = 'disabled'
```
ktomala, line 48: I don't like this. This is prone to errors when ordering will change. Would suggest doing ordered dict and then creating a list out of it.

scripts/concent_integration_tests/params.py
```diff
@@ -43,3 +43,9 @@ def get_datadir(role: str):
...
+
+PROVIDER_ARGS_NO_CONCENT = list(PROVIDER_ARGS_DEBUG)
+PROVIDER_ARGS_NO_CONCENT[1] = 'disabled'
```
ktomala, line 51: Same here

scripts/concent_integration_tests/tests/test_concent_disabled.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class RegularRun(NodeTestPlaybook):
+    provider_node_script = 'provider/no_concent'
+    requestor_node_script = 'requestor/no_concent'
+
+
+playbook = RegularRun.start()
+if playbook.exit_code:
+    print("exit code", playbook.exit_code)
```
ktomala, line 14: Should change to logging I think.

ktomala (approved):

scripts/concent_integration_tests/params.py
shadeofblue: it doesn't need to be an _ordered_ dict, actually ... but yeah, a good idea!

PR #3216, opened 2018-08-23
remove working dir

Krigpl:
It's useless.


requirements.txt
```diff
@@ -30,7 +30,7 @@ eth-tester==0.1.0-beta.24
 eth-utils==1.0.3
 ethereum==1.6.1
 Golem-Messages==2.12.2
-Golem-Verificator==1.3.0
+-e git://github.com/golemfactory/golem-verificator.git@84b71eab899b7831786e642ca407a04138aca802#egg=golem_verificator
```
maaktweluit: is this update related to the cleanup? else a separate PR would be nice

Should be released before merge

golem/task/taskthread.py
```diff
@@ -122,9 +120,6 @@ def __do_work(self):
         abs_res_path = os.path.abspath(os.path.normpath(self.res_path))
         abs_tmp_path = os.path.abspath(os.path.normpath(self.tmp_path))
 
-        self.prev_working_directory = os.getcwd()
-        os.chdir(os.path.join(abs_res_path,
```
maaktweluit, line 126: where will taskthread work now there is no `chdir`?

maaktweluit (changes requested):
LGTM!

2 small comments, will approve on reply or resolve :)

requirements.txt
maaktweluit: should have checked the linked PR, is related :)

golem/task/taskthread.py
Krigpl: Hmmmmmm, good catch. Maybe it wasn't completely useless. Although I'm pretty sure working directory here could be simply set to `/tmp/<whatever>`, right?

maaktweluit: haha, almost useless still :P

yes, that would work

i also see a tmp_path variable available, maybe that is a good working dir to `chdir` to?

Elfoniok: /tmp/ ?  I think its always good to remove chdir (like who use that?!, what if separate threads are saving files and someone just change its working dir?). I am pretty sure that this is working dir for local test, during adding task (but better confirm this) so datadir/network_name/tmp dir ?

Elfoniok (approved):

golem/task/taskthread.py
Elfoniok: Hey i checked this, so this is base class and its never used (i tried as provider and requestor), derived class which is DockerTaskThread is used all the time and currently i don't see when one could want to use base implementation.

PR #3220, opened 2018-08-23
WIP: market stub with fifo queue

maaktweluit:
**Do not merge yet please!!! the PR should be to develop first**

Last step after the big cleanup to add the stub and the queue.

### DIFF
--- a/golem/market/queue.py
+++ b/golem/market/queue.py
@@ -1,20 +1,38 @@
-class MarketQueue:
-    class __MarketQueue:
-        def __init__():
-            self.queue = queue.Queue()
+import logging
+import queue
 
-    instance = None
+logger = logging.getLogger(__name__)
+
+class MarketQueue:
+    """
+        The MarketQueue buffers WantToCompute messages and passes them to the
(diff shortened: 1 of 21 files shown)


golem/market/queue.py
```diff
@@ -0,0 +1,38 @@
...
+import queue
+
+logger = logging.getLogger(__name__)
+
+class MarketQueue:
+    """
+        The MarketQueue buffers WantToCompute messages and passes them to the
+        market algorithm on each tick, returning selected messages to reply to.
+    """
+    __instance = None
```
mfranciszkiewicz, line 11: The market algorithm should not understand what `WantToCompute` is, only take a set of input parameters. We need an additional layer which translates this kind of message into a parameter set.

Plus, _IMO_ we should be able to have multiple queues, this way it will be easier to manage timers and separate offers for multiple task ids.

What we also need:
- persistence
- scheduled queue sweeping / explicit removal from the queue when a task is deleted, restarted (basically deleted and created anew) or finished

maaktweluit:
Replaced by #3241

PR #3219, opened 2018-08-23
Separate DockerManager and Hypervisor classes (base: #3203)

mfranciszkiewicz:
Follow up on #3203

PR #3223, opened 2018-08-23
Refactor golem.docker.manager module (base: #3219)

mfranciszkiewicz:
- move CommandHandler classes to their own modules
- move Hypervisor classes to their own modules
- separate unit tests


golem/docker/config.py
```diff
@@ -1,11 +1,19 @@
...
+from golem.core.common import get_golem_path
 from golem.core.hardware import cpu_cores_available
 from golem.docker.task_thread import DockerTaskThread
 
-__all__ = ['DockerConfigManager']
 logger = logging.getLogger(__name__)
 
+ROOT_DIR = get_golem_path()
+APPS_DIR = os.path.join(ROOT_DIR, 'apps')
+IMAGES_INI = os.path.join(APPS_DIR, 'images.ini')
```
Wiezzel, line 13: I would encourage using `Path` class:
```python
ROOT_DIR = Path(get_golem_path())
APPS_DIR = ROOT_DIR / 'apps'
IMAGES_INI = APPS_DIR / 'images.ini'
```

golem/docker/config.py
```diff
@@ -23,6 +31,25 @@
               'sys_resource', 'sys_time', 'sys_tty_config']
 )
 
+CONSTRAINT_KEYS = dict(
+    mem='memory_size',
+    cpu='cpu_count',
+    cpu_cap='cpu_execution_cap'
+)
```
Wiezzel, line 38: What is the actual benefit of having this `CONSTRAINT_KEYS` dict? The programmer still has to remember `'mem'`, `'cpu'` and `'cpu_cap'` keys. Maybe using an enum would be a better idea?

Wiezzel (approved):
If this is only moving thing around I guess I'll just approve.

golem/docker/config.py
mfranciszkiewicz: The point was that `mem`, `cpu` and `cpu_cap` are guaranteed to be constant.

With Enum we would have to use something like `EnumName.property.name` or have enum values as keys.

golem/docker/config.py
mfranciszkiewicz: I'm not against `Path`, but to me it seems to be an overkill just to have a nicer dir join syntax.

golem/docker/config.py
Wiezzel: Aren't `memory_size`, `cpu_count`, and `cpu_execution_cap` guaranteed to be constant?

Wiezzel:
`int()` could also raise `ValueError`.

Issue #3075, opened 2018-07-23
Task create RPC call -> add "Concent enabled" parameter


shadeofblue:
There needs to be `concent_enabled` flag in the task header: https://github.com/golemfactory/golem/issues/3122

This flag should be set when creating task based on `concent_service.enabled` and flag in task description in the RPC call.

PR #3227, opened 2018-08-27
add concent_enabled flag to task definition

### LINKED ISSUE #3122: [Concent] Add `concent_enabled` flag to the task header
`TaskFixedHeader` dictionary needs to include the `concent_enabled` flag

etam:
Resolves #3122 #3075

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -565,11 +565,17 @@ class Client(HardwarePresetsMixin):
         _result = Deferred()
 
         # FIXME: Statement only for old DummyTask compatibility #2467
+        task: TaskBase
         if isinstance(task_dict, dict):
             logger.warning('enqueue_new_task called with deprecated dict type')
             task = task_manager.create_task(task_dict)
         else:
             task = task_dict
+
+        if task.header.fixed_header.concent_enabled and \
+                not self.concent_service.enabled:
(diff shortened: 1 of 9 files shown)


golem/client.py
```diff
@@ -565,12 +565,18 @@ def enqueue_new_task(self, task_dict) -> Tuple[Deferred, str]:
...
+        task: TaskBase
         if isinstance(task_dict, dict):
             logger.warning('enqueue_new_task called with deprecated dict type')
             task = task_manager.create_task(task_dict)
         else:
             task = task_dict
 
+        if task.header.fixed_header.concent_enabled and \
+                not self.concent_service.enabled:
+            raise Exception("Cannot create task with concent enabled when "
```
jiivan: Is it possible to use a more specific Exception here?

jiivan (commented):
LGTM, except I would like to see more specific exception when concent is requested but not available.

golem/client.py
etam: Of course it is. Should I use `RuntimeError` or a custom exception class?

jiivan: I vote for a custom exception.

jiivan (approved):

PR #3228, opened 2018-08-27
Verificator improvements

Elfoniok:


apps/blender/verification_task.py
```diff
@@ -0,0 +1,26 @@
+from typing import Optional
+from twisted.internet.defer import Deferred, inlineCallbacks
+from golem.core.common import deadline_to_timeout
+
+class VerificationTask:
```
mfranciszkiewicz: IMO we can move the `VerificationQueue` and `VerificationTask` out of `apps.blender` sometime in the future.

apps/blender/verification_queue.py
```diff
@@ -1,32 +1,24 @@
...
-                                 'deadline', 'kwargs', 'cb'])
-
     def __init__(self, concurrency: int = 1) -> None:
-
         self._concurrency = concurrency
         self._queue: queue.Queue = queue.Queue()
-
-        self._lock = threading.Lock()
         self._jobs: Dict[str, Deferred] = dict()
+        self.callbacks: Dict[VerificationTask, FunctionType] = dict()
```
mfranciszkiewicz, line 21: Any thoughts on making the queue persistable?

apps/blender/verification_queue.py
```diff
@@ -1,32 +1,24 @@
...
 
-    Entry = namedtuple('Entry', ['verifier_class', 'subtask_id',
-                                 'deadline', 'kwargs', 'cb'])
-
     def __init__(self, concurrency: int = 1) -> None:
-
         self._concurrency = concurrency
         self._queue: queue.Queue = queue.Queue()
-
-        self._lock = threading.Lock()
```
mfranciszkiewicz, line 26: Are you sure that race conditions won't occur in `can_run` / `run`?

apps/blender/blender_reference_generator.py
```diff
@@ -259,40 +258,75 @@ def render_crops(self,
...
+        defer.addErrback(crop_failure)
 
-        # FIXME issue #1955
         verification_context.computer.start_computation(
             root_path=verification_context.get_crop_path(crop_number),
-            success_callback=partial(crop_rendered,
-                                     verification_context=verification_context,
-                                     crop_number=crop_number),
-            error_callback=crop_render_failure,
+            success_callback=success,
```
mfranciszkiewicz, line 310: It would be helpful for `verification_context` to return a `Deferred` object, so we can decrease the number of callbacks we pass around. `start` could simply raise an exception.

apps/blender/verification_queue.py
Elfoniok: This is handled in one (main) thread only.

apps/blender/verification_task.py
Elfoniok: Ok but lets leave it for later, for example during actual persistence implementation.

apps/blender/blender_reference_generator.py
Elfoniok: Ok, this is done, verification context will handle deferred list, and start will return it, so the caller code can handle this.

apps/blender/verification_queue.py
Elfoniok: Yes, i dont think it is possible right now without requestor able to connect back to provider. Right now all verification is during provider connection session. When we close golem we will lose this connection. The idea is that we serialize all verification tasks, not the queue itself. On golem start we deserialize verification tasks, recreate the queue based on them and establish connection back with provider. This should not be part of this PR.

apps/blender/blender_reference_generator.py
```diff
@@ -243,56 +240,82 @@ def render_crops(self,
...
+            error_callback=failure,
             compute_task_def=task_definition,
             resources=verification_context.resources,
             additional_resources=[]
         )
 
+        return defer
```
jiivan, line 317: Would it be possible/feasible to decorate `schedule_crop_job()` with `inlineCallbacks` and `results, time_spent = yield verification_context.computer.start_computation()` inside?

requirements.txt
```diff
@@ -29,8 +29,8 @@ eth-keys==0.2.0b3
 eth-tester==0.1.0-beta.24
 eth-utils==1.0.3
 ethereum==1.6.1
+-e git://github.com/golemfactory/golem-verificator.git@task_interface#egg=golem_verificator
```
jiivan: `#FIXME cleanup`

requirements_to-freeze.txt
```diff
@@ -49,7 +49,8 @@ py-cpuinfo==3.3.0
 humanize==0.5.1
 raven
 Golem-Messages==2.12.3
-Golem-Verificator==1.3.1
+#Golem-Verificator==1.3.1
+-e git://github.com/golemfactory/golem-verificator.git@task_interface#egg=golem_verificator
```
jiivan: `#FIXME cleanup`

jiivan (approved):
LGTM, some suggestions

apps/blender/blender_reference_generator.py
Elfoniok: It would be possible, but this require a lot o changes since computer implementation need to be changed. It is not justified for now.

Issue #3234, opened 2018-08-27
StatisticsError: no median for empty data


ederenn:
#### Encountered on: 

Windows, GOLEM Version: 0.17.0+dev90.g823fdfe
during creating the task - funds for tasks were locked, but task was not created in task list

in `golem.log`

```
2018-08-27 17:35:28 INFO     golem.client                        Enqueue new task "'d31ef8c8-aa0e-11e8-bd4b-5122e4d1d697'"
2018-08-27 17:35:28 CRITICAL twisted                             Unhandled error in Deferred:
2018-08-27 17:35:28 CRITICAL twisted                             
Traceback (most recent call last):
  File "C:\BuildbotWorker\buildpackage_windows\build\.venv\lib\site-packages\twisted\internet\defer.py", line 653, in _runCallbacks
  File "C:\BuildbotWorker\buildpackage_windows\build\golem\client.py", line 596, in package_created
  File "C:\BuildbotWorker\buildpackage_windows\build\golem\client.py", line 652, in _get_mask_for_task
  File "C:\BuildbotWorker\buildpackage_windows\build\golem\network\p2p\p2pservice.py", line 282, in get_estimated_network_size
  File "C:\BuildbotWorker\buildpackage_windows\build\golem\network\p2p\peerkeeper.py", line 226, in get_estimated_network_size
  File "C:\BuildbotWorker\buildpackage_windows\build\golem\network\p2p\peerkeeper.py", line 207, in filter_outliers
  File "C:\Python36\lib\statistics.py", line 380, in median
statistics.StatisticsError: no median for empty data
```

etam:
This should be automatically closed by github after merging the #3235 PR. Weird.

PR #3235, opened 2018-08-27
Handle buckets with more peers than k

jiivan:
fixes: #3234


Wiezzel (approved):
Very well. 👍 

However the title is a little bit confusing for two reasons:
  * Buckets cannot have more than _k_ peers, they can have **exactly** _k_ peers.
  * It's not about "handling buckets with _k_ peers" but rather "handling a situation when **all** buckets have exactly 0 or _k_ peers"

PR #3238, opened 2018-08-28
Bump pycryptodome version

badb:
There is a vulnerabilit in 3.5.1 (https://nvd.nist.gov/vuln/detail/CVE-2018-15560). It shouldn't affect us, but it's better to keep libraries up to date (just in case).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -69,7 +69,7 @@ py-ubjson==0.11.0
 pyasn1==0.4.1
 pyasn1-modules==0.2.1
 pycparser==2.17
-pycryptodome==3.5.1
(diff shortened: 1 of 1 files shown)


Krigpl (approved):
verynicegreatsuccess

etam (changes requested):
what about `requirements_to-freeze.txt`?

etam (approved):

etam:
Actually, `requirements_to-freeze.txt` should contain only things, that are explicitly imported by golem. This is a requirement of requirement, so it's ok, that it's only in `requirements.txt`.

PR #3241, opened 2018-08-28
Marketplace: offer pool

mfranciszkiewicz:


golem/core/common.py
```diff
@@ -49,6 +49,12 @@ def is_linux():
     return sys.platform.startswith('linux')
 
 
+def each(func, *iterables):
```
maaktweluit, line 53: nit: its more like `call_each_each()` but not sure if that makes it more clear ;)

golem/market/offer_pool.py
```diff
@@ -0,0 +1,53 @@
...
+    def add(self, key: str, *items) -> None:
+        pool = self._pool(key)
+        each(pool.append, items)
+
+    def get(self, key: str) -> List[Any]:
+        if key in self:
+            return self._pools[key][:]
+        return []
+
+    def drain(self, key: str) -> List[Any]:
```
maaktweluit: think we will also need something like drain_x, getting x items from the pool

Can be added later when it will be used

maaktweluit (approved):
LGTM!

2 nit comments, still approved

golem/core/common.py
mfranciszkiewicz: I think it's just as clear as `map` but here we don't care about the results :) Could be renamed to `for_each` though.

golem/market/offer_pool.py
```diff
@@ -0,0 +1,53 @@
...
+from golem.core.common import each
+
+logger = logging.getLogger(__name__)
+
+
+class OfferPool:
+
+    __instance = None
+
+    def __new__(cls):
```
Krigpl: If it's meant to be a singleton then in this particular case we don't even need to create an instance at all and have all the methods static.

golem/market/offer_pool.py
```diff
@@ -0,0 +1,53 @@
...
+            cls.__instance = super().__new__(cls)
+        return cls.__instance
+
+    def __init__(self):
+        from twisted.internet import reactor
+
+        self._reactor = reactor
+        self._pools: Dict[str, List[Any]] = dict()
+
+    def __contains__(self, key: str) -> bool:
```
Krigpl: I dunno if it's beneficial to have, it's not going to be used externally (or is it?) but adds an indirection when reading the code.

golem/market/offer_pool.py
```diff
@@ -0,0 +1,53 @@
...
+        return key in self._pools
+
+    def add(self, key: str, *items) -> None:
```
Krigpl: When do you expect to call this with more than one item?

golem/market/offer_pool.py
mfranciszkiewicz: That's the thing - I also initially thought we would need something like a `get_n` function.

There are 2 approaches here:
- count based
    Wait until we get `n` offers. `n` could be bigger than the number of subtasks, but this may slow down the job assignment process.
- time based (this one)
    Every timeout we take all the offers. If `n` is greater than the number of subtasks, we choose the "best" ones. There is no guarantee that we will have a pool `>=` number of subtasks.

But maybe you're right and this structure should handle both scenarios.

golem/market/offer_pool.py
```diff
@@ -0,0 +1,53 @@
...
+        return []
+
+    def drain(self, key: str) -> List[Any]:
+        elements: List[Any] = self.get(key)
+        if key in self:
+            del self._pools[key]
+        return elements
+
+    def drain_after(self, key: str, timeout: float) -> Deferred:
+        if key not in self._pools:
```
Krigpl: Inconsistent with others `key not in self`

Krigpl (commented):
I find it hard to review without a real use case example. Especially the `drain` methods. I have a feeling that this class doesn't have to depend on `reactor` but real case would help to understand.

golem/core/common.py
maaktweluit: haha, indeed.. i'm fine keeping it as is :)

golem/market/offer_pool.py
maaktweluit: when the network is big enough to fill the pool within the timeout a `get_n` would be nice

for now we can leave it time based and add it when we need it.

golem/market/offer_pool.py
mfranciszkiewicz: E.g. to re-add items that were taken out of the pool.

golem/market/offer_pool.py
mfranciszkiewicz: That's true, we can have only class methods and import the reactor lazily.

golem/market/offer_pool.py
mfranciszkiewicz: `_pools` are private-ish, so this could be a common method instead. This indirection doesn't seem to be unclear though.

mfranciszkiewicz:
@Krigpl Please refer to https://github.com/golemfactory/golem/pull/3241#discussion_r213968763 . If this class didn't depend on timer management (`reactor`) it could simply be a dict of str: list.

I've removed the singleton logic and added a `get_n`-like method. Please re-review @maaktweluit.

golem/market/offer_pool.py
```diff
@@ -0,0 +1,102 @@
...
+            del cls._pools[key]
+        return elements
+
+    @classmethod
+    def drain_after(cls, key: str, timeout: Union[int, float]) -> Deferred:
+        from twisted.internet import reactor
+        return task.deferLater(reactor, timeout, cls.drain, key)
+
+    @classmethod
+    def take_when(cls,
```
Krigpl: isn't it `take_after` ?

Krigpl: Also what is `count`? How is it gonna be used/chosen? I thought we will just pass all the offers to the algorithm?

golem/market/offer_pool.py
```diff
@@ -0,0 +1,102 @@
...
+            return 0
+        return len(cls._pools[key])
+
+    @classmethod
+    def add(cls, key: str, *items) -> None:
+        pool = cls._pool(key)
+        each(pool.append, items)
+
+    @classmethod
+    def peek(cls, key: str, count: int = 0) -> PoolType:
```
Krigpl: what is this for?

golem/market/offer_pool.py
Krigpl: Nvm, I think I get it now.

golem/market/offer_pool.py
mfranciszkiewicz: It's a simple `get` but renamed to make it clear that it does not modify the pool.

Krigpl: yeah, but what's the expected use case?

mfranciszkiewicz: What's the use case for `get`? I don't understand the intent behind your question.

It's a convenience method and part of `drain`. Since it is safe to use and doesn't modify the pools, it became a part of the public interface.

mfranciszkiewicz: What's the use case for `get`? I don't get the intent behind your question.

It's a convenience method and a part of `drain`. Since it is safe to use and doesn't modify the pools, it became a part of the public interface.

Krigpl: > Since it is safe to use and doesn't modify the pools, it became a part of the public interface.

That's not really the reason why something should be a part of the public interface - and that was my question: when this method is expected to be used externally?
Because as I mentioned I find this code hard to review without knowing/seeing the requirements.

mfranciszkiewicz: Soft requirements were provided [here](https://github.com/golemfactory/golem-model/pull/14/files#diff-499ff97698fe2960557a736bf3c70197R53) and [here](https://github.com/golemfactory/golem-model/pull/14#discussion_r211654862). Implementation details are left to the marketplace developer(s), because the initial assumptions (i.e. constant pooling time) may prove impractical.

This method is a proper way to resolve `drain` + `add` situations. There is no specific use-case scenario behind it and I strongly disagree that we need one to make a method public. Refactor as you please.

Krigpl: I strongly believe there should be a clear use case in order to introduce a public method. Each public method adds a maintenance cost - as opposed to private method which you can modify as you wish at any time without any additional penalty.

> Refactor as you please.

I doubt this is a proper response. It will be way more beneficial and less wasteful if we all agree on the same conventions that we're all gonna follow instead of changing the code to your liking any time you touch it.

mfranciszkiewicz: > It will be way more beneficial and less wasteful if we all agree on the same conventions

I don't think this questionnaire is a productive and healthy way to propose changes.

> Each public method adds a maintenance cost 

We're discussing a blatantly common `get` method for a pretty generic collection wrapper. If you are thinking about __other__ methods and this is just an example, please speak your mind clearly. If not, I don't think we're producing any value here.

> I doubt this is a proper response

This PR is not the place to agree on conventions, also with others from @golemfactory/clay not participating. I'm leaving for vacation for the next days and won't be a part of the decision process. If you believe that "refactor as you please" is not appropriate, then please do not introduce changes to this PR until I'm back.

PR #3244, opened 2018-08-28
Use posix path in json files

badb:
Problem: normal representation of windows path like "C:\Something" was badly interpreted by json serializer as an escape sentence, as a result Docker tests that used those json task definition were failing on Windows. These tests were skipped in automatic tests. 

Solution: Use posix-like representation of windows paths.

### DIFF
--- a/tests/golem/docker/test_docker_task.py
+++ b/tests/golem/docker/test_docker_task.py
@@ -77,9 +77,8 @@ class DockerTaskTestCase(
         task_path = Path(__file__).parent / cls.TASK_FILE
         with open(task_path) as f:
             golem_path = get_golem_path()
-            if is_windows():
-                golem_path = golem_path.replace('\\', '\\\\')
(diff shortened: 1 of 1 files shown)


tests/golem/docker/test_docker_task.py
```diff
@@ -61,7 +61,8 @@ def tearDown(self) -> None:
     def _get_test_task_definition(cls) -> TaskDefinition:
         task_path = Path(__file__).parent / cls.TASK_FILE
         with open(task_path) as f:
```
etam: note: with `Path` object you can use `task_path.read_text()` instead of
```python
with open(task_path) as f:
    f.read()
```

etam (approved):
ok

Wiezzel (commented):
This will probably do as a temporal solution but in general making some hacky substitutions on serialized objects is just asking for troubles.
Is there some reason why it cannot be done after deserialization?

PR #3247, opened 2018-08-29
Wait for requestors deposit confirmation

### LINKED ISSUE #3212: Provider responds with `CannotComputeTask` reason: `InsufficientDeposit` after receiving `TaskToCompute`
affects only the latest `develop`

may be related to: https://github.com/golemfactory/golem/issues/3210

jiivan:
Requestor didn't wait for deposit confirmation/mining in blockchain. This bug surfaced when providers check for concent deposit was implementied.
fixes: #3212

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -3,7 +3,6 @@
 import collections
 import json
 import logging
-import random
 import sys
 import time
 import uuid
@@ -64,7 +63,6 @@ from golem.network.p2p.p2pservice import P2PService
 from golem.network.p2p.peersession import PeerSessionInfo
 from golem.network.transport.tcpnetwork import SocketAddress
 from golem.network.upnp.mapper import PortMapperManager
-from golem.ranking.helper.trust import Trust
 from golem.ranking.ranking import Ranking
 from golem.report import Component, Stage, StatusPublisher, report_calls
 from golem.resource.base.resourceserver import BaseResourceServer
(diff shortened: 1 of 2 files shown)


Krigpl:
That's gonna impact the processing time of the `create task` request which will need to include time for confirming the transaction. And that can be long. Iirc @jamuszyn was working on some mitigation for this from the UX POV. I dunno if we wanna merge this change just yet.

jamuszyn:
We should take into consideration two scenarios during task creation. 

First is when gas prices are low, and the confirmation time of deposit should not take more than 5 minutes. In this case, a task is being created with "Creating the deposit" status: https://invis.io/UGNICBIREA5#/314104068_Task_Settings_Main___Concents_Copy_10
The time of deposit confirmation should not be taken as a part of a task timeout that was set by a user.

Second is when gas prices are high, and the confirmation time of the deposit will be higher than 5 minutes. In this case, we strongly recommend creating this task without Concent Service: https://invis.io/Q4NICAKXEJM#/314103867_Task_Settings_Main___Concents_Copy_8

jiivan:
I've created issues for scenarios mentioned by @jamuszyn 
https://github.com/golemfactory/golem/issues/3279
https://github.com/golemfactory/golem-electron/issues/430

shadeofblue (approved):
looks good to me

PR #3259, opened 2018-08-30
Deadline already passed

### LINKED ISSUE #3230: Wrong task header received: Deadline already passed
Example:
```
2018-08-27 12:20:32 WARNING  golem.task.taskserver               Wrong task header received
  File "/data/golem/golem_git/golem/task/taskserver.py", line 315, in add_task_header
    TaskHeader.validate(th_dict_repr)
  File "/data/golem/golem_git/golem/task/taskbase.py", line 235, in validate
    return TaskFixedHeader.validate(fixed_header)
  File "/data/golem/golem_git/golem/task/taskbase.py", line 156, in validate
    raise ValueError(msg)
ValueError: Deadline already passed 
 task_id = cfc3db7a-a887-11e8-b953-65684fed3648 
 node name = [...]
```

What's worth noting: the task was running and explicitly deleted long ago (about 16 hours). For some reason the header is still propagated.

jiivan:
fixes: #3230

1. Harden TaskServer.sync_network() to ensure that all jobs are attempted even when previous one fails
1. Use custom exception for TaskHeader validation errors and don't include their tracebacks in logs

### DIFF
--- a/golem/task/taskbase.py
+++ b/golem/task/taskbase.py
@@ -13,6 +13,8 @@ from golem.network.p2p.node import Node
 from golem.task.masking import Mask
 from golem.task.taskstate import TaskState
 
+from . import exceptions
+
 logger = logging.getLogger("golem.task")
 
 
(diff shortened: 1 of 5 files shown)


jiivan:
It's an open question whether to additionally filter out timeouted headers in `PeerSession._react_to_get_tasks()`.

golem/task/taskbase.py
```diff
@@ -145,58 +147,83 @@ def validate(th_dict_repr: dict) -> None:
            defined parameters
          :param dict th_dict_repr: task header dictionary representation
         """
-        if not isinstance(th_dict_repr.get('task_id'), str):
-            raise ValueError('Task ID missing')
+        task_id = th_dict_repr.get('task_id')
+        task_owner = th_dict_repr.get('task_owner')
+
+        try:
+            node_name = task_owner['node_name']  # type: ignore
```
etam: This is wrong. Here you try to get something from `task_owner` and below you check if it's a `dict`. You should first make sure it's a `dict` and then get anything from it. It will also make mypy less angry.

golem/task/taskbase.py
```diff
@@ -145,58 +147,83 @@ def validate(th_dict_repr: dict) -> None:
...
-                  "node name = %s " % \
-                  (th_dict_repr['task_id'],
-                   th_dict_repr['task_owner']['node_name'])
-            raise ValueError(msg)
+            raise exceptions.TaskHeaderError(
+                "Deadline already passed",
+                task_id=task_id,
+                node_name=node_name,
+                deadline=th_dict_repr['deadline'],
+                now=common.get_timestamp_utc(),
```
etam: Get a timestamp once, assign it to a variable and use it in both places.

etam (changes requested):

etam:
> It's an open question whether to additionally filter out timeouted headers in `PeerSession._react_to_get_tasks()`.

We don't want to propagate outdated tasks, so they have to be dropped somewhere. This is a good place to filter the incoming headers, so I'm OK with that.

golem/task/taskbase.py
jiivan: It's just [EAFP](https://docs.python.org/3.6/glossary.html#term-eafp). What's wrong with that?

etam: Me being used to C++.

But on the other hand, you can move the line `node_name = task_owner['node_name']` under `if not isinstance(task_owner, dict):` and you won't need to catch `TypeError` nor silence the type checker. Then you could replace the whole try...catch thing with `node_name = task_owner.get('task_owner', task_owner)`. Wait, what?

etam: I don't understand this part:

```python
        except (TypeError, KeyError):
            node_name = task_owner
```
What sense does it make, if we expect `task_owner` to be a `dict`?

jiivan: We expect `task_owner` to be a `dict` and act accordingly, but when it raises `TypeError` it means it's something else and I would like to know what it was i.e. include it in `TaskHeaderError`.

I wanted to have all exceptions raised here to contain at least `task_id` and `node_name`. Could use something generic like `<unknown>` but thought that some actual/related value would be better.

golem/task/taskbase.py
```diff
@@ -147,58 +149,81 @@ def validate(th_dict_repr: dict) -> None:
            defined parameters
          :param dict th_dict_repr: task header dictionary representation
         """
-        if not isinstance(th_dict_repr.get('task_id'), str):
-            raise ValueError('Task ID missing')
+        task_id = th_dict_repr.get('task_id')
+        task_owner = th_dict_repr.get('task_owner')
+
+        try:
+            node_name = task_owner['node_name']  # type: ignore
```
etam: I'd change that to:
```python
try:
    node_name = task_owner.get('node_name')
except TypeError:
    ...
```
This way, if there's no `node_name` in `task_owner` it'll be handled by `if not isinstance(node_name, str):` below. (because in such case, error message 'Task owner missing' is misleading)

etam (approved):

PR #3262, opened 2018-08-31
ignore patch number while determining the version compatibility of go…

shadeofblue:
…lem nodes...


golem/task/taskkeeper.py
```diff
@@ -424,7 +424,7 @@ def check_version_compatibility(self, remote):
         local = Version(self.app_version, partial=True)
         if local.major != remote.major or local.minor != remote.minor:
             return False
-        return local.patch >= remote.patch
+        return True
```
badb: What about: `return local.major == remote.major and local.minor == remote.minor`?

badb (approved):

PR #3266, opened 2018-09-03
lock additional funds when restarting subtasks

### LINKED ISSUE #3265: restart of a frame in finished task is not working properly
#### Steps to reproduce

- start a multiframe task, at least two frames, wait for task to finish
- go to second preview window (click detailed view on preview image), to frames view 
- click resubmit on chosen frame
- on task view progress bar will jump back, and subtasks within frame will change status from finished to restarted
- task will end in timeout - no nodes are starting computations 

screen shot 2018-08-31 at 19 51 19

screen shot 2018-08-31 at 19 51 50

task seems to be restarted, but task status is not changed to restarted, no other node is starting computations for restarted subtasks
only result is - change in progress bar

etam:
Fixes #3265


etam:
Restarting frame, when task is finished, is working, but not quite. Task state is changed to computing, but there are no funds locked in fundslocker, task is removed from p2pservice and many other things are cleaned up when task is finished. It works, because providers that have the task header and are still asking for more work. To make it work properly we need to create a new task, like in `Client.restart_subtasks_from_task`.

etam:
I'll write tests that are reproducing problems before applying those fixes.

shadeofblue (approved):
approving for now, @etam please add a regression test

Issue #3263, opened 2018-08-31
failing provider nodes should be blocked for task session 


ederenn:
if computations fails on specific node, this node should be banned from further computations for that task. 

Bad Wolf has failed on first subtask but is allowed to compute again

```
VM-2-iMac1017:golem a$ golemcli subtasks show 8c2bc874-ad39-11e8-b104-7c3ec369ac11
description: 'path_root: /Users/a/Desktop, start_task: 1, end_task: 1, total_tasks:
  10, outfilebasename: scene-Helicopter-30, scene_file: /golem/resources/scene-Helicopter-27-cycles.blend'
node_id: 9d5cb6e59b287ba73223b5ee35307f074c90615463a11aa37c9ee1056e7e82a4c0de3a90fbf850df4648c68225d071ab2635f1675c70bd71e2b92bb1f3f02138
node_ip_address: 10.30.11.251
node_name: Bad Wolf
node_performance: '240.61350776529488'
node_port: 0
progress: 1.0
results: []
status: Failure
stderr: 'Error downloading resources: [Failure instance: Traceback (failure with no
  frames): <class ''requests.exceptions.HTTPError''>: {"error":"0ca1d8411757a34fd26fb1199c47ad790c5e1b43dea69adf33f94670b46db67a
  download timed out after 234 s"}

  ]'
stdout: ''
subtask_id: 8c2bc874-ad39-11e8-b104-7c3ec369ac11
time_remaining: 0.0
time_started: 1535732331.3517282
```

<img width="494" alt="screen shot 2018-08-31 at 18 36 19" src="https://user-images.githubusercontent.com/32061891/44925688-5848fc80-ad4f-11e8-9986-22f0c617b814.png">

ederenn:
it repeats on any OS
can lead to extreme situations:

```
C:\Users\ederenn>golemcli subtasks show 38d01ae4-ad51-11e8-8db6-649b8e85a011
description: 'path_root: C:\Users\ederenn\Desktop, start_task: 1, end_task: 1, total_tasks:
  1, outfilebasename: scene-Helicopter-27-1, scene_file: /golem/resources/scene-Helicopter-27-cycles.blend'
node_id: 7c3ec369ac1108e578389756216b9e790a133a5ca7dbffcba620f4ff4655ebd0f270f69c9b8acb302744950b35829e7417c7c784bb7a49ae4f50f7449d46f3fc
node_ip_address: 10.30.11.251
node_name: vm sierra 0.17.1
[...]

PR #3270, opened 2018-09-03
fix the issue that allows a provider node to be sent a TTC repeatedly…

shadeofblue:
… despite

having issued a "CannotComputeTask" for this task previously

fixes https://github.com/golemfactory/golem/issues/3263


golem/task/taskserver.py
```diff
@@ -631,6 +631,12 @@ def should_accept_provider(  # noqa pylint: disable=too-many-arguments,too-many-
             logger.info(f'network mask mismatch: {ids}')
             return False
 
+        if task.should_accept_client(node_id) != AcceptClientVerdict.ACCEPTED:
+            logger.info(f'provider {node_id} is not allowed'
```
tworec, line 635: would be great add a reason here! It will highly simplify debugging

scripts/concent_integration_tests/nodes/provider/cannot_compute.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
```
tworec, line 1: I do not know about those integrations tests... can not review it...

Wiezzel (approved):
Fix looks fine.
But I don't know anything about integration tests.

golem/task/taskserver.py
tworec: eg. like here https://github.com/golemfactory/golem/blob/ae70ab85a52c2a62dc12b62dc820a4939ddefd34/golem/task/taskmanager.py#L449-L451

tworec (approved):
looks ok, just style comment.

maaktweluit (approved):

golem/task/taskserver.py
shadeofblue: @tworec added -> though the reason can be twofold

tworec: thats more than enough. thx

PR #3273, opened 2018-09-04
Misc log improvements

maaktweluit:
During my last task: #3189 these fixes did not make it to the PR in time.

Split them in this separate PR:
- Add uniform way to print node_name + node_id
- Add log messages in `tasksession._react_to_wants_to_compute`

### DIFF
--- a/tests/golem/test_client.py
+++ b/tests/golem/test_client.py
@@ -5,6 +5,7 @@ import time
 import uuid
 from random import Random
 from types import MethodType
+from unittest import mock
 from unittest import TestCase
 from unittest.mock import call, Mock, MagicMock, patch
 
@@ -819,6 +820,7 @@ class TestTaskCleanerService(TestWithReactor):
         )
         service._run()
         client.clean_old_tasks.assert_called_once()
+
 
 @patch('golem.client.node_info_str')
 @patch('signal.signal')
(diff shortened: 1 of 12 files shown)


tests/golem/test_client.py
```diff
@@ -820,9 +822,11 @@ def test_run(self):
         client.clean_old_tasks.assert_called_once()
 
 
+@patch('golem.client.node_info_str')
 @patch('signal.signal')
 @patch('golem.network.p2p.node.Node.collect_network_info')
 class TestClientRPCMethods(TestWithDatabase, LogTestCase):
+    # pylint: disable=too-many-public-methods
```
shadeofblue, line 829: it would probably be good to refactor these tests instead...

shadeofblue (approved):

Wiezzel (approved):
I prefer using f-strings over `%` or `{}` formatting, but it's OK.

tests/golem/test_client.py
maaktweluit: Will have a look if i can split it up, if it takes too much time i'll make an issue to fix this later

maaktweluit:
@Wiezzel when using logs `%` is the best way, because it will format only when the level is active.

For this one i'll change to f'' strings, agree this will look neater.

tests/golem/test_client.py
maaktweluit: Looked at it and do not see an easy split, most of them are testing RPC methods..

shall i make a separate issue to look into this or leave as is, there are many more tests that need cleanup like this.

PR #3274, opened 2018-09-04
Add tests for subtask restart

### LINKED ISSUE #3271: add regression check for the "lock additional funds when restarting subtasks" issue
add regression test that checks for https://github.com/golemfactory/golem/pull/3266

etam:
Fixes #3271


tests/golem/test_client.py
```diff
@@ -348,27 +348,73 @@ def test_activate_hw_preset(self, *_):
...
-            transaction_system=Mock(),
+            transaction_system=ts,
             connect_to_known_hosts=False,
             use_docker_manager=False,
             use_monitor=False
         )
+        self.client.funds_locker.persist = False
+
+        task = Mock()
+        task.header.task_id = 'tid'
```
shadeofblue: @etam shouldn't this go into some helper factory method so that we don't need to repeat this sequence of operations in each test?

shadeofblue (changes requested):
please move those parts of the `test_taskmanager` tests to a common method so that code is not repeated in each of those tests...

tests/golem/test_client.py
etam: done

etam:
`test_taskmanager`? I changed a single line there.

shadeofblue:
@etam oops, sorry, I meant `test_client` ... my bad

etam:
I'm not sure if there is that much code repeated that makes it reasonable to pull out a common part. Such function wouldn't have even any meaningful name.

I was thinking about moving creation of `Client` object into `setUp`, but it wouldn't work with some of test cases. Then I found that there's also a `TestClientRPCMethods` class, which is initializing a `Client` object in `setUp`, but it's not clear what's the distinction between it and `TestClient` class.

shadeofblue:
> I'm not sure if there is that much code repeated that makes it reasonable to pull out a common part. Such function wouldn't have even any meaningful name.
> 
> I was thinking about moving creation of `Client` object into `setUp`, but it wouldn't work with some of test cases. Then I found that there's also a `TestClientRPCMethods` class, which is initializing a `Client` object in `setUp`, but it's not clear what's the distinction between it and `TestClient` class.

@etam well, what's stopping us from creating a new `TestClientRestartTask` TestCase? ... generally, imvho, the test cases should be separate unless there is a good reason for them to be part of the same class ... ... so, if we have more than one test that requires a similar setup, that - imvho - constitues a pretty good reason to create a common TestCase class for them...

etam:
done

shadeofblue (approved):

shadeofblue:
@etam approved -> the only issue is that we probably don't want to merge to b0.17.1 specifically...

PR #3278, opened 2018-09-05
RPC support for deposit transactions

### LINKED ISSUE #2980: [Concent] Add RPC support for deposit transactions
to allow the front-end to display the outgoing deposit transactions on the "Outgoing" and "Deposit" list

jiivan:
fixes: #2980 
Quite a few changes were required to add RPC support for deposit transactions:
 1. Create DB model of deposit transaction
 1. Actually store deposit transactions in database
 1. Refactor `concent_deposit` to actually use `inlineCallbacks` decorator
 1. Add `deposit_payments` command to _golemcli_


golem/database/schemas/021_create_deposit_payment.py
```diff
@@ -0,0 +1,29 @@
...
+
+import peewee as pw
+
+SCHEMA_VERSION = 21
+
+
+def migrate(migrator, database, fake=False, **kwargs):
+    @migrator.create_model  # pylint: disable=unused-variable
+    class DepositPayment(pw.Model):
+        dbid = pw.PrimaryKeyField()
```
shadeofblue: @jiivan is this needed? peewee should automatically create a primary key id in case none is specified... shouldn't it?

golem/ethereum/transactionsystem.py
```diff
@@ -279,9 +297,34 @@ def get_payments_list(self):
         """
         return self._payments_keeper.get_list_of_all_payments()
 
+    @classmethod
+    def get_deposit_payments_list(cls):
```
shadeofblue: I wonder -> shouldn't we think ahead and already implement an interface for pagination? iow, add `limit` and `offset` parameters to the call, filled by default with `1000` and `0` respectively, so that we can later implement the pagination on the front-end when we decide so?

I'm mentioning that because there already are concerns about lack of pagination in our RPC calls... so maybe it would be good to stay future-proof when we implement new calls...

golem/model.py
```diff
@@ -234,6 +276,31 @@ def __repr__(self) -> str:
             )
 
 
+class DepositPayment(BaseModel):
+    dbid = PrimaryKeyField()
```
shadeofblue: (same here - or rather - _actually_ here) I believe Peewee should add the identity field automatically...

golem/ethereum/transactionsystem.py
```diff
@@ -279,9 +297,34 @@ def get_payments_list(self):
         """
         return self._payments_keeper.get_list_of_all_payments()
 
+    @classmethod
+    def get_deposit_payments_list(cls):
+        result = []
+        query = model.DepositPayment.select().order_by('dbid').limit(1000)
+        for dpayment in query:
+            entry = {}
+            entry["pk"] = common.to_unicode(dpayment.dbid)
```
maaktweluit: I thought `common.to_unicode()` was only needed for python2, or is this something peewee related?

golem/ethereum/transactionsystem.py
```diff
@@ -6,25 +6,41 @@
...
 from golem.core.service import LoopingCallService
 from golem.ethereum.node import NodeProcess
 from golem.ethereum.paymentprocessor import PaymentProcessor
 from golem.ethereum.exceptions import NotEnoughFunds
 from golem.ethereum.incomeskeeper import IncomesKeeper
 from golem.ethereum.paymentskeeper import PaymentsKeeper
-from golem.model import GenericKeyValue, Payment, Income
 from golem.utils import privkeytoaddr
 
+from . import exceptions
```
maaktweluit, line 40: `NotEnoughFunds` is now imported twice

Maybe it is good to keep the imports consistent to make them easier to manage

maaktweluit (commented):
LGTM!

2 small comments, can approve on reply :)

golem/model.py
```diff
@@ -234,6 +276,31 @@ def __repr__(self) -> str:
...
+    value = HexIntegerField()
+    status = PaymentStatusField(index=True, default=PaymentStatus.awaiting)
+    fee = HexIntegerField(null=True)
+    block_hash = BlockchainHashField(null=True)
```
Krigpl: afaict this isn't used anywhere

golem/ethereum/transactionsystem.py
```diff
@@ -506,35 +549,55 @@ def concent_timelock(self, account_address: Optional[str] = None) -> int:
...
         tx_hash = self._sci.deposit_payment(max_possible_amount)
         log.info(
             "Requested concent deposit of %.6fGNT (tx: %r)",
             max_possible_amount / denoms.ether,
             tx_hash,
         )
+        dpayment = model.DepositPayment.create(
+            status=model.PaymentStatus.sent,
+            value=max_possible_amount,
+            tx=int(tx_hash, 16),
```
Krigpl: Transaction hashes are never conceptually used as integers so I find this counter-intuitive - they are just hex strings.

golem/ethereum/transactionsystem.py
```diff
@@ -506,35 +549,55 @@ def concent_timelock(self, account_address: Optional[str] = None) -> int:
...
+        )
 
-        def transaction_receipt(receipt):
-            if not receipt.status:
-                result.errback(Exception(f"Deposit failed. Receipt: {receipt}"))
-                return
-            result.callback(None)
+        receipt = yield transaction_receipt
+        if not receipt.status:
+            dpayment.status = model.PaymentStatus.awaiting
```
Krigpl: That will never be retried so it's better to just remove it completely imo

Krigpl (changes requested):
I believe I mentioned it before, but would you mind adding me to PRs touching transaction system by default?

golem/ethereum/transactionsystem.py
jiivan: Not really sure. It returns `None` for `None` but `str(value)` otherwise. I just used the same logic as in `get_payments_list()`.

golem/database/schemas/021_create_deposit_payment.py
jiivan: Good point. I didn't know peewee has default id field.

golem/ethereum/transactionsystem.py
jiivan: :+1:

golem/model.py
jiivan: Yes, `block_hash` and `block_number` are not displayed anywhere. Do you think they shouldn't be stored?

golem/ethereum/transactionsystem.py
jiivan: Done (simplified to strings).

golem/model.py
Krigpl: Yeah, no need for them anytime soon. And if they will ever be needed they can be obtained from tx hash anyway so there's no loss here.

golem/ethereum/transactionsystem.py
```diff
@@ -506,35 +554,52 @@ def concent_timelock(self, account_address: Optional[str] = None) -> int:
...
         tx_hash = self._sci.deposit_payment(max_possible_amount)
         log.info(
             "Requested concent deposit of %.6fGNT (tx: %r)",
             max_possible_amount / denoms.ether,
             tx_hash,
         )
+        dpayment = model.DepositPayment.create(
+            status=model.PaymentStatus.sent,
+            value=max_possible_amount,
+            tx=tx_hash,
```
Krigpl, line 570: `tx_hash` is 66 chars long (because of `0x` prefix) while `tx` field has `max_length=64`

golem/model.py
```diff
@@ -77,6 +87,35 @@ def python_value(self, value):
...
+                " not {should} characters".format(
+                    value=value,
+                    has=current_len,
+                    should=self.EXPECTED_LENGTH,
+                ),
+            )
+        return value
+
+
+class BlockchainHashField(FixedLengthHexField):
```
Krigpl: not used

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -322,35 +327,114 @@ def test_subscriptions(self):
...
+                expected=40,
+            )
+
+    def test_concent_deposit_transaction_failed(self):
+        self.sci.get_deposit_value.return_value = 0
+        self.ets._gntb_balance = 20
+        self.ets._eth_balance = denoms.ether
+        self.ets.lock_funds_for_payments(1, 1)
+        tx_hash = \
+            '5e9880b3e9349b609917014690c7a0afcdec6dbbfbef3812b27b60d246ca10ae'
```
Krigpl: sci returns hashes with `0x` prefix

golem/ethereum/transactionsystem.py
```diff
@@ -279,9 +296,37 @@ def get_payments_list(self):
...
+            .order_by('id') \
+            .limit(limit) \
+            .offset(offset)
+        for dpayment in query:
+            entry = {}
+            entry["pk"] = common.to_unicode(dpayment.id)
+            entry["value"] = common.to_unicode(dpayment.value)
+            entry["status"] = common.to_unicode(dpayment.status.name)
+            entry["fee"] = common.to_unicode(dpayment.fee)
+            entry["block_number"] = common.to_unicode(dpayment.block_number)
```
Krigpl: `block_number` doesn't exist anymore

golem/ethereum/transactionsystem.py
```diff
@@ -279,9 +296,37 @@ def get_payments_list(self):
...
+            try:
+                tx_hash = "{:x}".format(dpayment.tx)
```
Krigpl: it's already hex, isn't it ?

golem/ethereum/transactionsystem.py
```diff
@@ -279,9 +296,37 @@ def get_payments_list(self):
...
+    @classmethod
+    def get_deposit_payments_list(cls, limit=1000, offset=0):
+        result = []
+        query = model.DepositPayment.select() \
+            .order_by('id') \
+            .limit(limit) \
+            .offset(offset)
+        for dpayment in query:
+            entry = {}
+            entry["pk"] = common.to_unicode(dpayment.id)
```
Krigpl: What `pk` stands for? Is it primary key? isn't it an internal stuff that doesn't need to be exposed? Transaction hash is also a primary key.

Krigpl (changes requested):

golem/ethereum/transactionsystem.py
```diff
@@ -506,35 +548,52 @@ def concent_timelock(self, account_address: Optional[str] = None) -> int:
             account_address=account_address,
         )
 
-    def concent_deposit(self, required: int, expected: int) -> Deferred:
+    @defer.inlineCallbacks
+    def concent_deposit(self, required: int, expected: int) \
+            -> Generator[defer.Deferred, TransactionReceipt, Optional[str]]:
```
Krigpl, line 549: I may just misunderstand but why the return type isn't simply `Optional[str]`?

golem/ethereum/transactionsystem.py
```diff
@@ -279,9 +296,31 @@ def get_payments_list(self):
...
+            .order_by('id') \
+            .limit(limit) \
+            .offset(offset)
+        for dpayment in query:
+            entry = {}
+            entry["value"] = common.to_unicode(dpayment.value)
+            entry["status"] = common.to_unicode(dpayment.status.name)
+            entry["fee"] = common.to_unicode(dpayment.fee)
+            entry["transaction"] = common.to_unicode(dpayment.tx)
+            entry["created"] = common.datetime_to_timestamp_utc(
```
Krigpl: I don't think `created` and `modified` are used either - I know they exist in `payments_list` but probably shouldn't as well.

golem/ethereum/transactionsystem.py
```diff
@@ -279,9 +296,31 @@ def get_payments_list(self):
...
 
+    @classmethod
+    def get_deposit_payments_list(cls, limit=1000, offset=0):
+        result = []
+        query = model.DepositPayment.select() \
+            .order_by('id') \
+            .limit(limit) \
+            .offset(offset)
+        for dpayment in query:
+            entry = {}
```
Krigpl: I'd make it a structured object instead of arbitrary dict, it'd be easier to deal with. And if `to_unicode` is really needed for whatever reason then it should happen in `client` since it's related to presentation layer which doesn't belong here (same for `payments_list` of course, so it's not a perfect base to follow).

golem/ethereum/transactionsystem.py
jiivan: It's a [generator](https://docs.python.org/3.5/glossary.html#term-generator)/coroutine. [Typing specifications](https://docs.python.org/3.5/library/typing.html?highlight=typing#typing.Generator)

golem/ethereum/transactionsystem.py
jiivan: @Krigpl they'll be used in GUI: https://projects.invisionapp.com/share/ZHISL0F3TX9#/screens/297433337

golem/ethereum/transactionsystem.py
jiivan: IMHO db/peewee model is structured enough. I don't see a reason to create custom mapping class only for "presentation layer"/rpc.

Maybe it's a good idea to move rpc formatting to `client`, but client is overbloated (1.7k lines). I think it deserves a proper refactoring to smaller modules e.g. `client.payment`, `client.account` etc. but it's totally outside of this PR.

I think this PR is big enough not to implement this refactoring, don't you?

Krigpl: I didn't suggest any kind of refactoring here. I'm just saying that `TransactionSystem` should return structured data, just list of `model.DepositPayment` is fine and all this `to_unicode` transformations should happen in the `client`.
In other words - why would transaction system convert all values to strings by default, what's the benefit? IE.g. f you'd need to calculate the sum of all deposits you'd need to convert `value`s back to `int`.

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -322,35 +328,113 @@ def test_subscriptions(self):
...
+            )
+            cb(receipt)
+
+        self.sci.on_transaction_confirmed.side_effect = fail_it
+        with self.assertRaises(exceptions.DepositError):
+            self._call_concent_deposit(
                 required=10,
                 expected=40,
             )
+        deposit_value = 20 - 1
```
Krigpl: `20 - 1` looks random. I believe this is `gnt_balance - gnt_lock` so please use those values instead.

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -322,35 +328,113 @@ def test_subscriptions(self):
...
+            cb(receipt)
+
+        self.sci.on_transaction_confirmed.side_effect = confirm_it
+        db_tx_hash = self._call_concent_deposit(
             required=10,
             expected=40,
         )
-        self.sci.deposit_payment.assert_called_once_with(20 - 1)
+        self.assertEqual(tx_hash, db_tx_hash)
+        deposit_value = 20 - 1
```
Krigpl: ditto

shadeofblue (approved):
lgtm

golem/ethereum/transactionsystem.py
jiivan: I apologize. Must have misunderstood you then.

Krigpl (approved):
Looks good
obama_approves

PR #3280, opened 2018-09-05
 Another possible fix for the restart issue

etam:
Instead of making a special case for marking subtask as failed, do it early in `copy_results`


shadeofblue:
[ note ] the tests show that this fix causes the restarted tasks not to be computed by the network - needs more investigation - or maybe - even some refactoring of the task restart functionality and related RPC calls

shadeofblue (changes requested):
make sure that any changes are tested (with the semi-automated test + ideally using normal golem nodes) before we decide to merge it

etam:
fixed

shadeofblue (approved):

PR #3290, opened 2018-09-07
[Concent] RPC endpoint to provide the current deposit balance

### LINKED ISSUE #2978: [Concent] RPC endpoint to provide the current deposit balance
(possibly: modification of an existing endpoint)

The endpoint(s) should allow the front-end to display:
* current deposit balance
* locked/unlocked status
* lock deadline

jiivan:
fixes: #2978


golem/client.py
```diff
@@ -1130,6 +1132,28 @@ def get_balance(self):
...
+
+        class DepositStatus(msg_datastructures.StringEnum):
+            locked = enum.auto()
+            unlocking = enum.auto()
+            unlocked = enum.auto()
+
+        now = time.time()
+        if timelock == 0:
+            status = DepositStatus.locked
+        elif timelock < now:
```
shadeofblue, line 1166: @jiivan hmm... if both the `timelock` and `now` are expressed in seconds since epoch (and btw - are we sure that `timelock` is an int? asking because `time.time()` is a float)) wouldn't `timelock > now` mean that it's "unlocking" and `timelock < now` (which means that `timelock` is in the past relative to `now`) mean that it's already unlocked? ...

jiivan: :scream: Good catch. Added proper tests and fixed.

shadeofblue (approved):
cat cool

PR #3291, opened 2018-09-10
Verification restart after timeout

Elfoniok:


apps/blender/verification_queue.py
```diff
@@ -3,23 +3,28 @@
...
 class VerificationQueue:
 
+    VERIFICATION_TIMEOUT = 90
+    RESTART_DELAY = 5
+
     def __init__(self, concurrency: int = 1) -> None:
         self._concurrency = concurrency
         self._queue: queue.Queue = queue.Queue()
         self._jobs: Dict[str, Deferred] = dict()
+        self.tasks: Dict[VerificationTask, Deferred] = dict()
```
mfranciszkiewicz: `self.tasks` seems redundant - it's only used for writing.

apps/blender/verification_queue.py
```diff
@@ -3,23 +3,28 @@
...
         self.callbacks: Dict[VerificationTask, FunctionType] = dict()
         self._paused = False
+        self.timed_out = False
```
mfranciszkiewicz: Maybe `self._timed_out`?

apps/blender/verification_queue.py
```diff
@@ -79,12 +84,37 @@ def callback(*args):
...
+            fn_timeout = partial(self.stop_task, task=entry, event=result,
+                                 subtask_id=subtask_id,
+                                 verifier_cls=verifier_cls)
+
+            result.addTimeout(VerificationQueue.VERIFICATION_TIMEOUT, reactor,
+                              onTimeoutCancel=fn_timeout)
             self._jobs[subtask_id] = result
+            self.tasks[entry] = result
+
+    def stop_task(self, *_, **kwargs):
```
mfranciszkiewicz: `stop_task` is not the same as "timed out". I think it's worth renaming and hiding (e.g. as `_verification_timed_out`). 

Plus, maybe you can use named keyword arguments (`task`, `event`, etc.) for readability?

apps/blender/verification_queue.py
```diff
@@ -79,12 +84,37 @@ def callback(*args):
                 self._jobs.pop(subtask_id, None)
                 self._process_queue()
 
+        def errback(_):
+            if self.timed_out:
+                logger.debug("Timeout detected")
```
mfranciszkiewicz: A warning / error perhaps? E.g. "Verification timed out for subtask %r"

apps/blender/verification_queue.py
```diff
@@ -79,12 +84,37 @@ def callback(*args):
...
+                reactor.callLater(VerificationQueue.RESTART_DELAY,
+                                  self._process_queue)
+                self.timed_out = False
+            else:
+                logger.debug("Finishing verification with fail")
```
mfranciszkiewicz: I'm not sure if the message will be duplicated, but this one could use a higher log severity.

apps/blender/verification_queue.py
Elfoniok: done

apps/blender/verification_queue.py
Elfoniok: done

apps/blender/verification_queue.py
Elfoniok: done

apps/blender/verification_queue.py
Elfoniok: done

apps/blender/verification_queue.py
Elfoniok: done

apps/blender/verification_task.py
```diff
@@ -1,30 +1,34 @@
...
         else:
-            verifier.task_timeout(self.subtask_id)
+            self.verifier.task_timeout(self.subtask_id)
         return None
 
+    def get_results(self):
+        return self.verifier.verification_completed()
+
     # Currently golem protocol does not allow for partial verification,
     # therefore we have to wait for any ongoing verification.
```
jiivan, line 31: Is this comment still valid?

jiivan (approved):
LGTM

apps/blender/verification_task.py
Elfoniok: That is a good question, currently, verification task can be restarted after canceling, so this means that is not entirely valid. But I will remove it when proper verification queue persistence is implemented.

PR #3294, opened 2018-09-11
Don't return None during web3 requests

Krigpl:
Slightly different retry logic again.

### DIFF
--- a/golem/ethereum/web3/middleware.py
+++ b/golem/ethereum/web3/middleware.py
@@ -1,6 +1,5 @@
 import logging
 import socket
-from typing import ClassVar
 
 from types import MethodType
 
@@ -8,7 +7,6 @@ from web3.exceptions import CannotHandleRequest
 
 logger = logging.getLogger(__name__)
 
-MAX_ERRORS = 20
 RETRIES = 3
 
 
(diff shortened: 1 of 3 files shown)


golem/ethereum/web3/middleware.py
```diff
@@ -47,12 +42,9 @@ def middleware(method, params):
                         exc,
                     )
                     self._cur_errors += 1
-                    if self._cur_errors >= self._max_errors:
-                        self.reset()
-                        raise
                     if self._cur_errors % self._retries == 0:
                         self._err_listener()
-                        return
+                        raise
```
mfranciszkiewicz, line 46: Hm, if we reach `self._retries` and `raise` here during startup, Golem will shut down. I believe some previous change was addressing that issue. 

We could return a `{"error": str(exc)}` dict here instead of `None`.

golem/ethereum/web3/middleware.py
```diff
@@ -18,19 +16,16 @@ class RemoteRPCErrorMiddlewareBuilder:
...
-    _cur_errors: ClassVar[int] = 0
-
     def __init__(self,
                  error_listener: MethodType,
-                 max_errors: int = MAX_ERRORS,
                  retries: int = RETRIES) -> None:
         """
         :param error_listener: Function to execute when the maximum number of
         consecutive errors is reached
         :param max_errors: Maximum number of consecutive unrecoverable errors
```
mfranciszkiewicz: No longer here

golem/ethereum/web3/middleware.py
Krigpl: web3 will translate this dict into an exception anyway here https://github.com/ethereum/web3.py/blob/master/web3/manager.py#L112 so that won't help.

mfranciszkiewicz (approved):

PR #3302, opened 2018-09-12
More short node_id's in logs

maaktweluit:


tests/golem/task/test_concent_logic.py
```diff
@@ -317,6 +317,7 @@ def test_ack(self, send_mock, *_):
 
 
 @mock.patch('golem.task.tasksession.node_info_str')
+@mock.patch('golem.resource.resourcehandshake.short_node_id')
```
Krigpl: why do you need to patch it ?

tests/golem/task/test_tasksession.py
```diff
@@ -134,7 +134,8 @@ def _set_task_state(self):
         self.conn.server.task_manager.tasks_states[self.task_id] = task_state
         return task_state
 
-    def test_want_to_compute_task(self):
+    @patch('golem.resource.resourcehandshake.short_node_id')
```
Krigpl: same question

tests/golem/task/test_concent_logic.py
maaktweluit: because the `node_id` passed to this function is a `Mock` object.

Now i think about it i could also set a random node_id, would that be better?

Krigpl: I guess that's also the reason why the above patch exists?
imo it'd be better to set proper `node_id` in the `setUp` - otherwise we might end up patching more and more functions like that (and in general I find it better to mock inputs in tests rather than random functions across whole codebase that happen to be inconvenient at the moment - you wouldn't be able to do it in civilized languages /r).

maaktweluit:
@Krigpl Thanks for the comments, it makes the PR a lot cleaner!
Will keep this in mind when fighting more civilised languages ;)
Could you please have another look?

tests/golem/task/test_concent_logic.py
```diff
@@ -316,14 +316,14 @@ def test_ack(self, send_mock, *_):
...
 
-@mock.patch('golem.task.tasksession.node_info_str')
 @mock.patch('golem.task.tasksession.TaskSession.send')
 class ReactToWantToComputeTaskTestCase(unittest.TestCase):
     def setUp(self):
         super().setUp()
         self.requestor_keys = cryptography.ECCx(None)
         self.msg = factories.tasks.WantToComputeTaskFactory()
         self.task_session = tasksession.TaskSession(mock.MagicMock())
+        self.task_session.key_id = 'a'
```
Krigpl: In cases like this I usually like to use strings like `unittest_key_id` so it's clearer in the logs what it is, since standalone `a` is not self explanatory.
Or even use proper stub values, so like 64 hex characters here, because e.g. `short_node_id('a') == 'aa'` which isn't correct :)
Up to you though

Krigpl (approved):
Yeah, looks much better :)

tests/golem/task/test_concent_logic.py
maaktweluit: When picking the value i looked at other around, they are `ABC` and `DEF`, this is why i picked `a` since it meant just as much.

I like making it more clear, so updated it to your proposal :)

PR #3303, opened 2018-09-12
Add golemcli tasks purge command

mplebanski:


shadeofblue:
@mplebanski would you consider adding some tests for the purge_tasks method? both positive (where the task exists) and negative -> where the given id(s) are not found? ...

mplebanski:
@shadeofblue please clarify what do you mean by "given id" because purge_tasks does not take any input. It's a convenience wrapper over delete_task

shadeofblue:
> @shadeofblue please clarify what do you mean by "given id" because purge_tasks does not take any input. It's a convenience wrapper over delete_task

@mplebanski ah, yeah - right... my bad ... so just a positive test that ensures that indeed the `delete_task` is called for any tasks found should suffice...

mplebanski:
I added the test for purge. The control_test timed out though. It looks to me like a heavy load on buildbot.

maaktweluit:
@mplebanski re-build the `hook_pr` so it would try again, the timeout seems like a bug in one of the tests.

When you get an approval the `control_build` should trigger too :)

mplebanski:
*buildbot/control_test *— Build done. Succeeded

maaktweluit (changes requested):
LGTM!

Maybe add a info log on the core when the purge is called.

Currently CLI commands give no feedback, so this is consistent. Would be nice to print `X tasks deleted` . as user feedback ( optional )

will approve on reply or fix :)

mplebanski:
@maaktweluit As you said it is an inconsistent behaviour. Adding verbosity to CLI is a subject for another PR.

maaktweluit:
@mplebanski Is the reply for both the log and the optional cli feedback?

I think the log would be nice to add now, and leave the additional feedback for the other PR

mplebanski:
I misunderstood you. It's changed now.

maaktweluit (approved):

PR #3304, opened 2018-09-12
Take into consideration TaskHeader.concent_enabled

### LINKED ISSUE #3123: [Concent] Take task header into consideration when setting `WantToCompute`.`concent_enabled`
The `concent_enabled` flag in the `WantToCompute` message should be set based on:
* the information from the requestor's task header
* provider's own client setting

depends on: https://github.com/golemfactory/golem/issues/3122

jiivan:
fixes: #3123


golem/resource/resourcehandshake.py
```diff
@@ -80,6 +80,14 @@ def request_task(self, node_name, task_id, perf_index, price,
         """
 
         key_id = self.key_id
+        task_header = self.task_server.task_keeper.task_headers[task_id]
+        if not self.task_server.client.concent_service.enabled:
+            concent_enabled = False
+        elif not task_header.concent_enabled:
+            self._handshake_error(key_id, 'Concent required')
```
tworec, line 87: why do you returning handshake error here?

tworec (commented):
I'm rather aside concent. It is intentional to ask me for this revie?

shadeofblue (approved):
pear on a tree

jiivan:
@tworec You were suggested by github as a reviewer (because you've worked on files changed in this PR). I like the idea of spreading knowledge among the team, so yeah - intentional but nothing personal. :)

golem/resource/resourcehandshake.py
jiivan: Because if provider has concent enabled but requestor doesn't, then there is no reason to continue communication.

tworec: ok, didn't notice one of `not`'s

tworec:
selfie-0

PR #3313, opened 2018-09-14
Allow origins fix

mdtanrikulu:
To use multiple golem-electron for multiple instances, allowed origins fixed


golem/rpc/router.py
```diff
@@ -93,12 +93,10 @@ def _build_config(address: WebSocketAddress,
                       enable_webstatus: bool = False):
 
         allowed_origins = [
-            address.host,
-            address.host + ':*',
-            'http://' + address.host,
+            "http://localhost:*",
```
mfranciszkiewicz: Entries will be duplicated when `address.host == "localhost"`

mdtanrikulu: Yes, that's true, should we keep only localhost then as hardcoded?

mfranciszkiewicz: The thing is that `localhost` is the default value. You can pass a different address for crossbar to listen on, e.g. `-r 127.0.0.1:60111` or `-r myrig:60111` so that won't be enough.

mdtanrikulu: Given flag on golem electron;
```
npm run start -- --datadir /Users/USER_NAME/golem_2 --rpc-address 127.0.0.1:60002
```
for golem;
```
python golemapp.py -d /Users/USER_NAME/golem_2 -r 127.0.0.1:60002 
```
Result;
```
WARNING  [twisted] dropping connection to peer tcp4:127.0.0.1:50579 with abort=False: WebSocket connection denied: origin 'http://localhost:3002' not allowed
```
As you see above; crossbar gets `127.0.0.1` address as `localhost`, but CORS does not applies the  rule for `localhost` properly; because `address.host` variable is `127.0.0.1` by default not `localhost`. And it's probably doing a simple string check for CORS. 

The only way to get `address.host` as `localhost` is passing rpc-address flag as `localhost:60002`.

So my suggestion is to check if `address.host`  is either `localhost` or `127.0.0.1` and replace it with `localhost` all the time.

mfranciszkiewicz (approved):

Issue #3272, opened 2018-09-03
Add support for `WantToComputeTask` inside `TaskToCompute`


shadeofblue:
This is the Golem Core counterpart that depends on:

depends on: https://github.com/golemfactory/golem-messages/issues/244

rwrzesien:
@shadeofblue Could you please share more details, at least in which places I could possibly start to look? It will be my first Golem Core task so, every piece of information is appreciated :)

jiivan:
@rwrzesien sure, here it goes:

- [ ] As a requestor we only need to fill `TaskToCompute.want_to_compute_task`. See: https://github.com/golemfactory/golem/blob/d0986ebe2f896377c459d32c7e37d81c46124a64/golem/task/tasksession.py#L500-L512
- [ ] As a provider verify that `WantToComputeTask` attached to `TaskToCompute` is indeed correct (verify signature?) See: https://github.com/golemfactory/golem/blob/d0986ebe2f896377c459d32c7e37d81c46124a64/golem/task/tasksession.py#L545-L557

PR #3314, opened 2018-09-14
Feature add support for want to compute task inside task to compute

rwrzesien:
Resolves https://github.com/golemfactory/golem/issues/3272

Please take into consideration that this is my first pull request in Golem. All comments are appreciated.

### DIFF
--- a/golem/network/concent/helpers.py
+++ b/golem/network/concent/helpers.py
@@ -37,12 +37,13 @@ def verify_message_payment_address(
         ecc: cryptography.ECCx) -> bool:
     # Prevent self payments. This check deserve its own reject_reason but also
     # it belongs ealier in the flow rather then here.
-    if privtoaddr(ecc.get_privkey()) == decode_hex(
-            report_computed_task.eth_account):
+    decoded_provider_ethereum_address = decode_hex(
+        report_computed_task.task_to_compute.provider_ethereum_address)
+    if privtoaddr(ecc.get_privkey()) == decoded_provider_ethereum_address:
         logger.warning('Prevented self payment: %r', report_computed_task)
         return False
     # Prevent payments to zero address. Same as above.
-    if decode_hex(report_computed_task.eth_account) == b'\x00' * 20:
+    if decoded_provider_ethereum_address == b'\x00' * 20:
         logger.warning(
             'Prevented payment to zero address: %r',
             report_computed_task,
--- a/golem/resource/resourcehandshake.py
+++ b/golem/resource/resourcehandshake.py
@@ -98,8 +98,8 @@ class ResourceHandshakeSessionMixin:
             max_memory_size=max_memory_size,
             num_cores=num_cores,
             concent_enabled=concent_enabled,
-            provider_public_key=key_id,
-            provider_ethereum_public_key=key_id,
+            provider_public_key=self.task_server.get_key_id(),
+            provider_ethereum_public_key=self.task_server.get_key_id(),
         )
 
         if self._is_peer_blocked(key_id):
--- a/golem/resource/resourcehandshake.py
+++ b/golem/resource/resourcehandshake.py
@@ -98,6 +98,8 @@ class ResourceHandshakeSessionMixin:
             max_memory_size=max_memory_size,
             num_cores=num_cores,
             concent_enabled=concent_enabled,
+            provider_public_key=key_id,
+            provider_ethereum_public_key=key_id,
         )
 
         if self._is_peer_blocked(key_id):
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -7,6 +7,7 @@ import time
 from ethereum.utils import denoms
 from golem_messages import helpers as msg_helpers
 from golem_messages import message
+from golem_messages.exceptions import InvalidSignature
 
 from golem.core.common import HandleAttributeError, node_info_str
 from golem.core.keysauth import KeysAuth
@@ -205,15 +206,10 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin):
             task_to_compute = get_task_message(
                 'TaskToCompute', task_id, subtask_id)
 
-            eth_address = get_task_message(
-                'ReportComputedTask',
-                task_id,
-                subtask_id,
-            ).eth_account
             payment_processed_ts = self.task_server.accept_result(
                 subtask_id,
                 self.key_id,
-                eth_address,
+                task_to_compute.provider_ethereum_address,
             )
 
             response_msg = message.tasks.SubtaskResultsAccepted(
@@ -265,14 +261,12 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin):
             task_result,
             address,
             port,
-            eth_account,
             node_info):
         """ Send task results after finished computations
         :param WaitingTaskResult task_result: finished computations result
                                               with additional information
         :param str address: task result owner address
         :param int port: task result owner port
-        :param str eth_account: ethereum address (bytes20) of task result owner
         :param Node node_info: information about this node
         :return:
         """
@@ -308,7 +302,6 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin):
             port=port,
             key_id=self.task_server.get_key_id(),
             node_info=node_info.to_dict(),
-            eth_account=eth_account,
             extra_data=extra_data,
             size=task_result.result_size,
             package_hash='sha1:' + task_result.package_sha1,
@@ -506,12 +499,11 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin):
             )
             ttc = message.tasks.TaskToCompute(
                 compute_task_def=ctd,
+                want_to_compute_task=msg,
                 requestor_id=task.header.task_owner.key,
                 requestor_public_key=task.header.task_owner.key,
                 requestor_ethereum_public_key=task.header.task_owner.key,
                 provider_id=self.key_id,
-                provider_public_key=self.key_id,
-                provider_ethereum_public_key=self.key_id,
                 package_hash='sha1:' + task_state.package_hash,
                 concent_enabled=msg.concent_enabled,
                 price=price,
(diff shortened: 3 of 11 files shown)


requirements.txt
```diff
@@ -31,7 +31,7 @@ eth-tester==0.1.0-beta.24
 eth-utils==1.0.3
 ethereum==1.6.1
 Golem-Verificator==1.4.0
-Golem-Messages==2.13.0
+Golem-Messages==2.13.0+dev4.gd8cd1d3
```
cameel: I don't think you're supposed to put untagged development versions in the requirements.

tests/golem/network/concent/test_helpers.py
```diff
@@ -19,7 +19,7 @@ def test_self_payment(self):
         ecc = ECCx(privkey)
         ecc.verify = mock.Mock()
         msg = mock.Mock()
-        msg.eth_account = encode_hex(addr)
+        msg.task_to_compute.provider_ethereum_address = encode_hex(addr)
```
cameel, line 22: Wasn't `eth_account` supposed to be replaced by a property to avoid having this kind of changes?

golem/task/tasksession.py
```diff
@@ -550,8 +543,21 @@ def _react_to_want_to_compute_task(self, msg):
     @history.provider_history
     def _react_to_task_to_compute(self, msg):
         ctd = msg.compute_task_def
-        if ctd is None:
-            logger.debug('TaskToCompute without ctd: %r', msg)
+        want_to_compute_task = msg.want_to_compute_task
+        if ctd is None or want_to_compute_task is None:
+            logger.debug(
+                'TaskToCompute without ctd or want_to_compute_task: %r', msg)
+            self.task_computer.session_closed()
```
cameel, line 549: I think that golem-messages should have a validation against this.

requirements.txt
Krigpl: This is fine to check whether CI passes but should be changed to a proper version before merging.

golem/resource/resourcehandshake.py
```diff
@@ -89,6 +89,8 @@ def request_task(self, node_name, task_id, perf_index, price,
             max_memory_size=max_memory_size,
             num_cores=num_cores,
             concent_enabled=self.task_server.client.concent_service.enabled,
+            provider_public_key=key_id,
+            provider_ethereum_public_key=key_id,
```
Krigpl: Why is this needed? I have a feeling that ethereum public key shouldn't be necessary for resource handshake.

jiivan: Actually it's not really related to resources. It's just a method that generates `WantToComputeTask`. Reasoning for including provider eth details in WTC is in https://github.com/golemfactory/golem-messages/issues/244

requirements.txt
shadeofblue: exactly as @Krigpl said

golem/network/concent/helpers.py
```diff
@@ -38,11 +38,13 @@ def verify_message_payment_address(
     # Prevent self payments. This check deserve its own reject_reason but also
     # it belongs ealier in the flow rather then here.
     if privtoaddr(ecc.get_privkey()) == decode_hex(
-            report_computed_task.eth_account):
+            report_computed_task.task_to_compute.provider_ethereum_address):
```
shadeofblue: @rwrzesien hmm, I know this had been there before - but maybe it would be good to put the `decode_hex(address)` lookup first so that it doesn't need to be converted twice (here and in line 45 below)

golem/resource/resourcehandshake.py
Krigpl: Right, I got confused and thought this is some kind of Hello message instead of WantToComputetTask

golem/task/tasksession.py
shadeofblue: adding a way to ensure this: 
https://github.com/golemfactory/golem-messages/issues/262

requirements.txt
rwrzesien: Yes, also my point was to allow anyone who would like to test/run from this branch to have everything what is needed in place. I didn't assume that this commit would get merged along with this branch.

golem/resource/resourcehandshake.py
rwrzesien: Yes, exactly, this method accepts parameters that are passed directly to `WantToComputeTask`.

tests/golem/network/concent/test_helpers.py
rwrzesien: Yes, but please take a look on lines above, it is set on a `Mock` object to test `process_report_computed_task_no_time_check` function.

golem/network/concent/helpers.py
rwrzesien: @shadeofblue Sure, no problem done :+1:

golem/task/tasksession.py
```diff
@@ -550,8 +543,21 @@ def _react_to_want_to_compute_task(self, msg):
...
+        want_to_compute_task = msg.want_to_compute_task
+        if ctd is None or want_to_compute_task is None:
+            logger.debug(
+                'TaskToCompute without ctd or want_to_compute_task: %r', msg)
+            self.task_computer.session_closed()
+            self.dropped()
+            return
+
+        try:
+            msg.verify_signature(
```
shadeofblue: @rwrzesien please correct me if I'm wrong but `msg` here is a `TaskToCompute`, isn't it? so I believe it should be:

```
want_to_compute_task.verify_signature(
    decode_hex(msg.provider_public_key)  # we can use the helper here actually... 
)
```

am I missing something?

requirements.txt
shadeofblue: @rwrzesien btw - does it work this way? I mean - shouldn't this line be a link to github's commit? like this: https://github.com@4r/golemfactory/golem/pull/3317/files#diff-b4ef698db8ca845e5845c4618278f29aR34 ?

tests/golem/network/concent/test_helpers.py
shadeofblue: @rwrzesien @cameel (disregarding the fact that here something else is happening) hmm, well it would be better to just replace the occurences of `eth_account` with `ttc.provider_ethereum_address` anyway sooner or later... so if we can do it here, all the better...

requirements.txt
shadeofblue: @rwrzesien as I was afraid -> it doesn't work that way: https://buildbot.golem.network/buildbot/#/builders/9/builds/6583/steps/2/logs/install_requirements -> please replace the requirement with a link to the github commit (and once the `golem-messages` counterpart is released we'll be able to put the correct version here before merging)

golem/task/tasksession.py
shadeofblue: @rwrzesien on second thought -> it may make little sense to verify the signature of the `want_to_compute_task` against the included `provider_public_key` here - remember, we're the provider and the message had been generated by us - so, it would actually make more sense to verify that the attached `WantToComputeTask` is _our_ message, so, instead, we should probably do the following:

```
want_to_compute_task.verify_signature(self.task_server.keys_auth.ecc.raw_pubkey)
```

to ensure the message had been generated by us...

requirements.txt
rwrzesien: :+1: done

golem/task/tasksession.py
rwrzesien: :+1: done

requirements.txt
rwrzesien: :+1: done

golem/task/tasksession.py
rwrzesien: :+1: done

requirements.txt
cameel: @shadeofblue Just FYI, it would work if it was what `git describe` returns (which would look like this: `2.13.0-10-gd8cd1d3`). It does not work because it's from the version script and does not have  the original `<tag>-<n>-<hash>` format. I was surprised too when I saw it for the first time but git does allow you to use something like this as a reference to a commit and simply ignores everything but the hash.

cameel: Oh, and by the way - if this was just temporary change I think it would have been a good idea to have a comment in the code. Or have it in a separate commit. Otherwise it's not obvious and easy to miss later when you merge (hence my remark that started this thread).

golem/task/tasksession.py
```diff
@@ -550,8 +542,21 @@ def _react_to_want_to_compute_task(self, msg):
     @history.provider_history
     def _react_to_task_to_compute(self, msg):
         ctd = msg.compute_task_def
-        if ctd is None:
-            logger.debug('TaskToCompute without ctd: %r', msg)
+        want_to_compute_task = msg.want_to_compute_task
+        if ctd is None or want_to_compute_task is None:
```
shadeofblue, line 546: @rwrzesien so yeah, I believe the relevant validation _has_ been added to `golem-messages`

tests/golem/network/concent/test_received_handler.py
```diff
@@ -238,11 +238,12 @@ def get_vrct(self):
             requestor_public_key=msg_utils.encode_hex(
                 self.requestor_keys.raw_pubkey,
             ),
-            provider_public_key=msg_utils.encode_hex(
-                self.provider_keys.raw_pubkey,
-            ),
             sign__privkey=self.requestor_keys.raw_privkey,
         )
+        ttc.want_to_compute_task.provider_public_key = msg_utils.encode_hex(
```
shadeofblue: @rwrzesien I believe it would be safer to specify the relevant key in the call to `TaskToComputeFactory` above with:

```
want_to_compute_task__provider_ethereum_public_key = self.provider_keys.raw_pubkey,
want_to_compute_task__sign__privkey=self.provider_keys.raw_privkey
```

to ensure the `WantToComputeTask` is signed correctly... but maybe the handlers here don't check that condition...

shadeofblue (approved):
seems good - I added one comment above to make the test more futureproof...

golem/task/tasksession.py
rwrzesien: @shadeofblue Should I remove this check then? If yes, should I remove it along with check for CTD which was there before?

tests/golem/network/concent/test_received_handler.py
rwrzesien: :+1: done

jiivan (approved):
LGTM

PR #3315, opened 2018-09-17
Lock GNT during withdrawal

Krigpl:
So that it's substracted from the balance.

### DIFF
--- a/golem/ethereum/transactionsystem.py
+++ b/golem/ethereum/transactionsystem.py
@@ -75,6 +75,7 @@ class TransactionSystem(LoopingCallService):
         self._last_gnt_update: Optional[float] = None
         self._payments_locked: int = 0
         self._gntb_locked: int = 0
+        self._gntb_withdrawn: int = 0
         # Amortized gas cost per payment used when dealing with locks
         self._eth_per_payment: int = 0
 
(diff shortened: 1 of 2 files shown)


golem/ethereum/transactionsystem.py
```diff
@@ -483,7 +485,16 @@ def withdraw(
                 amount / denoms.ether,
                 destination,
             )
-            return self._sci.convert_gntb_to_gnt(destination, amount)
+            tx_hash = self._sci.convert_gntb_to_gnt(destination, amount)
+
+            def on_receipt(receipt) -> None:
+                if not receipt.status:
+                    log.error("Failed GNTB withdrawal: %r", receipt)
+                    return
```
maaktweluit: will this be re-tried? else the gntb is locked for ever.

maaktweluit (commented):
LGTM!

One small q, will approve on reply or fix

golem/ethereum/transactionsystem.py
Krigpl: Yeah, I noticed it right after submitting the PR and fixed it already, I hoped nobody has noticed yet :P

maaktweluit (approved):

PR #3317, opened 2018-09-17
use messages with strict slot verification (in `verify_slot`)

shadeofblue:
Golem Core counterpart to https://github.com/golemfactory/golem-messages/pull/264


Krigpl (approved):
waffle

PR #3321, opened 2018-09-19
Rename computation_finished to subtask_finished.

mplebanski:
Refactor a confusing method name very similiar to "finished_computation"
which is a method used to determine task completion.

### DIFF
--- a/apps/core/task/coretask.py
+++ b/apps/core/task/coretask.py
@@ -206,7 +206,7 @@ class CoreTask(Task):
     def computation_failed(self, subtask_id):
         self._mark_subtask_failed(subtask_id)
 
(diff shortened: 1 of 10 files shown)


mplebanski:
This pull requests is actually a question about what is our approach to such changes in the code? This method name is clearly confusing, but some other code might already depend on it.

Wiezzel:
1) What about `computation_failed`?
2) I think we will be getting rid of "subtask" at all, so maybe wait until our new nomenclature is settled? (summon @badb)

badb:
This one should wait until taski API is more precisely defined (and we decide if we want to inform about job being finished, child task being finished or other part of computation being finished)

PR #3322, opened 2018-09-19
Removal NAT type retrieval

mplebanski:
NAT type is not in use, yet it's retrieval takes significant amount of time resulting in a start up slowdown.

### DIFF
--- a/golem/core/hostaddress.py
+++ b/golem/core/hostaddress.py
@@ -139,10 +139,10 @@ def get_external_address(source_port=0):
     :return (str, int, str): tuple with host public address, public port that is
     mapped to local <source_port> and this host nat type
     """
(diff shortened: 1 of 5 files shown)


mplebanski:
I should fix a commit message to "Remove NAT type retrieval"

jiivan (approved):
LGTM, but two small comments:
1. I tried to add a review, but then you forced pushed... Please don't, it breaks things on github, and really there is no good reason for it. You can always squash when closing PR.
1. Here is my comment:
commit-missing

mplebanski:
1. Fine, starting from next commit 
2. Honestly I would inline get_nat_type code in get_ip_info. That will also require tests refactoring

PR #3330, opened 2018-09-20
Docker toolbox (Windows): fix pythoncom import

mfranciszkiewicz:
- fix pythoncom import
- fix init argument order
- fix CPU cap
- fix VirtualBox VM names

### DIFF
--- a/golem/docker/config.py
+++ b/golem/docker/config.py
@@ -35,16 +35,13 @@ DEFAULT_HOST_CONFIG = dict(
 CONSTRAINT_KEYS = dict(
     mem='memory_size',
     cpu='cpu_count',
-    cpu_cap='cpu_execution_cap'
 )
 MIN_CONSTRAINTS = dict(
     memory_size=1024,
-    cpu_execution_cap=10,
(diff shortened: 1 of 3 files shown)


maaktweluit (approved):
LGTM!

Wiezzel (commented):
I thinks this will work but I have some doubts regarding the design.
1) Checking if the environment is available should only require checking if the `pythoncom` module is available.
2) Calling import multiple times in unnecessary.

I would recommend a different (but maybe over-complicated) approach:

1) Use a metaclass and importlib to assign `pythoncom` module to a class variable of `VirtualBoxHypervisor` class. Checking availability would require only `cls.pythoncom is not None`.
2) Call `self.pythoncom.CoInitialize()` in `VirtualBoxHypervisor.__init__()`

mfranciszkiewicz:
As we've talked offline, `pythoncom.CoInitialize()` needs to be called on a per thread basis. I don't think that the second import is a bottleneck, so the whole initialization process can be packed into a single function.

PR #3332, opened 2018-09-20
Remove scipy dependency

inexxt:
The `scipy` dependency is responsible for almost a half of the Golem dependencies size (in terms of disk space - all requirements take `437M`, `scipy` - `183M`).
It appears that there is only one place, where one one-line scipy function was used.
This PR removes the dependency.

### DIFF
--- a/apps/rendering/resources/imgverifier.py
+++ b/apps/rendering/resources/imgverifier.py
@@ -4,88 +4,8 @@ import math
 
 from apps.rendering.resources.imgrepr import (ImgRepr, PILImgRepr)
 
-from golem_verificator.verifier import SubtaskVerificationState
-
-from ssim import compute_ssim
 
 logger = logging.getLogger("apps.rendering")
-
-
-class ImgStatistics(object):
-    def __init__(self, base_img: ImgRepr, img: ImgRepr):
-        if base_img.get_size() != img.get_size():
-            raise ValueError('base_img and img are of different sizes.')
-
-        self.img = img
-        self.ssim = compute_ssim(base_img.to_pil(), self.img.to_pil())
-        self.mse, self.norm_mse = \
-            self._calculate_color_normalized_mse(base_img, self.img)
-        self.mse_bw, norm_mse_bw = \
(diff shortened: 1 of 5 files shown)


maaktweluit (changes requested):
LGTM! nice cleanup!

There are 2 locations left in the project mentioning scipy:

https://github.com/golemfactory/golem/blob/develop/scripts/pyinstaller/hooks/hook-scipy.py

https://github.com/golemfactory/golem/blob/f567285f189285b04f4a5cd0bfe4eb2b2c6fc441/apps/rendering/resources/images/requirements.txt

I think the first one needs to be deleted, the second one is only for inside the docker images ( i think )

Will approve on reply!

maaktweluit:
When testing this branch i noticed `scipy` is still installed:
```
(.venv) ➜  golem git:(remove_scipy_dep) ✗ pip freeze | grep scipy
scipy==1.1.0
```

checking the pip dependency tree it shows:
```
pyssim==0.4
...
  - scipy [required: Any, installed: 1.1.0]
```

This change does not make much sense if we still keep the dependency indirectly, `pyssim` is only used in 1 place, maybe this can also be moved to the verification docker image an removed here.

inexxt:
Right, I haven't spotted the `pyssim` dependency. But it seems that it is only used in docker and in `imgverifier.py` - which I think is dead code (no unit tests break, I am able to act both as a provider and requestor after deleting this code).
But it would be better if @badb confirmed that.

mfranciszkiewicz:
`pyssim` is also a dependency in https://github.com/golemfactory/golem-verificator but doesn't seem to be required there either.

inexxt:
In `golem-verificator`, it is used in `lux_verifier.py`. So it seems that it is actually not possible to drop `scipy`, without dropping support for lux verification.

mfranciszkiewicz:
You're right, my bad.

maaktweluit:
Maybe this PR can continue after #3382 is merged

maaktweluit:
Merged in #3401

PR #3334, opened 2018-09-24
Update SCI to 1.6.0

Krigpl:
SOo Concent can provide different GNTDeposit addresses depending on cluster (it's not on this PR though).


mfranciszkiewicz (approved):
Uhm, ok

PR #3337, opened 2018-09-24
Add current user to hyperv group

maaktweluit:
screen shot 2018-09-24 at 16 47 07


Wiezzel (commented):
How is the group creation supposed to work? Will Windows somehow magically assign the appropriate SID and permissions based on the name we provide? It doesn't look possible to me. Also, the name should depend on system's language.

maaktweluit:
@Wiezzel good catch!

 found the group is created when windows is installed, so i think we can remove this block.

Wiezzel:
Also, I cannot see what happens if adding user to the group fails.

maaktweluit:
screen shot 2018-09-24 at 18 03 06

Updated without the create block . ^

Currently nothing happens when the `add` fails, should we fail the installer?

Wiezzel:
I believe that failure should stop the installer. This not an extra feature. This is really necessary to run Golem.

maaktweluit:
Sounds reasonable :) updated the script to fail if the check does not complete the second time:

screen shot 2018-09-25 at 13 11 50

Wiezzel (approved):
:+1:

PR #3341, opened 2018-09-25
Add directory support to resultpackage module

mplebanski:
This allows specifying a directory in "resources" key in task
definition, e.g.:

```
{
    ...
        "resources": [
            '/home/user/task_data'
        ]
    ...
}
```

All files and subdirectories will be zipped under 'task_data' directory
inside a package.

### DIFF
--- a/golem/task/result/resultpackage.py
+++ b/golem/task/result/resultpackage.py
@@ -153,7 +153,8 @@ class ZipPackager(Packager):
                     ZipPackager.zip_append(archive, os.path.join(root, d),
                                            os.path.join(subdirectory, d))
                 for f in files:
-                    archive.write(os.path.join(root, f), os.path.join(subdirectory, f))
+                    archive.write(os.path.join(root, f),
+                                  os.path.join(subdirectory, f))
                 break
         elif os.path.isfile(path):
             archive.write(path, os.path.join(subdirectory, basename))
--- a/golem/task/result/resultpackage.py
+++ b/golem/task/result/resultpackage.py
@@ -109,7 +109,6 @@ class Packager(object):
     def write_cbor_file(self, obj, file_name, cbor_data):
         pass
 
-
 class ZipPackager(Packager):
 
     ZIP_MODE = zipfile.ZIP_STORED  # no compression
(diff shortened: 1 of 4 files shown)


golem/task/result/resultpackage.py
```diff
@@ -141,6 +141,25 @@ def package_name(cls, file_path):
             return file_path
         return file_path + '.zip'
 
+    @staticmethod
+    def zip_append(ob, path, rel=""):
```
Wiezzel: I have following comments regarding this method:
1) Parameter names are not explanatory to me. I could guess that `path` points to the file/directory we want to add to a zip archive and `ob` is probably some object representing this archive but it's not obvious. Type annotations would help too. And the role of `rel` argument is still not clear to me after reading the code.
2) Are directory paths with trailing slashes handled well? As far as I can remember `os.path.basename()` returns empty string when called on such path.
3) This method could be unit-tested by creating some temporary files/directories and mocking `ob`. Please add some unit tests. Apart from assuring correctness they would help to understand the function's desired behaviour.
4) I'm not sure if silently skipping non-existent paths is a good choice. I'd rather go with raising an exception or at least displaying a warning.

Wiezzel (changes requested):

golem/task/result/resultpackage.py
mplebanski: Thanks for comments. 

Ad 1 
As suggested I refactored names `ob` to `archive` and `rel` to `subdirectory`. Is the `rel` paramter clear now? 

Ad 2
Fixed that with `rstrip('/')`. I am not sure if I should put this inside `zip_append` function or leave it where it is now. 

Ad 3 
Added 

Ad 4 
True, I added `RuntimeException` with appropriate explanation

tests/golem/task/result/test_resultpackage.py
```diff
@@ -82,6 +82,83 @@ def testExtract(self):
...
+        path, _ = zp.create(self.out_path, self.disk_files, None)
+
+        self.assertTrue(os.path.exists(path))
+
+    def testExtract(self):
+        zp = ZipPackager()
+        zp.create(self.out_path, self.disk_files, None)
+        files, _ = zp.extract(self.out_path)
+
+        self.assertTrue(len(files) == len(self.expected_results))
```
Wiezzel: Why compare only lengths? I'd say we should check if the directory structure is OK.

mplebanski:
(Alt Text

Wiezzel (approved):

PR #3343, opened 2018-09-26
fix Concent integration tests after introduction of the new RPC authentication

### LINKED ISSUE #3336: Provide helpful message when WAMPCRA secret is missing
Currently if you use custom datadir (`./golemapp.py -d <datadir>`) then `./golemcli.py` fails with `Maybe you forgot to create secrets?`. It would be more helpful to suggest `-d` parameter for `golemcli.py`.

Introduced in #3171

### LINKED ISSUE #3342: make Concent integration tests compatible with the new RPC authentication

shadeofblue:
+ provide a more meaningful exception message

closes #3342
closes #3336

### DIFF
--- a/golem/rpc/cert.py
+++ b/golem/rpc/cert.py
@@ -19,7 +19,7 @@ DH_PARAM_BITS_LOW = 1024
 KEY_BITS = 2048
 
 
-class CertificateException(Exception):
+class CertificateError(Exception):
     pass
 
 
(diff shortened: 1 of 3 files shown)


golem/rpc/cert.py
```diff
@@ -19,6 +19,10 @@
 KEY_BITS = 2048
 
 
+class CertificateException(Exception):
```
jiivan: `CertificateError`

jiivan (commented):
`*Error` is more pythonic, otherwise LGTM.

jiivan (approved):

PR #3345, opened 2018-09-26
Move repeated code to a common base in TaskCore

Krigpl:
Move repetitive piece of code to a common base.
This isn't non functional change for Lux, but I think way is more appropriate and Lux is going away anyway.

### DIFF
--- a/apps/dummy/task/dummytask.py
+++ b/apps/dummy/task/dummytask.py
@@ -101,8 +101,6 @@ class DummyTask(CoreTask):
 
     def accept_results(self, subtask_id, result_files):
         super().accept_results(subtask_id, result_files)
-        node_id = self.subtasks_given[subtask_id]['node_id']
-        TaskClient.assert_exists(node_id, self.counting_nodes).accept()
         self.num_tasks_received += 1
 
     def __get_result_file_name(self, subtask_id: str) -> str:
--- a/apps/dummy/task/dummytask.py
+++ b/apps/dummy/task/dummytask.py
@@ -15,7 +15,6 @@ from apps.dummy.task.dummytaskstate import DummyTaskDefaults, DummyTaskOptions
 from apps.dummy.task.dummytaskstate import DummyTaskDefinition
 from apps.dummy.task.verifier import DummyTaskVerifier
 from golem.task.taskbase import Task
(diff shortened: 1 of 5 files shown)


apps/dummy/task/dummytask.py
```diff
@@ -101,8 +101,6 @@ def query_extra_data(self,
 
     def accept_results(self, subtask_id, result_files):
         super().accept_results(subtask_id, result_files)
-        node_id = self.subtasks_given[subtask_id]['node_id']
-        TaskClient.assert_exists(node_id, self.counting_nodes).accept()
```
maaktweluit, line 105: you left an unused import here

mplebanski:
A more meaningful commit message would be useful

Wiezzel (commented):
Looks good, but:
1) I don't get the Lux part. I mean I have no idea why these instructions were placed inside that loop in the first place.
2) What about moving `self.num_tasks_received += 1` to `CoreTask` as well?

maaktweluit (changes requested):
LGTM!

same comment 3x on all non-core files

will approve after fix :)

maaktweluit:
@Wiezzel was thinking the same, but after i looked at the rendertask one it shows multiple results
https://github.com/golemfactory/golem/pull/3345/files#diff-0eb7ac38825687120c20b4fd70f9008bL205

Wiezzel:
@maaktweluit I think we've just established on #developers channel that `start_task` and `end_task` are always equal.

maaktweluit:
@Wiezzel i did not see that those 2 where related, then the loop can be ignored.

maaktweluit (approved):

Wiezzel:
@Krigpl Why did you ignore my comment? :crying_cat_face:

Krigpl:
@Wiezzel Sorry, I meant to answer you:
> What about moving self.num_tasks_received += 1 to CoreTask as well?

I wanted to do that, but then I thought that it looks innocent but may spark a larger discussion that I didn't want to conduct in this PR so I sticked to my original change (since I wanted to clean up the usage of `TaskClient` and not do general clean up). But otherwise yeah, this can be done.

Wiezzel:
@Krigpl OK, that's fine.

PR #3346, opened 2018-09-26
More short node_ids, part 3

maaktweluit:
Small improvements to logs:
- use `short_node_id` or `node_info_str` where possible
- make logs in the same function more consistent in data printed.

### DIFF
--- a/golem/task/taskkeeper.py
+++ b/golem/task/taskkeeper.py
@@ -183,7 +183,7 @@ class CompTaskKeeper:
 
     def add_request(self, theader: TaskHeader, price: int):
         # price is task_header.max_price
-        logger.debug('CT.add_request()')
+        logger.debug('CT.add_request(%r, %d)', theader, price)
         if price < 0:
             raise ValueError("Price should be greater or equal zero")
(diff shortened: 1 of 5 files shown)


golem/task/taskkeeper.py
```diff
@@ -183,7 +183,7 @@ def restore(self):
 
     def add_request(self, theader: TaskHeader, price: int):
         # price is task_header.max_price
-        logger.debug('CT.add_request()')
+        logger.debug('CT.add_request(%r, %d)', theader, price)
```
Krigpl, line 186: If you wanna display price you probably wanna divide it by 10^18 (`denoms.ether`) for readability

Krigpl (approved):
thumbupglitch

golem/task/taskkeeper.py
maaktweluit: good one!

In this case i'm not sure if we want to show the arguments or the price. Since its a debug log i think the first, so propose to leave it as is.

mfranciszkiewicz (approved):

PR #3348, opened 2018-09-26
Add --mainnet parameter to golemcli

### LINKED ISSUE #3347: Add a `--mainnet` parameter to `golemcli`
to resolve the issue of `golemcli` not working with golem on mainnet after the merge of https://github.com/golemfactory/golem/pull/3171

mplebanski:
If defined this parameter takes precedence over environment variable
GOLEM_ENVIRONMENT. This commit closes #3347.


golemcli.py
```diff
@@ -84,7 +87,9 @@ def start():
         logging.raiseExceptions = 0
         cli = CLI(main_parser=parser, main_parser_options=flag_options)
 
-    datadir = get_local_datadir('default', root_dir=parsed.datadir)
+    chaindir = 'mainnet' if parsed.mainnet else None
```
shadeofblue: @mplebanski I'm unsure if it's a good idea to hardcode `mainnet` data directory in one more place - it's already defined in golem.config.environments.mainnet.DATA_DIR ... so I'd expect that value to be used, not an arbitrary value that happens to be the same as that one... 

so, instead of adding another parameter to `get_local_datadir`, I'd rather just use `golem.config.environments.set_environment` exactly as it's used in `golemapp.py`...

golem/core/simpleenv.py
```diff
@@ -3,15 +3,17 @@
 import appdirs
 
 
-def get_local_datadir(name: str, root_dir=None) -> str:
+def get_local_datadir(name: str, root_dir=None, data_subdir=None) -> str:
```
shadeofblue: I wouldn't add the `data_subdir` parameter here (see the other comment)

shadeofblue (dismissed):
okay, after discussion, maybe it's the only workable solution... 

grumpy cat 2

mplebanski:
Unfortunately active environment in `golemcli` depends on when the modules are loaded. In this case it's the backtrace:

```
-> from golem.interface.client.account import Account  # noqa
-> from golem.node import ShutdownResponse
-> from apps.appsmanager import AppsManager
-> from golem.config.active import APP_MANAGER_CONFIG_FILES
-> if os.environ.get(GOLEM_ENVIRONMENT_VARIABLE) == MAINNET:
```

that is setting the active env long before the parsed arguments are checked so `set_environment` solution analogical to `golemapp` won't work. I marked `simpleenv.py` with `FIXME`

Krigpl:
Does anybody understand the PR that caused these issues in the first place and could explain it to me?
Because this looks like a bandaid solution to what seems like not properly designed "solution" to some other problem. Since when RPC connections require access to some local internals of the remote side?
What is happening here?

golem/core/simpleenv.py
```diff
@@ -1,17 +1,24 @@
+# pylint: skip-file
 import os
 
 import appdirs
 
-
-def get_local_datadir(name: str, root_dir=None) -> str:
+def get_local_datadir(name: str, root_dir=None, use_mainnet=False) -> str:
```
Krigpl: instead of changing this function you could try to use `set_environment` from `config.environments` just `golemapp` does

Krigpl (dismissed):

golem/core/simpleenv.py
shadeofblue: :D https://github.com/golemfactory/golem/pull/3348#discussion_r220587771

mplebanski:
The last change looks acceptable to me and the tests pass. I could put that config reload into set_environment function to hide this problem from developers in the future.

golem/core/simpleenv.py
```diff
@@ -1,16 +1,16 @@
+# pylint: skip-file
```
shadeofblue: is this needed?

shadeofblue (approved):
okay... I'm generally not in love with the fact that we need to do all these tweaks but I guess it's okay...

golem/core/simpleenv.py
mplebanski: No, not really. I will checkout `develop` for this file.

mplebanski:
The real problem was `set_environment` function and the way it depended on module loading. Now if someone wants to use it the config will be reloaded - as expected.

PR #3349, opened 2018-09-26
Remove perf from subtask info

Krigpl:
Looks unused and I think it's duplicated with `performance` in ComputeTaskDefinition


Wiezzel (approved):
:+1: 
This value was never read anyway...

maaktweluit (approved):
LGTM!

PR #3353, opened 2018-09-27
First draft on using dynamic memory for hyperv

maaktweluit:
Enable dynamic memory:
- Max memory is the configured max from app_cfg
- Min memory is the `MIN_CONSTRAINS[]` value
- Extended VMUtils to query max memory

### DIFF
--- a/golem/docker/hypervisor/hyperv.py
+++ b/golem/docker/hypervisor/hyperv.py
@@ -6,12 +6,13 @@ from subprocess import CalledProcessError, TimeoutExpired
 from typing import Optional, Union, Any, List, Dict, ClassVar
 
 from os_win.exceptions import OSWinException
+from os_win.utils import _wqlutils
 from os_win.utils.compute.vmutils import VMUtils
 
 from golem.core.common import get_golem_path
 from golem.docker import smbshare
 from golem.docker.client import local_client
-from golem.docker.config import CONSTRAINT_KEYS
+from golem.docker.config import CONSTRAINT_KEYS, MIN_CONSTRAINTS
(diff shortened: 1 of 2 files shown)


Wiezzel (commented):
My thoughts:
* I think it would be better to override `VMUtils.get_vm_summary_info()` to include an additional field in the result instead of introducing a new public method.
* Make sure to update `HyperVHypervisor._parse_create_params()` as well to have dynamic memory enabled and configured when creating the VM.

maaktweluit:
@Wiezzel Thanks for the comments!

> I think it would be better to override VMUtils.get_vm_summary_info() to include an additional field in the result instead of introducing a new public method.

In the underlying library the vm_summary_info does not contain this data, that is why i opted to make a new function to get the `vm_memory_data`.
I think this also makes it easier to merge back to the upstream `os_win` library. ( https://github.com/openstack/os-win/compare/master...maaktweluit:master )

> Make sure to update HyperVHypervisor._parse_create_params() as well to have dynamic memory enabled and configured when creating the VM.

This was where i started, but these arguments are not available. The machine is always created with static minimum of 1GB, so i think this is OK.

Wiezzel:
> This was where i started, but these arguments are not available. The machine is always created with static minimum of 1GB, so i think this is OK.

OK, so just remove the `--hyperv-disable-dynamic-memory` parameter.

maaktweluit:
Good one! fixed :)

Wiezzel (approved):
👍

mfranciszkiewicz (approved):
LGTM, cannot test it though.

PR #3354, opened 2018-09-27
fix the created task check in the (concent) integration tests

shadeofblue:


scripts/concent_integration_tests/tests/base.py
```diff
@@ -212,10 +212,11 @@ def on_success(result):
             new_tasks = tasks - self.known_tasks
             if len(new_tasks) != 1:
                 print("Cannot find the new task ({})".format(new_tasks))
-
-            self.task_id = list(new_tasks)[0]
-            print("Task id: {}".format(self.task_id))
-            self.next()
+                time.sleep(30)
+            else:
+                self.task_id = list(new_tasks)[0]
```
jiivan, line 217: `new_tasks.pop()`

scripts/concent_integration_tests/tests/base.py
```diff
@@ -212,10 +212,11 @@ def on_success(result):
             new_tasks = tasks - self.known_tasks
             if len(new_tasks) != 1:
                 print("Cannot find the new task ({})".format(new_tasks))
-
-            self.task_id = list(new_tasks)[0]
-            print("Task id: {}".format(self.task_id))
-            self.next()
+                time.sleep(30)
```
jiivan, line 215: I can't understand from context, why is this sleep needed? Maybe inline comment?

shadeofblue: sigh... there are more such sleep statements in the code there ... and they are used so that anyone running the tests is not spammed with repeated `Cannot find the new task ({})` messages but instead the test waits more patiently for the deposit to happen...

scripts/concent_integration_tests/tests/base.py
shadeofblue: hmm... unsure if it makes a difference here... I'm finding the used syntax more readable...

jiivan: Is really transforming one-element set to list and fetching element by index more readable than just poping element? :)

jiivan (approved):

PR #3358, opened 2018-09-27
React to cancelled offer

Krigpl:
As per marketplace doc.
Provider should not be banned when sending `OfferCancelled`.
TODO: timestamp validation

I'm not happy with this approach but I spent some time thinking how to do it differently and everything requires significant refactoring. Alternative solutions are welcomed.

### DIFF
--- a/apps/core/task/coretask.py
+++ b/apps/core/task/coretask.py
@@ -203,8 +203,12 @@ class CoreTask(Task):
     def finished_computation(self):
         return self.num_tasks_received == self.total_tasks
 
-    def computation_failed(self, subtask_id):
-        self._mark_subtask_failed(subtask_id)
+    def computation_failed(
+            self,
+            subtask_id,
+            offer_cancelled: bool = False):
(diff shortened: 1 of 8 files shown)


Wiezzel:
@Krigpl Could you explain why `computation_failed` and `on_computation_failed` need to be two separate methods?

Krigpl:
@Wiezzel Before `computation_failed` was overridden in derived classes so if I wanted to add this extra param I would need to add it everywhere. And I think it's not really relevant for derived classes. 
Current approach solves that. But still, I'm happy to find a better one.

Wiezzel:
@Krigpl OK, sounds reasonable. I can't think of any better solution. Maybe just make `on_computation_failed` a protected method to make clear that it's supposed to be called from `computation_failed` only.

Krigpl:
After offline discussion with @lukasz-glen OfferCancelled shouldn't be treated any differently than other reasons, so these changes are not required.

PR #3362, opened 2018-09-28
Providers to send offers to multiple requestors at the same time

Krigpl:
Until now provider would send `WantToComputeTask` and then block waiting for the response of the requestor.
Now he doesn't block anymore, so is able to send multiple offers to different requestors and reject any `TaskToCompute`s he wants. As per the marketplace doc.


mfranciszkiewicz:
If I understand correctly, task requests will be tracked by the marketplace module?

Krigpl:
Only after my last commit I actually understood your question.
Should it be tracked by the marketplace module? Right now I just implemented it simply in the TaskServer. I can't find any requestor selection algorithm for providers in the marketplace doc so I dunno it should belong in the markerplace module.

golem/task/taskcomputer.py
```diff
@@ -250,27 +233,20 @@ def task_computed(self, task_thread: TaskThread) -> None:
...
-        self.counting_task = None
         if self.finished_cb:
             self.finished_cb()
 
     def run(self):
         """ Main loop of task computer """
-        if self.counting_task:
-            if self.counting_thread is not None:
-                self.counting_thread.check_timeout()
+        if self.counting_thread is not None:
```
maaktweluit, line 241: is there a reason you did not use `self.is_computing()` here?

golem/task/taskcomputer.py
```diff
@@ -351,42 +329,19 @@ def session_timeout(self):
...
-        self.waiting_deadline = None
+        pass
 
     def __request_task(self):
-        with self.lock:
-            perform_request = not self.waiting_for_task and \
-                              (self.counting_task is None)
-
-        if not perform_request:
+        if self.has_assigned_task():
```
maaktweluit, line 335: the check for `is_computing()` is removed here, was that on purpose?

maaktweluit (commented):
LGTM!

2 small q's will approve on reply or fix

golem/task/taskcomputer.py
Krigpl: I don't think so

golem/task/taskcomputer.py
Krigpl: `has_assigned_task` is broader than `is_computing` (i.e. anytime `is_computing=True => has_assigned_task=True`

golem/task/taskcomputer.py
Krigpl: Actually yes - I changed it so it's properly thread safe

maaktweluit: Looking at it again it is even more like `self.is_computing()` now, why is this thread safe over calling the function?

Krigpl: If you write:
```
if self.is_computing():
  self.counting_thread.check_timeout()
```
then after returning from `is_computing` but before executing the next line other thread can change `self.counting_thread` - because it's not protected by the lock anymore.

maaktweluit: Thank you!

maaktweluit (approved):

golem/task/taskcomputer.py
Krigpl: I actually reverted this change now because linux and win builds were not happy, I dunno, something's wrong here

Krigpl: I reverted the revert. As I suspected the problem was in a different place.

PR #3367, opened 2018-10-01
Fix verification paths to work again on windows

maaktweluit:
- Update tests to install and set env for the DockerManager
- Update task_server and task_manager default path from `res` to `ComputerRes`

### DIFF
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -76,8 +76,8 @@ class TaskManager(TaskEventListener):
         pass
 
     def __init__(
-            self, node_name, node, keys_auth, listen_address="", listen_port=0,
-            root_path="ComputerRes", use_distributed_resources=True,
+            self, node_name, node, keys_auth, root_path,
+            use_distributed_resources=True,
             tasks_dir="tasks", task_persistence=True,
             apps_manager=AppsManager(), finished_cb=None):
         super().__init__()
@@ -96,8 +96,8 @@ class TaskManager(TaskEventListener):
         self.tasks_states: Dict[str, TaskState] = {}
         self.subtask2task_mapping: Dict[str, str] = {}
 
-        self.listen_address = listen_address
-        self.listen_port = listen_port
+        self.listen_address = ""
+        self.listen_port = 0
 
         self.task_persistence = task_persistence
 
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -76,8 +76,8 @@ class TaskManager(TaskEventListener):
         pass
 
     def __init__(
-            self, node_name, node, keys_auth, listen_address="",
-            listen_port=0, root_path="res", use_distributed_resources=True,
+            self, node_name, node, keys_auth, listen_address="", listen_port=0,
+            root_path="ComputerRes", use_distributed_resources=True,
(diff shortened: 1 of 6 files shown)


golem/task/taskmanager.py
```diff
@@ -76,8 +76,8 @@ class AlreadyRestartedError(Error):
         pass
 
     def __init__(
-            self, node_name, node, keys_auth, listen_address="",
-            listen_port=0, root_path="res", use_distributed_resources=True,
+            self, node_name, node, keys_auth, listen_address="", listen_port=0,
+            root_path="ComputerRes", use_distributed_resources=True,
```
mfranciszkiewicz: Maybe `"ComputerRes"` could be a constant?

Wiezzel:
@maaktweluit This is only a fix for the unit tests, right?

golem/task/taskmanager.py
Krigpl: Imo we could just get rid of the default value

maaktweluit:
@Wiezzel no, the second commit also fixes the verification on a real task

golem/task/taskmanager.py
maaktweluit: Haha, i was thinking about both options, could not choose so left as is :D

Will try to remove this default, if it breaks tests a lot i will make it a common CONST somewhere

maaktweluit:
@mfranciszkiewicz @Krigpl resolved your comments, please check again :)

golem/task/taskmanager.py
```diff
@@ -76,8 +76,8 @@ class AlreadyRestartedError(Error):
         pass
 
     def __init__(
-            self, node_name, node, keys_auth, listen_address="",
-            listen_port=0, root_path="res", use_distributed_resources=True,
```
Krigpl, line 80: Since you touched `listen_*` args: looks like they are not useful and they are used only in one place in TaskManager, so you may remove them completely. Either here or in a separate PR, if it'll turn out too big.

golem/testutils.py
```diff
@@ -30,12 +30,15 @@ def setUpClass(cls):
                 # to avoid issues with mounting directories in Docker containers
                 cls.root_dir = os.path.join(get_local_datadir('tests'))
                 os.makedirs(cls.root_dir, exist_ok=True)
+            elif is_windows():
+                import win32api  # noqa pylint: disable=import-error
+                base_dir = get_local_datadir('default')
+                cls.root_dir = os.path.join(base_dir, 'ComputerRes', 'tests')
```
mfranciszkiewicz, line 36: Is this directory also shared with SMB? AFAIK it's `%HOME%\AppData\Local\golem\golem\default\ComputerRes\tests` without the `testnet` or `mainnet` part.

maaktweluit: No, without `testnet` or `mainnet` it will not be shared

but, `cls.root_dir` will include `rinkeby` so this will be.

golem/task/taskmanager.py
maaktweluit: good one! will pick this up in a new PR

Wiezzel (approved):
It works.

mfranciszkiewicz (approved):

PR #3368, opened 2018-10-01
Unburden golem.client as a sole RPC provider

jiivan:
* Introduce `@expose` decorator for remote procedures
* Introduce `golemcli.py debug exposed_procedures` for dynamic procedures mapping
* Remove static procedures mapping file
* Remove unused event listening logic


golem/interface/client/environments.py
```diff
@@ -1,13 +1,13 @@
 from golem.core.deferred import sync_wait
 from golem.environments.minperformancemultiplier import MinPerformanceMultiplier
 from golem.interface.command import group, Argument, command, CommandResult
-from golem.rpc.session import Client
+from golem.rpc.session import ClientProxy
```
etam: https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking

golem/interface/client/terms.py
```diff
@@ -5,7 +5,7 @@
 
 from golem.core.deferred import sync_wait
 from golem.interface.command import group, command
-from golem.rpc.session import Client
+from golem.rpc.session import ClientProxy
```
etam: https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking

golem/node.py
```diff
@@ -395,6 +413,10 @@ def _run(self, *_) -> None:
                 self.client.connect(peer)
         except SystemExit:
             self._reactor.callFromThread(self._reactor.stop)
+            return
+
+        methods = self.client.get_wamp_rpc_mapping()
+        self.rpc_session.add_procedures(methods)  # type: ignore
```
etam, line 434: `# type: ignore` Why?

jiivan: `Optional[Session]` has no attribute `add_procedures`. Mypy not cool? :)

golem/node.py
```diff
@@ -163,7 +169,7 @@ def set_password(self, password: str) -> bool:
             # payments and identity this should be called only when
             # idendity was not just created above for the first time.
             self._ets.backwards_compatibility_privkey(
-                self._keys_auth._private_key,
+                self._keys_auth._private_key,  # noqa pylint: disable=protected-access
```
etam, line 192: Maybe `_private_key` shouldn't be a private variable? After all, it's used everywhere, but not in `KeysAuth`.

jiivan: True, it may be a good candidate for next refactoring PR.

golem/node.py
etam: If you add `if not self.rpc_session:`, like in other methods around, then `Optional` goes away and ignoring type should not be needed.

golem/client.py
```diff
@@ -234,6 +233,19 @@ def __init__(  # noqa pylint: disable=too-many-arguments
...
+        from apps.rendering.task import framerenderingtask
+        providers = (
+            self,
+            framerenderingtask,
+            self.task_server.task_manager,
+            self.environments_manager,
+        )
+        mapping = {}
+        for rpc_provider in providers:
+            mapping.update(rpc_utils.object_method_map(rpc_provider))
```
Elfoniok, line 245: I am not sure if this is a good way to collect exposed methods? basically we scan all methods of providers and check if its has uri attribute?

golem/node.py
```diff
@@ -24,14 +25,28 @@
...
+from golem.rpc.session import (
+    Publisher,
+    Session,
+)
 from golem.terms import TermsOfUse
 
 logger = logging.getLogger(__name__)
 
 
+def require_rpc_session():
```
etam:
```python
from typing import Any, cast, Callable, Optional, TypeVar
from functools import wraps

F = TypeVar('F', bound=Callable[..., Any])


def _require_rpc_session(f: F) -> F:
    @wraps(f)
    def wrapper(self: 'Node', *args, **kwargs):
        if self.rpc_session is None:
            self._error("RPC session is not available")  # noqa pylint: disable=protected-access
            return None
        return f(self, *args, **kwargs)
    return cast(F, wrapper)


class Node:
    ...
    @_require_rpc_session  # no '()' at the end here
    def foo():
```

Not ideal, but I think a bit better.
https://mypy.readthedocs.io/en/latest/generics.html#declaring-decorators

golem/rpc/router.py
```diff
@@ -8,7 +8,7 @@
 from crossbar.common import checkconfig
 from twisted.internet.defer import inlineCallbacks
 
-from golem.rpc.cert import CertificateManager
```
etam, line 11: What's wrong with that?

golem/rpc/session.py
```diff
@@ -183,32 +185,24 @@ def onDisconnect(self):
...
-    def unregister_events(self, event_names):
-        for event_name in event_names:
-            if event_name in self.subs:
-                yield self.subs[event_name].unsubscribe()
-                self.subs.pop(event_name, None)
-            else:
-                logger.error("RPC: Not subscribed to: {}".format(event_name))
+    def exposed_procedures(self):
+        exposed: typing.Dict[str, str] = {}
+        for registration in self._registrations.values():
```
Elfoniok, line 207: why do you need this? Is is allowed to use private fields of base class according to python conventions?

golem/rpc/session.py
```diff
@@ -183,32 +185,24 @@ def onDisconnect(self):
...
+            fn = registration.endpoint.fn
```
Elfoniok, line 208: I am not sure if we don't depend here on WAMP internals? Wouldn't be a problem when updating WAMP version?

Elfoniok (approved):

golem/rpc/session.py
```diff
@@ -53,13 +55,13 @@ def __init__(self, host, port, realm, ssl=True):
...
     # pylint: disable=too-many-arguments
-    def __init__(self, address, methods=None, events=None,
+    def __init__(self, address, mapping=None,
                  cert_manager=None, use_ipv6=False,
                  crsb_user=None, crsb_user_secret=None) -> None:
         self.address = address
-        self.methods = methods or []
-        self.events = events or []
-        self.subs = {}  # type: ignore
+        if mapping is None:
```
etam, line 68: Maybe just use `mapping={}` default value in arguments?

jiivan: Nope. Try this:
```python
def f(a=[]):
    a.append(1)
    print(a)

for x in range(100):
    f()
```

jiivan: http://www.effbot.org/zone/default-values.htm

golem/rpc/session.py
```diff
@@ -218,22 +212,47 @@ def is_closing(self):
...
-class Client(object):
-
-    def __init__(self, session, method_map, timeout=2):
+class ClientProxy():  # pylint: disable=too-few-public-methods
+    PREFIXES: typing.Tuple[str, ...] = (
+        'golem.client.Client.',
+        'golem.node.Node.',
+    )
 
+    def __init__(self, session, timeout=2):
```
etam: Can I haz sum type annotations? (for variables below too, especially `_mapping`)

golem/rpc/session.py
```diff
@@ -218,22 +212,47 @@ def is_closing(self):
...
     def _on_error(err):
-        logger.error("RPC: Session error: {}".format(err))
+        logger.error("RPC: Session error: %r", err)
         return err
 
 
-class Client(object):
-
-    def __init__(self, session, method_map, timeout=2):
+class ClientProxy():  # pylint: disable=too-few-public-methods
```
etam:
```python
class ClientProxy:
```

golem/rpc/session.py
```diff
@@ -218,22 +212,47 @@ def is_closing(self):
...
+            self._ready = Deferred()
+            return
+        txdeferred = self._call('sys.exposed_procedures')
+        txdeferred.addCallback(self._init_mapping)
+        txdeferred.addErrback(logger.error)
+        self._ready = txdeferred
+
+    def __getattr__(self, name):
+        if name.startswith('_'):
+            return super().__getattr__(name)  # pylint: disable=no-member
```
etam, line 245: `super()`?

golem/rpc/session.py
```diff
@@ -249,26 +268,18 @@ def _call(self, method_alias, *args, **kwargs):
...
     def _on_error(self, err):
         if not self._session.is_closing():
-            logger.error("RPC: call error: {}".format(err))
+            logger.error("RPC: call error: %r", err)
             return err
 
 
-class Publisher(object):
-
+class Publisher():  # pylint: disable=too-few-public-methods
```
etam:
```python
class Publisher:
```

golem/rpc/utils.py
```diff
@@ -0,0 +1,41 @@
+import inspect
+# SEE autobahn.twisted.wamp.Application.register
+
+
+def expose(uri=None):  # pylint: disable=unused-argument
+    def wrapper(f):
+        nonlocal uri
```
etam, line 8: Is this `nonlocal` necessary? `uri` is not used outside of `wrapper` scope.

golem/rpc/utils.py
```diff
@@ -0,0 +1,41 @@
+import inspect
+# SEE autobahn.twisted.wamp.Application.register
+
+
+def expose(uri=None):  # pylint: disable=unused-argument
+    def wrapper(f):
```
etam, line 7: https://docs.python.org/3/library/functools.html#functools.wraps

golem/rpc/utils.py
```diff
@@ -0,0 +1,41 @@
...
+        if isinstance(f, (staticmethod, classmethod)):
+            f.__func__.rpc_uri = uri
+        else:
+            f.rpc_uri = uri
+        return f
+    return wrapper
+
+
+def object_method_map(instance):
+    mapping = {}
```
etam: type annotation?

golem/node.py
```diff
@@ -395,6 +419,10 @@ def _run(self, *_) -> None:
                 self.client.connect(peer)
         except SystemExit:
             self._reactor.callFromThread(self._reactor.stop)
+            return
+
+        methods = self.client.get_wamp_rpc_mapping()
+        self.rpc_session.add_procedures(methods)  # type: ignore
```
etam, line 434: There's a short time between when RPC session is ready and procedures are registered. If in that time a client connects and calls something, it will get errors.

Maybe we should have a method (registered very early, returning bool) and send notification after methods are registered, so that a client could know when golem is ready?

etam: that would also require some changes from electron side

etam (changes requested):
I like the idea! But I left some notes around...

golem/client.py
jiivan: Yes, it's done once when `Node` initializes `Client`. I'm open any improvement suggestions to this one. haven't come with anything better.

golem/rpc/session.py
jiivan: I need this, because base class doesn't implement `exposed_procedures()` method and I really want it.

golem/rpc/session.py
jiivan: I get your point, but I see no better solution.

golem/rpc/session.py
jiivan: yes, object.

golem/rpc/utils.py
jiivan: For linters only.

golem/rpc/utils.py
jiivan: It would be usefull fo `inner` if defined. This function actually returns `f`.

golem/rpc/router.py
jiivan: It was done as part of some refactoring which I don't remember, but IMHO it's always better to use namespaces. Helps to minimize local namespace pollution.

golem/node.py
jiivan: IMHO it's really ugly, and... it doesn't work. :/ I don't want to skip one level of function nesting as it removes ability of accepting parameters of this decorator. And I really don't want to try to overcome mypy limits here.

golem/node.py
jiivan: AFAIK electron has some grace time when golem starts to wait for registering procedures. We could minimize the time gap of this race-condition, but I don't think it's worth the effort.

etam (approved):

PR #3372, opened 2018-10-01
Marketplace: track provider idle time

mfranciszkiewicz:


golem/task/timer.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def comp_finished(self) -> None:
+        """ Updates the state to keep track of idle time and increases
+            the accumulated computation time.
+        """
+
+        if self._last_comp_finished is None:
+            self._last_comp_finished = time.time()
+        else:
+            logger.debug("Computation is not running")
```
maaktweluit: maybe this can be a warn or error, to be consistent with the `comp_started()` error.

golem/task/timer.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def time_idle(self) -> float:
+        """ Returns the total idle time. If not computing, returns the
+            accumulated value enlarged by the time since last computation.
+        """
+        if self._last_comp_finished is None:
+            return self._time_idle
+        return self._time_idle + time.time() - self._last_comp_finished
+
+    def comp_started(self) -> None:
```
maaktweluit, line 23: unused function?

maaktweluit (commented):
LGTM!

The new class could use some debug statements at the start of the functions, like `logger.debug('comp_finished()')`

Left some small comments, will approve on reply.

golem/task/taskcomputer.py
```diff
@@ -442,6 +446,7 @@ def __compute_task(self, subtask_id, docker_images,
             if self.finished_cb:
                 self.finished_cb()
 
+            ProviderIdleTimer.comp_finished()
```
Krigpl: I have a feeling that `ProviderIdleTimer.comp_finished()` and `self.finished_cb()` should always be called at the same time (which is not the case e.g. in `task_resource_failure`). Maybe a helper function which is always called when task is completed would be nice.

Krigpl (commented):
Somewhat arbitrary but I'm fine with merging it directly to `develop`. This change isn't breaking or anything but can only introduce merge conflicts when it's time to merge the parent branch.

golem/task/timer.py
mfranciszkiewicz: The idea behind it was that you can safely call `comp_finished` as many times as you want and won't be spammed with log messages. I'll change it and hopefully that won't be the case.

golem/task/timer.py
mfranciszkiewicz: Whoops, the call was not committed

golem/task/timer.py
maaktweluit: can agree with that

but it would be better to call it exactly enough

Krigpl:
Disregard my previous comment about merging to develop - I guess there'll be fewer conflicts while keeping everything in the feature branch.

One question though: what exactly is considered idle time here? I'm thinking that anytime that `has_assigned_task = False` - is that right?

mfranciszkiewicz:
Idle is the time **not** spent on downloading resources, computing the task and packaging the result.

golem/task/taskcomputer.py
```diff
@@ -391,16 +406,21 @@ def __compute_task(self, subtask_id, docker_images,
...
+            self.__task_finished()
             return
 
         with self.lock:
             self.counting_thread = tt
 
         tt.start().addBoth(lambda _: self.task_computed(tt))
 
+    def __task_finished(self):
+        self.counting_task = None
```
maaktweluit: did you mean `counting_thread`?

should this use `self.lock` then as well?

golem/task/taskcomputer.py
mfranciszkiewicz: Unfortunately, that is not the case. We want to track the time between receiving `ctd` and sending `ReportComputedTask`.

golem/task/taskcomputer.py
mfranciszkiewicz: Yes, these changes were merged in the meantime.

maaktweluit (approved):

golem/task/taskcomputer.py
```diff
@@ -391,16 +402,23 @@ def __compute_task(self, subtask_id, docker_images,
...
+            self.__task_finished()
             return
 
         with self.lock:
             self.counting_thread = tt
 
         tt.start().addBoth(lambda _: self.task_computed(tt))
 
+    def __task_finished(self):
+        ProviderIdleTimer.comp_finished()
```
Krigpl, line 383: This is called before packaging the results, right? But you mentioned we wanna include that in the not idle time.

mfranciszkiewicz: Actually, it's called after `self.task_server.send_results` which is responsible for packaging the results

Krigpl: Neat. I like the fact we call `ProviderIdleTimer` only from this class.
And not I feel like the timer is off iff `has_assigned_task=True` after all. And that would make sense. Might be worth keeping it in mind when doing some clean up (e.g. to change `self.assigned_subtasks` from dict to `Optional[str]`).

Krigpl (approved):
lgtm

Krigpl:
How about persistance? isn't it needed?

mfranciszkiewicz:
Not really, the market document only states a need for the idle time since last computation.

Krigpl (dismissed):
Please fix end line characters in `tasksession.py`

mfranciszkiewicz:
@Krigpl Will do, I guess my IDE did not know better when resolving conflicts.

PR #3374, opened 2018-10-02
Handle any errors on accepting results

Elfoniok:
Ensure that confirmation is sent back

### DIFF
--- a/apps/core/task/coretask.py
+++ b/apps/core/task/coretask.py
@@ -237,11 +237,14 @@ class CoreTask(Task):
         return []
 
     def verification_finished(self, subtask_id, verdict, result):
-        if verdict == SubtaskVerificationState.VERIFIED:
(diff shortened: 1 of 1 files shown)


apps/core/task/coretask.py
```diff
@@ -217,7 +217,10 @@ def computation_finished(self, subtask_id, task_result,
         result_files = self.results.get(subtask_id)
 
         def verification_finished_(subtask_id, verdict, result):
-            self.verification_finished(subtask_id, verdict, result)
+            try:
+                self.verification_finished(subtask_id, verdict, result)
+            except Exception as exc:
+                logger.warning("Failed during accepting results %s", exc)
```
mfranciszkiewicz: If the issue lies with accepting results, maybe this block should wrap the `self.accept_results` call instead?

mfranciszkiewicz (approved):

PR #3375, opened 2018-10-02
Handle run_benchmark RPC errors

mfranciszkiewicz:
GUI tries to run benchmarks for `BLENDER_NVGPU` even when it's not available. The following benchmarks won't be run.

This PR catches and logs all exceptions.

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -1415,11 +1415,15 @@ class Client(HardwarePresetsMixin):
     @inlineCallbacks
     def run_benchmark(self, env_id):
         deferred = Deferred()
+        result = 0
 
-        self.task_server.benchmark_manager.run_benchmark_for_env_id(
-            env_id, deferred.callback, deferred.errback)
+        try:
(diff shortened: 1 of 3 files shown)


Krigpl:
I somewhat prefer explicit exception in cases like this - after all you're not supposed to run a benchmark on an env that you don't support.
Why not just fix electron instead?

mfranciszkiewicz:
It's the same with CLI when explicitly running an unsupported benchmark. The problem is that the RPC connection dies.

tworec:
I'd rather check if env is supported and log error instead of trowing exception

tworec:
sorry, misclick

mfranciszkiewicz:
@Krigpl @tworec I've reverted the changes.
Electron is being fixed, independently. CLI now catches benchmark errors. Clients are responsible for proper error handling, otherwise the connection will become unusable. This goes for all of the RPC methods.

PR #3384, opened 2018-10-03
Marketplace: requestor tracks providers' computation time

mfranciszkiewicz:
Note: `ComputeTimers.remove` will be called after updating `R` (as specified in the doc).


maaktweluit (approved):
LGTM!

The `ComputeTimers` could have the same logs ( warn / debug ) as the other PR.

PR #3386, opened 2018-10-04
Timeout restart regression test

### LINKED ISSUE #3293: Add a regression test to test the correct restart of a task in case of a subtask timeout

shadeofblue:
+ `concent_integration_test` refactor

closes: #3293


scripts/concent_integration_tests/tests/playbooks/base.py
```diff
@@ -281,28 +308,23 @@ def run(self):
...
-        )
+        if task_package:
+            self.task_package = task_package
 
-        self.provider_output_queue = helpers.get_output_queue(
-            self.provider_node)
-        self.requestor_output_queue = helpers.get_output_queue(
-            self.requestor_node)
+        for attr, val in kwargs.items():
+            setattr(self, attr, val)
```
etam, line 320: Is this used anywhere? Looks quite dangerous...

etam (approved):
ok

I added one question (if it's reasonable, you can remove some code). Apart from that I think it's ok.

PR #3388, opened 2018-10-04
Exact scene file

### LINKED ISSUE #3327: Add a way to specify the exact scene file to use through RPC / CLI
Currently Golem uses some mechanism to select which possible `.blend` file to use when given a directory containing more than one `.blend` file - we need to provide a way to specify exactly which file should be the main scene file

etam:
fixes #3327

### DIFF
--- a/apps/rendering/task/renderingtask.py
+++ b/apps/rendering/task/renderingtask.py
@@ -251,7 +251,7 @@ class RenderingTask(CoreTask):
         return path.replace("\\", "/")
 
 
-class RenderingTaskBuilderException(Exception):
+class RenderingTaskBuilderError(Exception):
     pass
 
 
@@ -278,7 +278,7 @@ class RenderingTaskBuilder(CoreTaskBuilder):
         candidates = [res for res in resources if any(res.lower().endswith(ext.lower())
                                             for ext in extensions)]
         if not candidates:
-            raise RenderingTaskBuilderException("Scene file was not found.")
+            raise RenderingTaskBuilderError("Scene file was not found.")
 
         candidates.sort(key=len)
         return candidates[0]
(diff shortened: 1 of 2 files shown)


apps/rendering/task/renderingtask.py
```diff
@@ -251,6 +251,10 @@ def __get_path_windows(self, path):
         return path.replace("\\", "/")
 
 
+class RenderingTaskBuilderException(Exception):
```
shadeofblue: `RenderingTaskBuilderError` perhaps?

apps/rendering/task/renderingtask.py
```diff
@@ -274,11 +278,22 @@ def _scene_file(type, resources):
...
-            raise Exception("Scene file was not found.")
+            raise RenderingTaskBuilderException("Scene file was not found.")
 
         candidates.sort(key=len)
         return candidates[0]
 
+    @staticmethod
+    def _scene_file_with_key(resources: Iterable[str], key: str) -> str:
+        main_scene_file_candidates = [res for res in resources
+                                      if res.endswith(key)]
```
shadeofblue: why are we looking for partial matches instead of just verifying equality? plus, why would there ever be more than one candidate? shouldn't we just check for exact equality and only ever find one match or none?

tests/apps/rendering/task/test_renderingtask.py
```diff
@@ -367,72 +367,141 @@ def __init__(self, default_subtasks=13, min_subtasks=3,
...
-               path.join("/dir3/dir4", "MY task.txt")
+    def test_main_scene_file(self):
+        # given
+        self.task_dict['resources'] = {
+            "/path/to/file1.png",
+            "/path/to/file2_longer_name.txt",
+            "/path/to/file3.jpg",
+            "/path/to/file4.txt",
+        }
+        self.task_dict['main_scene_file'] = "file4.txt"
```
shadeofblue: again, I'm unsure why we're specifying a partial path -> I think it should be exactly the same path as is specified in `resources`

apps/rendering/task/renderingtask.py
```diff
@@ -303,7 +318,10 @@ def build_minimal_definition(cls, task_type, dictionary):
         resources = dictionary['resources']
 
         definition = parent.build_minimal_definition(task_type, dictionary)
-        definition.main_scene_file = cls._scene_file(task_type, resources)
+        definition.main_scene_file = \
+            cls._scene_file_with_key(resources, dictionary['main_scene_file']) \
```
shadeofblue: I would just add an optional `main_scene_file` parameter to `_scene_file` and do the check inside that function instead of performing this conditional assignement... but that's up to taste...

apps/rendering/task/renderingtask.py
etam: I thought it might be easier to copy just a filename, if you know it's unique and have it already in your directory browser.

etam: Ok, I removed partial matching.

shadeofblue (approved):

PR #3390, opened 2018-10-04
Offer buffering

Krigpl:
This is my proposal for offer buffering.

We freeze in `react_to_want_to_compute_task` and let marketplace module do its magic.
And marketplace module sorts the offers from best to worst and resumes `react_to_want_to_compute_task` in order - meaning that best offers will receive subtasks and when we assign all available subtasks the rest of providers will receive `CannotAssignTask.NoMoreSubtasks`.


golem/marketplace/offerpool.py
```diff
@@ -0,0 +1,39 @@
...
+
+class Offer:
+    def __init__(self, msg: Any, deferred: Deferred) -> None:
+        self.msg = msg
+        self.deferred = deferred
+
+
+class OfferPool:
+
+    _INTERVAL: ClassVar[float] = 30.0  # s
```
maaktweluit: 30 seconds interval sounds a bit long, on mainnet you get a lot of offers in the first few seconds

golem/task/tasksession.py
```diff
@@ -507,61 +507,68 @@ def _cannot_assign(reason):
...
+            )
+            history.add(
+                msg=ttc,
+                node_id=self.key_id,
+                local_role=Actor.Requestor,
+                remote_role=Actor.Provider,
+            )
+            self.send(ttc)
+
+        from golem.marketplace.offerpool import OfferPool
```
maaktweluit: any reason why the import is not at the top of the file?

maaktweluit (approved):
LGTM!

Some small q's, not breaking so approved!

golem/marketplace/offerpool.py
Krigpl: Hm, maybe, the doc proposes 30 s as default 
https://github.com/golemfactory/golem-model/blob/master/market/doc/OfferBufferingProposal.md

golem/task/tasksession.py
Krigpl: Yeah, no, this was just quick RFC so I still need to do cleanup and add tests :)

maaktweluit: sounds good, let me know when to check again!

golem/marketplace/offerpool.py
```diff
@@ -0,0 +1,39 @@
...
+        if task_id not in cls._pools:
+            cls._pools[task_id] = []
+            from twisted.internet import reactor
+            task.deferLater(reactor, cls._INTERVAL, cls._choose_offers, task_id)
+        deferred = Deferred()
+        cls._pools[task_id].append(Offer(msg, deferred))
+        return deferred
+
+    @classmethod
+    def _choose_offers(cls, task_id: str) -> None:
```
mfranciszkiewicz: How many?

golem/task/tasksession.py
```diff
@@ -507,61 +507,68 @@ def _cannot_assign(reason):
...
+        OfferPool.add(msg.task_id, msg).addCallback(_offer_chosen)
```
mfranciszkiewicz, line 583: Providers may disconnect in the meantime

Krigpl: Can't we make the assumption that they won't disconnect? And if they do they are not taken into consideration?

mfranciszkiewicz: They shouldn't if they want to compute a task. But if they do (for whatever reason) and we assign a task to them, `self.send(ttc)` will fail and we'll be basically waiting for the subtask to time out.

mfranciszkiewicz: A simple sanity check for the current session should do the trick

Krigpl: Good point.
Although it isn't marketplace related, marketplace only increases chances of it happening :)
In general we could(/should) check whether `self.send(ttc)` succeeded and react on it. Otherwise it's basically the same as provider getting the task but not computing it - we'll wait for the timeout and decrease reputation.

Krigpl:
@maaktweluit @mfranciszkiewicz This is ready for review now.

golem/task/tasksession.py
```diff
@@ -507,61 +508,71 @@ def _cannot_assign(reason):
                            msg.task_id, node_name_id)
             return
 
-        # TODO: Queue requests here
+        def _offer_chosen(is_chosen: bool) -> None:
+            if not is_chosen:
+                logger.info("Provider not chosen by marketplace")
```
maaktweluit: this message could include `...provider=%r', node_name_id) `

maaktweluit (approved):
LGTM!

Only some log improvements possible. Think the "error" cases should include the ID/name for ease of debugging

mfranciszkiewicz (approved):

PR #3392, opened 2018-10-05
Fix fundslocker to use Task API properly

mplebanski:
This fixes fundslocker to use appropriate API method from Task class instead
derived class). Task base class does not have the `total_tasks`


mplebanski:
Now I see that there are more cases like this.

PR #3396, opened 2018-10-05
Names refactor

mplebanski:
It's an important step to simplify current task API. I removed mappings that caused different identifiers point to the same data/objects. 

This change requires `golem-electron` to use `subtasks_count` instead of `subtasks` when creating/getting a task through golem RPC. @mdtanrikulu


mplebanski:
I resolved the conflicts that appeared because it was hanging for over a week.

golem/client.py
```diff
@@ -1,1900 +1,1900 @@
-# pylint: disable=too-many-lines
```
jiivan: What happened here? Please revert this change.

jiivan (changes requested):
Full body of `golem/client.py` changed...

golem/client.py
etam: Looks like change of end-of-line encoding.

@mplebanski I know we have some files with `\r\n` and some with `\n`, but we have to stick with what we have.

mplebanski:
@jiivan I fixed the line endings problem. Weird.

golem/client.py
mplebanski: This happened because I used github web gui to merge the golem/client.py into feature branch

etam:
@mplebanski when merging this PR (not now, obviously) please use the option to squash all commits.

mplebanski:
Can we give it a go? It's easy for this refactor to become a conflict (many files affected).

Wiezzel (commented):
This seems to affect RPC (`comp.task.create`). Do we have coordinated interface changes?

mplebanski:
I only plan to do this regarding Electron client (@mdtanrikulu). Is there anything else I should take into consideration?

jiivan (approved):

jiivan:
> I only plan to do this regarding Electron client (@mdtanrikulu). Is there anything else I should take into consideration?

Yes, have a look at *golemcli*

mplebanski:
Closing due to too many conflicts

PR #3397, opened 2018-10-08
Decoupling crop generation

Elfoniok:


jiivan (approved):
I really don't like name mangling (double underscore). I'm yet to find valid use-case for them in python. Besides that - looks good.

Krigpl:
Doesn't apply anymore

Issue #3300, opened 2018-09-12
AttributeError: 'RenderingTaskDefinition' object has no attribute 'concent_enabled'


ederenn:
```
CRITICAL [golem.client                       ] Can't start network. Giving up.
Traceback (most recent call last):
  File "/Users/ederenn/projects/golem/golem/client.py", line 293, in start
    self.start_network()
  File "/Users/ederenn/projects/golem/golem/client.py", line 381, in start_network
    self.clean_old_tasks()
  File "/Users/ederenn/projects/golem/golem/client.py", line 1350, in clean_old_tasks
    for task in self.get_tasks():
  File "/Users/ederenn/projects/golem/golem/client.py", line 1062, in get_tasks
    return list(filter(None, tasks))
  File "/Users/ederenn/projects/golem/golem/client.py", line 1060, in <genexpr>
    tasks = (self.get_task(task_id) for task_id in task_ids)
  File "/Users/ederenn/projects/golem/golem/client.py", line 1022, in get_task
    task_dict = self.task_server.task_manager.get_task_dict(task_id)
  File "/Users/ederenn/projects/golem/golem/task/taskmanager.py", line 984, in get_task_dict
    self.get_task_definition_dict(task))
  File "/Users/ederenn/projects/golem/golem/task/taskmanager.py", line 155, in get_task_definition_dict
    return task_type.task_builder_type.build_dictionary(definition)
  File "/Users/ederenn/projects/golem/apps/blender/task/blenderrendertask.py", line 662, in build_dictionary
    dictionary = parent.build_dictionary(definition)
  File "/Users/ederenn/projects/golem/apps/rendering/task/framerenderingtask.py", line 515, in build_dictionary
    dictionary = parent.build_dictionary(definition)
  File "/Users/ederenn/projects/golem/apps/rendering/task/renderingtask.py", line 295, in build_dictionary
    dictionary = parent.build_dictionary(definition)
  File "/Users/ederenn/projects/golem/apps/core/task/coretask.py", line 618, in build_dictionary
    return definition.to_dict()
  File "/Users/ederenn/projects/golem/apps/core/task/coretaskstate.py", line 122, in to_dict
    'concent_enabled': self.concent_enabled
AttributeError: 'RenderingTaskDefinition' object has no attribute 'concent_enabled'
INFO     [golem.client                       ] Shutting down ...
INFO     [golem.client                       ] Stopping network ...
INFO     [golem.network.concent.client       ] Waiting for received messages queue to empty
[...]

Issue #3405, opened 2018-10-09
Enable migrations for tasks persistence


jiivan:
Currently tasks persistence is achieved by pickling *TaskDefinition* instances. When attribute is added to *TaskDefinition* it requires migrating it from old pickles (see: #3300).

One way is to use database models for task persistence. *TaskDefinition* should be a model in our database. We have database migration mechanism in place.

Other solution is suggested in pickle documentation. Use `golem.__version__` in `__getstate__` and write migrations in `__setstate__`.

etam:
related: #2768

PR #3406, opened 2018-10-09
Migrate task definition pickles from 0.17.1

jiivan:
fixes: #3300 

Migrate task definitions pickled in *0.17.1*. It's ad hoc solution for release only. Full solution suggested in #3405

### DIFF
--- a/apps/core/task/coretaskstate.py
+++ b/apps/core/task/coretaskstate.py
@@ -133,9 +133,10 @@ class TaskDefinition(object):
         subtask_timeout = timeout_to_string(self.subtask_timeout)
         output_path = self.build_output_path()
 
-        result = {
+        return {
             'id': self.task_id,
             'type': self.task_type,
+            'compute_on': self.compute_on,
             'name': self.task_name,
             'timeout': task_timeout,
             'subtask_timeout': subtask_timeout,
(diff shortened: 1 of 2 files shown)


etam (changes requested):
This is not a migration. This is asking for forgiveness. If you look at `TaskManager.restore_tasks` there is already one migration. You should try to go this way.

apps/core/task/coretaskstate.py
```diff
@@ -54,6 +56,30 @@ def __init__(self):
 
         self.concent_enabled: bool = False
 
+    def __getstate__(self):
+        return golem.__version__, self.__dict__
+
+    def __setstate__(self, state):
```
etam, line 62: I like that even more than a migration, that I wrote in TaskManager. :+1:

apps/core/task/coretaskstate.py
```diff
@@ -121,8 +145,10 @@ def to_dict(self) -> dict:
             'options': {
                 'output_path': output_path
             },
-            'concent_enabled': self.concent_enabled
+            'compute_on': self.compute_on,
+            'concent_enabled': self.concent_enabled,
         }
+        return result
```
etam: this change is not needed anymore

etam (approved):
ok

PR #3416, opened 2018-10-10
downgrade log severity of the "faucet error" from `error` to `warning`

shadeofblue:


Krigpl (approved):
thisisgine

This could be `info` even imo

PR #3424, opened 2018-10-11
Remove unnecessary short_desc

Krigpl:
This isn't used anywhere.
Also imo short description should be associated with environment, not with a particular task instance making this redundant anyway.


maaktweluit (changes requested):
LGTM!

A search in the repo still shows quite some occurances of `short_desc` are you sure all task related ones are cleaned up now?

Krigpl:
@maaktweluit Yeah, I'm not removing `short_desc` completely (that'd include removing it from ComputeTaskDefinition), this is just that `TaskThread` doesn't use it at all, so this is obviously useless.

maaktweluit (approved):

PR #3426, opened 2018-10-11
Refactor assigned_subtasks

Krigpl:
`assigned_subtasks` was a map holding at most one value, so the refactor is self explanatory.
Golem doesn't support multiple computations at the same time.


golem/monitor/model/taskcomputersnapshotmodel.py
```diff
@@ -10,4 +10,6 @@ def __init__(self, meta_data, task_computer):
         self.counting_task = task_computer.counting_task
         self.task_requested = task_computer.task_requested
         self.compute_task = task_computer.compute_tasks
-        self.assigned_subtasks = list(task_computer.assigned_subtasks.keys())
+        self.assigned_subtask = ''
+        if task_computer.assigned_subtask:
+            task_computer.assigned_subtask['subtask_id']
```
maaktweluit: what python magic is this? it looks like there is an assignment missing ( or knowledge on my side )

golem/task/taskcomputer.py
```diff
@@ -103,55 +102,52 @@ def __init__(self, task_server: 'TaskServer', use_docker_manager=True,
...
-                    subtask_id,
-                    subtask['docker_images'],
-                    subtask['src_code'],
-                    subtask['extra_data'],
-                    subtask['short_description'],
-                    subtask['deadline'])
-                return True
+        subtask = self.assigned_subtask
+        if not subtask or subtask['task_id'] != task_id:
+            logger.error("Resource colledted for a wrong task, %s", task_id)
```
maaktweluit: typo, `colledted  => collected`

tests/golem/monitor/model/test_taskcomputersnapshotmodel.py
```diff
@@ -13,7 +13,7 @@ def test_channel(self):
         computer_mock.counting_task = counting_task = random.random() > 0.5
         computer_mock.task_requested = task_requested = random.random() > 0.5
         computer_mock.compute_tasks = compute_tasks = random.random() > 0.5
-        computer_mock.assigned_subtasks = assigned_subtasks = dict((x, None) for x in range(100))
+        computer_mock.assigned_subtask = assigned_subtask = 'test_subtask_id'
```
maaktweluit: mock is not matching type, not sure if that matters: `Optional[Dict[str, Any]]`

maaktweluit (changes requested):
LGTM!

Some small comments, will approve on reply!

golem/monitor/model/taskcomputersnapshotmodel.py
Krigpl: Good catch, my mistake

golem/task/taskcomputer.py
```diff
@@ -90,8 +90,7 @@ def __init__(self, task_server: 'TaskServer', use_docker_manager=True,
 
         self.stats = IntStatsKeeper(CompStats)
 
-        self.assigned_subtasks = {}
-        self.task_to_subtask_mapping = {}
+        self.assigned_subtask: Dict[str, Any] = {}
```
Wiezzel: I'd say that empty value should represented as `None` instead of `{}`.

golem/task/taskcomputer.py
```diff
@@ -169,7 +164,8 @@ def task_computed(self, task_thread: TaskThread) -> None:
         work_wall_clock_time = task_thread.end_time - task_thread.start_time
         subtask_id = task_thread.subtask_id
         try:
-            subtask = self.assigned_subtasks.pop(subtask_id)  # ComputeTaskDef
+            subtask = self.assigned_subtask
+            self.assigned_subtask = {}
```
Wiezzel: See prev. comment.

golem/task/taskcomputer.py
```diff
@@ -402,7 +398,8 @@ def __compute_task(self, subtask_id, docker_images,
                               task_timeout)
         else:
             logger.error("Cannot run PyTaskThread in this version")
-            subtask = self.assigned_subtasks.pop(subtask_id)
+            subtask = self.assigned_subtask
+            self.assigned_subtask = {}
```
Wiezzel: See prev. comment.

golem/task/taskcomputer.py
Krigpl: I started with that. But then it's easier to do `self.assigned_subtask.get('task_id') == task_id` when it's always a map. I can revert back to it.

Wiezzel: Maybe add something like this?
```python
@property
def asssigned_task_id(self) -> Optional[str]:
    return self.assigned_subtask.get('task_id') if self.assigned_subtask else None
```
And then just use `self.assigned_task_id == task_id`.

Wiezzel (approved):

maaktweluit (approved):

PR #3428, opened 2018-10-11
Accept SRA without `comp_task_keeper` record

etam:


golem/task/tasksession.py
```diff
@@ -128,30 +134,27 @@ def interpret(self, msg):
...
     #######################
 
     @property
-    def my_private_key(self):
-        if self.task_server is None:
-            logger.error("Task Server is None, can't sign a message.")
-            return None
-        return self.task_server.keys_auth.ecc.raw_privkey
+    def my_private_key(self) -> bytes:
+        return self.task_server.keys_auth._private_key  # noqa pylint: disable=protected-access
```
shadeofblue, line 146: @etam I'm unsure I understand the removal of the check here... especially since you don't seem to be checking for the possible `AttributeError` that may result from uninitialized `task_server` in any places that use this property...

golem/task/tasksession.py
```diff
@@ -751,7 +754,13 @@ def _react_to_get_resource(self, msg):
         ))
 
     @history.provider_history
-    def _react_to_subtask_result_accepted(self, msg):
+    def _react_to_subtask_result_accepted(
+            self, msg: message.tasks.SubtaskResultsAccepted):
+        if self.key_id is None:
+            logger.error("received SubtaskResultsAccepted, but I don't know "
+                         "from who")
```
shadeofblue: whoM ...

golem/task/tasksession.py
```diff
@@ -751,7 +754,13 @@ def _react_to_get_resource(self, msg):
         ))
 
     @history.provider_history
-    def _react_to_subtask_result_accepted(self, msg):
+    def _react_to_subtask_result_accepted(
+            self, msg: message.tasks.SubtaskResultsAccepted):
+        if self.key_id is None:
```
shadeofblue: on the other hand, can `self.key_id` be `None` in within a correctly initialized `TaskSession` ? isn't a session always initialized with the other party's node id?

golem/task/tasksession.py
etam: I removed the check, because it's not possible for `self.task_server` to be `None`. It was in the past, but it's not anymore. But those checks were left, so I removed them.

golem/task/tasksession.py
etam: done

golem/task/tasksession.py
etam: `self.key_id` is set in `_react_to_hello`.

https://github.com/golemfactory/golem/blob/158d31afecd59ff850d507a819aa11ea4edd4e2c/golem/task/tasksession.py#L879

golem/task/tasksession.py
```diff
@@ -751,7 +754,13 @@ def _react_to_get_resource(self, msg):
         ))
 
     @history.provider_history
-    def _react_to_subtask_result_accepted(self, msg):
+    def _react_to_subtask_result_accepted(
+            self, msg: message.tasks.SubtaskResultsAccepted):
+        if self.key_id is None:
+            logger.error("received SubtaskResultsAccepted, but I don't know "
```
shadeofblue: @etam in that case, I think this message it not really useful... it should rather say e.g.

`SubtaskResultsAccepted` received within an improperly intialized `TaskSession` - `key_id` is empty.

or something like that ... otherwise, it's completely unclear what the meaning is without looking at the specific line of code here...

etam: My idea was, in higher level of abstraction, that if someone connects and don't send `hello` message, then golem doesn't know who he is.

golem/task/tasksession.py
Krigpl: I believe `self.key_id` will always be set - `SRA` message has to be verified, before reacting to it, and verification uses `self.key_id`. So this `if` won't happen. Similar for other messages in general.

golem/task/tasksession.py
```diff
@@ -761,9 +770,28 @@ def _react_to_subtask_result_accepted(self, msg):
...
+            )
+        except (msg_exceptions.InvalidSignature,
+                msg_exceptions.OwnershipMismatch) as e:
+            logger.error("SubtaskResultAccepted has %s: %s",
+                         e.__class__.__name__, e)
+            self.dropped()
+            return
+
+        transaction_system = self.task_server.client.transaction_system
+        if not transaction_system.is_income_expected(
```
shadeofblue: @etam @Krigpl correct me if I'm wrong but I don't think we can completely eliminate the `check_requestor_for_subtask` and just replace it with `is_income_expected` since we specifically modified Golem _not_ to generate the `Income` record before `SRA` is issued... so this is mostly a fix for _older_ awaiting transactions for which the `comp_task_keeper` record has already been removed ... but for the currently processed subtasks, we should still rely on that check and not on existence of the expected payment... Am I right, @Krigpl ?

golem/task/tasksession.py
shadeofblue: yep, but that requires extensive knowledge of `TaskSession`s internal functioning... and won't tell anything to a less informed user...

golem/task/tasksession.py
```diff
@@ -761,9 +770,28 @@ def _react_to_subtask_result_accepted(self, msg):
...
+                subtask_id=msg.subtask_id,
+                payer_address=pubkeytoaddr(self.key_id)
+        ):
+            logger.error("Unexpected income from %r for subtask %r",
```
Krigpl: I dunno if it's an error - provider can't prevent from receiving these messages and can't do anything about it. Malicious requestor shouldn't generate error messages on the provider side.

golem/task/tasksession.py
```diff
@@ -761,9 +770,28 @@ def _react_to_subtask_result_accepted(self, msg):
...
+                msg_exceptions.OwnershipMismatch) as e:
+            logger.error("SubtaskResultAccepted has %s: %s",
+                         e.__class__.__name__, e)
+            self.dropped()
+            return
+
+        transaction_system = self.task_server.client.transaction_system
+        if not transaction_system.is_income_expected(
+                subtask_id=msg.subtask_id,
+                payer_address=pubkeytoaddr(self.key_id)
```
Krigpl: I think `SRA` has `task_to_compute` property which has `requestor_ethereum_address` property which should be used here.

golem/ethereum/incomeskeeper.py
```diff
@@ -96,6 +96,17 @@ def expect(
             accepted_ts=accepted_ts,
         )
 
+    @staticmethod
+    def is_expected(
+            subtask_id: str,
+            payer_address: str,
+    ) -> bool:
+        return Income.select().where(
```
Krigpl, line 104: fwiw, I find
```
.where(
  cond1,
  cond2,
)
```
more legible than spamming with parentheses.

Krigpl (changes requested):

golem/task/tasksession.py
Krigpl: I think so, yes. Imo the `is_income_expected` check should live inside `check_requestor_for_subtask`'s if body.

golem/task/tasksession.py
etam: Actually, a "less informed user" wouldn't be much more informed by a message "key_id uninitialized".

golem/task/tasksession.py
etam: I had to somehow silence `mypy` complaining about `self.key_id` being `Optional[str]`.
Should I put `assert self.key_id is not None` then?

Krigpl: If you apply my comment from line 788 you won't have this issue at all

golem/ethereum/incomeskeeper.py
etam: done

golem/task/tasksession.py
etam: So `assert` it is.

golem/task/tasksession.py
etam: Anyway, the message is gone now.

golem/task/tasksession.py
etam: done

golem/task/tasksession.py
etam: changed to debug

golem/task/tasksession.py
etam: done

golem/task/tasksession.py
```diff
@@ -761,9 +768,32 @@ def _react_to_subtask_result_accepted(self, msg):
             self.disconnect(message.base.Disconnect.REASON.BadProtocol)
             return
 
-        if not self.check_requestor_for_subtask(msg.subtask_id):
+        try:
+            msg.verify_owners(
```
Krigpl, line 772: Why did you add this?

golem/task/tasksession.py
```diff
@@ -761,9 +768,32 @@ def _react_to_subtask_result_accepted(self, msg):
...
+        except (msg_exceptions.InvalidSignature,
+                msg_exceptions.OwnershipMismatch) as e:
+            logger.error("SubtaskResultAccepted has %s: %s",
+                         e.__class__.__name__, e)
+            self.dropped()
+            return
+
+        transaction_system = self.task_server.client.transaction_system
+
+        if (
```
Krigpl, line 785: Woah, this `if` is ugly

Krigpl (approved):
waffle

golem/task/tasksession.py
etam: Yes, it is. I spent a minute with jiivan trying to find a way that's readable and acceptable by pylint.

golem/task/tasksession.py
etam: To make sure the message is properly signed? Or is it checked anywhere else?

Krigpl: I imagine this should happen for all messages and not just this particular one and should be moved to some other place, somewhere before calling `react_to_*` methods. Afaict this is not the case right now, but I wouldn't add this check here now.
Unless @shadeofblue or @jiivan have something planned for that already.

Krigpl: > To make sure the message is properly signed?

Also checking whether message is properly signed is already done before reaching here. This is for something else.

shadeofblue: indeed, it's not to check whether the message itself is properly signed - that's done during the deserialization - but rather whether the whole message chain is correct and that all messages belong to the expected parties

shadeofblue (approved):
I believe it's okay

PR #3432, opened 2018-10-12
fix the contract addresses in the Concent Acceptance tests

shadeofblue:


golem/config/environments/testnet.py
```diff
@@ -41,7 +41,9 @@ class EthereumConfig:  # pylint:disable=too-few-public-methods
     CONTRACT_ADDRESSES = {
         contracts.GNT: '0x924442A66cFd812308791872C4B242440c108E19',
         contracts.GNTB: '0x123438d379BAbD07134d1d4d7dFa0BCbd56ca3F3',
-        contracts.GNTDeposit: '0x694667D7787CFca1892606E81734860a617537B2',
+        contracts.GNTDeposit: CONCENT_CHOICES[
+            os.environ[CONCENT_ENVIRONMENT_VARIABLE]
+        ].get('deposit_contract_address'),
```
Krigpl: That'll return `None` when concent is disabled and SCI will fail, because it expects strings or not to have the key at all.
So for disabled concent this entry shouldn't appear here at all.

Krigpl (approved):
ok_thumb

PR #3433, opened 2018-10-12
Move enqueue_new_task & friends to separate module

jiivan:


golem/client.py
```diff
@@ -229,16 +213,24 @@ def __init__(  # noqa pylint: disable=too-many-arguments
...
+    @property
+    def task_manager(self):
+        return self.task_server.task_manager
+
     def set_rpc_publisher(self, rpc_publisher):
         self.rpc_publisher = rpc_publisher
 
     def get_wamp_rpc_mapping(self):
+        from apps.rendering.task import framerenderingtask
+        from golem.task import rpc as task_rpc
```
etam, line 225: why import here?

golem/interface/command.py
```diff
@@ -75,9 +75,17 @@ def command(name=None, root=False, **kwargs):
     def wrapper(func):
         parent = kwargs.get('parent', None)
 
+        nonlocal name
```
etam: `name` is not used outside `wrapper` scope

golem/task/rpc.py
```diff
@@ -0,0 +1,532 @@
...
+            return result
+        return curry
+    return wrapped
+
+
+class CreateTaskError(Exception):
+    pass
+
+
+def _validate_task_dict(client, task_dict) -> None:
```
etam, line 52: I suggest decoupling `_validate_task_dict` from `client`. In this case it would be enough to pass a `concent_enabled: bool` argument.

golem/interface/client/tasks.py
```diff
@@ -2,19 +2,23 @@
...
 import json
+import typing
 from typing import Any, Optional, Tuple
 
 from apps.core.task.coretaskstate import TaskDefinition
 from golem.core.deferred import sync_wait
 from golem.interface.command import doc, group, command, Argument, CommandResult
 
+if typing.TYPE_CHECKING:
+    from golem.rpc.session import ClientProxy
```
etam: don't you need `# noqa pylint:disable=unused-import` here?

golem/client.py
jiivan: It's the only place it's used. IMO those imports shouldn't pollute module's namespace.

golem/interface/client/tasks.py
jiivan: lintdiff didn't complain, but I added ignore anyway.

golem/client.py
etam: https://www.python.org/dev/peps/pep-0008/#imports
https://wiki.python.org/moin/PythonSpeed/PerformanceTips#Import_Statement_Overhead

I think your argument is invalid.

golem/interface/client/tasks.py
```diff
@@ -110,7 +114,7 @@ def subtasks(self, id, sort):
 
     @command(argument=id_req, help="Restart a task")
     def restart(self, id):
-        deferred = Tasks.client.restart_task(id)
+        deferred = Tasks.client._call('comp.task.restart', id)  # noqa pylint: disable=protected-access
```
etam, line 117: maybe `_call` shouldn't be private anymore?

tests/golem/task/test_rpc.py
```diff
@@ -0,0 +1,394 @@
...
+mock_task = mock.MagicMock()
+mock_task.header.task_id = 'task_id'
+
+
+@mock.patch('signal.signal')
+@mock.patch('golem.network.p2p.node.Node.collect_network_info')
+@mock.patch('golem.task.rpc.enqueue_new_task')
+@mock.patch(
+    'golem.task.taskmanager.TaskManager.create_task',
+    return_value=mock_task,
```
etam: Doesn't that causes `mock_task` object to be reused in all tests? Is it possible that one test changes the object and another test sees those changes? (I'm not sure that's the case, but I have a feeling, that it might be. Just asking)

golem/client.py
jiivan: I believe those quote validate my argument:
> It's often useful to place them inside functions to restrict their visibility [...]
> Note that putting an import in a function can speed up the initial loading of the module [...]

This method is called only once in `golem.Node`.
https://github.com/golemfactory/golem/blob/a62ee28a12b8abb642dbed9b6abfd0334e991ae3/golem/node.py#L416-L438

etam: :+1:

golem/interface/client/tasks.py
jiivan: IMHO whole idea of using *ClientProxy* should be refactored out in favour of using wamp uris. But for now `_call` has to start with underscore because otherwise it can collide with proxied methods:
https://github.com/golemfactory/golem/blob/a62ee28a12b8abb642dbed9b6abfd0334e991ae3/golem/rpc/session.py#L243-L245

tests/golem/task/test_rpc.py
jiivan: Looks like it will be fine:

> Patch can be used as a TestCase class decorator. It works by decorating each test method in the class. This reduces the boilerplate code when your test methods share a common patchings set. patch() finds tests by looking for method names that start with patch.TEST_PREFIX. By default this is 'test', which matches the way unittest finds tests. You can specify an alternative prefix by setting patch.TEST_PREFIX.

source: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

golem/interface/client/tasks.py
etam: :+1:

tests/golem/task/test_rpc.py
etam: This quote is not relevant. And I checked that this **is** the case.

golem/interface/command.py
```diff
@@ -73,14 +73,21 @@ def command(name=None, root=False, **kwargs):
     """
 
     def wrapper(func):
-        parent = kwargs.get('parent', None)
+        parent_cmd = kwargs.get('parent', None)
 
+        if name is None:
+            if hasattr(func, '__func__'):
+                given_name = func.__func__.__name__
```
Wiezzel, line 80: What is the reason for this? To make classmethods work with `@command`?

golem/task/rpc.py
```diff
@@ -0,0 +1,532 @@
...
+    client.task_manager.start_task(task.header.task_id)
+
+
+@defer.inlineCallbacks
+def enqueue_new_task(client, task) \
+            -> typing.Generator[defer.Deferred, typing.Any, taskbase.Task]:
+    """
+    :return: (deferred, task_id) - deferred returns Task object when it's
+    successfully created.
+    """
```
Wiezzel: This docstring seems outdated.

tests/golem/task/test_rpc.py
jiivan: Good catch, thanks @etam !

Wiezzel (approved):

golem/interface/command.py
jiivan: Yep. It'll also work with bound-method and whatever python proxy object you'll feed it.

golem/task/rpc.py
jiivan: True it just duplicates typing annotation.

etam (approved):
ok

PR #3434, opened 2018-10-12
Remove broken check from react to sra

Krigpl:


golem/task/tasksession.py
```diff
@@ -996,19 +985,24 @@ def check_provider_for_subtask(self, subtask_id) -> bool:
...
+            return None
         if node_id != self.key_id:
             logger.warning('Received message about task %r from diferrent '
                            'node %r than expected %r. %s', task_id,
                            self.key_id, node_id, additional_msg)
             return False
         return True
 
-    def check_requestor_for_subtask(self, subtask_id) -> bool:
+    def check_requestor_for_subtask(self, subtask_id) -> Optional[bool]:
```
jiivan, line 1001: Would be nice to rewrite this to raise exceptions instead of this quirky triple return, but prolly outside of this PR.

jiivan (approved):
Sadly approved

golem/task/tasksession.py
etam: In my opinion, exceptions should be raised in unexpected situations. This is totally expected. An enum would be better.

etam (approved):
:hankey:

PR #3437, opened 2018-10-12
Sign all messages before adding to DB

### LINKED ISSUE #3435: Check why TaskToCompute embedded in SubtaskResultsAccepted has no signature
After it's fixed, add check (`.verify_owners()`) to `TaskSession.react_to_subtaskResults_accepted()`

jiivan:
fixes: #3435


golem/task/tasksession.py
```diff
@@ -215,6 +218,7 @@ def verification_finished():
             task_to_compute = get_task_message(
                 'TaskToCompute', task_id, subtask_id)
 
+            # FIXME Remove in 0.20
```
shadeofblue, line 235: @jiivan do we mean that it will be checked during deserialization?

golem/task/tasksession.py
```diff
@@ -322,13 +326,18 @@ def send_report_computed_task(
             options=client_options.__dict__,
         )
 
+        self.send(report_computed_task)
+        if report_computed_task.sig is None:
```
shadeofblue: it would probably make sense to pull that copy + sign operation to a separate function method since it seems it can be used more than in just one place

shadeofblue (approved):
mckayla okayish

golem/task/tasksession.py
jiivan: I mean that, by then all messages in history will be signed (enforced by a check in `history.py`)

PR #3443, opened 2018-10-15
Use decorator for post initialization checks in TransactionSystem

### LINKED ISSUE #3190: Use decorator for post initialization checks in TransactionSystem
Use a decorator instead of multiple copy pastes like this:
https://github.com/golemfactory/golem/blob/b1a0e640c3b5b062d48ea360bd40aa2542650f1f/golem/ethereum/transactionsystem.py#L183-L184

Decorator could look like this:
```python
def sci_required/start_required():
    def _wrapper(f):
        @functools.wraps(f)
        def _curry(self, *args, **kwargs):
            if not self._sci:
            return f(self, *args, **kwargs)
        return _curry
    return _wrapper
```

jiivan:
fixes: #3190

### DIFF
--- a/golem/ethereum/transactionsystem.py
+++ b/golem/ethereum/transactionsystem.py
@@ -1,4 +1,5 @@
 import calendar
+import functools
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


Krigpl (approved):
obama_approves

PR #3444, opened 2018-10-15
Handle 16 bits images

bidzyyys:


apps/rendering/resources/imgrepr.py
```diff
@@ -41,6 +42,49 @@ def close(self):
...
+        # image must be saved with extension,
+        # then rename to path
+        file_path = '{}_{}.{}'.format(path,
+                                      "tmp",
+                                      output_format.lower())
+        self.save_fullname(file_path)
+        os.replace(file_path, path)
+
+    def save_fullname(self, path):
+        cv2.imwrite(path, self.img)
```
Elfoniok: Can this throw exception?

apps/rendering/resources/renderingtaskcollector.py
```diff
@@ -26,106 +27,39 @@ def add_img_file(self, img_file):
...
-            e = EXRImgRepr()
-            e.load_from_file(img)
-            l_im = e.to_l_image()
-            final_alpha = ImageChops.add(final_alpha, l_im)
-            l_im.close()
-
-        final_img.putalpha(final_alpha)
-        final_alpha.close()
-
-    def finalize_exr(self, exr_repr):
```
Elfoniok, line 76: well as far as i know we are still supporting writing to EXR format, could you check this with badb or blue?

apps/rendering/resources/renderingtaskcollector.py
```diff
@@ -26,106 +27,39 @@ def add_img_file(self, img_file):
...
-    def finalize_pil(self):
+    def finalize_img(self):
         res_x, res_y = 0, 0
 
         for name in self.accepted_img_files:
-            with Image.open(name) as img:
-                res_x, img_y = img.size
-                res_y += img_y
+            img = cv2.imread(name, cv2.IMREAD_UNCHANGED)
+            img_y, res_x = img.shape[:2]
```
Elfoniok: Some error handling would be nice :)

Elfoniok (changes requested):

apps/rendering/resources/imgrepr.py
```diff
@@ -3,7 +3,8 @@
 import logging
 from copy import deepcopy
 from typing import Optional
-
+import numpy as np
```
magdasta: I prefer simply importing "numpy" over importing it as "np" (but it's totally subjective - I'm suggesting a change, not insisting on it)

apps/rendering/resources/imgrepr.py
```diff
@@ -41,6 +42,49 @@ def close(self):
         return
 
 
+class OpenCVImgRepr:
```
magdasta, line 49: Should inherit from `ImgRepr`

apps/rendering/resources/imgrepr.py
```diff
@@ -41,6 +42,49 @@ def close(self):
         return
 
 
+class OpenCVImgRepr:
+    def __init__(self):
+        self.img = None
+
+    def __enter__(self):
```
magdasta, line 53: I think we don't need `__enter__` and `__exit__`

apps/rendering/resources/imgrepr.py
```diff
@@ -41,6 +42,49 @@ def close(self):
...
+        return self
+
+    def __exit__(self, type, value, traceback):
+        pass
+
+    def empty(self, width, height, channels, dtype):
+        self.img = np.zeros((height, width, channels),
+                            dtype)
+
+    def get_shape(self):
```
magdasta: this method is used only in tests

apps/rendering/resources/imgrepr.py
```diff
@@ -41,6 +42,49 @@ def close(self):
...
+        return self.img.shape
+
+    def get_dtype(self):
```
magdasta: this method is used only in tests

apps/rendering/resources/imgrepr.py
```diff
@@ -41,6 +42,49 @@ def close(self):
...
+    def save(self, path, output_format):
+        # image must be saved with extension,
+        # then rename to path
+        file_path = '{}_{}.{}'.format(path,
+                                      "tmp",
+                                      output_format.lower())
+        self.save_fullname(file_path)
+        os.replace(file_path, path)
+
+    def save_fullname(self, path):
```
magdasta: please rename to `save`

apps/rendering/resources/imgrepr.py
```diff
@@ -41,6 +42,49 @@ def close(self):
...
+    def get_shape(self):
+        return self.img.shape
+
+    def get_dtype(self):
+        return self.img.dtype
+
+    def paste_image(self, img, x, y):
+        self.img[y:y + img.shape[0], x:img.shape[1]] = img
+
+    def save(self, path, output_format):
```
magdasta: Please rename to `save_with_extension`

apps/rendering/resources/imgrepr.py
```diff
@@ -41,6 +42,49 @@ def close(self):
...
+
+    def save_fullname(self, path):
+        cv2.imwrite(path, self.img)
+
+    @property
+    def size(self):
+        return self.get_size()
+
+    def get_size(self):
+        return self.img.shape[1], self.img.shape[0]
```
magdasta: Do we really need both `size` and `get_size` methods?

apps/rendering/resources/renderingtaskcollector.py
```diff
@@ -1,23 +1,24 @@
-
 import logging
 import math
 import os
```
magdasta: This import is unused, you can remove it on the occasion of working on this file :)

apps/rendering/resources/renderingtaskcollector.py
magdasta: Opencv supports EXRs, collector should now transparently handle them

magdasta (commented):
Before merging, we have to manually check if Golem still correctly assembles images:
*PNGs
*JPEGs
*EXRs
in both 8- and 16-bits.

apps/rendering/resources/imgrepr.py
bidzyyys: It isn't mentioned in the official docs.

apps/rendering/resources/imgrepr.py
bidzyyys: ImgRepr has some abstract methods that we don't want to implement.

apps/rendering/resources/imgrepr.py
bidzyyys: It's required for ```with``` statement.

magdasta: my bad, I didn't notice the `with` which uses it

apps/rendering/resources/imgrepr.py
magdasta: Let's leave it for now, we'll see in the future if we should add error handling

Elfoniok (approved):

Elfoniok:
It works well for PNG, although blender scripts will have to be extended at some point, for example
bpy.context.scene.render.image_settings.color_depth = '16'
Otherwise blender will simply use whatever is set in the scene file, but if it is 16 bit, then collecting images save it in 16 bit format.

Elfoniok:
EXR and JPEG are assembled well as well.

PR #3446, opened 2018-10-15
Estimate gas price/deposit time

### LINKED ISSUE #3335: Estimate gas price/deposit time
If estimated deposit time is larger than 5 minutes then advise user to turn off concent for a current task.

https://projects.invisionapp.com/share/Q4NICAKXEJM#/screens/314103867_Task_Settings_Main___Concents_Copy_8

copied from: https://github.com/golemfactory/golem-electron/issues/430

jiivan:
Prevent user from creating new task when transaction time could be long

depends on: #3433 
fixes: #3335

### DIFF
--- a/tests/golem/ethereum/test_transactionsystem.py
+++ b/tests/golem/ethereum/test_transactionsystem.py
@@ -35,7 +35,7 @@ class TransactionSystemBase(testutils.DatabaseFixture):
         self.sci.GAS_BATCH_PAYMENT_BASE = 30000
         self.sci.get_gate_address.return_value = None
         self.sci.get_block_number.return_value = 1223
-        self.sci.get_current_gas_price.return_value = 10 ** 9
+        self.sci.get_current_gas_price.return_value = self.sci.GAS_PRICE - 1
         self.sci.get_eth_balance.return_value = 0
         self.sci.get_gnt_balance.return_value = 0
         self.sci.get_gntb_balance.return_value = 0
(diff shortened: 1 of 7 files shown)


tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -521,6 +523,32 @@ def confirm_it(tx_hash, cb):
                 ('tx', tx_hash),):
             self.assertEqual(getattr(dpayment, field), value)
 
+    def test_gas_price_skyrocketing(self):
+        self.sci.get_deposit_value.return_value = 0
+        self.sci.get_gntb_balance.return_value = 20
+        self.sci._eth_balance = denoms.ether
```
Krigpl:
```suggestion
        self.sci.get_eth_balance.return_value = denoms.ether
```

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -131,6 +131,11 @@ def test_get_withdraw_gas_cost(self):
         cost = self.ets.get_withdraw_gas_cost(200, dest, 'GNT')
         assert cost == self.sci.GAS_WITHDRAW * gas_price
 
+    def test_withdraw_unknown_currency(self):
+        dest = '0x' + 40 * 'd'
+        with self.assertRaises(ValueError):
```
Krigpl: Since you touched it, please do `assertRaisesRegex` for specific exception, I just noticed that `dest` is not checksum'd address so this method could fail with a different exception as well.
I don't think this is the case right now, but it's better to check against specific exception anyway.

Krigpl (approved):
lgtm

PR #3448, opened 2018-10-16
change the Concent Integration test datadir report so that it can be …

shadeofblue:
…immediately used as the input to run the nodes again more easily


etam (approved):
ok

PR #3452, opened 2018-10-17
Fix order of help entries in golemcli

### LINKED ISSUE #3021: correct order of arguments
After `golemcli network` additional arguments are printed before the primary list

dybi:
resolves: https://github.com/golemfactory/golem/issues/3021

### DIFF
--- a/golem/interface/cli.py
+++ b/golem/interface/cli.py
@@ -285,8 +285,8 @@ class CLI(object):
     def rearrange_action_groups(self):
         for parser in self.subparsers.choices.values():
             action_groups = parser._action_groups  # pylint: disable=protected-access
-            # if there are more action groups than just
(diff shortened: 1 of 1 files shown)


etam:
This touches private parts of standard library. This smells bad. Isn't it possible to add those things to argument parser in correct order?

dybi:
@etam, quite frankly, I wasn't able to figure out how to do it **without** touching the internals. The thing is, that options for parent parser (`self.shared_parser`) are normally added first. Even though, I tried to tinker with internals as carefully as possible ;)

shadeofblue (approved):
strawberry leaf

etam:
@dybi you can trigger rebuilds directly from buildbot web interface

PR #3453, opened 2018-10-17
Implement provider ordering algorithm

Krigpl:
Straight from the marketplace doc. Using `scaled_price` which is `price / max_price` as it's the only context `price` is used in the algorithm and this way we don't need to pass `max_price` along.
Also writing tests for it seems awkward


maaktweluit (approved):
LGTM!

mfranciszkiewicz (approved):

PR #3454, opened 2018-10-17
Fix fundslocker to use Task API properly

mplebanski:
This fixes fundslocker to use appropriate API method from Task class instead

### DIFF
--- a/tests/golem/ethereum/test_fundslocker.py
+++ b/tests/golem/ethereum/test_fundslocker.py
@@ -17,7 +17,7 @@ def make_mock_task(*_, task_id: str = 'tid', subtask_price: int = 100,
     task.header.deadline = timeout_to_deadline(timeout)
     task.header.task_id = task_id
     task.subtask_price = subtask_price
-    task.get_total_tasks = mock.Mock(return_value=total_tasks)
+    task.get_total_tasks.return_value = total_tasks
     return task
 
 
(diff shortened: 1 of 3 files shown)


tests/golem/ethereum/test_fundslocker.py
```diff
@@ -17,7 +17,7 @@ def make_mock_task(*_, task_id: str = 'tid', subtask_price: int = 100,
     task.header.deadline = timeout_to_deadline(timeout)
     task.header.task_id = task_id
     task.subtask_price = subtask_price
-    task.total_tasks = total_tasks
+    task.get_total_tasks = mock.Mock(return_value=total_tasks)
```
Krigpl: or
`task.get_total_tasks.return_value = total_tasks`

Krigpl (approved):

tests/golem/test_client.py
```diff
@@ -810,8 +810,8 @@ def test_enqueue_new_task_from_type(self, *_):
         task_mock = MagicMock()
         task_mock.header.max_price = 1 * 10**18
         task_mock.header.subtask_timeout = 158
-        task_mock.total_tasks = 3
-        price = task_mock.header.max_price * task_mock.total_tasks
+        task_mock.get_total_tasks = mock.Mock(return_value=3)
```
etam:
```suggestion
        task_mock.get_total_tasks.return_value=3
```

etam (approved):
ok

mplebanski:
Thanks for the comments

PR #3457, opened 2018-10-17
Fix Golem Client to use Task API properly

mplebanski:
This fixes Golem Client to use appropriate API method from Task class instead


mplebanski:
I will open another bulk PR for problems like this.

PR #3458, opened 2018-10-17
Fix TaskManager dependency on derived class

mplebanski:
`to_dictionary` method was not defined for abstract Task class (golem/task/taskmanager.py:1028)
`get_preview` method was not defined for abstract TaskTypeInfo class (golem/task/taskmanager.py:1024)
`build_dictionary` method was not defined for abstract TaskBuilder class (golem/task/taskmanager.py:164)

Fix Golem Client to use Task API properly

This fixes Golem Client to use appropriate API method from Task class instead

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -632,7 +632,7 @@ class Client(HardwarePresetsMixin):
                 task.header.mask = Mask()
 
             estimated_fee = self.transaction_system.eth_for_batch_payment(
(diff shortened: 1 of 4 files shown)


Wiezzel (approved):

maaktweluit (approved):

mplebanski:
It happened again. Github web gui changes line endings...

PR #3459, opened 2018-10-17
Marketplace: statistics, requestor side

mfranciszkiewicz:
- [x] ~replace `max_price` with provider offers.~


golem/ethereum/incomeskeeper.py
```diff
@@ -55,6 +56,7 @@ def received_batch_transfer(
                     event='confirmed',
                     node_id=e.sender_node,
                     amount=e.value_received,
+                    delay=time.time() - datetime_to_timestamp(e.created_date),
```
Krigpl: It might be more accurate to use transaction's timestamp in place of `time.time()` - especially when transaction gets mined when client was off

golem/ethereum/incomeskeeper.py
```diff
@@ -101,6 +103,12 @@ def expect(
         if not inserted and not income.accepted_ts:
             income.accepted_ts = accepted_ts
             income.save()
+            dispatcher.send(
```
Krigpl, line 102: I think that should happen iff `inserted`

golem/ethereum/paymentprocessor.py
```diff
@@ -100,11 +103,24 @@ def _on_batch_confirmed(self, payments: List[Payment], receipt) -> None:
...
+            payment.details.fee / denoms.ether
+        )
+
+        delay = time.time() - datetime_to_timestamp(payment.created_date)
+        dispatcher.send(
+            signal="golem.payment",
+            event="confirmed",
+            subtask_id=payment.subtask,
+            payee=payment.payee,
+            delay=delay,
```
Krigpl, line 124: Same, `delay` might not be accurate

golem/ethereum/incomeskeeper.py
mfranciszkiewicz: You're right, this happened due to some external changes

golem/database/schemas/023_schema.py
```diff
@@ -15,8 +15,8 @@ def migrate(migrator, *_args, **_kwargs):
         provider_efficacy=pw.ProviderEfficacyField(
             default=ProviderEfficacy(0, 0, 0, 0)),
         provider_efficiency=pw.FloatField(default=1.0),
-        requestor_paid_sum=pw.FloatField(default=0.0),
-        requestor_assigned_sum=pw.FloatField(default=0.0)
+        requestor_paid_sum=pw.IntegerField(default=0),
```
Krigpl: These should be `HexIntegerField`

golem/ethereum/incomeskeeper.py
```diff
@@ -50,11 +51,14 @@ def received_batch_transfer(
             e.save()
 
             if e.value_expected == 0:
+                reference_ts = e.settled_ts or e.accepted_ts
```
Krigpl: Just use `accepted_ts`, settled is not reliable is only Concent specific.

golem/ethereum/paymentprocessor.py
```diff
@@ -100,11 +103,25 @@ def _on_batch_confirmed(self, payments: List[Payment], receipt) -> None:
...
+
+    @staticmethod
+    def _payment_confirmed(payment: Payment) -> None:
+        log.debug(
+            "- %s confirmed fee %.6f",
+            payment.subtask,
+            payment.details.fee / denoms.ether
+        )
+
+        delay = payment.processed_ts - \
```
Krigpl: `processed_ts` is set just after payment is created, so this isn't what you want. `on_batch_confirmed`'s `receipt` argument has timestamp field, use it instead.

Krigpl (changes requested):

maaktweluit (approved):
LGTM!

golem/ethereum/incomeskeeper.py
```diff
@@ -51,8 +51,7 @@ def received_batch_transfer(
             e.save()
 
             if e.value_expected == 0:
-                reference_ts = e.settled_ts or e.accepted_ts
-                delay = reference_ts - datetime_to_timestamp(e.created_date)
+                delay = e.accepted_ts - datetime_to_timestamp(e.created_date)
```
Krigpl: This measures how long verification process took.
I think you're looking for `block.timestamp - e.accepted_ts`

mfranciszkiewicz: I've removed the `delay` from this branch (requestor stats) and will cover the issue in provider's one

Krigpl (approved):

PR #3462, opened 2018-10-18
B0.18.0

jiivan:


requirements.txt
```diff
@@ -30,7 +30,7 @@ eth-keys==0.2.0b3
 eth-tester==0.1.0-beta.24
 eth-utils==1.0.3
 ethereum==1.6.1
-Golem-Verificator==1.5.0
+Golem-Verificator==1.5.2
```
Krigpl, line 33: afaict develop needs changes from `1.5.1` which are not included in `1.5.2` (say whaaat) dunno how this should be resolved @Elfoniok

Elfoniok: There will be 1.5.3 merging both paths.

PR #3463, opened 2018-10-18
Names refactor

mplebanski:


apps/core/task/coretaskstate.py
```diff
@@ -32,20 +30,24 @@ def subtask_timeout(self):
 class TaskDefinition(object):
     """ Task description used in GUI and in save file format"""
 
+    PICKLED_VERSION_0 = 0
+    PICKLED_VERSION_1 = 1
+    PICKLED_VERSION_2 = 2
```
jiivan: Isn't it self-explanatory?

jiivan: IMO you could place this int in `golem.core.variables.PICKLE_ID`

golem/task/rpc.py
```diff
@@ -138,7 +138,7 @@ def _restart_subtasks(client, old_task_id, task_dict, subtask_ids_to_copy):
     @defer.inlineCallbacks
     @safe_run(
         lambda e: logger.error(
-            'Restarting subtasks failed. task_dict=%r, subtask_ids_to_copy=%r',
+            'Restarting subtasks_countfailed. task_dict=%r, subtask_ids_to_copy=%r',
```
jiivan: ?

tests/apps/core/task/test_coretaskstate.py
```diff
@@ -101,3 +101,32 @@ def test_missing_concent_enabled(self, *_):
...
+        del self.task_definition.name
+        del self.task_definition.timeout
+        del self.task_definition.subtasks_count
+        setattr(self.task_definition, 'task_name', 'some_name')
+        setattr(self.task_definition, 'full_task_timeout', '00:01:00')
+        setattr(self.task_definition, 'total_subtasks', 1)
+        with mock.patch(
+            'apps.core.task.coretaskstate.TaskDefinition.__getstate__',
+            side_effect=lambda: self.task_definition.__dict__,
+        ):
```
jiivan, line 128: We also need to test what happens when `__getstate__` returns `golem.__version__, __dict__`

jiivan (commented):
Have verified that task related commands in *golemcli* work?
`tasks show`
`tasks create`
`tasks restart` etc

apps/core/task/coretaskstate.py
mplebanski: Not really. If you take a look at:

```
    def __getstate__(self):
        return self.PICKLED_VERSION_2, self.__dict__
``` 
only, then 2 in place of the identifier is just a magic number

golem/task/rpc.py
mplebanski: My bad

apps/core/task/coretaskstate.py
jiivan: WIll it be less magic if we label it `MAGIC_NUMBER_1`? :)

mplebanski:
Regarding question about CLI: create, show, restart all work.

apps/core/task/coretaskstate.py
mplebanski: I don't understand. It clearly states that this is a version information

jiivan (approved):
:+1:

golem/task/rpc.py
```diff
@@ -98,6 +98,11 @@ def prepare_and_validate_task_dict(client, task_dict):
         'concent_enabled',
         client.concent_service.enabled,
     )
+    # TODO #3474
+    if 'subtasks' in task_dict:
+        logger.warning("Using soon to be deprecated data format for\
+                        input JSON change `subtasks` to `subtasks_count`")
```
jiivan:
```suggestion

```

golem/task/rpc.py
```diff
@@ -98,6 +98,11 @@ def prepare_and_validate_task_dict(client, task_dict):
         'concent_enabled',
         client.concent_service.enabled,
     )
+    # TODO #3474
+    if 'subtasks' in task_dict:
+        logger.warning("Using soon to be deprecated data format for\
```
jiivan:
```suggestion
        logger.warning(
            "Using soon to be deprecated data format for input JSON."
            " Change `subtasks` to `subtasks_count`",
        )
```

jiivan (commented):
Lotta whitespace

golem/task/rpc.py
```diff
@@ -98,6 +98,13 @@ def prepare_and_validate_task_dict(client, task_dict):
         'concent_enabled',
         client.concent_service.enabled,
     )
+    # TODO #3474
+    if 'subtasks' in task_dict:
+        logger.warning(
+            "Using soon to be deprecated data format for input JSON."
+            "Change `subtasks` to `subtasks_count`",
```
jiivan:
```suggestion
            " Change `subtasks` to `subtasks_count`",
```

PR #3466, opened 2018-10-18
Fix log message

Krigpl:


Wiezzel (approved):
Maybe also limit the number of decimal places (e.g. `%.6f`)?

Krigpl:
@Wiezzel It limits to 6 by default, so no need for that.

maaktweluit (approved):

PR #3467, opened 2018-10-18
Check for NoMoreSubtasks earlier

Krigpl:
When there no more subtasks but provider who is waiting acceptance wants another task he will receive `WaitingForResults` and then try again after a while. Instead we can just respond `NoMoreSubtasks` to let them know there is no more work needed anymore.


jiivan (approved):

shadeofblue (approved):
girl a okay stockphoto

Krigpl:
Anybody knows what is Windows upset about?

PR #3471, opened 2018-10-19
Always set default 'concent_enabled' before _validate_task_dict

jiivan:


mplebanski:
@jiivan Maybe we could move this dict patching into a separate function?

mplebanski (approved):

PR #3475, opened 2018-10-19
Disable benchmark computation on each test

mplebanski:
It speeds up each test run by ~60 seconds.


maaktweluit (approved):
LGTM!

Maybe also invite someone from CGI team to review? since it relates to their work

magdasta (approved):

Issue #3020, opened 2018-07-12
unifying setting results


jamuszyn:
right now cli prints "0" or "False" when disabled and "1" or "True" when enabled. It should be unified

dybi:
@jamuszyn, is there any direction preferred? I.e. You would like to have `0` and `1` or `True` and `False` when showing `settings`, `envs`, etc.?

dybi:
I have picked up :
(`0`,`1`) -> (`False`, `True`)
option

PaweuB:
Pivotal:
https://www.pivotaltracker.com/story/show/161383623

PR #3481, opened 2018-10-22
Chore displaying booleans in golemcli

dybi:
(`0`,`1`) -> (`False`, `True`)
Resolves: https://github.com/golemfactory/golem/issues/3020

### DIFF
--- a/golem/interface/client/settings.py
+++ b/golem/interface/client/settings.py
@@ -28,6 +28,10 @@ class Setting(namedtuple('Setting', ['help', 'type', 'converter', 'validator',
 
 @group(help="Manage settings")
 class Settings(object):
+    BOOL_CONVERTIBLE_KEY_PREFIXES = [
+        'accept_', 'debug_', 'use_', 'enable_',
+        'in_shutdown', 'net_masking_enabled'
+    ]
 
     client = None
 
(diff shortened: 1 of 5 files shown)


golem/interface/client/settings.py
```diff
@@ -194,7 +198,15 @@ def fmt(k, v):
                 return self.settings[k].format(v)
             return v
 
+        def convert(k, v):
+            if k in self.settings:
+                return self.settings[k].converter(v)
+            elif any(k.startswith(prefix) for prefix in Settings.BOOL_CONVERTIBLE_KEY_PREFIXES):  # pylint: disable=line-too-long
```
etam:
```suggestion
            elif any(k.startswith(prefix) for prefix
                     in Settings.BOOL_CONVERTIBLE_KEY_PREFIXES):
```

golem/interface/client/settings.py
```diff
@@ -28,6 +28,10 @@ def format(self, value):
 
 @group(help="Manage settings")
 class Settings(object):
+    BOOL_CONVERTIBLE_KEY_PREFIXES = [
```
etam, line 31: `send_pings` is missing?

dybi: @etam, not really - `send_pings` has its converter.
`BOOL_CONVERTIBLE_KEY_PREFIXES` is only for fields that are defined outside `Settings` class and thus cannot have own converter.

etam (approved):

PR #3484, opened 2018-10-23
Reorganize modules

Krigpl:
Small reorg to make things cleaner.
Main lib.rs will only contain the python bindings and any extra to/from python conversion will be under `bindings`


rust/golem/src/bindings/marketplace.rs
```diff
@@ -0,0 +1,23 @@
+extern crate cpython;
```
maaktweluit: This line is not needed because it is already in `lib.rs`

maaktweluit (changes requested):
LGTM!

One small removal and i will approve :)

maaktweluit (approved):

PR #3486, opened 2018-10-23
[WIP] Feature add blender script package to ctd

### LINKED ISSUE #3066: Replace scripts embedded in TaskToCompute with individual parameters
TaskToCompute currently contains parameters for Blender in the form of a script (`script_src` field). Here's a fragment from a sample message:

``` python
for scene in bpy.data.scenes:

    scene.render.tile_x = 0
    scene.render.tile_y = 0
    scene.render.resolution_x = 1000
    scene.render.resolution_y = 600
    scene.render.resolution_percentage = 100
    scene.render.use_border = True
    scene.render.use_crop_to_border = True
    scene.render.border_max_x = 1.0
    scene.render.border_min_x = 0.0
    scene.render.border_min_y = 0.25
    scene.render.border_max_y = 0.375
    scene.render.use_compositing = bool(False)
```

There are two big problems with this approach:
- To render a task you have to run untrusted code supplied by the client
- There's no foolproof way of finding out what these parameters are without executing or parsing the script

We talked about it with @lukasz.glen and he agreed that it's something that should be changed in the protocol. Sending individual parameters and letting the client put them in the template would be safer and simpler. But we understand that a change to the protocol needs to be discussed first. We'd like to ask for some input and suggestions on the matter.

Once we agree on the right solution we offer to implement it and submit a pull request for approval.

kbeker:
Resolves #3066

### DIFF
--- a/golem/task/taskcomputer.py
+++ b/golem/task/taskcomputer.py
@@ -1,6 +1,8 @@
 import logging
 from typing import Any, Dict, Optional, TYPE_CHECKING
 
+from copy import copy
+
 import os
 import time
 import uuid
@@ -8,6 +10,9 @@ from threading import Lock
 
 from pydispatch import dispatcher
 from twisted.internet.defer import Deferred, TimeoutError
+
+from golem_messages.message import ComputeTaskDef
+from apps.blender.resources.scenefileeditor import generate_blender_crop_file
 
 from golem.core.common import deadline_to_timeout
 from golem.core.deferred import sync_wait
@@ -113,6 +118,23 @@ class TaskComputer(object):
         )
         return True
 
+    def _extract_extra_data_from_subtask(self, subtask: ComputeTaskDef) -> dict:  # noqa pylint:disable=no-self-use
+        if subtask['task_type'] == 'Blender':
+            extra_data = copy(subtask['extra_data'])
+            extra_data['frames'] = subtask['meta_parameters']['frames']
+            extra_data['output_format'] = \
+                subtask['meta_parameters']['output_format']
+            extra_data['script_src'] = generate_blender_crop_file(
+                resolution=subtask['meta_parameters']['resolution'],
+                borders_x=subtask['meta_parameters']['borders_x'],
+                borders_y=subtask['meta_parameters']['borders_y'],
+                use_compositing=subtask['meta_parameters']['use_compositing'],
+                samples=subtask['meta_parameters']['samples'],
+            )
+        else:
+            raise RuntimeError('Task Type is set to None')
+        return extra_data
+
     def task_resource_collected(self, task_id, unpack_delta=True):
         subtask = self.assigned_subtask
         if not subtask or subtask['task_id'] != task_id:
@@ -123,11 +145,14 @@ class TaskComputer(object):
             self.task_server.unpack_delta(rs_dir, self.delta, task_id)
         self.delta = None
         self.last_task_timeout_checking = time.time()
+
+        extra_data = self._extract_extra_data_from_subtask(subtask)
+
         self.__compute_task(
             subtask['subtask_id'],
             subtask['docker_images'],
             subtask['src_code'],
-            subtask['extra_data'],
+            extra_data,
             subtask['deadline'])
         return True
 
--- a/apps/blender/task/blenderrendertask.py
+++ b/apps/blender/task/blenderrendertask.py
@@ -463,12 +463,14 @@ class BlenderRenderTask(FrameRenderingTask):
         min_y = numpy.float32(min_y)
         max_y = numpy.float32(max_y)
 
-        script_src = generate_blender_crop_file(
-            resolution=(self.res_x, self.res_y),
-            borders_x=(0.0, 1.0),
-            borders_y=(min_y, max_y),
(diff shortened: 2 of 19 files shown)


ederenn:
@kbeker it's impossible to upload a task from GUI. Golem is stuck on " testing local render" after clicking button "next."

error from console:
``ERROR    [twisted                            ] WAMP message serialization error: unsupported type: <class 'AttributeError'>`


[golem.log](https://github.com/golemfactory/golem/files/2519252/golem.log)

mplebanski:
Why files like:

`apps/core/task/coretask.py`

mplebanski:
Why files like:
```
apps/core/task/coretask.py
apps/dummy/task/dummytask.py
golem/task/localcomputer.py
golem/task/taskcomputer.py
```

Need to know anything about blender specifically?

kbeker:
@ederenn 
> @kbeker it's impossible to upload a task from GUI. Golem is stuck on " testing local render" after clicking button "next."
> 
> error from console:
> ``ERROR [twisted ] WAMP message serialization error: unsupported type: <class 'AttributeError'>`
> 
> [golem.log](https://github.com/golemfactory/golem/files/2519252/golem.log)

From logs:
```
  File "/home/ederenn/projects/golem/golem/task/localcomputer.py", line 69, in run
    ctd = self.get_compute_task_def()
  File "/home/ederenn/projects/golem/apps/blender/task/blenderrendertask.py", line 560, in query_extra_data_for_test_task
    output_format="PNG",
  File "/home/ederenn/projects/golem/apps/core/task/coretask.py", line 358, in _new_blender_script_package
    return golem_messages.message.tasks.BlenderScriptPackage(
AttributeError: module 'golem_messages.message.tasks' has no attribute 'BlenderScriptPackage'
```
I can see here that it is posibility that you have not installed newest `golem_messages` which are not tagged yet. 
Did you installed new golem_messages where `BlenderScriptPackage` is already implemented?

kbeker:
> Why files like:
> 
> ```
> apps/core/task/coretask.py
> apps/dummy/task/dummytask.py
> golem/task/localcomputer.py
> golem/task/taskcomputer.py
> ```
> Need to know anything about blender specifically?

@mplebanski 

-`coretask.py` needs to create new `compute_task_def` with package for specific task (in this case it is `Blender`)
-`dummytask.py` same here as above. When `ComputeTaskDef` is created it needs for proper work script_package for specific task.
-`localcomputer.py` to create `script_src` we need to convert everything what is now in `meta_parameters` in `compute_task_def` and from all this info generate blender script for proper rendering. That's why there is added extractor for this data.
-`taskcomputer.py` same as above.

mplebanski:
To me those classes seem pretty abstract, yet here they are using concrete implementations. For each of the files I mentioned I believe blender specific stuff should be removed. Especially DummyTask which is about POW now knows blender stuff. Do you see my point?

kbeker:
> To me those classes seem pretty abstract, yet here they are using concrete implementations. For each of the files I mentioned I believe blender specific stuff should be removed. Especially DummyTask which is about POW now knows blender stuff. Do you see my point?

I think yes, but then how `dummytask` would know that it should be blender task if it won't be there? In actual GM configuration when `CTD` is created it should be specified what `task_type` it is <- in this case it will be Blender so it has to be there. 

About `coretask.py` <- I can move it to `blenderrendertask.py` if you want.

About 
```
golem/task/localcomputer.py
golem/task/taskcomputer.py
```
where do you think should blender script be generated if these two files should not have any blender specifically?

golem/task/taskcomputer.py
```diff
@@ -123,11 +145,14 @@ def task_resource_collected(self, task_id, unpack_delta=True):
...
         self.last_task_timeout_checking = time.time()
+
+        extra_data = self._extract_extra_data_from_subtask(subtask)
+
         self.__compute_task(
             subtask['subtask_id'],
             subtask['docker_images'],
             subtask['src_code'],
-            subtask['extra_data'],
+            extra_data,
```
mplebanski, line 155: What changed that previously it was fine and now we need to modify 'extra_data' ?

mplebanski:
> I think yes, but then how dummytask would know that it should be blender task if it won't be there? 

When such case happens?

mplebanski (commented):
If I understood correctly then all that blender specific parameters should be nested in 'extra_data' field.

golem/task/taskcomputer.py
rwrzesien: Have you even read the related issue?

mplebanski: No, sorry for that one.

mplebanski:
OK now I have read the related issue. What about putting `meta_parameters` inside 'extra_data' and rendering a template on provider side?

kbeker:
> OK now I have read the related issue. What about putting `meta_parameters` inside 'extra_data' and rendering a template on provider side?

It was set at the very beginning that `meta_parameters` should be in the same level as `task_id`, `task_type` etc. 

What do you mean by saying that rendering a template should be on provider side?

Krigpl:
Is this still relevant? If not please close.

Krigpl:
Assuming no.

PR #3487, opened 2018-10-23
Put offer_pooling_intervall in config

Krigpl:
So it can be set to lower values, e.g. for testing.
Also some minor clean up.


maaktweluit (approved):
LGTM!

PR #3488, opened 2018-10-23
Fix create_task_error Exception override

mplebanski:


jiivan (approved):
Is it possible to test this scenario?

mplebanski:
@jiivan I added the tests but there is a problem with them. They depend on stuff they are not testing like internal implementation of e.g. create_task. IMHO exception handling should not be done using decorators but there should be one big exception handler on a higher level.

Issue #3490, opened 2018-10-24
Verification fails and it never stops


etam:
- verification fails with timeouts
- golem created a bunch of docker images, all bound to a single cpu core
- 
```
$ ./golemcli.py tasks subtasks 44294512-d6d2-11e8-976e-d276138057ae
node           id                                    ETA      status     completion
-------------  ------------------------------------  -------  ---------  ------------
someone0       8e5cede8-d6d2-11e8-b2c9-d276138057ae  0:00:00  Failure    100.00 %
someone1       06dadd1c-d6d3-11e8-ab49-d276138057ae  0:00:00  Verifying  0.00 %
someone2       9165d79c-d6d5-11e8-829c-d276138057ae  0:00:00  Verifying  0.00 %
someone3       f983fe62-d6d8-11e8-84a5-d276138057ae  0:00:00  Failure    100.00 %
someone4       7efc9326-d6d9-11e8-822b-d276138057ae  0:00:00  Verifying  0.00 %
someone5       3b1cd8be-d6da-11e8-8ad0-d276138057ae  0:00:00  Verifying  0.00 %
someone6       af62d7f8-d6da-11e8-b21a-d276138057ae  0:00:00  Verifying  0.00 %
someone7       bbf5459a-d6da-11e8-a813-d276138057ae  0:00:00  Verifying  0.00 %
someone8       6c7303dc-d6dc-11e8-8d9a-d276138057ae  0:00:00  Verifying  0.00 %
someone9       9bcdd40c-d6dc-11e8-91d1-d276138057ae  0:00:00  Failure    0.00 %
someone10      c462b2a4-d6dc-11e8-a0fd-d276138057ae  0:00:00  Verifying  0.00 %
someone11      ca043674-d6dc-11e8-8143-d276138057ae  0:00:00  Failure    100.00 %
someone12      ab00075a-d6e1-11e8-804c-d276138057ae  0:00:00  Verifying  0.00 %
someone13      28c244ca-d700-11e8-9daa-d276138057ae  0:00:00  Failure    0.00 %
someone14      b6628f54-d723-11e8-b73f-d276138057ae  0:00:00  Verifying  0.00 %
```
- 
```
$ ag 3b1cd8be-d6da-11e8-8ad0-d276138057ae golem.error.log 
[...]

PR #3493, opened 2018-10-24
Kill containers on stop

Elfoniok:
closes https://github.com/golemfactory/golem/issues/3490

### DIFF
--- a/apps/blender/verification_queue.py
+++ b/apps/blender/verification_queue.py
@@ -24,6 +24,7 @@ class VerificationQueue:
         self.callbacks: Dict[VerificationTask, FunctionType] = dict()
         self._paused = False
         self._timed_out = False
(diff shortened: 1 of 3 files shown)


apps/blender/verification_queue.py
```diff
@@ -14,16 +14,14 @@
 
 class VerificationQueue:
 
-    VERIFICATION_TIMEOUT = 90
-    RESTART_DELAY = 5
+    VERIFICATION_TIMEOUT = 1800
```
jiivan, line 22: Could add a comment about this value? And that it should be related to local performance in the future?

jiivan (approved):

apps/blender/verification_queue.py
Elfoniok: Done

PR #3494, opened 2018-10-24
Remove persistance from FundsLocker

Krigpl:
The state of locks is trivially retrievable from tasks state so no need for a separate storage.
adam_pickles

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -353,24 +353,7 @@ class Client(HardwarePresetsMixin):
             task_finished_cb=self._task_finished_cb,
         )
 
-        tm = self.task_server.task_manager
-        for task_id, task_state in tm.tasks_states.items():
-            if not task_state.status.is_completed():
-                unfinished_subtasks = 0
-                for subtask_state in task_state.subtask_states.values():
-                    if not subtask_state.status.is_finished():
-                        unfinished_subtasks += 1
-                task = tm.tasks[task_id]
-                try:
-                    self.funds_locker.lock_funds(
-                        task_id,
-                        task.subtask_price,
-                        task.get_total_tasks(),
-                        task.header.deadline,
-                    )
-                except NotEnoughFunds as e:
-                    # May happen when gas prices increase, not much we can do
-                    logger.info("Not enough funds to restore old locks: %r", e)
+        self._restore_locks()
 
         monitoring_publisher_service = MonitoringPublisherService(
             self.task_server,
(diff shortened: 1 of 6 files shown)


golem/client.py
```diff
@@ -353,6 +353,25 @@ def start_network(self):
...
+                task = tm.tasks[task_id]
+                try:
+                    self.funds_locker.lock_funds(
+                        task_id,
+                        task.subtask_price,
+                        task.get_total_tasks(),
+                        task.header.deadline,
+                    )
+                except NotEnoughFunds as e:
+                    # May happen when gas prices increase, not much we can do
```
maaktweluit: maybe we can lock all the funds available? the user should not be allowed create a bigger "lock depth"

maaktweluit (approved):
LGTM!
One small comment..
And looks like the builds are having some issues :(

golem/client.py
```diff
@@ -353,6 +353,25 @@ def start_network(self):
             task_finished_cb=self._task_finished_cb,
         )
 
+        tm = self.task_server.task_manager
+        for task_id, task_state in tm.tasks_states.items():
```
mfranciszkiewicz: Could you move this block to a separate method?

golem/client.py
Krigpl: I'm somewhat reluctant to do this - I mean you're correct, but imo it doesn't help that much (user is in debt anyway) to justify the code changes required.

golem/ethereum/fundslocker.py
```diff
@@ -1,48 +1,48 @@
...
 
 class TaskFundsLock:
-    def __init__(self, task):
-        self.price = task.subtask_price
-        self.num_tasks = task.get_total_tasks()
-        self.task_deadline = task.header.deadline
+    def __init__(self, subtask_price: int, num_tasks: int, deadline) -> None:
+        self.price = subtask_price
+        self.num_tasks = num_tasks
+        self.task_deadline = deadline
```
etam, line 18: what's the type of `deadline`?

Krigpl: I dunno, that's why I didn't include it

golem/client.py
```diff
@@ -466,6 +468,26 @@ def terminate(*exceptions):
         self.task_server.start_accepting(listening_established=task.callback,
                                          listening_failure=task.errback)
 
+    def _restore_locks(self) -> None:
+        tm = self.task_server.task_manager
```
etam, line 473: there is a shortcut above:
```python
    @property
    def task_manager(self):
        return self.task_server.task_manager
```

golem/ethereum/fundslocker.py
etam: https://github.com/golemfactory/golem/blob/f2c00b6312e5851cfcd7c18cfb2e9f77000b3939/golem/task/taskbase.py#L192

etam:
One more optional idea: maybe we should remove old pickles files?

golem/client.py
```diff
@@ -466,6 +468,27 @@ def terminate(*exceptions):
         self.task_server.start_accepting(listening_established=task.callback,
                                          listening_failure=task.errback)
 
+    def _restore_locks(self) -> None:
+        assert self.task_server is not None
+        tm = self.task_server.task_manager
+        for task_id, task_state in tm.tasks_states.items():
+            if not task_state.status.is_completed():
```
etam, line 475: Codecov says this is not tested. Is it correct?

mfranciszkiewicz (approved):

etam (approved):
pickle

etam:
Shame on us all, who reviewed this: #3581

PR #3502, opened 2018-10-26
Remove unnecessary dependency on CompTaskKeeper

Krigpl:
Those checks are unnecessary as same validation happens in the received message.


Wiezzel (approved):
Very nice.

mfranciszkiewicz (approved):
:put_litter_in_its_place:

PR #3508, opened 2018-10-29
Remove unused imports golemcli.py

mplebanski:


etam:
Does golemcli work with this change?

AFAIK these imports are important. Imported objects are not directly used, but they are automatically registered as handlers.

PR #3510, opened 2018-10-31
Add helpful debug logs in PP and minor cleanup

jiivan:
In develop `TransactionSystem.get_funds_needed` will be exposed via rpc

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -145,7 +145,7 @@ class PaymentProcessor:
                     'Insufficient GNTB balance.'
                     ' value=%(value).6f, subtask_id=%(subtask)s',
                     {
-                        'value': p.value,
+                        'value': p.value / denoms.ether,
                         'subtask': p.subtask,
                     },
                 )
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -1,9 +1,12 @@
 import calendar
 import logging
 import time
-from collections import defaultdict
-from typing import List
+from typing import (
+    NamedTuple,
+    List,
+)
 
+from collections import defaultdict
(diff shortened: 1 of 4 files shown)


Krigpl (commented):
What is this needed for?

jiivan:
@Krigpl for a requestor to know how much funds he needs to pay his debts

golem/ethereum/transactionsystem.py
```diff
@@ -345,6 +344,11 @@ def get_locked_gnt(self) -> int:
             raise Exception('Start was not called')
         return self._gntb_locked + self._payment_processor.reserved_gntb
 
+    def get_funds_needed(self) -> dict:
```
Krigpl: If you introduced a named tuple then just stick to it

Krigpl (commented):
> In develop TransactionSystem.get_funds_needed will be exposed via rpc

I don't get something - it's gonna be exposed in develop, but the PR is to b0.18. How is it gonna be used in this branch and why it's not gonna be exposed via rpc here?

golem/ethereum/transactionsystem.py
jiivan: NamedTuple won't go through RPC. It's only reason for this pass through method to make results RPC ready.

golem/ethereum/paymentprocessor.py
```diff
@@ -141,14 +141,31 @@ def __get_next_batch(self, closure_time: int) -> int:
                 break
             gntb_balance -= p.value
             if gntb_balance < 0:
+                log.debug(
+                    'Insufficient GNTB balance.'
+                    ' value=%(value).6f, subtask_id=%(subtask)s',
+                    {
+                        'value': p.value,
```
Krigpl: `/ denoms`

Krigpl (approved):

PR #3523, opened 2018-11-06
0.18.0 port status always available

shadeofblue:
Make port statuses always available in RPC/CLI output (for open and timed-out ports alike).

That way:
1) the users will have an easier way of telling which ports their golem instances are listening on
2) the integration tests and other automated tools will have an easy way to query the golem instance for that information (again, irrespective of whether those ports are actually reachable or not reachable from the outside world)

for some reason, previously, Golem only made that information available when the ports were _NOT_ reachable...

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -236,10 +236,15 @@ class Client(HardwarePresetsMixin):
     def p2p_listener(self, event='default', **kwargs):
         if event == 'unreachable':
             self.on_unreachable(**kwargs)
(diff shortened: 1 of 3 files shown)


Wiezzel (approved):
image

mfranciszkiewicz (approved):

mdtanrikulu:
@shadeofblue Could you add list of statuses of the port parameter as description?

PR #3526, opened 2018-11-07
Implement provider price function

Krigpl:
As per marketplace doc


mfranciszkiewicz (approved):

mfranciszkiewicz:
Now we need to store those prices in place of current `max_price * subtask_timeout` calculations (i.e. `compute_subtask_value` function, `TaskServer.requested_tasks`)

PR #3537, opened 2018-11-09
Added some debug logs for resource pulling

maaktweluit:


golem/resource/client.py
```diff
@@ -145,18 +145,31 @@ def _retry(self, method: MethodType,
...
+                         retries, method, args, kwargs, exc)
 
             try:
                 result = method(*args, **kwargs)
             except Exception as exc:
-                logger.error('Error executing %r (%r, %r): %r',
-                             method, args, kwargs, exc)
 
                 if exc.__class__ not in self.retry_exceptions:
+                    logger.error('Error executing, raising. '
```
mfranciszkiewicz, line 157: Just a small remark: we don't know what we're `executing` (sync / async), unless we have the debug mode on :)

maaktweluit: Good one! resolved by adding `async` to the async logs

mfranciszkiewicz (approved):

Issue #3414, opened 2018-10-10
No secret for `golemcli` when flag --mainnet is missing


etam:
If `golemcli` has wrong parameters and fails to connect to `golemapp`, it just shows an error "WAMP-CRA signature is invalid".

This is very confusing. It should also print hints to check if `-d` and `--mainnet` flags are correct.

See also: #3171 #3348

jiivan:
https://github.com/golemfactory/golem/issues/3477#issuecomment-431384800

PR #3540, opened 2018-11-09
Added warning on cli and app net mismatch

kmazurek:
Closes: #3414

Note: this PR was initially split in two (see: #3549).

Adds a check in golemcli which attempts to detect whether an
instance of golem app is running on the system with a different target
chain (e.g. CLI is started with --mainnet flag, but there's a golem
instance running on testnet).

To allow for earlier detection of a running instance, the logic for acquiring a lock on datadir has been moved up to `golemapp.py`.


golem/core/simpleenv.py
```diff
@@ -2,16 +2,24 @@
 
 import appdirs
 
+APP_NAME = 'golem'
+DEFAULT_DATA_DIR = 'default'
 
-def get_local_datadir(name: str, root_dir=None) -> str:
+
+def get_local_datadir(name=None, root_dir=None, env_suffix=None) -> str:
```
etam: Type annotations for arguments plz.

golemcli.py
```diff
@@ -105,6 +108,20 @@ def start():
     ws_cli.execute(forwarded, interactive=interactive)
 
 
+def check_golem_running(datadir: str, cli_in_mainnet: bool):
+    net_to_check = 'rinkeby' if cli_in_mainnet else 'mainnet'
```
etam: you have proper name in `golem.config.active.DATA_DIR`

etam:
Please add tests.

golemcli.py
etam: I'm a bit confused. Why do you need to pass 'rinkeby' or 'mainnet' to `get_local_datadir` as additional `env_suffix` argument? What's wrong with `from golem.config.active import DATA_DIR` that was in there?

etam (changes requested):
Try to make `mainnet.py` and `testnet.py` in `golem/config/environments/` stateless.

golemcli.py
kmazurek: That's because the environment needs to be cross-checked here - when we're running on mainnet, we want to inspect the datadir for testnet (and vice versa).

kmazurek:
Okay, will take a look.

etam (approved):
ok

golemapp.py
```diff
@@ -43,6 +46,56 @@ def monkey_patched_getLogger(*args, **kwargs):
...
 
 
+def set_active_environment(func: Callable) -> Callable:
+    """
+    Function decorator which sets Golem's active configuration before calling
+    the decorated function.
+
+    Note: this decorator overrides the ``datadir`` keyword argument to provide
+    the full path to the app's data directory.
+    """
```
jiivan: I would prefer this to be a function that takes `(mainnet, net, concent)` as arguments and returns `datadir`. Then I would call this in `@lock_datadir`

golemapp.py
```diff
@@ -122,21 +181,10 @@ def start(monitor, concent, datadir, node_address, rpc_address, peer, mainnet,
         start_crossbar_worker(m)
         return
 
-    set_environment('mainnet' if mainnet else net, concent)
-
-    # Import AFTER the active configuration has been set
```
jiivan, line 127: I would leave some comment in the new version, explaining why import is not global (at top of the file)

golemcli.py
```diff
@@ -2,11 +2,14 @@
 
 import os
 import argparse
+import logging
 import sys
 
 from multiprocessing import freeze_support
 import click
 
+from portalocker import lock, unlock, LOCK_EX, LOCK_NB, LockException
```
jiivan: Maybe just `import portalocker` to avoid namespace pollution?

golemcli.py
```diff
@@ -115,5 +134,27 @@ def delete_reactor():
...
+    """ Checks if a lock file exists in the given data dir and whether
+    that file is currently locked by another process.
+    If both conditions are true we assume that an instance of Golem is running
+    and using the specified data dir.
+    """
+    datadir = get_local_datadir(root_dir=root_dir, env_suffix=net_name)
+    lock_path = os.path.join(datadir, 'LOCK')
+
+    if os.path.isfile(lock_path):
+        lock_file = open(lock_path, 'w')
```
jiivan: `with open(lock_path, 'w') as lock_file:` might be more readable and there won't be a need to close file in `finally` block. Maybe `lock` could also be used as a context manager?

tests/golem/interface/test_golemcli.py
```diff
@@ -19,3 +20,36 @@ def test_golem_cli(self, *_):
...
+
+    @patch('builtins.open', mock_open())
+    @patch('os.path.isfile', side_effect=[True])
+    @patch('golemcli.lock', side_effect=LockException)
+    @patch('golemcli.unlock')
+    @patch('golemcli.logger')
+    def test_is_app_running_positive(self, logger, *_):
+        with patch.object(sys, 'argv', ['program', '-d', 'datadir']):
+            start()
+            assert logger.warning.called
```
jiivan: `logger.warning.assert_called()`

tests/golem/interface/test_golemcli.py
```diff
@@ -19,3 +20,36 @@ def test_golem_cli(self, *_):
...
+    def test_check_golem_running(self, logger, *_):
+        with patch.object(sys, 'argv', ['program', '-d', 'datadir', '-m']):
+            start()
+            args, kwargs = logger.warning.call_args
+            assert 'removing' in args[0]
+
+        with patch.object(sys, 'argv', ['program', '-d', 'datadir']):
+            start()
+            args, kwargs = logger.warning.call_args
+            assert 'adding' in args[0]
```
jiivan: `self.assertIn('adding', args[0])`

tests/golem/interface/test_golemcli.py
```diff
@@ -19,3 +20,36 @@ def test_golem_cli(self, *_):
...
         with patch.object(sys, 'argv', ["program", "--some_forwarded_flag"]):
             start()
+
+    @patch('golemcli.is_app_running', side_effect=[True, True])
+    @patch('golemcli.logger')
+    def test_check_golem_running(self, logger, *_):
+        with patch.object(sys, 'argv', ['program', '-d', 'datadir', '-m']):
+            start()
+            args, kwargs = logger.warning.call_args
+            assert 'removing' in args[0]
```
jiivan: `self.assertIn('removing', args[0])`

golemcli.py
kmazurek: Good idea, will do

golemapp.py
kmazurek: Alright, I guess it makes sense since `@lock_datadir` depends on environment being set anyway.

jiivan (approved):

PR #3542, opened 2018-11-12
Add status filter for golemcli incomes | payments | deposit_payments

### LINKED ISSUE #3105: Add "awaiting" and "confirmed" additional arguments
for "payments" and "incomes" to allow the user filtering the list

EDIT:
for "deposit_payments" as well

dybi:
resolves: #3105

### DIFF
--- a/golem/interface/client/payments.py
+++ b/golem/interface/client/payments.py
@@ -71,7 +71,7 @@ def incomes(sort, status):
     return CommandResult.to_tabular(incomes_table_headers, values, sort=sort)
 
 
-@command(arguments=(sort_payments, status_filter,),
+@command(arguments=(sort_payments, status_filter),
          help="Display payments", root=True)
 def payments(sort, status):
 
@@ -104,7 +104,7 @@ def payments(sort, status):
 
 
 @command(
-    arguments=(sort_deposit_payments, status_filter,),
+    arguments=(sort_deposit_payments, status_filter),
     help="Display deposit payments",
     root=True,
 )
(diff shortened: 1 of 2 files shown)


tests/golem/interface/test_client_commands.py
```diff
@@ -404,45 +403,142 @@ def setUpClass(cls):
...
             assert result.data[1][0] == [
-                'node_1', 'waiting', '1.00000000 GNT'
+                'node_1', 'sent', '1.00000000 GNT'
+            ]
+
+    def test_incomes_awaiting(self):
+        with client_ctx(incomes, self.client):
+            result = incomes(None, "awaiting")
+
+            assert isinstance(result, CommandResult)
```
marmistrz: Please use `self.assertEquals` instead of `assert` n tests. It produces more informative output if the assertion fails.

golem/interface/client/payments.py
```diff
@@ -55,13 +71,16 @@ def incomes(sort):
     return CommandResult.to_tabular(incomes_table_headers, values, sort=sort)
 
 
-@command(argument=sort_payments, help="Display payments", root=True)
-def payments(sort):
+@command(arguments=(sort_payments, status_filter,),
```
marmistrz: I think you don't need the trailing comma after `status_filter`

marmistrz (changes requested):

tests/golem/interface/test_client_commands.py
dybi: sure. BTW: Do you have defined convention for `assertEquals`:
- `assertEquals(expected, result)` or
- `assertEquals(result, expected)` ?

golem/interface/client/payments.py
dybi: ok

tests/golem/interface/test_client_commands.py
shadeofblue: @dybi I use the latter but I'm unsure if that's a organization-wide rule

jiivan (approved):

PR #3543, opened 2018-11-12
Add 'current' option for `golemcli tasks show` to display only current (ongoing) tasks

### LINKED ISSUE #3106: Add `current` additional argument in `tasks` cli call
Print current task id, task state (progress; is it computing or being computed), task cost, task settings (resolution; frames; subtask; timeouts)

dybi:
resolves: https://github.com/golemfactory/golem/issues/3106

### DIFF
--- a/golem/interface/client/tasks.py
+++ b/golem/interface/client/tasks.py
@@ -8,6 +8,7 @@ from typing import Any, Optional, Tuple
 from apps.core.task.coretaskstate import TaskDefinition
 from golem.core.deferred import sync_wait
 from golem.interface.command import doc, group, command, Argument, CommandResult
+from golem.task.taskstate import TaskStatus
 
 if typing.TYPE_CHECKING:
     from golem.rpc.session import ClientProxy  # noqa pylint: disable=unused-import
@@ -25,6 +26,13 @@ class Tasks:
     subtask_table_headers = ['node', 'id', 'ETA', 'status', 'completion']
     unsupport_reasons_table_headers = ['reason', 'no of tasks',
                                        'avg for all tasks']
+
+    current_task_states = [
+        TaskStatus.sending.name,
+        TaskStatus.waiting.name,
+        TaskStatus.starting.name,
+        TaskStatus.computing.name,
+    ]
 
     id_req = Argument('id', help="Task identifier")
     id_opt = Argument.extend(id_req, optional=True)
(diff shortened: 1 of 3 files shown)


dybi:
@jamuszyn - could you please check what format is displayed in GUI for status, ie.:
"notStarted" or "Not started", "waiting" or "Waiting", etc.

EDIT:
checked offline - everything's crystal clear now.

golem/interface/client/tasks.py
```diff
@@ -26,6 +27,13 @@ class Tasks:
     unsupport_reasons_table_headers = ['reason', 'no of tasks',
                                        'avg for all tasks']
 
+    current_task_states = [
```
etam: `TaskStatus` has `is_completed` method. Maybe you could add a new, similar one, and use it instead?

dybi: I was wondering if it would be ok to do this. Sure :+1:

dybi: done

golem/interface/client/tasks.py
```diff
@@ -57,20 +58,34 @@ class Tasks:
...
+    def show(self, id, sort, current):
 
         deferred = Tasks.client.get_tasks(id)
         result = sync_wait(deferred)
 
         if not id:
             values = []
 
+            if current:
+                result = [t for t in result
```
Wiezzel, line 86: `result or []` in line 89 suggests that `result` might be `None`. It should be handled here as well.

etam (approved):
ok

golem/interface/client/tasks.py
etam: Or line 80 can be changed to `result = sync_wait(deferred) or []`

etam (changes requested):
@Wiezzel had a good point

golem/interface/client/tasks.py
dybi: quite frankly, it shouldn't be; if you thoroughly analyze `def get_tasks(self, task_id: Optional[str] = None)` from `golem.client`, you will see that `None` can be returned only when `task_id` has `True` value in boolean context. But when this is the case, whole `if` from line 82 is skipped. 

So it seems that `or []` from line 89 should be removed.
If you insist, I can add handling of `None`, but so far I feel more inclined towards removing `or []`. @Wiezzel, what would be your preference here?

Anyway, thanks for pointing this out :)

Wiezzel: If you're sure this won't happen, just remove the `or []`. I'm fine with this.

dybi: @Wiezzel @etam removed; also I have added a unit test to make sure we will notice if logic connected with `None` ever changes

etam (approved):

PR #3546, opened 2018-11-13
Remove unnecessary ComputerState

Krigpl:
It's quite useless the way it is now.

### DIFF
--- a/tests/golem/task/test_taskmanager.py
+++ b/tests/golem/task/test_taskmanager.py
@@ -1324,6 +1324,8 @@ class TestCopySubtaskResults(TwistedTestCase):
         old_subtask_state = SubtaskState()
         new_subtask_state = SubtaskState()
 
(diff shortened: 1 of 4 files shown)


tests/golem/task/test_taskmanager.py
```diff
@@ -1387,12 +1372,6 @@ def verify(_):
             self.assertEqual(new_subtask_state.stderr, 'stderr')
             self.assertEqual(new_subtask_state.results, ['result'])
 
-            self.assertEqual(old_subtask_state.computer.node_id, 'node_id')
```
maaktweluit, line 1390: maybe keep this assert for the new property `node_id`?

maaktweluit (changes requested):
LGTM!

One q about both tests, will approve on reply!

maaktweluit (approved):

PR #3547, opened 2018-11-13
"Waiting for task" status removed on provider state

mdtanrikulu:
Due to unnecessary usage, `Waiting for task` status removed.

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -128,8 +128,7 @@ class Client(HardwarePresetsMixin):
         self.apps_manager = apps_manager
         self.datadir = datadir
         self.__lock_datadir()
-        self.task_tester:
-            Optional[TaskTester] = None
+        self.task_tester: Optional[TaskTester] = None
 
         self.task_archiver = TaskArchiver(datadir)
 
@@ -172,8 +171,7 @@ class Client(HardwarePresetsMixin):
             variant=concent_variant,
         )
 
-        self.task_server:
-            Optional[TaskServer] = None
+        self.task_server: Optional[TaskServer] = None
         self.port_mapper = None
 
         self.nodes_manager_client = None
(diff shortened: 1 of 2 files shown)


golem/client.py
```diff
@@ -128,7 +128,8 @@ def __init__(  # noqa pylint: disable=too-many-arguments
         self.apps_manager = apps_manager
         self.datadir = datadir
         self.__lock_datadir()
-        self.task_tester: Optional[TaskTester] = None
+        self.task_tester:
+            Optional[TaskTester] = None
```
etam: Why?

golem/client.py
```diff
@@ -274,9 +277,11 @@ def taskmanager_listener(self, sender, signal, event='default', **kwargs):
             self._publish(Task.evt_subtask_status, kwargs['task_id'],
                           kwargs['subtask_id'], op.value)
         else:
-            op_class_name: str = op.__class__.__name__ \
-                                 if op is not None else None
-            op_value: int = op.value if op is not None else None
+            op_class_name:
+                str = op.__class__.__name__ \
+                    if op is not None else None
```
etam: I think this notation makes it very confusing.

(In C this would mean that `op_class_name` is a label, that you can `jump` to, and you're assigning value to `str` variable)

etam: If you want a newline so much, this way would be more legible:

```python
op_class_name: str = \
    op.__class__.__name__ if op is not None else None
```

golem/client.py
mdtanrikulu: Auto formatter force to do that, working on it.

golem/client.py
```diff
@@ -577,7 +582,8 @@ def enqueue_new_task(self, task_dict) -> Tuple[Deferred, str]:
         _result = Deferred()
 
         # FIXME: Statement only for old DummyTask compatibility #2467
-        task: TaskBase
+        task:
+            TaskBase
```
etam: ?

etam:
I'd approve that, but there are some very odd whitespace changes. I added comments to some of them, but then I realized there are more, so I just gave up.

golem/client.py
mdtanrikulu: Same reason as above

golem/client.py
mdtanrikulu: Same.

etam (approved):
clean

Please squash commits when merging.

PR #3549, opened 2018-11-13
Moved datadir locking from client to node

kmazurek:
#3414

Moves the logic for acquiring a lock on datadir to happen
before client authentication. This allows for earlier detection of a Golem instance
running on the machine.


jiivan (approved):
LGTM

etam:
Idea: implement locking as a context manager (example: https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) and in `golemapp.py` wrap creating and starting `Node` object in `with` statement.

kmazurek:
Good idea, this can actually be replaced with [this library](https://github.com/WoLpH/portalocker), which already includes a context manager for file locking. The [original implementation](https://github.com/golemfactory/golem/blob/449c018a1ec3734be3f9d43395ea91cb7f3050c4/golem/tools/filelock.py#L9-L10) of file locking is based on that library.

tests/golem/interface/test_golemcli.py
```diff
@@ -19,3 +19,6 @@ def test_golem_cli(self, *_):
             start()
         with patch.object(sys, 'argv', ["program", "--some_forwarded_flag"]):
             start()
+
+    def test_stuff(self):
```
jiivan: ?

jiivan (commented):
Also if you update `requirements_to-freeze.txt` tests might pass :)

tests/golem/interface/test_golemcli.py
kmazurek: you know, just stuff

PR #3560, opened 2018-11-15
Simplify price in CompTaskInfo

Krigpl:
`price` setter and getter ain't no used


golem/task/taskkeeper.py
```diff
@@ -55,32 +55,21 @@ class WrongOwnerException(Exception):
 class CompTaskInfo:
     def __init__(self, header: TaskHeader, price: int):
         self.header = header
-        self._price, self.subtask_price = 0, 0  # lints and typing
-        self.price = price
-        self.requests = 1
-        self.subtasks = {}
-        # TODO Add concent communication timeout. Issue #2406
```
maaktweluit, line 62: Is this also no longer needed? Is the issue closed?

maaktweluit (changes requested):
LGTM!

One small q, approve on reply!

golem/task/taskkeeper.py
Krigpl: This isn't relevant to this PR and I didn't touch that, the issue is not closed (as you can check by yourself) and I dunno if it's still needed. I might have missed the point of your question :P

maaktweluit: my question was: "Why is this removed"

Now i see its moved.. /shy

No questions remain ;)

maaktweluit (approved):

PR #3569, opened 2018-11-16
Set environment in acceptance tests

### LINKED ISSUE #3210: Concent reports "too small requestor deposit" in the "Force Accept" scenario's tests
```$ pytest scripts/concent_acceptance_tests/force_accept --maxfail=1 -vs --lf```

results in:

```
ServiceRefused(header=MessageHeader(type_=4000, timestamp=1534953973, encrypted=True), sig=b'\x7f<\x98Q\xab\x88w\xb6\x9e\xa1L\x17?f*-&\xbbK2\xbc\xbeD\x8a\xbc\x8c\x9f\x17\x1a\xae\x02\x02cH\x07\x83\xb0J\x84Y\xb1\x9a\x84t\xa8\xd5M\x05=R\x89l\xbe\x00\x05\xe3\xaf\x9a\xba\x85K]\x916\x00', slots=[['task_to_compute', None], ['reason', 'TOO_SMALL_REQUESTOR_DEPOSIT']])
```

### PR DESCRIPTION
**Pull request by jiivan**
fixes: #3210

### DIFF
--- a/golem/network/concent/client.py
+++ b/golem/network/concent/client.py
@@ -31,17 +31,6 @@ def verify_response(response: requests.Response) -> None:
         raise exceptions.ConcentUnavailableError('response is None')
 
     logger.debug('Headers received from Concent: %s', response.headers)
-    concent_version = response.headers['Concent-Golem-Messages-Version']
-    if not utils.is_version_compatible(
-            theirs=concent_version,
-            spec=gconst.GOLEM_MESSAGES_SPEC,):
-        raise exceptions.ConcentVersionMismatchError(
-            'Incompatible version',
-            ours=gconst.GOLEM_MESSAGES_VERSION,
-            theirs=concent_version,
-        )
-    if response.status_code == 200:
-        return
 
     if response.status_code % 500 < 100:
         raise exceptions.ConcentServiceError(
@@ -51,7 +40,7 @@ def verify_response(response: requests.Response) -> None:
             )
         )
 
-    if response.status_code % 400 < 100:
+    if response.status_code != 200:
         logger.warning('Concent request failed with status %d and '
                        'response: %r', response.status_code, response.text)
 
@@ -60,6 +49,19 @@ def verify_response(response: requests.Response) -> None:
                 response.status_code,
                 response.text
             )
+        )
+
+    concent_version = response.headers.get(
+        'Concent-Golem-Messages-Version',
+        None,
+    )
+    if not utils.is_version_compatible(
+            theirs=concent_version,
+            spec=gconst.GOLEM_MESSAGES_SPEC,):
+        raise exceptions.ConcentVersionMismatchError(
+            'Incompatible version',
+            ours=gconst.GOLEM_MESSAGES_VERSION,
+            theirs=concent_version,
         )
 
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by maaktweluit — APPROVED** [review]
LGTM!

I'm not very familiar with the concent acceptance tests, please get another approval too :)

File: scripts/concent_acceptance_tests/force_accept/test_requestor_doesnt_send.py [hunk 1]
```diff
@@ -90,6 +90,10 @@ def provider_send_force(
             rct_kwargs = {}
         price = random.randint(1 << 20, 10 << 20)
         self.requestor_put_deposit(price)
+        import time
+        print('sleep 120')
+        time.sleep(120)
+        print('woken')
```
dybi: maybe `'awake'` ? ;)

golem/network/concent/client.py
```diff
@@ -51,7 +40,7 @@ def verify_response(response: requests.Response) -> None:
             )
         )
 
-    if response.status_code % 400 < 100:
+    if response.status_code != 200:
```
dybi: what about `BadRequest` or `NotFound` responses (4XX in general)? They are now indistinguishable...

And sometimes concent responds with `202` (i.e `handle_send_force_report_computed_task` case)

dybi (commented):
Please consider the comments.

golem/network/concent/client.py
pawelkisielewicz: I spoke with @cameel and all responses between 200 and 299 are OK. Please change this line to
 ` if not 200 <= response.status_code <= 299:`
This will resolve problem with https://github.com/golemfactory/golem/issues/3210

cameel: Anything between 200 and 299 is a success. In practice Concent uses:
- HTTP 200: request accepted, no errors, response sent
- HTTP 202: request accepted, no errors but the requested operation has not finished yet - the response will be delivered asynchonously (via the `receive/` endpoint).
- HTTP 204: Like HTTP 200 but the request does not require any response

As for 4xx and 5xx - it would be best to distinguish them but all in all they're both errors. HTTP 4xx is an error in the request (malformed request, validation error, etc.) which suggests a bug on the client side, 5xx is an error on the Concent side that's out of client's control (timeout, unexpected database error, bug in Concent code, etc.).

Note that the code refers only to the transport layer (HTTP). When a protocol message is delivered successfully it happens with HTTP 200 even it it's a message like `ServiceRefused`.

scripts/concent_acceptance_tests/force_accept/test_requestor_doesnt_send.py
```diff
@@ -90,6 +90,10 @@ def provider_send_force(
             rct_kwargs = {}
         price = random.randint(1 << 20, 10 << 20)
         self.requestor_put_deposit(price)
+        import time
+        print('sleep 120')
+        time.sleep(120)
```
dybi: one more thing - wouldn't be better to uncomment 
```python
# time.sleep(120)
```
in `SCIBaseTest.put_deposit()` method (`scripts/concent_acceptance_tests/base.py`) ?

dybi (approved):
:+1:

PR #3589, opened 2018-11-23
Tweaked docker-machine settings:

maaktweluit:
- disk size default is now 5gb, was 20gb
- `pad_memory` added as hypervisor specific function
- Fixed old virtualbox error message with recovery steps
- Added catch on start_vm for windows to apply minimum constraints
- Removed dynamic memory, it broke more then it fixed

### DIFF
--- a/golem/docker/hypervisor/docker_machine.py
+++ b/golem/docker/hypervisor/docker_machine.py
@@ -110,8 +110,8 @@ Ensure that you try the following before reporting an issue:
  2. The proper environment is set. On windows powershell please run:
     & "C:\Program Files\Docker Toolbox\docker-machine.exe" ^
     env golem | Invoke-Expression
- 3. virtualbox driver is available:
-    docker-machine.exe create --driver virtualbox golem
+ 3. hyperv driver is available:
+    docker-machine.exe create --driver hyperv golem
  4. Restart Windows machine"""
             logger.error(typical_solution_s)
             raise
--- a/golem/docker/hypervisor/docker_machine.py
+++ b/golem/docker/hypervisor/docker_machine.py
@@ -107,12 +107,9 @@ Ensure that you try the following before reporting an issue:
 
  1. The virtualization of Intel VT-x/EPT or AMD-V/RVI is enabled in BIOS
     or virtual machine settings.
- 2. The proper environment is set. On windows powershell please run:
-    & "C:\Program Files\Docker Toolbox\docker-machine.exe" ^
-    env golem | Invoke-Expression
- 3. hyperv driver is available:
-    docker-machine.exe create --driver hyperv golem
- 4. Restart Windows machine"""
+ 2. The windows feature 'Hyper-V' is enabled
+ 3. docker-machine is in your path `docker-machine --version` in ps or cmd
+ 4. docker-machine ls has no errors `docker-machine  ls` in ps or cmd"""
             logger.error(typical_solution_s)
             raise
 
--- a/golem/docker/hypervisor/hyperv.py
+++ b/golem/docker/hypervisor/hyperv.py
@@ -44,6 +44,17 @@ class HyperVHypervisor(DockerMachineHypervisor):
         super().__init__(*args, **kwargs)
         self._vm_utils = VMUtilsWithMemFix()
 
+
+    def setup(self) -> None:
+        try:
+            # The windows VM fails to start when too much memory is assigned
+            super().setup(self)
+        except Exception:
+            logger.warning("Failed to setup VM, retry with minimum values")
+            self.constrain(self.vm_name, MIN_CONSTRAINTS)
+            super().setup(self)
+
+
     @classmethod
     def is_available(cls) -> bool:
         command = "@(Get-Module -ListAvailable hyper-v).Name | Get-Unique"
--- a/golem/docker/hypervisor/hyperv.py
+++ b/golem/docker/hypervisor/hyperv.py
@@ -97,8 +97,6 @@ class HyperVHypervisor(DockerMachineHypervisor):
         mem_key = CONSTRAINT_KEYS['mem']
         mem = params.get(mem_key)
         assert isinstance(mem, int)
-        # Dynamic memory works best in blocks of 128
-        mem = (mem // 128) * 128
         cpu = params.get(CONSTRAINT_KEYS['cpu'])
 
         min_mem = MIN_CONSTRAINTS[mem_key]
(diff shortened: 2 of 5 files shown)


golem/docker/manager.py
```diff
@@ -128,8 +129,15 @@ def build_config(self, config_desc):
         try:
             memory_size = max(int(config_desc.max_memory_size) // 1024,
                               memory_size)
+            max_mem_in_mb = psutil.virtual_memory().free // 1024 // 1024
```
Wiezzel: Shouldn't we use `available` instead of `free`? Psutil docs state that:
>available: the memory that can be given instantly to processes without the system going into swap. This is calculated by summing different memory values depending on the platform and it is supposed to be used to monitor actual memory usage in a cross platform fashion.

golem/docker/manager.py
```diff
@@ -128,8 +129,15 @@ def build_config(self, config_desc):
...
+            if self.hypervisor and self.hypervisor.vm_running():
+                max_mem_in_mb += self.hypervisor.constraints()['memory_size']
```
Wiezzel: I'm afraid we might get tricked here since we're using the dynamic memory feature. Imagine the following scenario:
1. The host machine has 8 GB of RAM.
1. There's a VM running with 6 GB memory limit.
1. The VM is not running any computations so the OS allocates only 1 GB of RAM to it.
1. Other processes on the machine are using 2 GB of RAM.
I suspect that in such case we would get information about 6 GB of free memory from `psutil.virtual_memory()` and 6 GB from `hypervisor.constraints()` thus ending up with `max_mem_in_mb` equal to 12 GB.

I am not sure if this would actually happen but it needs to be checked and somehow rectified if true.

golem/docker/hypervisor/docker_machine.py
```diff
@@ -110,8 +110,8 @@ def _set_env(self, retried=False):
  2. The proper environment is set. On windows powershell please run:
     & "C:\Program Files\Docker Toolbox\docker-machine.exe" ^
     env golem | Invoke-Expression
```
Wiezzel: Point 2. is deprecated as well. And actually, I believe that this message should be somewhere else.

golem/docker/hypervisor/hyperv.py
```diff
@@ -93,6 +97,8 @@ def constrain(self, name: Optional[str] = None, **params) -> None:
         mem_key = CONSTRAINT_KEYS['mem']
         mem = params.get(mem_key)
         assert isinstance(mem, int)
+        # Dynamic memory works best in blocks of 128
+        mem = (mem // 128) * 128
```
Wiezzel: Wouldn't this cause the hypervisor try to change the memory every single time Golem starts? Since the actual memory limit would be different than the config value I believe that it would.

As far as I remember this was why I put this line in `DockerManager` and not in `HyperVHypervisor`:
```python
memory_size = memory_size // 2 * 2
```

golem/docker/hypervisor/hyperv.py
```diff
@@ -211,3 +217,35 @@ def _create_volume(self, hostname: str, shared_dir: Path) -> str:
         )
 
         return volume_name
+
+class VMUtilsWithMemFix(VMUtils):
+
+    # TODO: Fix override!
```
Wiezzel: Let me know if I get this right. Is this a bug in `os-win` that we're fixing by overriding?

golem/docker/manager.py
maaktweluit: Good one, will look into this!

golem/docker/manager.py
maaktweluit: I think you are right.. i need to move this code to "start_vm" so will adopt it accordingly to this case.

golem/docker/hypervisor/docker_machine.py
maaktweluit: agree, will fix it or make an issue out of it :)

golem/docker/hypervisor/hyperv.py
maaktweluit: good one! will move it up.

Thought it was nice to only make this 128 rule inside windows

golem/docker/hypervisor/hyperv.py
maaktweluit: i am not sure.

i think some machines might be able to dynamicaly change "GuestVisibleMemory". so for them OS-win works now.

For our docker it does not update after start, once set to 1GB it stays 1GB, so we need this to be as high as we can.

os-win people are super nice, so will make an issue when this branch is working again

golem/docker/hypervisor/hyperv.py
Wiezzel: > Thought it was nice to only make this 128 rule inside windows

Agreed. Maybe try to get the best of both worlds by introducing some new method like `pad_memory` in the `Hypervisor` abstract class?

golem/docker/manager.py
maaktweluit: done

golem/docker/manager.py
maaktweluit: stopped the VM while building config for now, open to better solves :)

golem/docker/hypervisor/docker_machine.py
maaktweluit: done

golem/docker/hypervisor/hyperv.py
maaktweluit: nice idea! move it up now, will move like you proposed once it works

golem/docker/manager.py
Wiezzel: Stopping the VM looks fine to me. It would be stopped anyway a moment later when the config is applied.

Wiezzel:
@maaktweluit Could you explain the recent changes?

maaktweluit:
@Wiezzel sure!

- Reverted dynamic memory, it brought more edge cases then it solved
- Catch errors on `start_vm` and try to start the VM with the minimum constrains

golem/docker/manager.py
```diff
@@ -128,8 +129,18 @@ def build_config(self, config_desc):
         try:
             memory_size = max(int(config_desc.max_memory_size) // 1024,
                               memory_size)
-            # Hyper-V expects a multiple of 2 MB
-            memory_size = memory_size // 2 * 2
+            vmem = psutil.virtual_memory()
```
mfranciszkiewicz: Can we move this code to the Hyper-V hypervisor class? This workaround will also affect macOS.

mfranciszkiewicz (changes requested):

golem/docker/manager.py
maaktweluit: good idea!

Wiezzel: That's weird. I thought it was already moved to `pad_memory` method.

Wiezzel (approved):
OK

golem/docker/manager.py
maaktweluit: Now it is, first try! ;)

mfranciszkiewicz (approved):

Issue #3580, opened 2018-11-21
Not unique task_id and subtask_id in test_already_processed 


dybi:
**DEPENDS ON**:
https://github.com/golemfactory/golem/pull/3569#pullrequestreview-176624275
the above issue must be fixed first (`response.status_code` from concent can be in [200, 300) )

**DESCRIPTION**:
`scripts/concent_acceptance_tests/force_accept/test_requestor_doesnt_send.py::RequestorDoesntSendTestCase::test_already_processed` test should use the same `task_id` and `subtask_id` for subsequent `ForceSubtaskResults` messages to make concent respond with `DuplicatedRequest` for second msg. This is not the case.

**HINT**:
Problem is caused by the `post_generation` functions (`task_id()` and `subtask_id()`)  in `TaskToComputeFactory`. 

**TO REPRODUCE**:
`$ pytest scripts/concent_acceptance_tests/force_accept --maxfail=1 -vs --lf` after #3569 is merged

PaweuB:
Pivotal: 
https://www.pivotaltracker.com/story/show/162116452

dybi:
Updated the description with the latest findings

PR #3591, opened 2018-11-23
Workaround for problem with non unique task_id and subtask_id in `test_already_processed`

dybi:
This is a workaround for: https://github.com/golemfactory/golem/issues/3580


dybi:
A proper solution will be proposed

PR #3594, opened 2018-11-23
Marketplace: misc fixes

mfranciszkiewicz:
- changes the way payment delays are calculated
- sends provider stats to the monitor

### DIFF
--- a/tests/golem/task/test_taskkeeper.py
+++ b/tests/golem/task/test_taskkeeper.py
@@ -510,7 +510,7 @@ class TestCompSubtaskInfo(TestCase):
         self.assertIsInstance(csi, CompSubtaskInfo)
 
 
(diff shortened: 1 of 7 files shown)


tests/golem/task/test_taskkeeper.py
```diff
@@ -510,6 +510,7 @@ def test_init(self):
         self.assertIsInstance(csi, CompSubtaskInfo)
 
 
+@mock.patch('golem.task.taskkeeper.ProviderStatsManager')
```
Krigpl: You can just pass `mock.Mock()` as the second argument and you won't need `*_` in test methods. Also, test_conformance won't need to change (fwiw I'd remove `test_conformance` altogether from the codebase - the coverage is very sporadic and we have other tools in place to cover that).

Krigpl (approved):

PR #3611, opened 2018-11-29
Remove unused stuff from TaskManager

Krigpl:
Simply unused.

### DIFF
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -93,7 +93,6 @@ class TaskManager(TaskEventListener):
 
         self.node = node
         self.keys_auth = keys_auth
-        self.key_id = keys_auth.key_id
 
         self.tasks: Dict[str, Task] = {}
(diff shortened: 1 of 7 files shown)


golem/task/taskmanager.py
```diff
@@ -977,10 +975,6 @@ def get_frame_subtasks(self, task_id: str, frame) \
             return None
         return task.get_subtasks(frame)
 
-    def change_config(self, root_path, use_distributed_resource_management):
-        self.dir_manager = DirManager(root_path)
```
Wiezzel, line 981: Are you sure this line is unnecessary? I see `dir_manager` being used in a few methods.

Krigpl: Yeah, but `dir_manager` is set in the constructor and `change_config` is only called with exactly the same argument which changes nothing, so this is no-op.

Wiezzel (approved):

PR #3613, opened 2018-11-29
Remove last_checking variable from TaskHeader

### LINKED ISSUE #2407: Remove last_checking variable from TaskHeader
This label is used by task_keeper, but it should be part of some external structure wrapping TaskHeader and  not TaskHeader itself.

jiivan:
fixes: #2407

### DIFF
--- a/golem/task/taskkeeper.py
+++ b/golem/task/taskkeeper.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 import pathlib
 import pickle
@@ -323,7 +324,9 @@ class TaskHeaderKeeper:
         # be added again to task_headers
         self.removed_tasks = {}
         # task ids by owner
-        self.tasks_by_owner = {}
+        self.tasks_by_owner: typing.Dict[str, set] = {}
+        # Keep track which tasks were checked when
+        self.last_checking: typing.Dict[str, datetime.datetime] = {}
 
         self.min_price = min_price
         self.app_version = app_version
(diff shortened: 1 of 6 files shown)


golem/task/taskkeeper.py
```diff
@@ -556,15 +560,34 @@ def remove_task_header(self, task_id) -> bool:
...
-                self.tasks_by_owner[owner_key_id].discard(task_id)
-        if task_id in self.supported_tasks:
-            self.supported_tasks.remove(task_id)
-        if task_id in self.support_status:
-            del self.support_status[task_id]
+            self.tasks_by_owner[owner_key_id].discard(task_id)
+        except KeyError:
+            pass
+
+        for container in (
```
kmazurek, line 569: I think it would be more readable to use `if task_id in container` for dicts (since that's `O(1)`) and handle `supported_tasks` array separately, something like this:
```
for container in (
    self.task_headers,
    self.support_status,
    self.last_checking
):
    if task_id in container:
        del container[task_id]

try:
    self.support_status.remove(task_id)
except ValueError:
    pass
```

What do you think?

etam (approved):
seems_cool

golem/task/taskkeeper.py
jiivan: https://docs.python.org/3/glossary.html#term-eafp :)

kmazurek: LBYL ¯\_(ツ)_/¯

PR #3615, opened 2018-11-29
Clean up add_payment

Krigpl:
So that TransactionSystem doesn't create `Payment` object

### DIFF
--- a/golem/ethereum/paymentprocessor.py
+++ b/golem/ethereum/paymentprocessor.py
@@ -5,7 +5,7 @@ import time
 from collections import defaultdict
 from typing import List
 from sortedcontainers import SortedListWithKey
-from eth_utils import encode_hex
+from eth_utils import decode_hex, encode_hex
 from ethereum.utils import denoms
 from twisted.internet import threads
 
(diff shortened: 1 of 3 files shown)


maaktweluit (approved):

golem/ethereum/paymentprocessor.py
```diff
@@ -106,23 +112,22 @@ def _on_batch_confirmed(self, payments: List[Payment], receipt) -> None:
...
-            payment.subtask,
-            encode_hex(payment.payee),
-            payment.value / denoms.ether))
-
-        if not payment.processed_ts:
-            payment.processed_ts = get_timestamp()
-            payment.save()
+    def add(self, subtask_id: str, eth_addr: str, value: int) -> Payment:
+        log.info(
+            "Adding payment for %s to %s (value: %f)",
```
maaktweluit, line 117: `%.6f` formatter is removed, was this meant?

Krigpl: The default is `.6` anyway, so yeah

golem/ethereum/paymentprocessor.py
```diff
@@ -106,23 +112,22 @@ def _on_batch_confirmed(self, payments: List[Payment], receipt) -> None:
...
+            subtask_id,
+            eth_addr,
+            value / denoms.ether,
+        )
+        payment = Payment.create(
+            subtask=subtask_id,
+            payee=decode_hex(eth_addr),
```
mfranciszkiewicz, line 124: Why remove the `len` check?

Krigpl: I'd say we either do full validation or none at all and we didn't validate subtask_id nor value.
These args are coming from SRA message and are validated there (afaik).

PR #3616, opened 2018-11-29
Don't check created_time for overdue incomes

Krigpl:
`accepted_ts` is always not null now, so no need to check created_time.


Wiezzel:
I can seee it still being nullable:
https://github.com/golemfactory/golem/blob/10ec48f0d62dd2f8d222efcd20d3213f3dce2c8c/golem/model.py#L296

Wiezzel (approved):
It's fine, I guess.

PR #3619, opened 2018-11-29
Users without Hyper-V can install Golem

Wiezzel:
* Hyper-V prerequisite changed to non-obligatory
* Added `CheckHyperVAvailability` custom action which sets property `HYPER_V_AVAILABLE`
* Added `MessageBox_HyperVUnavailable` custom action which triggers on condition `HYPER_V_AVAILABLE = ""` and displays a warning that Docker Toolbox must be installed manually.
* Changed condition of action `ExecuteUninstallDockerTb` by adding ` AND (HYPER_V_AVAILABLE <> "")`
* Changed condition of action `Execute_PrepareDocker` by adding ` AND (HYPER_V_AVAILABLE <> "")`

Bonus: Fixed issue with expiring golem-docker user


Wiezzel:
@maaktweluit The diff of the .aip file is much bigger than I expected. I'm not sure if I didn't mess up anything.

maaktweluit (approved):
LGTM!

Maybe the inline script `AI_DATA_SETTER_11` could be a file? nitty q so still approve :)

Wiezzel:
@maaktweluit It's a one-liner so I decided no to make into a separate file.

PR #3620, opened 2018-11-30
Put order back in requirements

maaktweluit:
And add a linter check to ensure they are sorted like:

```
LC_ALL=C sort --ignore-case -c requirements.txt


mfranciszkiewicz (approved):
Because why not

Wiezzel (approved):
As I've already said sorting only *package names* instead of lines would be better but this will do as well.

PR #3621, opened 2018-11-30
Fix locks in crossbar for binaries

maaktweluit:
The binary version of golem was no longer working after: #3540

This was caused by the locking code called twice, `start()` is called twice when golem runs as binary.

This patch moves the locking to the ( non crossbar ) version of `start()` only.

### DIFF
--- a/golemapp.py
+++ b/golemapp.py
@@ -4,9 +4,7 @@ import os
 import platform
 import sys
 import logging
-from functools import wraps
 from multiprocessing import freeze_support
-from typing import Callable
 
 import click
 import humanize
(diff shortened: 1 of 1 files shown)


golemapp.py
```diff
@@ -184,59 +142,73 @@ def start(monitor, concent, datadir, node_address, rpc_address, peer, mainnet,
...
-
-    if accept_concent_terms:
-        node.accept_concent_terms()
-
-    if accept_all_terms:
-        node.accept_terms()
-        node.accept_concent_terms()
-
-    node.start()
+    from golem.core.simpleenv import get_local_datadir
```
kmazurek: This is already imported at the top level, no need to re-import locally.

golemapp.py
```diff
@@ -46,46 +46,6 @@ def monkey_patched_getLogger(*args, **kwargs):
...
 
 
-def set_active_environment(
-        mainnet: bool,
-        net: str,
-        concent: str,
-        datadir: str,
-        **kwargs
-) -> str:
-    set_environment('mainnet' if mainnet else net, concent)
```
kmazurek, line 56: This needs to be called from `start` before getting the full datadir path.

mfranciszkiewicz, line 56: This line is missing

golemapp.py
maaktweluit: good one! fixed

golemapp.py
maaktweluit: good one! fixed

maaktweluit: good one! fixed

mfranciszkiewicz (approved):

kmazurek (approved):

PR #3622, opened 2018-12-03
Handle create / start / update config failure in docker

maaktweluit:
- warn when any docker machine fails on create
- cleanup the machine and fail when hyper-v fails to create
- cleanup dynamic memory remains
- added shutdown state to hyper-v machine so it does not save RAM on the host disk

### DIFF
--- a/golem/docker/hypervisor/docker_machine.py
+++ b/golem/docker/hypervisor/docker_machine.py
@@ -40,7 +40,7 @@ class DockerMachineHypervisor(Hypervisor, metaclass=ABCMeta):
         return [self.DRIVER_PARAM_NAME, self.DRIVER_NAME]
 
     @report_calls(Component.hypervisor, 'vm.create')
-    def create(self, vm_name: Optional[str] = None, **params) -> bool:
+    def create(self, vm_name: Optional[str] = None, **params):
         vm_name = vm_name or self._vm_name
         constraints = {
             k: params.pop(v, None)
@@ -52,11 +52,22 @@ class DockerMachineHypervisor(Hypervisor, metaclass=ABCMeta):
 
         try:
             self.command('create', vm_name, args=command_args)
-            return True
+            return
+        except subprocess.CalledProcessError as exc:
+            logger.exception(
+                f'{self.DRIVER_NAME}: error creating VM "{vm_name}"" '
+                f'stdout={exc.stdout.decode('utf8')}, '
+                f'stderr={exc.stderr.decode('utf8')}')
+
+        # Only runs when create fails. Removes a possibly corrupt machine
+        try:
+            self.command('rm', vm_name, args='-f')
         except subprocess.CalledProcessError:
             logger.exception(
-                f'{self.DRIVER_NAME}: error creating VM "{vm_name}"')
-            return False
+                'Failed to clean up a (possible) corrupt machine, '
+                f'please run `docker-machine rm -y -f {vm_name}`')
+
+        raise exc
 
     @contextmanager
     @report_calls(Component.hypervisor, 'vm.recover')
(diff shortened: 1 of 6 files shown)


golem/docker/hypervisor/docker_machine.py
```diff
@@ -28,8 +28,8 @@ def __init__(self,
 
     def setup(self) -> None:
         if self._vm_name not in self.vms:
-            logger.info("Creating Docker VM '%r'", self._vm_name)
-            self.create(self._vm_name, **self._get_config())
+            if not self.create(self._vm_name, **self._get_config()):
```
mfranciszkiewicz, line 31: Could you log `Creating Docker VM` in `parent().create`?

golem/docker/hypervisor/hyperv.py
```diff
@@ -83,19 +79,36 @@ def _parse_create_params(
...
         if mem is not None:
             args += [self.OPTIONS['mem'], str(mem)]
 
-        args += [self.OPTIONS['disk'], self.VOLUME_SIZE]
-
-        logger.debug('raw hyperv create args: args=%r', args)
         return args
 
+    def create(self, vm_name: Optional[str] = None, **params) -> bool:
+        if not super().create(vm_name, **params):
```
mfranciszkiewicz: IMO we could do as you suggested - let's move this logic to the `DockerMachineHypervisor` class. But I'd suggest moving to `setup` and make `create` responsible only for creating the VM (i.e. no recovery process).

golem/docker/hypervisor/docker_machine.py
maaktweluit: I removed line 31 here because it was a duplicate of the one inside create:

https://github.com/golemfactory/golem/blob/epic/docker-for-windows/golem/docker/hypervisor/docker_machine.py#L51

Does this cover your question?

golem/docker/hypervisor/hyperv.py
maaktweluit: ok, will do!

maaktweluit: Updated, please check

golem/docker/hypervisor/hyperv.py
```diff
@@ -83,19 +79,33 @@ def _parse_create_params(
...
-        logger.debug('raw hyperv create args: args=%r', args)
         return args
 
+    def _failed_to_create(self, vm_name: Optional[str] = None):
+        name = vm_name or self._vm_name
+        logger.error(
+            f'{ self.DRIVER_NAME}: VM failed to create, this can be '
+            'caused by insufficient RAM or HD free on the host machine')
+        try:
+            self.command('rm', self._vm_name, args=['-f'])
```
Wiezzel: I was just wondering if this shouldn't be the default behavior in parent class. 🤔 We don't want corrupt VMs for other hypervisors as well, do we?

golem/docker/hypervisor/docker_machine.py
mfranciszkiewicz: Yes, I missed that - thanks

golem/docker/hypervisor/hyperv.py
maaktweluit: - for mac we do not control the VM, so there this will never help
- for virtualbox/toolbox i'm not sure if it is needed, afaik when create fails it can be run again without rm

Wiezzel: Alright, just leave it like this then.

maaktweluit: i'll make an issue to confirm this with toolbox, it is a good idea to check :)

mfranciszkiewicz (approved):

PR #3626, opened 2018-12-03
Prepare new docker images

bidzyyys:
Prepare Golem for running new docker images


apps/core/task/coretask.py
```diff
@@ -122,6 +122,7 @@ def __init__(self,
 
         # src_code stuff
         self.main_program_file = self.environment.main_program_file
+        print(self.main_program_file)
```
nieznanysprawiciel: Is it something created for debugging ?
Maybe it shouldn't pass to final version ???

apps/blender/task/blenderrendertask.py
```diff
@@ -458,28 +458,27 @@ def query_extra_data(self, perf_index: float, num_cores: int = 0,
...
+             "borders_y": [min_y, max_y],
+             }
+        ]
 
-        extra_data = {"path_root": self.main_scene_dir,
-                      "start_task": start_task,
-                      "end_task": end_task,
-                      "total_tasks": self.total_tasks,
-                      "outfilebasename": self.outfilebasename,
+        extra_data = {"RESOURCES_DIR": RESOURCES_DIR_RT,
```
nieznanysprawiciel: RESOURCES_DIR, WORK_DIR and OUPUT_DIR is added to this dictinary by golem in later steps. Why is it necessary here ??

bidzyyys: Ok, I will remove it.

apps/core/task/coretask.py
bidzyyys: I've forgotten to remove it.

apps/entrypoint.sh
```diff
@@ -3,12 +3,12 @@
 if [ "$LOCAL_USER_ID" != "" ]; then
```
nieznanysprawiciel, line 3: Duplicated file.
New entrpoint script was added as apps/core/resources/images/scripts/entrypoint.sh

apps/rendering/task/renderingtask.py
```diff
@@ -213,9 +220,25 @@ def _get_scene_file_rel_path(self):
         else:
             return ''
 
+    @staticmethod
+    def short_extra_data_repr(extra_data):
+        return "RESOURCES_DIR: {RESOURCES_DIR}, " \
```
nieznanysprawiciel: RESOURCES_DIR, WORK_DIR and OUTPUT_DIR are added by Golem automatically.
Are they really necessary here ?

bidzyyys: Nope

nieznanysprawiciel (approved):

apps/blender/blender_reference_generator.py
```diff
@@ -204,7 +201,7 @@ def clear(self):
     # pylint: disable=R0914
     def generate_crops_data(self,
                             resolution: Tuple[int, int],
-                            subtask_border: List[float],
```
magdasta, line 207: please change the docstring accordingly

apps/blender/resources/images/scripts/blender_render.py
```diff
@@ -0,0 +1,230 @@
...
+            sh_file.write(shell_cmd)
+
+        # Add permissions to execute script.
+        st = os.stat(shell_script_file)
+        os.chmod(shell_script_file, st.st_mode | stat.S_IEXEC)
+
+        output_info.append(shell_script_file)
+        crop_counter += 1
+
+    return output_info
```
magdasta, line 235: Pycharm claims: `Expected type 'List[dict]', got 'List[Union[bytes, str]]' instead`

apps/blender/resources/images/scripts/blender_render.py
```diff
@@ -0,0 +1,230 @@
...
+#       {
+#           "outfilebasename" : "output_1.png",
+#           "border_x" : [ 0.0, 1.0 ],
+#           "border_y" : [ 0.0, 1.0 ]
+#       }
+#   ]
+# }
+
+
+def params_to_dict(params) -> dict:
```
magdasta, line 66: Cant this be done using `params.__dict__` or something?

apps/core/resources/images/scripts/install_py_requirements.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
magdasta: this file is suspiciously similar to `apps/core/resources/images/scripts/install_py_libs.sh`

bidzyyys: install_py_libs - for manually provided libs
install_py_requirements - installs libs from file

apps/blender/resources/images/scripts/blender_render.py
bidzyyys: Params are being refactored to be json file, so it will change.

apps/blender/resources/images/scripts/blender_render.py
bidzyyys: Should be `List[str]` :+1:

apps/blender/task/blenderrendertask.py
```diff
@@ -446,35 +435,30 @@ def query_extra_data(self, perf_index: float, num_cores: int = 0,
...
-                      "start_task": start_task,
-                      "total_tasks": self.total_tasks,
-                      "outfilebasename": self.outfilebasename,
-                      "scene_file": scene_file,
-                      "script_src": script_src,
+        crops = [
+            {"outfilebasename": "{}_{}".format(
+                self.outfilebasename, start_task),
+             "borders_x": [0.0, 1.0],
+             "borders_y": [min_y, max_y]}
```
mfranciszkiewicz, line 450: What is the impact of removing the `numpy.float32(min_y)` conversion?

apps/blender/resources/images/blender.Dockerfile
```diff
@@ -18,13 +18,24 @@ RUN apt-get update && \
...
+ENV PATH=/blender:/usr/bin/:$PATH
+ENV PYTHONPATH=/golem/scripts:/golem:$PYTHONPATH
+
+# Create symbolic link to python. I don't know where, something removes it.
+RUN ln -s /usr/bin/python3.6 /usr/bin/python3
+
+COPY scripts/ /golem/scripts/
 
-ENV PATH=/opt/blender:$PATH
+WORKDIR /golem/work/
```
mfranciszkiewicz: Maybe move this line to `golemfactory/base`?

This directory is actually enforced by Golem while creating the container, but since `WORKDIR` is already here, maybe it's a good idea to set the default value.

apps/core/resources/images/scripts/install_py_requirements.sh
mfranciszkiewicz: The only difference is the `-r` flag, the initialization and cleanup are copy-pasted. IMO these should be merged into a single script.

apps/blender/task/blenderrendertask.py
magdasta: Golem-messages can't serialize `float32`s, and we believe this whole `double`-`float32` conversion isn't really necessary - it should be removed completely in the future

mfranciszkiewicz: AFAIR this was introduced because sometimes the border was off by 1 pixel. It's OK if that's no longer the case.

bidzyyys: Ok, but it will be done inside docker - it causes problem with serialization while sending via the internet.

apps/blender/blender_reference_generator.py
```diff
@@ -116,7 +114,7 @@ def create_from_pixel_region(crop_id: str, pixel_region: PixelRegion,
         crop.crop_region = crop.calculate_borders()
         return crop
 
-    def __init__(self, crop_id: str, subimage: SubImage, crops_path: str)\
+    def __init__(self, crop_id: str, subimage: SubImage, crops_path: str) \
```
Krigpl, line 117: ~~Please set your editor to trim whitespace at the line endings.~~
Nvm, I was blind to the `\` character.

mfranciszkiewicz:
I've encountered issues with building Docker images with Golem during startup (macOS). One problem lies in moving the `entrypoint.sh` script.

apps/blender/resources/scripts/docker_blendertask.py
```diff
@@ -1,61 +1,20 @@
...
-    if exit_code is not 0:
-        sys.exit(exit_code)
-
-
-run_blender_task(params.outfilebasename, params.scene_file, params.script_src,
-                 params.start_task, params.frames, params.output_format)
+import blender_render as blender
+
+
+def run_blender_task():
```
Krigpl: So this script doesn't do anything useful anymore and the main script is in the Docker image - wouldn't it make sense to get rid of this file completely?

golem/docker/job.py
```diff
@@ -132,6 +132,10 @@ def _prepare(self):
 
         # Setup volumes for the container
         client = local_client()
+        if is_osx():
```
Krigpl: How come it wasn't needed before?

mfranciszkiewicz: @Krigpl This was accidentally removed in a previous PR and effectively disabled the `su-exec` (privilege drop) part in `entrypoint.sh`.

Krigpl: So this is a bug that made it to 0.19 branch, but this PR most likely won't, right? If not then it should be a separate bug fix PR to the 0.19 branch.

apps/blender/resources/images/blender.Dockerfile
```diff
@@ -2,7 +2,7 @@
 # Blender setup is based on
 # https://github.com/ikester/blender-docker/blob/master/Dockerfile
 
-FROM golemfactory/base:1.2
+FROM golemfactory/base:1.3
 
 MAINTAINER Artur Zawłocki <artur.zawlocki@imapp.pl>
```
Krigpl: Let's change maintainer to Golem Tech

apps/blender/resources/scripts/docker_blendertask.py
nieznanysprawiciel: Blender docker is designed to have internal rendering api (blender_render.py)
This script forwards arguments from params.json. But it's not the only one use case.
Verificator will generate crops rendering parameters itself and then pass them to blender_render.

So it is useful to have entrypoint separated even if it is simple code as this.

golem/docker/job.py
mfranciszkiewicz: @Krigpl Yes, this should be patched for 0.19

mfranciszkiewicz: @bidzyyys This code wasn't removed but has been moved to `docker/task/task_thread.py`

mfranciszkiewicz: Please remove the newly added lines from this file

nieznanysprawiciel: Some tests won't pass after removing these lines.
There's problem with removing files from directories that are created during tests, because they need root privileges as far as I remember @bidzyyys

apps/blender/resources/images/blender.Dockerfile
nieznanysprawiciel: ok but maybe next time, because this image is already published

magdasta: But AFAIK it's not used by anyone yet? We can re-publish it

golem/docker/job.py
bidzyyys: @mfranciszkiewicz I've added these lines because of failing tests on Linux during removal of files created by docker(it works on other platforms). Please discuss with @nieznanysprawiciel whether keep these lines or remove them and fix failing tests.

apps/blender/resources/images/blender.Dockerfile
bidzyyys: This image is not in use currently

PR #3628, opened 2018-12-04
DockerForWindows - Mem check on start_vm

maaktweluit:


golem/docker/hypervisor/hyperv.py
```diff
@@ -50,6 +50,13 @@ def __init__(self, *args, **kwargs):
 
 
     def start_vm(self, name: Optional[str] = None) -> None:
+        constr = self.constraints()
+        if not self._check_memory(constr):
+            logger.warning('Not enough memory to start the VM, lowering memory')
```
mfranciszkiewicz, line 55: The problem here is that this max memory does not reflect the settings in the UI. In fact the user is not in control of the assigned resources. 

I believe this is both a UX issue and a one with max memory calculation based on total memory. ~IMO we should try to force-swap some memory (as we've discussed).~

golem/docker/hypervisor/hyperv.py
```diff
@@ -220,15 +227,32 @@ def create_volumes(self, binds: Iterable[DockerBind]) -> dict:
         }
 
     def pad_memory(self, memory: int) -> int:
+        max_mem_in_mb = self._get_max_memory()
+
+        memory_size = min(memory, max_mem_in_mb - max_mem_in_mb // 10)
```
mfranciszkiewicz, line 232: Maybe this could be useful (in the future):
> The memory cost associated with running a guest operating system on a Hyper-V virtual machine was observed to be approximately 300 MB for the hypervisor, plus 32 MB for the first GB of RAM allocated to each virtual machine, plus another 8 MB for every additional GB of RAM allocated to each virtual machine.

golem/docker/hypervisor/hyperv.py
maaktweluit: As discussed we will add a message to the GUI when this warning is triggered, this way the user is informed and can update its settings or close RAM hungry apps

golem/docker/hypervisor/hyperv.py
maaktweluit: Indeed!

Want to add it now or make a follow up issue to start using this?

mfranciszkiewicz: I believe a follow up issue will be fine

maaktweluit: Ok, made the issue

mfranciszkiewicz (approved):

PR #3632, opened 2018-12-04
Change params.py to params.json

Krigpl:
Replace `params.py` with `params.json` so it isn't so Python-centric and could be used by other technologies as well.


apps/dummy/resources/scripts/docker_dummytask.py
```diff
@@ -1,17 +1,20 @@
 from __future__ import print_function
 
 import imp
+import json
 import os
 
-import params  # This module is generated before this script is run
+
+with open('/golem/work/params.json', 'r') as params_file:
```
maaktweluit: Should this also use the same const `WORK_DIR` as in the file above?

maaktweluit (changes requested):
LGTM!

Will approve when `Golem-Verificator` is updated.

apps/blender/resources/scripts/docker_blendertask.py
```diff
@@ -1,14 +1,19 @@
...
 import subprocess
 import sys
 from multiprocessing import cpu_count
 
-# pylint: disable=import-error
-import params  # This module is generated before this script is run
 
 BLENDER_COMMAND = "blender"
+# FIXME WORK_DIR shoudn't be hardcoded here if it's set in Dockerfile
+WORK_DIR = '/golem/work'
```
mfranciszkiewicz: `WORK_DIR` is exported as an environment variable (see `docker/job.py`)

Krigpl: Also it is explicitly set during container creation so this is not needed at all

maaktweluit (approved):

PR #3636, opened 2018-12-05
Remove end_task param

Krigpl:
This is some artifact from the ancient times and now is always equal to `start_task`


mplebanski (approved):

Wiezzel (approved):
This is fine but it should go with renaming `start_task` because without `end_task` this name doesn't make any sense. It is the ordinal number of a subtask within a task and the name "`start_task`" doesn't express it clearly.

Krigpl:
@Wiezzel Yes.

Krigpl:
But in a separate PR.

PR #3638, opened 2018-12-05
Remove short description

### LINKED ISSUE #2460: Refactor and better usage of short_extra_data representation
Idea behind it was to create short description of what exactly was computed in this subtask. String represenation and a place for metadata. This data may be displayed to user in the interface. Right now displaying is not implemented and the whole varaible itself is pretty messy.

Krigpl:
There's no value in this, #2460 mentions some ideas behind this, but those are neither implemented nor even specified. So it's just useless mess.
Closes #2460


maaktweluit (approved):
Headshot!

mplebanski (approved):

Issue #3624, opened 2018-12-03
Add RPC to report current gas price and the gas limit in SCI


shadeofblue:
The result should be included:

{'current_gas_price': xxxx, 'gas_price_limit': xxxx }

shadeofblue:
@mdtanrikulu could you put any details re: the currently used RPC endpoint that you're using for gas price estimation into this issue here?

mdtanrikulu:
Currently used RPC for estimated tx fee is; `comp.tasks.estimated.cost`. 
And required parameters are; 
```js
{
    type                    // type of task
    options: {
         price              // your bid
         num_subtasks       // given subtask number
         subtask_time       // given subtask timeout
    }
}
```

And the RPC result is; 
```js
{
    GNT,   // Total GNT price
    ETH    // ETH TX fee to lock
}
```

PR #3641, opened 2018-12-05
Created RPC endpoint for current gas price

kmazurek:
Fixes: #3624

### DIFF
--- a/tests/golem/test_client.py
+++ b/tests/golem/test_client.py
@@ -140,8 +140,8 @@ class TestClient(TestClientBase):
 
         result = self.client.get_gas_price()
 
-        self.assertEquals(result["current_gas_price"], str(test_gas_price))
-        self.assertEquals(result["gas_price_limit"], str(test_price_limit))
+        self.assertEqual(result["current_gas_price"], str(test_gas_price))
+        self.assertEqual(result["gas_price_limit"], str(test_price_limit))
 
     def test_get_payments(self, *_):
         ets = self.client.transaction_system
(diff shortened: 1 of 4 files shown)


mdtanrikulu:
But why we have another RPC for it? I was expecting to have the parameters in [this RPC](https://github.com/golemfactory/golem/issues/3624#issuecomment-443711243).

kmazurek:
@mdtanrikulu I created a separate endpoint since the data returned by `comp.tasks.estimated.cost` is specific to tasks, while the added `pay.gas_price` is general and could be used independently.

tests/golem/test_client.py
```diff
@@ -131,6 +131,18 @@ class TestClient(TestClientBase):
...
+        test_gas_price = 1234
+        test_price_limit = 12345
+        ets = self.client.transaction_system
+        ets.gas_price = test_gas_price
+        ets.gas_price_limit = test_price_limit
+
+        result = self.client.get_gas_price()
+
+        assert result["current_gas_price"] == str(test_gas_price)
+        assert result["gas_price_limit"] == str(test_price_limit)
```
jiivan: I generally prefer `self.assert*` methods from `unittest.TestCase`.

jiivan (approved):

tests/golem/test_client.py
```diff
@@ -140,8 +140,8 @@ def test_get_gas_price(self, *_):
 
         result = self.client.get_gas_price()
 
-        assert result["current_gas_price"] == str(test_gas_price)
-        assert result["gas_price_limit"] == str(test_price_limit)
+        self.assertEquals(result["current_gas_price"], str(test_gas_price))
```
jiivan: `assertEquals` is deprecated in favor of `assertEqual` (without s)

kmazurek: Good catch, sorry

PR #3645, opened 2018-12-06
Check if SMB port (445) is not blocked by firewall

Wiezzel:


golem/docker/commands/docker_machine.py
```diff
@@ -15,6 +15,7 @@ class DockerMachineCommandHandler(DockerCommandHandler):
         inspect=['docker-machine', 'inspect'],
         regenerate_certs=['docker-machine', 'regenerate-certs', '--force'],
         restart=['docker-machine', 'restart'],
+        execute=['docker-machine', 'ssh'],
```
maaktweluit, line 18: i think the name is a bit inconsistent, `execute` for `ssh`
any reason behind this?

maaktweluit (approved):
LGTM!

One nitty q

golem/docker/commands/docker_machine.py
Wiezzel: Well, this is about executing a command on a VM. I think it wasn't a good decision for docker-machine developers to name it `ssh`. Docker itself uses `exec` for this kind of command. Therefore I decided to rename it in our code.

maaktweluit: Ok i understand now but still do not agree

looking at the list only `ls` is replaced by `list`. the rest is named as the `docker-machine` command.

for `docker excec` i think execute would be a good name, but this is `docker-machine ssh`

Nitty comment so can leave it, what you perfer :)

Wiezzel: >looking at the list only ls is replaced by list. the rest is named as the docker-machine command.

Well, that's because other docker-machine commands are named reasonably.

maaktweluit: I think the most reasonable name is the one consistent with the underlying command, not the one we subjectively think is more applicable

PR #3646, opened 2018-12-06
Initialize hardware presets in Node.start

mfranciszkiewicz:
Should significantly improve Docker VM startup times. The VM is no longer restarted after the first configuration change in `Client`


maaktweluit (approved):
LGTM!

PR #3647, opened 2018-12-06
Empty process' working sets to try and free some memory

mfranciszkiewicz:


.gitignore
```diff
@@ -23,6 +23,7 @@ ui_*.py
 
 # Rust
 rust/golem/target/*
+rust/golem/Cargo.lock
```
Wiezzel: This should **not** be ignored but committed. ([Ref](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries))

Wiezzel (dismissed):
Please restore `Cargo.lock`

mfranciszkiewicz:
@Wiezzel I'm unsure how the conditional dependencies will be handled on other OSes though

golem/docker/hypervisor/hyperv.py
```diff
@@ -48,13 +47,23 @@ def __init__(self, *args, **kwargs):
...
     def start_vm(self, name: Optional[str] = None) -> None:
+
+        if not self._check_memory():
+            try:
+                from golem.os import windows_empty_working_sets
+                windows_empty_working_sets()
+            except (ImportError, OSError) as e:
+                logger.error('Unable to free memory: %r', e)
+
         constr = self.constraints()
```
maaktweluit, line 68: this can be done first, so it is only done once

`empty_..()` will not change the VM configuration so 1x is enough

golem/docker/hypervisor/hyperv.py
```diff
@@ -226,12 +234,9 @@ def create_volumes(self, binds: Iterable[DockerBind]) -> dict:
             for bind in binds
         }
 
-    def pad_memory(self, memory: int) -> int:
```
maaktweluit, line 243: is this meant to be no longer done on `build_config`?

maaktweluit (approved):
LGTM!

2 small comments, will approve on reply

maaktweluit (changes requested):
Oops, selected the wrong option by accident

golem/docker/hypervisor/hyperv.py
mfranciszkiewicz: Well, it's clearer when `pad_memory` doesn't do more than its name says. Besides, I can see no benefit in storing a volatile configuration - it will almost always differ from user's settings and cause a VM restart.

Actually, there is one more thing to do for comparing configurations - to make them aware of padding.

maaktweluit: fully agree

and since start_vm is called after build_config the free check is still done

resolved

golem/docker/hypervisor/hyperv.py
```diff
@@ -226,12 +233,9 @@ def create_volumes(self, binds: Iterable[DockerBind]) -> dict:
...
+    def _memory_cap(self, memory: int) -> int:
         max_mem_in_mb = self._get_max_memory()
```
Wiezzel, line 256: I dunno how costly it is to call `_get_max_memory()` but if it takes long we could get rid of this call (we already retrieve the max memory in `_check_memory()`). This however would probably make our code uglier so only if `_get_max_memory()` is really costly.

rust/golem/src/os/windows.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub fn new() -> Self {
+        Processes {
+            array: [0; MAX_PROCESS_COUNT],
+            count: 0,
+        }
+    }
+
+    pub fn enumerate(&mut self) -> Result<(), OSError> {
+        // A pointer to an array that receives the list of process identifiers
+        let lpid_process = &mut self.array[0];
```
Wiezzel, line 32: This name is kinda cryptic. Why not simply `process_ids`?

rust/golem/src/os/windows.rs
```diff
@@ -0,0 +1,84 @@
...
+        // The size of the pProcessIds array, in bytes
```
Wiezzel: What is `pProcessIds` array? Do you mean `lpid_process`?

rust/golem/src/os/windows.rs
```diff
@@ -0,0 +1,84 @@
...
+        let cb: DWORD = (size_of::<DWORD>() * MAX_PROCESS_COUNT) as DWORD;
+        // The number of bytes returned in the pProcessIds array
+        let mut cb_needed: DWORD = 0;
+        // A pointer to the number of bytes returned
+        let lpcb_needed: LPDWORD = &mut cb_needed;
```
Wiezzel, line 38: Can't we just use `cb_needed`? Creating a separate variable for the reference kinda obscures the code for me.

rust/golem/src/os/windows.rs
Wiezzel: Okay, I see now that this is WinAPI's Hungarian convention. But do we really need to follow it? It's totally confusing for me.

rust/golem/src/os/windows.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl Processes {
+    pub fn new() -> Self {
+        Processes {
+            array: [0; MAX_PROCESS_COUNT],
+            count: 0,
+        }
+    }
+
+    pub fn enumerate(&mut self) -> Result<(), OSError> {
```
Wiezzel: `enumerate` seems like a name of a function that returns some iterator. Maybe just `init`? Or even put this in `new`?

rust/golem/src/os/windows.rs
```diff
@@ -0,0 +1,84 @@
...
+                return Err(OSError::new("Unable to enumerate processes"));
+            }
+
+            self.count = (*lpcb_needed) / size_of::<DWORD>() as u32;
+        }
+
+        Ok(())
+    }
+
+    pub fn open(&self, idx: usize) -> HANDLE {
```
Wiezzel: I'd go with a name like `get_process_handle` instead of `open`.

rust/golem/src/os/windows.rs
```diff
@@ -0,0 +1,84 @@
...
+        let id = self.array[idx];
+        if id == 0 {
+            return NULL;
+        }
+
+        unsafe {
+            OpenProcess(
+                PROCESS_QUERY_INFORMATION | PROCESS_VM_READ | PROCESS_ALL_ACCESS,
```
Wiezzel: Doesn't `PROCESS_ALL_ACCESS` cover all permissions? If so why use the other two flags?

golem/docker/hypervisor/hyperv.py
```diff
@@ -48,13 +47,22 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._vm_utils = VMUtils()
 
-
     def start_vm(self, name: Optional[str] = None) -> None:
         constr = self.constraints()
+
+        if not self._check_memory(constr):
+            try:
```
Wiezzel, line 73: Maybe add some `logger.info()` here to inform about the performed operation?

maaktweluit (approved):

golem/docker/hypervisor/hyperv.py
```diff
@@ -48,13 +47,22 @@ def __init__(self, *args, **kwargs):
...
+                from golem.os import windows_ews
+                windows_ews()
+            except (ImportError, OSError) as e:
+                logger.error('Unable to free memory: %r', e)
```
Wiezzel: Using `logger.exception()` you don't need to pass the error explicitly. Just a convenience.

rust/golem/src/os/windows.rs
```diff
@@ -0,0 +1,82 @@
...
+use winapi::shared::ntdef::HANDLE;
+use winapi::shared::ntdef::NULL;
+use winapi::um::processthreadsapi::OpenProcess;
+use winapi::um::psapi::EmptyWorkingSet;
+use winapi::um::psapi::EnumProcesses;
+use winapi::um::winnt::PROCESS_ALL_ACCESS;
+use winapi::um::winnt::PROCESS_QUERY_INFORMATION;
+use winapi::um::winnt::PROCESS_VM_READ;
+
+const MAX_PROCESS_COUNT: usize = 2048;
```
Wiezzel, line 15: What will happen if there are more? Will `EnumProcesses` fail or just limit the result?

Wiezzel (commented):
I have some doubts about naming conventions used in the Rust code.

rust/golem/src/os/windows.rs
mfranciszkiewicz: Taken straight from ms docs, will update

rust/golem/src/os/windows.rs
mfranciszkiewicz: It will limit the result

golem/docker/hypervisor/hyperv.py
mfranciszkiewicz: I don't believe it's something we should optimize for, even if it takes 20ms.

rust/golem/src/os/windows.rs
mfranciszkiewicz: I cannot see why it obscures code since it strictly follows the API and naming conventions.

rust/golem/src/os/windows.rs
mfranciszkiewicz: Sure, I can rename it. It's called `enumerate` because the main thing it does is call `EnumProcesses`

rust/golem/src/os/windows.rs
mfranciszkiewicz: Same as above - it calls `OpenProcess`

rust/golem/src/os/windows.rs
Wiezzel: I just meant that maybe we don't need a separate variable just to hold a reference to another variable. We could inline it. But if you feel like it's necessary feel free to ignore this remark.

Wiezzel (approved):

PR #3648, opened 2018-12-06
Speed up Golem startup

Wiezzel:
### Use saving VM state to speed up Golem startup

If possible make use of saving/restoring VM state. Otherwise, fall back to regular start/stop. Restoring a suspended Hyper-V machine is about 10x faster then running `docker-machine start`.

### Custom script to replace docker-machine start

Docker-machine start command is very slow (~60 s) because it establishes SSH connection to the VM *thrice*. Using a custom powershell script we can start the VM in under 30 s.

###  Check if there's enough disk space to dump VM memory

Setting "save" as host shutdown action for VM requires pre-allocating disk space for memory dump. The VM will not start if there is not enough space on system drive. Therefore we check the available disk space
during VM reconfiguration and fall back to "shutdown" action if there is not enough.

### Check disk space before VM creation

Docker-machine sets host shutdown action to 'save' and there is no option to override it. This is the preferred action but using it requires pre-allocating disk space for memory dump on VM startup. If
there is not enough space we should try to create the VM with minimal amount of RAM and reconfigure it later to use 'shutdown' host shutdown action.


mfranciszkiewicz:
Is it possible to change the amount of memory assigned when the VM is suspended?

Wiezzel:
@mfranciszkiewicz No. This is why we still use `stop_vm()` and `start_vm()` in `restart_ctx()`.

golem/docker/hypervisor/hyperv.py
```diff
@@ -138,7 +161,7 @@ def constrain(self, name: Optional[str] = None, **params) -> None:
                 vcpus_per_numa_node=0,
                 limit_cpu_features=False,
                 dynamic_mem_ratio=1,
-                host_shutdown_action=HOST_SHUTDOWN_ACTION_SHUTDOWN,
+                host_shutdown_action=HOST_SHUTDOWN_ACTION_SAVE,
```
maaktweluit: This is the default option, so it could be removed too, both is fine for me

golem/docker/hypervisor/hyperv.py
```diff
@@ -49,11 +50,33 @@ def __init__(self, *args, **kwargs):
         self._vm_utils = VMUtils()
 
 
+    def save_vm(self, vm_name: Optional[str] = None) -> None:
```
maaktweluit, line 72: Since `host_shutdown_action` action is now save_state, maybe we do not have to manually save it.

I think it is worth to test the difference to make sure

maaktweluit (changes requested):
LGTM!

I removed the save state last PR all together because it will lock the full RAM amount on the host disk in a save-state file. 
Is this worth that amount of disk space?
If the answer is yes, maybe we also need to check disk space before starting like memory.

Some questions, will approve on reply

golem/docker/hypervisor/hyperv.py
Wiezzel: I believe `host_shutdown_action` is the action triggered when host OS is being shut down. We still need to do it ourselves when shutting Golem.

maaktweluit: makes total sense!

ok, then this is the way to go

golem/docker/hypervisor/hyperv.py
Wiezzel: I'll leave explicit.

Wiezzel:
Update: Test show that suspending VMs also prevents IP changes. Which kinda makes sense because the VM would only make DHCP request during boot.

Wiezzel:
Running `docker-machine start` actually makes **three** SSH connections to the VM after it is started:
1) To check if SSH is available:
https://github.com/docker/machine/blob/eae38c196a2e10e4e11cc7b8d6dfe8863fb60519/libmachine/provision/provisioner.go#L104-L107
2) To check what OS is running inside them VM so that appropriate provisioner class could be chosen:
https://github.com/docker/machine/blob/eae38c196a2e10e4e11cc7b8d6dfe8863fb60519/libmachine/provision/provisioner.go#L109-L114
3) To check if docker daemon is listening:
https://github.com/docker/machine/blob/7a9ce457496353549916e840874012a97e3d2782/libmachine/provision/utils.go#L257-L258

A single `docker-machine ssh` command takes ~12 seconds. This is quite suspicious but still none of these ssh connections is really necessary. To check if Docker daemon is running one can simply scan the appropriate port from outside the VM (if the env is set properly a simple `docker info` will do).

Wiezzel:
@maaktweluit @mfranciszkiewicz 
This PR is ready for final review. I updated the description to cover all the changes.

maaktweluit (approved):
LGTM!

mfranciszkiewicz (approved):

Issue #3378, opened 2018-10-03
Exception after adding a task on Requestor's end


shadeofblue:
Doesn't happen every time but every ~3-4 tasks...

```
Traceback (most recent call last):
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/twisted/python/log.py", line 103, in callWithLogger
    return callWithContext({"system": lp}, func, *args, **kw)
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/twisted/python/log.py", line 86, in callWithContext
    return context.call({ILogContext: newCtx}, func, *args, **kw)
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/twisted/python/context.py", line 122, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/twisted/python/context.py", line 85, in callWithContext
    return func(*args,**kw)
--- <exception caught here> ---
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/twisted/internet/posixbase.py", line 614, in _doReadOrWrite
    why = selectable.doRead()
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/twisted/internet/tcp.py", line 205, in doRead
    return self._dataReceived(data)
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/twisted/internet/tcp.py", line 211, in _dataReceived
    rval = self.protocol.dataReceived(data)
  File "/home/blue/.virtualenvs/golem/lib/python3.6/site-packages/twisted/internet/endpoints.py", line 125, in dataReceived
    return self._wrappedProtocol.dataReceived(data)
  File "/home/blue/PycharmProjects/golem/golem/network/transport/tcpnetwork.py", line 333, in dataReceived
    self._interpret(data)
  File "/home/blue/PycharmProjects/golem/golem/network/transport/tcpnetwork.py", line 359, in _interpret
    self.session.interpret(m)
  File "/home/blue/PycharmProjects/golem/golem/network/p2p/peersession.py", line 117, in interpret
    BasicSafeSession.interpret(self, msg)
  File "/home/blue/PycharmProjects/golem/golem/network/transport/session.py", line 75, in interpret
    action(msg)
[...]

PR #3649, opened 2018-12-06
Pause p2p and task sessions until node is ready (#3378)

kmazurek:
Fixes: #3378

This fixes a race condition which occured in rare cases on node startup. This was caused by P2P messages being received and interpreted before the node has restored its resources from HyperG. The problem occured only when the resources were not cached and needed to be re-uploaded.


jiivan (approved):
i-think-its-ok

mfranciszkiewicz (approved):

PR #3655, opened 2018-12-07
Move TaskHeader to golem-messages

### LINKED ISSUE #3557: Move TaskHeader to golem-messages
Also remove `FixedTaskHeader` and leave only `TaskHeader`. Since both can only be modified by requestor there is no need for both entities.

related: #80 #2407

jiivan:
fixes: #3557 
depends on: https://github.com/golemfactory/golem-messages/pull/285


apps/core/task/coretask.py
```diff
@@ -137,16 +137,17 @@ def __init__(self,
...
             environment=self.environment.get_id(),
             task_owner=owner,
             deadline=self._deadline,
-            subtask_timeout=task_definition.subtask_timeout,
+            subtask_timeout=float(task_definition.subtask_timeout),
             subtasks_count=total_tasks,
             resource_size=self.resource_size,
             estimated_memory=task_definition.estimated_memory,
-            max_price=task_definition.max_price,
+            max_price=int(task_definition.max_price),
```
etam: Why is this cast required? From glancing quickly at the code it seems to me that it should always be int. If it's not, then I think it's a bug and should be corrected elsewhere.

etam (commented):
Why are you changing all timeouts from ints to floats?

apps/rendering/task/renderingtaskstate.py
```diff
@@ -13,11 +13,11 @@ def __init__(self):
         self._pixel_to_seconds = 384
 
     @property
-    def subtask_timeout(self):
+    def subtask_timeout(self) -> float:
         return self.resolution[0] * self.resolution[1] / self._pixel_to_seconds
```
etam: Everywhere else timeouts were ints. They come from task definition in format "hh:mm:ss" and are converted to ints.

I think you should add here conversion to int and revert all conversions to floats you did in this PR.

etam (changes requested):

apps/rendering/task/renderingtaskstate.py
jiivan: I generally agree that decimal precision is not needed in timestamps, but would like to get another opinion.
Also about:
 1. `deadline`
 1. `subtask_timeout`
 1. `timestamp`

jiivan: This default value influenced my decision to float conversion: https://github.com/golemfactory/golem/pull/3655/files#diff-798280b9faa873b5f61d16829c133666L68

etam: `timeout` and `subtask_timeout` are given by requesting user and are a relative number of seconds. Therefore ints.

`timestamp` and `deadline` are absolute timestamp values. On the one hand I think that we don't need sub-second granularity and int would be sufficient, but on the other hand `time.time()` and our `get_timestamp_utc` are returning floats and changing that would require lots of changes all over the place, so I'm willing to accept status quo.

apps/core/task/coretask.py
jiivan: Fixed

etam (approved):
ok

kmazurek (approved):

prekucki:
$ golemcli network show
Exception: success return value from invoked procedure \\"net.peers.connected\\" could not be serialized: WAMP message serialization error: unsupported type: <class \'golem_messages.datastructures.p2p.Node\'>

jiivan:
> $ golemcli network show
> Exception: success return value from invoked procedure \"net.peers.connected\" could not be serialized: WAMP message serialization error: unsupported type: <class 'golem_messages.datastructures.p2p.Node'>

https://github.com/golemfactory/golem/pull/3676

PR #3662, opened 2018-12-12
Missing memory padding

mfranciszkiewicz:


maaktweluit (approved):
LGTM!

golem/docker/manager.py
```diff
@@ -5,6 +5,7 @@
 from threading import Thread
 from typing import Optional, Callable, Any, Iterable
 
+from golem import hardware
```
maaktweluit, line 8: `golem.hardware` is now imported "twice", maybe we can use on or the other?

PR #3665, opened 2018-12-13
[Concent] Soft-enable/disable toggle through RPC

### LINKED ISSUE #3623: [Concent] Soft-enable/disable toggle through RPC
+ accepting Concent ToS shouldn't prevent anyone from launching `golemapp`

jiivan:
fixes: #3623 

Added remote procedures:
`golem.concent.switch() -> bool` Concent soft switch state
`golem.concent.switch.turn(bool) -> None` Modify switch state

### DIFF
--- a/golem/interface/command.py
+++ b/golem/interface/command.py
@@ -432,8 +432,11 @@ class CommandHelper(object):
 
     @staticmethod
     def _public_method(entry):
-        return inspect.ismethod(entry) or inspect.isfunction(entry) \
-            and not entry.__name__.startswith('_')
+        if not (inspect.ismethod(entry) or inspect.isfunction(entry)):
+            return False
+        if entry.__name__.startswith('_'):
+            return False
+        return True
 
     @classmethod
     def debug(cls, elem, level=0):
--- a/golem/network/concent/client.py
+++ b/golem/network/concent/client.py
@@ -21,6 +21,7 @@ from golem.core import variables
 from golem.network.concent import exceptions
 from golem.network.concent.handlers_library import library
 
+from . import soft_switch
 from .helpers import ssl_kwargs
 
 logger = logging.getLogger(__name__)
@@ -200,7 +201,14 @@ class ConcentClientService(threading.Thread):
 
     @property
     def enabled(self):
+        """Indicates whether this client will communicate with
+            Remote Concent Service"""
         return None not in self.variant.values()
+
+    @property
+    def fully_enabled(self) -> bool:
+        """Indicates whether this client is enabled and user turned it on"""
+        return self.enabled and soft_switch.is_on()
 
     def run(self) -> None:
         last_receive = 0.0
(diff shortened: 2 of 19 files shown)


golem/interface/command.py
```diff
@@ -432,8 +432,11 @@ def is_callable(elem):
 
     @staticmethod
     def _public_method(entry):
-        return inspect.ismethod(entry) or inspect.isfunction(entry) \
-            and not entry.__name__.startswith('_')
+        if not (inspect.ismethod(entry) or inspect.isfunction(entry)):
+            return False
+        if entry.__name__.startswith('_'):
+            return False
+        return True
```
Wiezzel, line 439: IMO the previous version was more readable.

Wiezzel (approved):
Looks fine to me. There are only two issues:
1. I disapprove of the way `_public_method()` was refactored.
1. "Enabled" and "Fully enabled" is kinda awkward naming. I'd rather go with "available" & "enabled" or something like that.

golem/interface/command.py
jiivan: Really? Did you notice it didn't work? :)

Wiezzel: OK, there's missing parenthesis in the first line but still looks more intuitive.

golem/network/concent/client.py
```diff
@@ -163,21 +165,14 @@ def receive_from_concent(
     return response.content or None
 
 
-class ConcentRequest(msg_datastructures.FrozenDict):
```
shadeofblue, line 166: @jiivan why did we get rid of the structure? I think it does make sense to wrap those kinds of pieces of functionality in self-contained classes like this...

golem/network/concent/soft_switch.py
```diff
@@ -0,0 +1,32 @@
...
+        GenericKeyValue.key == KEY,
+    ).limit(1)
+    try:
+        value = SoftSwitch(query[0].value)
+    except IndexError:
+        value = DEFAULT
+    return value == SoftSwitch.ON
+
+
+@rpc_utils.expose('golem.concent.switch.turn')
```
shadeofblue, line 28: wouldn't it be better to have explicit on/off calls instead of a toggle?

golem/task/rpc.py
```diff
@@ -85,11 +84,17 @@ def _validate_task_dict(client, task_dict) -> None:
                 )
             )
 
-    if task_dict['concent_enabled'] and \
-            not client.concent_service.enabled:
-        raise CreateTaskError(
-            "Cannot create task with concent enabled when "
-            "concent service is disabled")
+    if task_dict['concent_enabled']:
+        if not client.concent_service.available:
```
shadeofblue: I'd prefer something like this here:

```
if not client.concent_service.enabled:  # `enabled` implies `available`
    raise CreateTaskError(
        "Cannot create task with concent enabled when "
        "Concent Service is " + \
        ('switched off' if available else 'disabled')
    )
```

golem/network/concent/client.py
jiivan: My point is that there was no functionality in this structure. Only a message. The key could probably be factored out, because it provides functionality that's not hypothetical (not used).

PR #3669, opened 2018-12-14
Fixed linter and add missing mock

maaktweluit:


Wiezzel (approved):
If buildbot says it's OK then I say it's OK.

PR #3672, opened 2018-12-17
Removed default of concent_variant from client init params

maaktweluit:
This default value was only used in unit tests and it broke the linter.


jiivan (approved):
trisie3

mfranciszkiewicz (approved):

PR #3673, opened 2018-12-17
Removed union, it is no longer required for the linter.

maaktweluit:


Wiezzel (approved):
If mypy says it's fine then I say it's fine.

PR #3676, opened 2018-12-17
DictSerializer aka DictCoder learns about p2p.Node

jiivan:
```
$ golemcli network show
Exception: success return value from invoked procedure \"net.peers.connected\" could not be serialized: WAMP message serialization error: unsupported type: <class 'golem_messages.datastructures.p2p.Node'>
```

### DIFF
--- a/golem/core/simpleserializer.py
+++ b/golem/core/simpleserializer.py
@@ -6,7 +6,7 @@ import types
 from abc import ABCMeta, abstractmethod
 from typing import Optional
 
-from golem_messages.datastructures import p2p as dt_p2p
+from golem_messages import datastructures
 
 from golem.core.common import to_unicode
 
@@ -80,7 +80,7 @@ class DictCoder:
             return obj.__class__(
                 [cls._to_dict_traverse_obj(o, typed) for o in obj]
             )
-        elif isinstance(obj, dt_p2p.Node):
+        elif isinstance(obj, datastructures.Container):
             return obj.to_dict()
         elif cls.deep_serialization:
             if hasattr(obj, '__dict__') and not cls._is_builtin(obj):
--- a/golem/core/simpleserializer.py
+++ b/golem/core/simpleserializer.py
@@ -6,6 +6,8 @@ import types
 from abc import ABCMeta, abstractmethod
 from typing import Optional
 
+from golem_messages.datastructures import p2p as dt_p2p
+
 from golem.core.common import to_unicode
 
 logger = logging.getLogger('golem.core.simpleserializer')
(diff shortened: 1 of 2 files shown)


golem/core/simpleserializer.py
```diff
@@ -102,7 +112,9 @@ def _is_class(cls, obj):
 
     @classmethod
     def _is_builtin(cls, obj):
-        return type(obj) in cls.builtin_types and not isinstance(obj, types.InstanceType)
+        # pylint: disable=unidiomatic-typecheck
```
etam, line 115: you can change `builtin_types` to be a tuple
```python
builtin_types = tuple(i for i in types.__dict__.values() if isinstance(i, type))
```
and then instead of
```python
type(obj) in cls.builtin_types
```
you can use
```python
isinstance(obj, cls.builtin_types)
```

golem/core/simpleserializer.py
```diff
@@ -102,7 +112,9 @@ def _is_class(cls, obj):
 
     @classmethod
     def _is_builtin(cls, obj):
-        return type(obj) in cls.builtin_types and not isinstance(obj, types.InstanceType)
+        # pylint: disable=unidiomatic-typecheck
+        return type(obj) in cls.builtin_types \
+            and not isinstance(obj, types.InstanceType)
```
etam, line 117: `types.InstanceType` seems to not exist in python3
https://stackoverflow.com/questions/9548936/how-to-determine-the-type-of-class-instance

golem/core/simpleserializer.py
```diff
@@ -70,8 +73,15 @@ def _to_dict_traverse_obj(cls, obj, typed=True):
...
-                logger.warning('set/frozenset have known problems with umsgpack: %r', obj)
-            return obj.__class__([cls._to_dict_traverse_obj(o, typed) for o in obj])
+                logger.warning(
+                    'set/frozenset have known problems with umsgpack: %r',
+                    obj,
+                )
+            return obj.__class__(
+                [cls._to_dict_traverse_obj(o, typed) for o in obj]
+            )
+        elif isinstance(obj, dt_p2p.Node):
```
etam: `datastructures.Container`?

mfranciszkiewicz (approved):

golem/core/simpleserializer.py
jiivan: Nice idea, but outside of this PR (I've only changed indentation to pass linter tests). IMHO these serializers need a total rewrite so I'm reluctant to work to much on this implementation.

PR #3681, opened 2018-12-18
Hyper-V RPC events

mfranciszkiewicz:
~Currently for `epic/docker-for-windows`, will be retargeted at `develop` after that branch is merged.~

To be merged into b0.19.


golem/docker/hypervisor/hyperv.py
```diff
@@ -16,10 +17,33 @@
...
+    'lowered_memory': {
+        'component': Component.hypervisor,
+        'method': 'start_vm',
+        'stage': Stage.warning,
+        'data': MESSAGES['lowered_memory'],
+    },
+}
+
+
+def publish_event(event: Dict) -> None:
```
Wiezzel, line 65: Maybe `publish_event` could just take event name instead of dict?

golem/docker/hypervisor/hyperv.py
```diff
@@ -16,10 +17,33 @@
...
+MESSAGES = {
+    'lowered_memory': 'Not enough free RAM to start the VM, lowering memory',
+}
+EVENTS = {
+    'lowered_memory': {
+        'component': Component.hypervisor,
+        'method': 'start_vm',
+        'stage': Stage.warning,
+        'data': MESSAGES['lowered_memory'],
+    },
```
Wiezzel, line 61: I think that the amount of memory assigned *after* lowering is an important information and should be included in the event.

golem/report.py
```diff
@@ -62,9 +65,31 @@ def last_status(cls):
...
+
+    @classmethod
+    def _publish_listener(cls, event: str = 'default', **kwargs) -> None:
+        if event != 'publish':
+            return
+
+        cls.publish(kwargs['component'],
+                    kwargs['method'],
+                    kwargs['stage'],
+                    kwargs.get('data'))
```
Wiezzel, line 103: Could you explain why are these changes to `StatusPublisher` necessary

Wiezzel (commented):
Please change base of this PR to `develop`.

maaktweluit (approved):

golem/docker/hypervisor/hyperv.py
mfranciszkiewicz: It could be, at first I implemented it like that. Somehow it feels like the event as an argument fits `publish_event` better, but that's my opinion

golem/report.py
mfranciszkiewicz: - `set_publisher` was the initialization function, called once. Now it does more, so it's been renamed
- registering the event handler here keeps the relevant logic in one file
- we avoid coupling using the dispatcher

golem/docker/hypervisor/hyperv.py
mfranciszkiewicz: It proved useful for implementing your suggestion below.

golem/docker/hypervisor/hyperv.py
```diff
@@ -108,15 +133,14 @@ def start_vm(self, name: Optional[str] = None) -> None:
...
             max_memory = self._memory_cap(constr[mem_key])
             constr[mem_key] = hardware.cap_memory(constr[mem_key], max_memory,
                                                   unit=hardware.MemSize.mebi)
-            logger.debug('Memory capped by "free - 10%%": %r', constr[mem_key])
 
-        # Always constrain to set the appropriate shutdown action
-        self.constrain(name, **constr)
+            self._lowered_memory_event(constr[mem_key])
+            # Always constrain to set the appropriate shutdown action
+            self.constrain(name, **constr)
```
Wiezzel: Constrain should be **outside** this if.

Wiezzel (dismissed):

golem/docker/hypervisor/hyperv.py
mfranciszkiewicz: Yes, it should. It was mistakenly auto-indented by the editor.

Wiezzel (approved):

PR #3682, opened 2018-12-18
when creating task with concent enabled, make sure deposit is locked

### LINKED ISSUE #3612: transaction system's `concent_deposit` function doesn't check if the existing deposit is locked
When checking for the existing deposit, we're not ensuring the deposit is in fact locked.

1) When doing my deposit, check if it's locked (in `concent_deposit`)
2) When as provider checking requestor deposit also check if it's locked.

etam:
resolves #3612

### DIFF
--- a/golem/ethereum/transactionsystem.py
+++ b/golem/ethereum/transactionsystem.py
@@ -111,12 +111,14 @@ class TransactionSystem(LoopingCallService):
     @property   # type: ignore
     @sci_required()
     def gas_price(self) -> int:
+        self._sci: SmartContractsInterface
         return self._sci.get_current_gas_price()
 
     @property   # type: ignore
     @sci_required()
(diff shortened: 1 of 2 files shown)


golem/task/rpc.py
```diff
@@ -216,6 +216,7 @@ def _ensure_task_deposit(client, task, force):
             expected=opt_amount,
             force=force,
         )
+        client.transaction_system.concent_relock()
```
jiivan: This shouldn't be called here.

jiivan (changes requested):

golem/task/rpc.py
etam: Why?

etam:
I've force-pushed a completely new change, so you can safely ignore what was before.

jiivan (approved):
trisie3

PR #3684, opened 2018-12-18
Hide withdraw in golecli account if not mainnet

### LINKED ISSUE #3012: `account withdraw` should not be presented in golemcli for testnet

dybi:
resolves: https://github.com/golemfactory/golem/issues/3012


golem/interface/cli.py
```diff
@@ -40,6 +41,19 @@ def exit(self, status=0, message=None):
         raise ParsingException(message, self)
 
 
+def adapt_children(children: Dict[Text, Callable]) -> Dict[Text, Callable]:
```
shadeofblue: @dybi  ... on the surface, this function seems generic but really it's only applicable to a very specific case and hardly reusable... wouldn't it make sense to just get rid of that key specifically in the `_build_parser` without having a separate function for this task?

dybi: @shadeofblue, you've got a point - the name of the function is too generic. However, I would be for having it separated from the rest of the code for the sake of better modularity and unit-testing.

golem/interface/cli.py
```diff
@@ -40,6 +41,20 @@ def exit(self, status=0, message=None):
         raise ParsingException(message, self)
 
 
+def adapt_parser_children_to_target_net(
+        children: Dict[Text, Callable]) -> Dict[Text, Callable]:
+    """
+    This function adapts children of an interface: if golemcli is not run on
+    mainnet, there should be no option to `withdraw` for `golemcli account`
+    """
+    new_children = {k: v for k, v in children.items()}
```
shadeofblue: wouldn't `copy.deepcopy` seem more natural here?

golem/interface/cli.py
```diff
@@ -40,6 +42,20 @@ def exit(self, status=0, message=None):
...
 
 
+def disable_withdraw(
+        children: Dict[Text, Callable]) -> Dict[Text, Callable]:
+    """
+    This function adapts children of an interface: if golemcli is not run on
+    mainnet, there should be no option to `withdraw` for `golemcli account`
+    """
+    new_children = deepcopy(children)
+    from golem.config.active import IS_MAINNET
```
Krigpl: Please import and use `WITHDRAWALS_ENABLED` instead

Krigpl (changes requested):

shadeofblue (approved):
looks good as far as I can tell.

@Krigpl , could you confirm?

jiivan (commented):
IMHO it would be nice to explicitly disable a method in `golem.interface.client.account.Account.withdraw()`. Possibly by adding a modifier to `@command` decorator.

golem/interface/command.py
```diff
@@ -243,6 +251,19 @@ def sort(headers, values, key):
...
+def filter_out_disabled(
+        methods: List[Tuple[Text, Callable]]
+) -> List[Tuple[Text, Callable]]:
+    filtered: List[Tuple[Text, Callable]] = []
+    for name, m in methods:
+        if hasattr(m, 'enabled'):
+            if m.enabled:  # type: ignore
+                filtered.append((name, m))
+        else:
+            filtered.append((name, m))
```
jiivan: `if getattr(m, 'enabled', False):`

jiivan: `return [(name, m) for name, m in methods if getattr(m, 'enabled', False)]`
I'm not sure if this list comprehension is worthy of a separate function. I would just use it in `CommandHelper.get_methods()`

dybi:
@jiivan the problem with the proposed approach is that `@group` and `@command` decorators are evaluated at import and whether we are running on miannet or not is determined later (after arguments are parsed)

golem/interface/command.py
dybi: it is no longer there - as stated above: 
> `@group` and `@command` decorators are evaluated at import and whether we are running on miannet or not is determined later (after arguments are parsed)

so it seems not to work properly: at the time `command_enabled` is evaluated  `IS_MAINET` flag is still `False` - it gets to `True` later (if `-m` is provided) when arguments are parsed

jiivan (approved):
Yeah, `CommandHelper` needs major refactoring, but it's not top priority right now.

PR #3690, opened 2018-12-20
Fixed HyperVHypervisor.restore_vm()

Wiezzel:
Restoring VM would fail in case of insufficient RAM available. A memory check was added to shut down the VM in such situation (i.e. remove the saved state) and boot it normally.

### DIFF
--- a/tests/golem/docker/test_hyperv.py
+++ b/tests/golem/docker/test_hyperv.py
@@ -3,10 +3,11 @@ import subprocess
 
 from pathlib import Path
 from unittest import TestCase
-from unittest.mock import Mock, patch, ANY
+from unittest.mock import Mock, patch, ANY, call
 
 from os_win.constants import HOST_SHUTDOWN_ACTION_SAVE, \
-    HOST_SHUTDOWN_ACTION_SHUTDOWN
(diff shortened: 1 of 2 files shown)


tests/golem/docker/test_hyperv.py
```diff
@@ -267,3 +268,67 @@ def test_check_smb_port_error(self, command, logger):
         self.hyperv._check_smb_port()
         command.assert_called_once()
         logger.error.assert_called_once()
+
+    def test_restore_vm_ok(self):
+        with patch.object(self.hyperv, '_check_memory') as check_memory, \
```
maaktweluit, line 273: i think the `@patch` decorator is neater here then `with patch`

maaktweluit (approved):
LGTM!

one nitty comment so still approved

tests/golem/docker/test_hyperv.py
Wiezzel: How do I use `@patch` on `self.something`?

maaktweluit: good one, did not see that..

can only think of patching `HyperVHypervisor._check_memory`, but i am not sure if that works

mfranciszkiewicz (approved):

PR #3699, opened 2019-01-03
Clean up CoreTask a bit

Krigpl:
Just a bunch of unused member variables.


Krigpl:
Meh, `root_path` is used in one random place

mplebanski:
Yes it is used by DirManager IIRC

PR #3700, opened 2019-01-03
argv-like invocation in entrypoint.sh

mfranciszkiewicz:
- `entrypoint.sh` has been modified to execute `$@`ll given arguments
- fixed Docker image building with Golem (+ setup.py)
- bumped image versions


apps/images.ini
```diff
@@ -1,5 +1,5 @@
-golemfactory/base core/resources/images/base.Dockerfile 1.3 core/resources/images/
```
nieznanysprawiciel, line 1: I would prefer previous solution. Why building docker images needs full apps/ directory as build context?

mfranciszkiewicz: That's because `entrypoint.sh` resided in `core/resources/images/scripts` and the scripts required by `image_metrics` are placed in the provided root `rendering/resources/images`. It was not possible to build the image.

nieznanysprawiciel: But blender doesn't need this script and you changed this path too.

mfranciszkiewicz: This file is derived from `develop`

golem/docker/job.py
```diff
@@ -125,9 +129,10 @@ def _prepare(self):
             json.dump(self.parameters, params_file)
 
         # Save the script in work_dir/TASK_SCRIPT
-        task_script_path = self._get_host_script_path()
-        with open(task_script_path, "wb") as script_file:
-            script_file.write(bytearray(self.script_src, "utf-8"))
+        payload = self.payload
+        if isinstance(payload, Source):
+            payload = payload.save(self._get_host_script_path(),
+                                   self._get_container_script_path())
```
Krigpl: `container_script_path` returns Python script path (`job.py`) so this isn't ideal.
Instead, it should either pass the script path in e.g. extra_data for wait for my change to remove passing the script file altogether (imo preferred).

mfranciszkiewicz: This reflects the current functionality, other changes may be introduced after this PR is merged.

Krigpl: The whole `Payload` class won't be necessary when we stop passing the script via the message and instead pass the `argv`-like invocation. So making changes in this order is imo wasteful.

mfranciszkiewicz: This PR changes the entrypoint without touching anything non-Docker, say apps, with 25 files modified. `Payload` is already here and trivial to remove.

Krigpl: Since this is just about the entrypoint wouldn't it be better to just change the command on line 147 to `'python ' + container_script_path`? Leaving the `src_code` untouched this would be the least invasive way to support generic entrypoint and without giving a false feeling that we do support non Python payloads (which we don't just yet because of the above).

mfranciszkiewicz: `'python ' + container_script_path` is hacky. `DockerJob` in this state supports non Python payloads.

Why removing `Payload` is so important?

mfranciszkiewicz: I'll prepare some other approach, without the `Payload` class. I don't want DockerJob to know anything about the command.

golem/docker/job.py
```diff
@@ -130,7 +130,7 @@ def _prepare(self):
             image=self.image.name,
             volumes=self.volumes,
             host_config=host_cfg,
-            command=[self.script_filepath],
+            command=[f'/usr/bin/python3 {self.script_filepath}'],
```
Krigpl: Ideally, we will pass this whole command via message, right?

Krigpl (approved):
lgtm

jiivan (approved):

golem/docker/job.py
mfranciszkiewicz: Yes, but I'm leaving this for now as it is.

Krigpl:
Why not publish the images before merging?

mfranciszkiewicz:
I'm doing that right now.

It's develop, not release. The images are automatically built by Golem.

golem/docker/job.py
Krigpl: `image_metrics` doesn't work with that
```
$ docker run golemfactory/image_metrics:1.12 /usr/bin/python3 --version
/usr/local/bin/entrypoint.sh: line 13: /usr/bin/python3: No such file or directory
```

Issue #3697, opened 2019-01-02
Change the meaning of `success` in the Front-end's footer


shadeofblue:
Verify where the front-end takes the `success` count from and modify the reported count to only be updated when `SubtaskResultsAccepted` message is actually sent by the requestor.

Otherwise (if the `SRA` never arrives before the timeout, or, if `SubtaskResultsRejected` is received - increase the `error` count instead).

Possibly, include the "local success" count in a separate field - e.g. `computation_completed` so that both can theoretically be shown to the user.

PR #3703, opened 2019-01-04
Updated stats RPC with accepted/rejected subtasks

kmazurek:
#3697

This adds two new fields to the RPC call 'comp.tasks.stats' These fields contain the total
count of SRA and SRR messages received by the provider (stored as stats
counters in local db).

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -872,32 +872,17 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             'provider_state': self.get_provider_status(),
             'in_network': self.get_task_count(),
             'supported': self.get_supported_task_count(),
-            'subtasks_computed': self.get_computed_task_count(),
-            'subtasks_accepted': self.get_accepted_task_count(),
-            'subtasks_rejected': self.get_rejected_task_count(),
-            'subtasks_with_errors': self.get_error_task_count(),
-            'subtasks_with_timeout': self.get_timeout_task_count(),
+            'subtasks_computed': self.get_comp_stat('computed_tasks'),
+            'subtasks_accepted': self.get_provider_stat('provider_sra_cnt'),
+            'subtasks_rejected': self.get_provider_stat('provider_srr_cnt'),
+            'subtasks_with_errors': self.get_comp_stat('tasks_with_errors'),
+            'subtasks_with_timeout': self.get_comp_stat('tasks_with_timeout'),
         }
-
-    def get_accepted_task_count(self):
-        return self.get_task_provider_stat('provider_sra_cnt')
-
-    def get_rejected_task_count(self):
-        return self.get_task_provider_stat('provider_srr_cnt')
 
     def get_supported_task_count(self) -> int:
         if self.task_server:
             return len(self.task_server.task_keeper.supported_tasks)
         return 0
-
-    def get_computed_task_count(self):
-        return self.get_task_computer_stat('computed_tasks')
-
-    def get_timeout_task_count(self):
-        return self.get_task_computer_stat('tasks_with_timeout')
-
-    def get_error_task_count(self):
-        return self.get_task_computer_stat('tasks_with_errors')
 
     @rpc_utils.expose('comp.tasks.unsupport')
     def get_unsupport_reasons(self, last_days):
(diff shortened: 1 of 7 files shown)


golem/task/taskproviderstats.py
```diff
@@ -17,6 +19,9 @@ def __init__(self, **kwargs) -> None:
         # WantToComputeTask [tx] to TaskToCompute [rx] delta time
         self.provider_wtct_to_ttc_delay_sum: int = 0
         self.provider_wtct_to_ttc_cnt: int = 0
+        # Verification messages count
+        self.provider_sra_cnt: int = 0
```
mfranciszkiewicz, line 23: We might want to update the monitor with these new fields

kmazurek: Good one, I'll add a separate task for that. I discussed it with @tworec - merging this PR won't break monitor backend (new fields will be ignored).

tworec: yup, [golem-monitor-replicator](https://github.com/golemfactory/golem-monitor-replicator) will ignore unknown requests and fields. As for now, `ProviderStats` is a completely new request type from `gmr` perspective, so it will be ignored as a whole.

kmazurek: I see, so this is entirely separate from the changes I made here. Thanks for taking a look!

golem/client.py
```diff
@@ -873,10 +873,18 @@ def get_task_stats(self) -> Dict[str, Any]:
...
             'supported': self.get_supported_task_count(),
             'subtasks_computed': self.get_computed_task_count(),
+            'subtasks_accepted': self.get_accepted_task_count(),
+            'subtasks_rejected': self.get_rejected_task_count(),
             'subtasks_with_errors': self.get_error_task_count(),
-            'subtasks_with_timeout': self.get_timeout_task_count()
+            'subtasks_with_timeout': self.get_timeout_task_count(),
         }
 
+    def get_accepted_task_count(self):
```
jiivan: is this method really necessary? Client is already overbloated... :/

tests/golem/task/test_tasksession.py
```diff
@@ -543,14 +543,21 @@ def __call_react_to_srr(self, srr):
             self.task_session._react_to_subtask_results_rejected(srr)
         dropped.assert_called_once_with()
 
-    def test_result_rejected(self):
+    @patch('golem.task.tasksession.dispatcher.send')
+    def test_result_rejected(self, send_mock):
```
jiivan: Maybe instead of mocking `dispatcher.send` just connect a `Mock` instance to dispatcher signal?

jiivan (approved):

golem/client.py
kmazurek: I've slimmed it down by removing these one-liner methods, see the updated version

tests/golem/task/test_tasksession.py
```diff
@@ -544,13 +545,19 @@ def __call_react_to_srr(self, srr):
...
+
         srr = self._get_srr()
         self.__call_react_to_srr(srr)
+
         self.task_session.task_server.subtask_rejected.assert_called_once_with(
             sender_node_id=self.task_session.key_id,
             subtask_id=srr.report_computed_task.subtask_id,  # noqa pylint:disable=no-member
         )
 
+        dispatch_listener.assert_called_once()
```
jiivan: `assert_called_once_with()`?

kmazurek: Updated, when using `assert_called_once_with` the assertion failed on messages in both tests. However, when comparing with the message stored in `call_args_list` this works as expected.

jiivan: Look now

kmazurek: Ugh, I missed the `sender` field there. Or better show the difference like unittests assertEqual

jiivan: Yeah it would be more readable if mock used [pprint](https://docs.python.org/3/library/pprint.html#pprint.pprint) Stop editing my stuff dude

mdtanrikulu (approved):

Issue #3643, opened 2018-12-05
RPC create_task should give feedback about insufficient funds for concent deposit 


mplebanski:
Branch: develop

Additional prerequisites: concent enabled

Right now, if one creates a task and has no GNT to supply the deposit the following error will be logged:

```
ERROR    [golem.task.rpc                     ] Cannot create task {'bid': 1.0, 'subtask_timeout': '00:10:00',
... 
Not enough GNT available. Required: 0.166667, available: 0.000000
```

but RPC call will return valid `task_id` and no `error_message`. This is not consistent with e.g. when one provides an invalid task` type` in task dict.

pawelkisielewicz:
I have investigated this task and the reason of problem is creating task asynchronicity. In function `create_task` (`rpc.py`) there is some validations and there is really `None` and error message returned in case of failed validation. 

But in next step task is created asynchronicity, so if we want to return consistent output we have to **wait** until task is created, check if there were errors and then return valid `task id`. This way all errors from function `enqueue_new_task` will be handled, but it will make the whole flow longer. 

My idea is to **wait** for results of enqueue_new_task and I will start implementation this way. If there is not acceptable for you please let me know.

mplebanski:
Looks good to me. What about you @shadeofblue?

mdtanrikulu:
@pawelkisielewicz It was working like that afaik, but the concept had to change because of UX related issues. When RPC waits until task creation it was taking a lot on GUI to redirect user to the task list screen or give a feedback about the related error. So instead of waiting all process we made this change. 
If we wait whole flow, this issue will pop-up again. Is there any alternative or an idea to shortened the waiting time tolerably?

mplebanski:
@mdtanrikulu How long is this wait time?

mdtanrikulu:
@mplebanski Only thing I remember, it was depend on the task complexity.

pawelkisielewicz:
I spoke with @jiivan on rocket and general suggestion is to move checking of available GNT and ETH before call  asynchronic function - as "expected" fails. I try to extract as much as I can of fast validations to cover as much cases as possible. I'm not sure if it will be 100% perfect, but it seems to be the best approach.

PaweuB:
Pivota:
https://www.pivotaltracker.com/story/show/163016112

shadeofblue:
yeah, @pawelkisielewicz @jiivan the actual deposit can be made asynchronously but the check whether there are funds for the deposit can indeed be made immediately

[...]

PR #3715, opened 2019-01-08
Move validations from enqueue_new_task method before it in create_task

pawelkisielewicz:
Resolves https://github.com/golemfactory/golem/issues/3643


tests/golem/ethereum/test_fundslocker.py
```diff
@@ -122,3 +123,11 @@ def test_add_subtask(self):
         self.ts.lock_funds_for_payments.assert_called_with(subtask_price,
                                                            num)
         assert fl.task_lock[task_id].num_tasks == num_tasks + num
+
+    def test_validate_lock_funds_possibility_raises_if_not_enough_gnt(self):
+        fl = FundsLocker(self.ts)
+        fl.transaction_system.get_available_gnt.return_value = 10000000000000
```
dybi: please first setup the mock, and then feed it to `FundsLocker`

tests/golem/ethereum/test_fundslocker.py
```diff
@@ -122,3 +123,11 @@ def test_add_subtask(self):
...
+        with self.assertRaises(exceptions.NotEnoughFunds) as e:
+            fl.validate_lock_funds_possibility(50000000000000, 1)
+        x = 'Not enough GNT available. Required: 0.000050, available: 0.000010'
```
dybi: please use f-strings and the values above (as variables)

tests/golem/task/test_rpc.py
```diff
@@ -133,6 +144,19 @@ def test_create_task_fail_on_illegal_character_in_name(self, *_):
...
+                'validate_lock_funds_possibility',
+                side_effect=_raise_not_enough_funds)
+    def test_create_task_fail_if_not_enough_gnt_available(self, *_):
+        t = dummytaskstate.DummyTaskDefinition()
+        t.name = "test"
+
+        result = self.provider.create_task(t.to_dict())
+        rpc.enqueue_new_task.assert_not_called()
+
+        assert result == (None, 'Not enough GNT available. Required: '
```
dybi: please use `self.assertEqual`

tests/golem/task/test_rpc.py
```diff
@@ -133,6 +144,19 @@ def test_create_task_fail_on_illegal_character_in_name(self, *_):
                           "Task name can only contain letters, numbers, "
                           "spaces, underline, dash or dot.")
 
+    @mock.patch('golem.ethereum.fundslocker.FundsLocker.'
+                'validate_lock_funds_possibility',
+                side_effect=_raise_not_enough_funds)
+    def test_create_task_fail_if_not_enough_gnt_available(self, *_):
```
dybi: you could also check, that the mock was actually called

dybi (commented):
Seems legit, but some polish would be appreciated.

tests/golem/ethereum/test_fundslocker.py
pawelkisielewicz: done

tests/golem/ethereum/test_fundslocker.py
pawelkisielewicz: done

tests/golem/task/test_rpc.py
pawelkisielewicz: changed. I did it like this at first, but in all other tests there is `assert result ==` so I decided to unify. But changed :)

tests/golem/task/test_rpc.py
pawelkisielewicz: done

golem/ethereum/transactionsystem.py
```diff
@@ -380,17 +380,7 @@ def lock_funds_for_payments(self, price: int, num: int) -> None:
         if not self._payment_processor:
             raise Exception('Start was not called')
         gnt = price * num
-        if gnt > self.get_available_gnt():
```
Krigpl: Leave these checks, `lock_funds_for_payments` can be called from wherever and if should still do the validation.

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -261,7 +262,8 @@ def test_locking_funds(self):
         assert self.ets.get_locked_gnt() == 0
 
         with self.assertRaisesRegex(NotEnoughFunds, 'GNT'):
-            self.ets.lock_funds_for_payments(gnt_balance, 2)
+            fl = FundsLocker(transaction_system=self.ets)
```
Krigpl: revert this

tests/golem/ethereum/test_fundslocker.py
```diff
@@ -122,3 +124,15 @@ def test_add_subtask(self):
         self.ts.lock_funds_for_payments.assert_called_with(subtask_price,
                                                            num)
         assert fl.task_lock[task_id].num_tasks == num_tasks + num
+
+    def test_validate_lock_funds_possibility_raises_if_not_enough_gnt(self):
+        required = 50000000000000
```
Krigpl: use smaller numbers or `denoms.ether`

golem/ethereum/fundslocker.py
```diff
@@ -28,9 +29,24 @@ def __init__(
             transaction_system: TransactionSystem,
             interval_seconds: int = 60) -> None:
         super().__init__(interval_seconds)
-        self.task_lock = {}
+        self.task_lock: dict = {}
         self.transaction_system = transaction_system
 
+    def validate_lock_funds_possibility(
```
Krigpl: Remove this method from `FundsLocker` - it doesn't use any FundsLocker's internals, just the transaction system.

Krigpl (changes requested):
Please include me as a reviewer anytime you touch payments/transaction system/ethereum/etc

golem/task/rpc.py
```diff
@@ -322,16 +330,9 @@ def _start_task(client, task, resource_server_result):
 
 
 @defer.inlineCallbacks
-def enqueue_new_task(client, task, force=False) \
+def enqueue_new_task(client, task) \
             -> typing.Generator[defer.Deferred, typing.Any, taskbase.Task]:
     """Feed a fresh Task to all golem subsystems"""
-    if client.config_desc.in_shutdown:
```
jiivan, line 322: What happens when `in_shutdown` was `False` during create task, and became true in async function? race condition

golem/task/rpc.py
```diff
@@ -436,12 +436,23 @@ def create_task(self, task_dict, force=False) \
         """
 
         prepare_and_validate_task_dict(self.client, task_dict)
+        validate_client(self.client)
```
jiivan, line 441: What about other places `enqueue_new_task` is called

golem/task/rpc.py
```diff
@@ -214,7 +223,6 @@ def _ensure_task_deposit(client, task, force):
         yield client.transaction_system.concent_deposit(
             required=min_amount,
             expected=opt_amount,
-            force=force,
```
jiivan, line 211: Why?

jiivan (dismissed):
Please fix typo in description url

golem/ethereum/transactionsystem.py
pawelkisielewicz: right, done

tests/golem/ethereum/test_fundslocker.py
pawelkisielewicz: done

golem/ethereum/fundslocker.py
pawelkisielewicz: done, moved to `rpc.py` as inner function

tests/golem/ethereum/test_transactionsystem.py
pawelkisielewicz: reverted and added assert for function validating possibility of locking founds

golem/task/rpc.py
pawelkisielewicz: yes... right. I left this check before calling `enqueue_new_task` as well, there is very high possibility that in this case it will be caught before. But it is called also in `enqueue_new_task` to prevent race condition

golem/task/rpc.py
pawelkisielewicz: :+1: added to other places

golem/task/rpc.py
pawelkisielewicz: It is used to check possibility to lock founds. In "first" version I moved this validation out of `concent_deposit`, it was the reason. Now it is back and this parameter has to stay

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -22,6 +22,7 @@
     tETH_faucet_donate,
 )
 from golem.ethereum.exceptions import NotEnoughFunds
+from golem.task.rpc import _validate_lock_funds_possibility
```
Krigpl: Nope, `test_transactionsystem` which tests `TransactionSystem` shouldn't test a method from a different module, move it to an appropriate test file.

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -568,12 +570,34 @@ def test_gas_price_skyrocketing_forced(self):
...
+            required=10,
+            force=True
+        )
         self._call_concent_deposit(
             required=10,
             expected=40,
-            force=True
         )
 
+    @patch(
```
Krigpl: Don't patch methods from a class that is being tested, patch external dependencies instead.

Krigpl (changes requested):

tests/golem/ethereum/test_transactionsystem.py
pawelkisielewicz: moved to `test_rpc.py`

tests/golem/ethereum/test_transactionsystem.py
pawelkisielewicz: after moving to `test_rpc.py` there is no need to patch any more. Thanks for comment anyway

Krigpl:
Payments stuff looks good, leaving the rest to @jiivan

golem/task/rpc.py
```diff
@@ -363,7 +389,7 @@ def enqueue_new_task(client, task, force=False) \
         yield _ensure_task_deposit(
             client=client,
             task=task,
-            force=force,
+            force=force
```
jiivan:
```suggestion
            force=force,
```

golem/task/rpc.py
jiivan: Maybe `validate_client()` should be called before `prepare_and_validate...`?

golem/ethereum/transactionsystem.py
```diff
@@ -566,13 +567,27 @@ def concent_timelock(self, account_address: Optional[str] = None) -> int:
...
 
+    @sci_required()
+    def validate_concent_deposit_possibility(
+            self,
+            required: int,
+            force: bool = False) -> None:
+        gntb_balance = self.get_available_gnt()
+        if gntb_balance < required:
+            raise exceptions.NotEnoughFunds(required, gntb_balance, 'GNTB')
+        if self.gas_price >= self._sci.GAS_PRICE:  # type: ignore
```
jiivan:
```suggestion
        if self.gas_price >= self.gas_price_limit:
```

tests/golem/task/test_rpc.py
```diff
@@ -1,27 +1,38 @@
...
 from golem.task import taskstate
 from golem.task import tasktester
 from tests.golem import test_client
-
+from tests.golem.test_client import TestClientBase
 
 fake = faker.Faker()
 
 
+def _raise_not_enough_funds(*_, **__):
```
jiivan: No need to define function only to raise exception. SEE: https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock

> Alternatively side_effect can be an exception class or instance. In this case the exception will be raised when the mock is called.

golem/ethereum/transactionsystem.py
```diff
@@ -566,13 +567,27 @@ def concent_timelock(self, account_address: Optional[str] = None) -> int:
             account_address=account_address,
         )
 
+    @sci_required()
+    def validate_concent_deposit_possibility(
```
jiivan, line 594: Should this method check for `required - self.concent_balance()`?

golem/task/rpc.py
```diff
@@ -436,9 +462,21 @@ def create_task(self, task_dict, force=False) \
...
-
+        _validate_lock_funds_possibility(
+            self.client.transaction_system,
+            task.subtask_price,
+            task.get_total_tasks(),
+        )
+        min_amount, _ = msg_helpers.requestor_deposit_amount(
+            task.price,
+        )
+        self.client.transaction_system.validate_concent_deposit_possibility(
```
jiivan: What if concent is disabled for this task?

tests/golem/task/test_rpc.py
pawelkisielewicz: :+1: thanks. Removed

golem/ethereum/transactionsystem.py
pawelkisielewicz: done

golem/task/rpc.py
```diff
@@ -436,9 +462,21 @@ def create_task(self, task_dict, force=False) \
...
 
         prepare_and_validate_task_dict(self.client, task_dict)
+        validate_client(self.client)
 
         task: taskbase.Task = self.task_manager.create_task(task_dict)
-
+        _validate_lock_funds_possibility(
+            self.client.transaction_system,
+            task.subtask_price,
+            task.get_total_tasks(),
```
jiivan: You could use `task.price`
https://github.com/golemfactory/golem/blob/fd77534db079df22eb68090dddab5ad38bb8bf1e/golem/task/taskbase.py#L118-L120

golem/task/rpc.py
pawelkisielewicz: done

golem/ethereum/transactionsystem.py
pawelkisielewicz: It should. Changed

golem/task/rpc.py
pawelkisielewicz: Check added to `_validate_enough_founds_to_pay_for_task`. This function is called in 3 places:
- `_validate_enough_founds_to_pay_for_task` - check added
- `_ensure_task_deposit` - there was check before
- `ask_for_verification`in `_react_to_subtask_results_rejected`- there is check before call as well.

golem/task/rpc.py
pawelkisielewicz: OK, thanks, didn't know about this. `task.subtask_price`. But `task.get_total_tasks()` has to stay to check ETH

golem/ethereum/transactionsystem.py
```diff
@@ -585,6 +587,15 @@ def validate_concent_deposit_possibility(
                 'Gas price is high. It can take some time to mine deposit.',
             )
 
+        eth_for_batch_payment_for_task = self.eth_for_batch_payment(tasks_num)
+        eth_to_put_deposit = self.gas_price * self._sci.GAS_TRANSFER_AND_CALL
```
jiivan: You can now use `transaction_system.eth_for_deposit()`

golem/task/tasksession.py
```diff
@@ -889,6 +889,7 @@ def ask_for_verification(_):
             self.task_server.client.transaction_system.\
                 validate_concent_deposit_possibility(
                     required_deposit_for_task=amount,
+                    tasks_num=1,
```
pawelkisielewicz, line 907: While restarting subtask we should ensure eth for batch transfer for only one task. Is it correct?

golem/ethereum/transactionsystem.py
pawelkisielewicz: done

golem/ethereum/transactionsystem.py
```diff
@@ -589,14 +590,40 @@ def concent_timelock(self, account_address: Optional[str] = None) -> int:
...
 
+    @sci_required()
+    def validate_concent_deposit_possibility(
+            self,
+            required: int,
+            tasks_num: int,
+            force: bool = False,
+    ) -> None:
+        current_concent_deposit = self.concent_balance()
+        required_deposit_difference = required - current_concent_deposit
```
kmazurek: `current_concent_deposit` is not used anywhere else but here, just do: `required_deposit_difference = required - self.concent_balance()`.

golem/task/rpc.py
```diff
@@ -448,6 +474,26 @@ def create_task(self, task_dict, force=False) \
         )
         return task_id, None
 
+    def _validate_enough_founds_to_pay_for_task(
```
kmazurek: founds -> funds

golem/task/rpc.py
```diff
@@ -147,6 +158,22 @@ def on_error(*args, **kwargs):
     client.task_tester.run()
 
 
+def _validate_lock_funds_possibility(
+        transaction_system: TransactionSystem,
```
kmazurek: Passing the transaction system here is redundant, you can just refer to `self.client.transaction_system` directly.

golem/task/rpc.py
```diff
@@ -147,6 +158,22 @@ def on_error(*args, **kwargs):
...
+        total_price_gnt: int,
+        number_of_tasks: int) -> None:
+    eth = transaction_system.eth_for_batch_payment(number_of_tasks)
```
kmazurek: This should be moved to after the check for available GNT (next to where `eth_available` is initialised).

golem/task/rpc.py
```diff
@@ -448,6 +474,26 @@ def create_task(self, task_dict, force=False) \
...
 
+    def _validate_enough_founds_to_pay_for_task(
+            self, task: taskbase.Task, force: bool
+    ):
+        _validate_lock_funds_possibility(
+            self.client.transaction_system,
+            task.price,
+            task.get_total_tasks(),
+        )
+        min_amount, _ = msg_helpers.requestor_deposit_amount(
```
kmazurek: Please merge this into a single line

golem/task/rpc.py
```diff
@@ -205,10 +233,14 @@ def _ensure_task_deposit(client, task, force):
     # fails. This case is most common but, the better way it to always
     # unlock them when the task fails regardless of the reason.
     try:
+        client.transaction_system.validate_concent_deposit_possibility(
+            required_deposit_for_task=min_amount,
```
kmazurek: There is no such kwarg in `validate_concent_deposit_possibility`, either change this to `required` or rename the expected argument.

golem/task/rpc.py
```diff
@@ -448,6 +474,26 @@ def create_task(self, task_dict, force=False) \
...
+            task.get_total_tasks(),
+        )
+        min_amount, _ = msg_helpers.requestor_deposit_amount(
+            task.price,
+        )
+        concent_enabled = task.header.concent_enabled
+        concent_available = self.client.concent_service.available
+        if concent_enabled and concent_available:
+            self.client.transaction_system.validate_concent_deposit_possibility(
+                required_deposit_for_task=min_amount,
```
kmazurek: See previous comment on this, this method expects a different argument name.

golem/task/tasksession.py
```diff
@@ -901,6 +901,11 @@ def ask_for_verification(_):
                     msg=srv,
                 )
 
+            self.task_server.client.transaction_system.\
+                validate_concent_deposit_possibility(
+                    required_deposit_for_task=amount,
```
kmazurek: Same as above, this method expects a different argument name.

golem/task/tasksession.py
kmazurek: Sounds correct, but is this part related to restarting subtasks?

kmazurek (changes requested):

golem/task/tasksession.py
pawelkisielewicz: checking if client has enough ETH for deposit is and was related. Now it is additionally checked gas for transfer, so by restarting I assume that there is one task.

golem/ethereum/transactionsystem.py
pawelkisielewicz: done

golem/task/rpc.py
pawelkisielewicz: done

golem/task/rpc.py
pawelkisielewicz: done, converted from function to method

golem/task/rpc.py
pawelkisielewicz: done

golem/task/rpc.py
pawelkisielewicz: done

golem/task/rpc.py
pawelkisielewicz: good catch, done

golem/task/rpc.py
pawelkisielewicz: done

golem/task/tasksession.py
pawelkisielewicz: done

kmazurek (approved):
lgtm

Krigpl (approved):

Krigpl (approved):

PR #3719, opened 2019-01-08
Merge Golem-Verificator back to the core

Krigpl:
It doesn't make sense as a stand-alone repo.
Simply moved all the state from 1.7.0
No extra clean up at this stage.


Krigpl:
catping

golem/verificator/docker/blender/images/Dockerfile
```diff
@@ -0,0 +1,48 @@
+# Use an official Python runtime as a parent image
```
nieznanysprawiciel, line 1: This is dockerfile that produces image_metrics image (this is the newest version). It is duplicated in golem/apps/rendering/resources/images/ (here resides old version, that isn't used anymore).
Remove old version.

mfranciszkiewicz (approved):

nieznanysprawiciel (changes requested):

golem/verificator/docker/blender/images/Dockerfile
mfranciszkiewicz: I'd suggest moving the latest to `golem/apps/rendering/resources/images`, so all images can be built by Golem automatically

Krigpl: I stated in the summary that this is supposed to be a simple code move, not a cleanup. There's much more to be cleaned, I'd rather not do it in the same PR.

nieznanysprawiciel: Moving these files should be simple move too ;)
But ok....
I approve this PR.

nieznanysprawiciel (approved):

PR #3723, opened 2019-01-09
Remove PyTaskThread

Krigpl:
And one of the DummyTasks that were using it.
This also enables further cleanup.


maaktweluit (approved):
LGTM!

Nice cleanup

mfranciszkiewicz (changes requested):
- `test_dummy_task_computation` fails on macOS
- `test_dummy_task_computation` will no longer be run on windows and macOS buildbot workers (no Docker)

jiivan:
@Krigpl is this still relevant?

Krigpl:
@jiivan Somewhat. This change requires an ability to use Docker on macos/windows buildbots, something that is not supported right now, but @maaktweluit is looking into that.

mplebanski:
Shall we revive this refactor?

PR #3727, opened 2019-01-09
Hide console during installation

maaktweluit:
Hide the console windows while configuring hyperv / docker for windows

### DIFF
--- a/scripts/docker/create-share.ps1
+++ b/scripts/docker/create-share.ps1
@@ -61,7 +61,7 @@ $Output = (New-TemporaryFile).FullName
 
 $Process = Start-Process -FilePath "powershell.exe" `
     -ArgumentList "-Command $Command 2>&1 | Out-File -FilePath $Output -Encoding UTF8" `
-    -Wait -PassThru -Verb RunAs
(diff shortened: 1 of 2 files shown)


Wiezzel:
@maaktweluit There is one thing you should double-check: If the SMB share is not created and you start Golem does it prompt to create it?

maaktweluit:
@Wiezzel tested this and it seems the share is not created at all on boot.

Golem only calls `apply_config` from node.py, while `update_config` (called later from client.py and friends) creates the share.

Think you found a new bug, should i try to fix it in this PR or make a new issue?

Wiezzel:
@maaktweluit 
> Think you found a new bug, should i try to fix it in this PR or make a new issue?

It's up to you. But the fix must be applied before merging this PR. My concern is that even when the script is run on Golem boot the `-WindowStyle hidden` option might cause the prompt not to appear.

maaktweluit:
@Wiezzel after some more tests the `-WindowStyle` does not affect the prompt, it still appeared even if the console is hidden

mfranciszkiewicz (approved):

Wiezzel (approved):

PR #3740, opened 2019-01-11
Information about Hyper-V in the installer

Wiezzel:
* If Hyper-V is not available just inform the user to download and install Docker Toolbox manually.
obraz

* If Hyper-V is available ask the user if they want to 'install' it.
obraz

  If the response is 'Yes':
   - copy `docker.exe` and `docker-machine.exe` binaries
   - turn on Hyper-V
   - execute `prepare-docker-for-windows.ps1` script

  If the response is 'No' do none of these.

Uninstalling Docker Toolbox was disabled.


maaktweluit (approved):
LGTM!

One small q:
Why is "docker toolbox uninstall" fully disabled? it could be done when hyperv is yes
When it is not needed at all anymore maybe we can remove the step?

mfranciszkiewicz (approved):

Wiezzel:
@maaktweluit 
The deal with uninstalling Docker Toolbox is that we're not sure if it's a good decision. Our UX experts decided that we shouldn't do that because if something goes wrong with Hyper-V users could easier go back to Docker Toolbox. But I wasn't sure about it so I only disabled this step in case wanted it in the future.

PR #3741, opened 2019-01-11
Move jobpy to Docker images

Krigpl:
Stop supporting `src_code` as it was before.
The path to the script is passed through `extra_data`.

Blender and image_metrics images still require an update on Docker hub (will push them after initial approval).


maaktweluit (commented):
LGTM! Nice cleanup!

Will approve after the 2 tests are fixed ( or an issue is added to the # fixme )

Krigpl:
@maaktweluit Right, these tests are also failing for me on develop so I just removed "fixme" and see what happens. I might have broken environment :)

maaktweluit (approved):

mfranciszkiewicz:
@Krigpl Why has this already been merged?

Krigpl:
Why not?

tworec:
https://hub.docker.com/r/golemfactory/blender/dockerfile
> `MAINTAINER Artur Zawłocki <artur.zawlocki@imapp.pl>`

golem/docker/task_thread.py
```diff
@@ -190,7 +188,7 @@ def _run_docker_job(self) -> Optional[int]:
 
         params = dict(
             image=self.image,
-            script_src=self.src_code,
+            script_filepath=self.extra_data['script_filepath'],
```
nieznanysprawiciel, line 191: Why script_filepath is part of extra_data ?
As far as I know, in extra_data are all parameters needed inside docker container. script_filepath is used only as command to run docker, so we don't need to pass it to params.json.

Krigpl: Yeah, this is a slight abuse of `extra_data`, but there were no better alternatives at the time. Ideally `src_code` from ctd would be replaced with `script_filepath` (or even better with some environment config dict, e.g. to include argv as well) but this requires changes to Golem-Messages so this will be done separately.

PR #3738, opened 2019-01-10
store proper type in msg_date in network history

etam:
This causes swipe to remove all records, no matter how old they are.

Fixes #3730

PR #3749, opened 2019-01-14
store proper type in msg_date in network history b0.18

etam:
The same as #3738, but to b0.18.

### DIFF
--- a/golem/network/history.py
+++ b/golem/network/history.py
@@ -4,7 +4,6 @@ import operator
 import pickle
 import queue
 import threading
-import time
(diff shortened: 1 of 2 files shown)


etam:
Houston! We have a problem!

Buildbot tries to run rust tests, but there are none in b0.18, so it fails.

jiivan (approved):

shadeofblue:
@etam is this pull request relevant anymore?

etam:
If we're not going to release more 0.18 versions, then no.

Issue #3601, opened 2018-11-28
pay.deposit.relock/unlock:  'NoneType' object has no attribute 'functions'


mdtanrikulu:
- pay.deposit.relock
- pay.deposit.unlock

Calling both RPC's above gives the same error; 
```
ERROR    [twisted                            ] AttributeError: 'NoneType' object has no attribute 'functions': Traceback (most recent call last):
  File "/Users/mdt/projects/golem/venv/lib/python3.6/site-packages/autobahn/twisted/websocket.py", line 162, in _onMessage
    self.onMessage(payload, isBinary)
  File "/Users/mdt/projects/golem/venv/lib/python3.6/site-packages/autobahn/wamp/websocket.py", line 95, in onMessage
    self._session.onMessage(msg)
  File "/Users/mdt/projects/golem/venv/lib/python3.6/site-packages/autobahn/wamp/protocol.py", line 895, in onMessage
    on_reply = txaio.as_future(endpoint.fn, *invoke_args, **invoke_kwargs)
  File "/Users/mdt/projects/golem/venv/lib/python3.6/site-packages/txaio/tx.py", line 417, in as_future
    return maybeDeferred(fun, *args, **kwargs)
--- <exception caught here> ---
  File "/Users/mdt/projects/golem/venv/lib/python3.6/site-packages/twisted/internet/defer.py", line 150, in maybeDeferred
    result = f(*args, **kw)
  File "/Users/mdt/projects/golem/golem/ethereum/transactionsystem.py", line 625, in concent_unlock
    if self.concent_balance() == 0:
  File "/Users/mdt/projects/golem/golem/ethereum/transactionsystem.py", line 55, in curry
    return f(self, *args, **kwargs)
  File "/Users/mdt/projects/golem/golem/ethereum/transactionsystem.py", line 543, in concent_balance
    account_address=account_address,
  File "/Users/mdt/projects/golem/venv/lib/python3.6/site-packages/golem_sci/implementation.py", line 755, in get_deposit_value
    return self._call(self._gntdeposit.functions.balanceOf(account_address))
builtins.AttributeError: 'NoneType' object has no attribute 'functions'
```

shadeofblue:
@mdtanrikulu are you sure that you aren't trying to run a mainnet node?

mdtanrikulu:
@shadeofblue nope, it was mainnet. but I think it shouldn't raise any error at all, even if it's not available on mainnet now.

jiivan:
[...]

PR #3750, opened 2019-01-14
Don't ask for gntdeposit function when deposit address is unknown

jiivan:
related: #3601

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -21,6 +21,7 @@ from apps.appsmanager import AppsManager
 import golem
 from golem.appconfig import TASKARCHIVE_MAINTENANCE_INTERVAL, AppConfig
 from golem.clientconfigdescriptor import ConfigApprover, ClientConfigDescriptor
+from golem.core import variables
 from golem.core.common import (
     datetime_to_timestamp_utc,
     get_timestamp_utc,
@@ -39,7 +40,7 @@ from golem.diag.service import DiagnosticsService, DiagnosticsOutputFormat
 from golem.diag.vm import VMDiagnosticsProvider
 from golem.environments.environmentsmanager import EnvironmentsManager
 from golem.manager.nodestatesnapshot import ComputingSubtaskStateSnapshot
-from golem.ethereum.exceptions import NotEnoughFunds
+from golem.ethereum import exceptions as eth_exceptions
 from golem.ethereum.fundslocker import FundsLocker
 from golem.ethereum.paymentskeeper import PaymentStatus
 from golem.ethereum.transactionsystem import TransactionSystem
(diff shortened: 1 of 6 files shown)


Krigpl (changes requested):
I'd really rather expect these functions not to be called at all when Concent is not supported - this is more direct approach. Returning `None` which has unknown meaning is hacky. Throwing is more sensible as that's what you'd expect when trying to use a service which is offline.

jiivan:
> I'd really rather expect these functions not to be called at all when Concent is not supported - this is more direct approach. Returning `None` which has unknown meaning is hacky. Throwing is more sensible as that's what you'd expect when trying to use a service which is offline.

OK @Krigpl , now exception is raised. Also concent is disabled when deposit contract is unavailable.

golem/ethereum/transactionsystem.py
```diff
@@ -198,7 +217,10 @@ def _init(self) -> None:
             if required_eth > self._eth_balance:
                 self._eth_per_payment = self._eth_balance // recipients_count
 
-        self._subscribe_to_events()
+        try:
```
Krigpl, line 220: Let's just check whether concent is available and only then subscribe to concent events (inside `_subscribe_to_events` method)

golem/client.py
```diff
@@ -139,6 +140,12 @@ def __init__(  # noqa pylint: disable=too-many-arguments,too-many-locals
 
         self.p2pservice = None
         self.diag_service = None
+
+        if not transaction_system.deposit_contract_available:
```
Krigpl, line 144: When do you expect this will be triggered? My understanding is that the transaction system will have concent address iff it's not disabled, so when can that happen?

jiivan: I believe that traceback from https://github.com/golemfactory/golem/issues/3601 answers your question. :) We don't have contract on mainnet.

https://github.com/golemfactory/golem/blob/fd77534db079df22eb68090dddab5ad38bb8bf1e/golem/core/variables.py#L33-L59

golem/ethereum/transactionsystem.py
jiivan: This would require `TransactionSystem` to learn about either `Client` or `ConcentClientService` (get them in `__init__`). I don't think it's justified.

Now it's just [EAFP](https://docs.python.org/3/glossary.html?highlight=eafp#term-eafp)

golem/client.py
Krigpl: Does it mean that it can go away once the contract is deployed on mainnet? If so please annotate this with such comment (or file an issue).

golem/ethereum/transactionsystem.py
Krigpl: At this point in the code you are able to just check `if self.deposit_contract_available` so no need for any external dependencies. But I'm not insisting.

Krigpl (approved):

Issue #3268, opened 2018-09-03
Add an RPC call to verify VT/VT-x status


ktomala:
On Windows and MacOS when VT/VT-x is not enabled there will be no indication by Golem if the Virtualization is enabled. Golem will proceed with creating e.g. docker-machine on Windows which will fail in such scenario, without giving any obvious reason except for "Problem with starting Docker". 

We can add checking for VT/VT-x flag if it's enabled in CPU flags. This can be done in various operating systems in different ways, however should be simple enough to add. The logic when virtualization flag is not found in CPU should prevent Golem from starting whatsoever and perhaps give meaningful message in Golem-Electron regarding requirement for enabling VT.

Here are some examples of how to check for existence of this flag:
Linux: https://stackoverflow.com/questions/11116704/check-if-vt-x-is-activated-without-having-to-reboot-in-linux

Mac: https://apple.stackexchange.com/questions/224870/how-to-check-vt-x-status-on-macbook-pro

Windows (paragraph with systeminfo): https://www.shaileshjha.com/how-to-find-out-if-intel-vt-x-or-amd-v-virtualization-technology-is-supported-in-windows-10-windows-8-windows-vista-or-windows-7-machine/

Also some additional resources:
https://github.com/pydata/numexpr/blob/master/numexpr/cpuinfo.py
https://github.com/workhorsy/py-cpuinfo
https://github.com/Blosc/python-blosc2/blob/master/cpuinfo.py
https://www.cyberciti.biz/faq/linux-xen-vmware-kvm-intel-vt-amd-v-support/
https://stackpointer.io/unix/linux-check-running-virtual-machine/553/
http://www.nongnu.org/dmidecode/

maaktweluit:
Please note clay team is working on docker for windows, this will switch the hypervisor from VirtualBox to HyperV

Not sure if it applies to this issue, just FYI :)

ktomala:
@maaktweluit Yes it also applies to Hyper-V

PR #3751, opened 2019-01-14
Created RPC call for checking VT/VT-x status

kmazurek:
Resolves: #3268

This adds logic for checking whether the machine's CPU supports hardware virtualization (currently only checks for Intel VT/VT-x). On Windows, the result is based on the output of systeminfo command. On all other platforms, the py-cpuinfo library is used.

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -222,9 +222,11 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             MinPerformanceMultiplier
         from golem.network.concent import soft_switch as concent_soft_switch
         from golem.task import rpc as task_rpc
+        from golem.core import virtualization
         task_rpc_provider = task_rpc.ClientProvider(self)
         providers = (
             self,
+            virtualization,
             concent_soft_switch,
             framerenderingtask,
             MinPerformanceMultiplier,
--- a/golem/client.py
+++ b/golem/client.py
@@ -1374,11 +1374,6 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             logger.info('change hw config result: %r', result)
             return self.environments_manager.get_performance_values()
 
-    @rpc_utils.expose('env.hw.virtualization')
-    @staticmethod
-    def get_virtualization_support():
-        return is_virtualization_enabled()
-
     @staticmethod
     def enable_talkback(value):
         enable_sentry_logger(value)
--- a/golem/client.py
+++ b/golem/client.py
@@ -222,11 +222,9 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             MinPerformanceMultiplier
         from golem.network.concent import soft_switch as concent_soft_switch
         from golem.task import rpc as task_rpc
-        from golem.core import virtualization
         task_rpc_provider = task_rpc.ClientProvider(self)
         providers = (
             self,
-            virtualization,
             concent_soft_switch,
             framerenderingtask,
             MinPerformanceMultiplier,
--- a/golem/client.py
+++ b/golem/client.py
@@ -34,6 +34,7 @@ from golem.config.active import EthereumConfig
 from golem.core.keysauth import KeysAuth
 from golem.core.service import LoopingCallService
 from golem.core.simpleserializer import DictSerializer
+from golem.core.virtualization import is_virtualization_enabled
 from golem.database import Database
 from golem.diag.service import DiagnosticsService, DiagnosticsOutputFormat
 from golem.diag.vm import VMDiagnosticsProvider
@@ -1372,6 +1373,10 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             result = yield deferred
             logger.info('change hw config result: %r', result)
             return self.environments_manager.get_performance_values()
+
+    @rpc_utils.expose('env.hw.virtualization')
+    def get_virtualization_support(self):
+        return is_virtualization_enabled()
 
     @staticmethod
     def enable_talkback(value):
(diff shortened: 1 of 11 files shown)


golem/client.py
```diff
@@ -1373,6 +1374,11 @@ def activate_hw_preset(self, name, run_benchmarks=False):
             logger.info('change hw config result: %r', result)
             return self.environments_manager.get_performance_values()
 
+    @rpc_utils.expose('env.hw.virtualization')
+    @staticmethod
+    def get_virtualization_support():
```
etam: You don't need to expose this through Client anymore. Just put `rpc_utils.expose` directly on `is_virtualization_enabled`.

golem/core/virtualization.py
```diff
@@ -0,0 +1,48 @@
...
+        subprocess.run('systeminfo', check=True, stdout=subprocess.PIPE)
+
+    # https://stackoverflow.com/a/9228117
+    try:
+        output = sys_info.stdout.decode(locale.getpreferredencoding())
+    except ValueError:
+        output = sys_info.stdout.decode()
+
+    key_vt_supported = 'VM Monitor Mode Extensions'
+    key_vt_enabled = 'Virtualization Enabled In Firmware'
```
etam: Does this work if system is configured to use other language?

tests/golem/core/test_virtualization.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+@patch('golem.core.virtualization.is_windows', side_effect=lambda: False)
+class VirtualizationTestUnix(unittest.TestCase):
+
+    @patch('golem.core.virtualization.get_cpu_info')
+    def test_vt_enabled(self, cpu_info_mock, *_):
+        cpu_info_mock.return_value = get_mock_cpuinfo_output()
+
+        self.assertTrue(is_virtualization_enabled())
```
etam, line 26: How about
```python
def test_vt_enabled(self, *_):
    with patch('golem.core.virtualization.get_cpu_info',
            return_value=get_mock_cpuinfo_output()):
        self.assertTrue(is_virtualization_enabled())
```
?

I know, it's bikeshedding. Anyway, I think it's a bit more tidy. Change it if you want, but I don't mind if you don't.

golem/core/virtualization.py
kmazurek: I tested it on Windows 10 with Polish locale. The output was in English, but encoded using CP1250 (i.e. appropriate for the system's language in Windows terms). As far as I can tell, the output does not get localized. Then again, it's not like I can take a look at the source code of that command.
One safeguard I can think of is to ensure the script is called via PowerShell and not the standard Windows command line. PowerShell uses UTF encoding by default (yay).

shadeofblue: indeed... especially, that we take care to correctly decode the string using the system's locale and then just check the string directly against English...

tests/golem/core/test_virtualization.py
shadeofblue: I believe both syntaxes are no more or no less "correct" ... just one applies the patch against the whole method and the other is more useful if the patch is supposed to be active in a specific context...

shadeofblue: iow, I wouldn't change it in this case to the other syntax

kmazurek: I already updated this, putting `return_value` in the decorator makes these tests more concise.

golem/core/virtualization.py
kmazurek: That's not really an issue here, since the output of the command doesn't get localized. I was thinking more of the encoding used in the shell itself.

tests/golem/core/test_virtualization.py
shadeofblue: ah, that... I thought @etam preferred the CM to decorator...

golem/core/virtualization.py
etam: Python standard library is calling `popen` directly and does not use any encoding/decoding: https://github.com/python/cpython/blob/3.7/Lib/platform.py#L455
Maybe that's the correct way?

etam: Well, in other place it does: https://github.com/python/cpython/blob/3.7/Lib/platform.py#L798 and it uses `latin-1`

etam: Soooo.... What is the conclusion? What is the right thing to do? I feel incompetent to approve or reject this part.

mdtanrikulu (approved):

golem/core/windows.py
```diff
@@ -0,0 +1,57 @@
...
+from typing import List, Optional
+
+DEFAULT_SCRIPT_TIMEOUT = 5  # seconds
+
+
+def run_powershell(
+        script: Optional[str] = None,
+        command: Optional[str] = None,
+        args: Optional[List[str]] = None,
+        use_profile: Optional[bool] = False,
```
etam: why not just `bool`?

scripts/virtualization/get-virtualization-state.ps1
```diff
@@ -0,0 +1,5 @@
+$ErrorActionPreference = "Stop"
+[Console]::OutputEncoding = [Text.UTF8Encoding]::UTF8
+
+$SystemInfo = (GWMI Win32_Processor)
+return $SystemInfo.VMMonitorModeExtensions -and $SystemInfo.VirtualizationFirmwareEnabled
```
etam, line 5: Cthulhu
Ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn

etam (approved):
ok

scripts/virtualization/get-virtualization-state.ps1
ktomala: :joy_cat:

PR #3761, opened 2019-01-16
Use new but pretty much the same tETH faucet

Krigpl:
And move faucet related code its own files.


shadeofblue (approved):
ba90134f42960c6092b0d8f97198fcf3

PR #3767, opened 2019-01-17
Initial Concent soft switch status is off

mdtanrikulu:


jiivan (approved):
ivlq_useless_light_switch

PR #3769, opened 2019-01-17
Remove get_task_env

Krigpl:
Since TaskHader is now included in the messages.


mplebanski:
What was it for anyway?

Krigpl:
For `_set_env_params` in tasksession except that now we can get it from the message itself instead of our local state.

mplebanski (approved):

PR #3778, opened 2019-01-18
 Hyper-V process detection on Windows

kmazurek:
On Windows, if Hyper-V is enabled, WMI queries report lack of
virtualization support. To correctly detect virtualization capabilities,
we need to first check if Hyper-V process is running on the host.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -78,7 +78,7 @@ Pillow==4.2.1
 pluggy==0.6.0
 portalocker==1.2.1
 priority==1.3.0
-psutil==5.2.2
+psutil==5.4.8
 py-cpuinfo==4.0.0
 py-ubjson==0.11.0
 pyasn1-modules==0.2.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -78,7 +78,7 @@ Pillow==4.2.1
 pluggy==0.6.0
 portalocker==1.2.1
 priority==1.3.0
-psutil==5.4.8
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -78,7 +78,7 @@ Pillow==4.2.1
 pluggy==0.6.0
 portalocker==1.2.1
 priority==1.3.0
-psutil==5.2.2
+psutil==5.4.8
```
mfranciszkiewicz: I'd suggest not to update dependencies now unless there is a strong reason to do so

Wiezzel:
You can check much easier if Hyper-V is enabled. It is actually already done in `HyperVHypervisor.is_available()` method: https://github.com/golemfactory/golem/blob/6ec4e13fba79b011864d1b453afc78716164c31c/golem/docker/hypervisor/hyperv.py#L194-L202
So just modify `_check_vt_windows()` or `get-virtualization-state.ps1` to include this command.

kmazurek:
@Wiezzel Thanks, didn't notice that. I saw another solution using `Get-WindowsOptionalFeature`, but that one requires running as admin. Let me update and test this.

requirements.txt
kmazurek: Got it, turns out it won't be required most likely (see comment below).

Wiezzel (approved):
👍

mfranciszkiewicz (approved):

PR #3788, opened 2019-01-21
Aio monitor

### LINKED ISSUE #3583: Migrate monitor to asyncio
With **asyncio** monitor probably wouldn't need `SenderThread`, just an [executor pool](https://docs.python.org/3/library/asyncio-eventloop.html#executing-code-in-thread-or-process-pools).

jiivan:
fixes: #3583

# How to test
This PR introduces changes in internal mechanism of reporting status to monitor. General communication between golem and stats.golem.network should be checked.

Also some care should be taken to check overall golem stability. Possibly issues with golem shutdown (does it take significantly more time than standard develop?). Lastly long running test should be performed to confirm wether golem takes significantly more resource than pure develop version.


mfranciszkiewicz:
But why?

Are you planning a transition from `twisted` to `asyncio`?

jiivan:
@mfranciszkiewicz gradually - yes. I treat this PR as a kind of POC, that aio can work together with twstd.

golem/core/golem_async.py
```diff
@@ -97,3 +102,159 @@ def curry(*args, **kwargs):
...
+
+async def _asyncio_wait_for_tasks():
+    while True:
+        task = await _ASYNCIO_TASKS.get()
+        await asyncio.wait({task})
+        _ASYNCIO_TASKS.task_done()
+
+async def asyncio_main():
+    wait_task = asyncio.ensure_future(_asyncio_wait_for_tasks())
+    while _ASYNCIO_RUN.is_set():
```
mplebanski: From looking at `golem_async.py` I can't deduce when happens the race condition that is being synchronized here. `asyncio_start` first calls `_ASYNCIO_RUN.set()` then schedules `asyncio_main`. What is synchronized here?

golem/monitor/monitor.py
```diff
@@ -17,39 +20,13 @@
...
-            except queue.Empty:
-                # send ping message
-                self.sender.send(self.node_info)
-
-    def join(self, timeout=None):
-        self.stop_request.set()
-        self.queue.put(None)
-        super(SenderThread, self).join(timeout)
+@golem_async.run_at_most_every(datetime.timedelta(minutes=10))
+def log_sometimes(msg, d):
```
mplebanski: Why `log_sometimes` name?

golem/core/golem_async.py
jiivan: No race condition. It's used in `golem_async.asyncio_stop()`.

golem/monitor/monitor.py
jiivan: Because it's run at most once every ten minutes. Do you find this name misleading? Have any suggestion?

mplebanski: Yes because log interval is already explicitly stated in the decorator. Maybe `log_routine` is a better fit?

golem/core/golem_async.py
mplebanski: OK

golem/core/golem_async.py
```diff
@@ -97,3 +102,159 @@ def curry(*args, **kwargs):
...
+        await _asyncio_process_thread_queue()
```
mplebanski: If an exception is thrown inside the task being processed, is this thread completing it's execution?

jiivan: Good point, thanks

jiivan:
Linux build: https://buildbot.golem.network/artifacts/aio_monitor/golem-0.19.0+dev320.g37c0695-linux.tar.gz
Windows: https://buildbot.golem.network/artifacts/aio_monitor/golem-0.19.0+dev320.g37c0695-windows.exe https://buildbot.golem.network/artifacts/aio_monitor/golem-0.19.0+dev320.g37c0695-windows.zip
MacOs: https://buildbot.golem.network/artifacts/aio_monitor/golem-0.19.0+dev320.g37c0695-macOS.tar.gz

golem/core/golem_async.py
```diff
@@ -97,3 +102,162 @@ def curry(*args, **kwargs):
...
     return wrapped
+
+
+##
+# ASYNCIO
+##
+
+_ASYNCIO_RUN = threading.Event()
+_ASYNCIO_ID = 'Thread-aio'
+_ASYNCIO_THREAD_QUEUE: queue.Queue = queue.Queue()
```
etam: this type hint is redundant

golem/core/golem_async.py
```diff
@@ -97,3 +102,162 @@ def curry(*args, **kwargs):
...
+_ASYNCIO_TASKS = None
```
etam: and this requires one

golem/core/golem_async.py
```diff
@@ -97,3 +102,162 @@ def curry(*args, **kwargs):
...
+        def curry(*args, **kwargs):
+            if not in_asyncio_thread():
+                _ASYNCIO_THREAD_QUEUE.put_nowait((f, args, kwargs))
+                return None
+            return f(*args, **kwargs)
+        return curry
+    return wrapper
+
+
+def run_at_most_every(delta: datetime.timedelta):
```
etam: from reading just the decorator name it's unclear whether it drops calls that are too frequent (which I thought it did), or if it delays them.

golem/core/golem_async.py
```diff
@@ -97,3 +102,162 @@ def curry(*args, **kwargs):
...
+    last_run = datetime.datetime.min
+
+    def wrapped(f):
+        @functools.wraps(f)
+        async def curry(*args, **kwargs):
+            nonlocal last_run
```
etam, line 157: You can also store this value as an attribute of `f`. I think it might be a bit more clear.

golem/core/golem_async.py
```diff
@@ -97,3 +102,162 @@ def curry(*args, **kwargs):
...
+                        'args': args,
+                        'kwargs': kwargs,
+                    },
+                )
+                await asyncio.sleep(sleep_for.total_seconds())
+                current_delta = datetime.datetime.now() - last_run
+            last_run = datetime.datetime.now()
+            result = f(*args, **kwargs)
+            if asyncio.iscoroutine(result):
+                return await result
```
etam, line 164: shouldn't this be done by caller?

tests/golem/test_opt_node.py
```diff
@@ -648,11 +667,14 @@ def test_start_creates_client(self, _ets, mock_gather_results, reactor, *_):
         assert self.node.client
         assert self.node.client.datadir == self.path
         assert self.node.client.config_desc == self.node_kwargs['config_desc']
-        assert reactor.addSystemEventTrigger.call_count == 2
-        assert reactor.addSystemEventTrigger.call_args_list[0][0] == (
-            'before', 'shutdown', self.node.rpc_router.stop)
-        assert reactor.addSystemEventTrigger.call_args_list[1][0] == (
-            'before', 'shutdown', self.node.client.quit)
+        self.assertAddSystemEvent(reactor)
+        reactor.addSystemEventTrigger.assert_has_calls(
```
etam: isn't this a duplication of `self.assertAddSystemEvent(reactor)`?

golem/core/golem_async.py
jiivan: It's not a property of `f` but of an object returned by a decorator.

golem/core/golem_async.py
jiivan: You're right. Calls should be dropped

golem/core/golem_async.py
jiivan: AFAIK it won't work like that

tests/golem/test_opt_node.py
jiivan: it is, removed

golem/core/golem_async.py
jiivan:
```
New findings:
golem/core/golem_async.py:113: error: Need type annotation for '_ASYNCIO_THREAD_QUEUE'
Changed lines findings:
golem/core/golem_async.py:113: error: Need type annotation for '_ASYNCIO_THREAD_QUEUE'\n
```

mfranciszkiewicz:
I'm in favour of using the `twisted.internet.asyncioreactor` (PoC PR https://github.com/golemfactory/golem/pull/4417)

The reactor substitution itself is -6/+2 lines of code. Other changes come mainly from upgraded dependencies (Twisted, crossbar, autobahn).

The downside is that `asyncioreactor` cannot use a proactor event loop on Windows (I/O completion ports), at least for now.

jiivan:
@mfranciszkiewicz is it possible to use `async def`, `await`, asyncio processing pools etc with twisted reactor?

mfranciszkiewicz:
@jiivan Yes, it's running an asyncio event loop underneath, basically translating Twisted reactor calls to aio event loop calls.

golem/monitor/monitor.py
```diff
@@ -122,91 +88,140 @@ def ping_request(self, ports) -> Optional[Dict]:
...
+            self.p2p_listener,
+            signal='golem.p2p',
+        )
 
-    def on_config_update(self, meta_data):
+    @golem_async.run_in_thread()
+    def send(self, model, loop):
+        url = self.config['HOST']
+        request_timeout = self.config['REQUEST_TIMEOUT']
+        # sender_thread_timeout = self.config['SENDER_THREAD_TIMEOUT']
```
mfranciszkiewicz: Remove?

golem/monitor/monitor.py
mfranciszkiewicz: `log_per_10_min`?

Besides that, this function could do a bit more and take care of `asyncio.ensure_future(...)` stuff.

golem/core/golem_async.py
mfranciszkiewicz: Maybe
```
while asyncio.iscoroutine(result):
    result = await result
```

golem/core/golem_async.py
```diff
@@ -97,3 +100,102 @@ def curry(*args, **kwargs):
...
+                    *args,
+                    **kwargs,
+                    loop=asyncio.get_event_loop(),
+                ),
+            )
+        return curry
+    return wrapped
+
+
+def locked():
```
mfranciszkiewicz: `synchronized`?

golem/monitor/test_helper.py
```diff
@@ -26,3 +27,8 @@ def meta_data():
 class MonitorTestBaseClass(TestCase):
     def setUp(self):
         self.monitor = SystemMonitor(meta_data(), MONITOR_CONFIG)
+        self.loop = asyncio.new_event_loop()
```
mfranciszkiewicz, line 30: What do you think of using `pytest-asyncio` and marking with `@pytest.mark.asyncio`?

golem/core/golem_async.py
jiivan: Like in java? brrr ;)

golem/core/golem_async.py
jiivan: IMHO it should a responsibility of a caller implement this kind of logic.

golem/monitor/monitor.py
jiivan: IMHO 10min shouldn't be a part of a function name as it's a decorators parameter. It would mean changing function name if we only changed timedelta.

golem/monitor/test_helper.py
jiivan: I'm not sure if it's worth introducing a new package.

golem/monitor/monitor.py
mfranciszkiewicz: IMO renaming / creating a new function when changing the delta would be better. `sometimes` is ambiguous.

mfranciszkiewicz (approved):

Issue #3791, opened 2019-01-21
RPC for estimated Concent deposit payment and tx fee


mdtanrikulu:
https://projects.invisionapp.com/share/FNISX71SRXK#/screens/297805835

In user interface prototype, we have estimated concent deposit payment and tx fee of it. 
Need an RPC to provide them.

PR #3801, opened 2019-01-23
RPC for estimated Concent deposit payment and tx fee

jiivan:
* Moved get_estimed_cost() from Client to tasks.rpc.ClientProvider
 * Refactored out TaskManager.get_estimated_cost()
 * All output numerical values are casted to strings
 * Renamed subtask_time to subtask_timeout
 * Renamed num_subtasks to subtasks
 * subtask_timeout is string timedelta as in task creation
 * Added deposit subdictionary
Example input:
```python
{
    'price': '150',  # str(int)
    'subtask_timeout': '00:00:02',  # str(timedelta)
    'subtasks': '5',  # str(int)
}
```
Example output:
```python
            {                                                                          
                "GNT": '5',                                      
                'ETH': '10000',                                         
                'deposit': {                                                           
                    'GNT_required': '10',                         
                    'GNT_suggested': '20',                        
                    'ETH': '20000',                                     
                },                                                                     
            }
```
Actual GNT deposit amount will be inside `<GNT_required, GNT_suggested>`.

fixes: #3791

### DIFF
--- a/golem/task/rpc.py
+++ b/golem/task/rpc.py
@@ -567,3 +567,38 @@ class ClientProvider:
         )
         # Don't wait for _deferred
         return True
+
+    @rpc_utils.expose('comp.tasks.estimated.cost')
+    def get_estimated_cost(self, _task_type: str, options: dict) -> dict:
+        # FIXME task_type is unused
+        options['price'] = float(options['price'])
+        options['subtask_time'] = float(options['subtask_time'])
+        options['num_subtasks'] = int(options['num_subtasks'])
+        def float_to_str(f: float) -> str:
+            return "{0:.18f}".format(f)
+
+        estimated_gnt: float = options['num_subtasks'] \
+            * options['price'] * options['subtask_time']
+        estimated_eth: float = self.client \
+            .transaction_system.eth_for_batch_payment(
+                options['num_subtasks'],
+            ) / denoms.ether
+        estimated_gnt_deposit: typing.List[float] = [
+            value / denoms.ether for value in
+            msg_helpers.requestor_deposit_amount(
+                int(estimated_gnt * denoms.ether),
+            )
+        ]
+        estimated_deposit_eth: float = self.client.transaction_system \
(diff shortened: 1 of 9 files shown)


golem/task/rpc.py
```diff
@@ -567,3 +567,38 @@ def run_test_task(self, task_dict) -> bool:
...
+    @rpc_utils.expose('comp.tasks.estimated.cost')
+    def get_estimated_cost(self, _task_type: str, options: dict) -> dict:
+        # FIXME task_type is unused
+        options['price'] = float(options['price'])
+        options['subtask_time'] = float(options['subtask_time'])
+        options['num_subtasks'] = int(options['num_subtasks'])
+        def float_to_str(f: float) -> str:
+            return "{0:.18f}".format(f)
+
+        estimated_gnt: float = options['num_subtasks'] \
```
Krigpl: I'd strongly encourage to use `int`s instead of `float`s. They are safer to work with and there's absolutely no need for `float`s here. Also you're unnecessarily going back and forth below with `denoms.ether`

tests/golem/task/test_rpc.py
```diff
@@ -23,6 +23,7 @@
 
 
 class ProviderBase(test_client.TestClientBase):
+    maxDiff = None
```
Krigpl, line 26: What's that ?

jiivan: https://docs.python.org/3/library/unittest.html#unittest.TestCase.maxDiff

golem/task/rpc.py
jiivan: I agree but didn't want to mess with backwards compatibility. What do you think @mdtanrikulu ? Can we change all `options` to `int`? Then I'll change all output values to ints also.

mdtanrikulu: @jiivan @Krigpl  if it's fine for the API representation, fine for me too.

Krigpl (approved):
rickandmorty_yes

PR #3804, opened 2019-01-23
bring back detection of task header beyond the deadline to golem core

shadeofblue:


golem/task/taskserver.py
```diff
@@ -388,6 +388,12 @@ def add_task_header(self, task_header: dt_tasks.TaskHeader) -> bool:
                 task_header.signature,
             )
             return False
+        if task_header.deadline < time.time():
+            logger.info(
+                "Task's deadline already in the past. task_header: %r",
```
jiivan, line 393:
```suggestion
                "Task's deadline already in the past. task_header=%r",
```

https://github.com/golemfactory/golem-internal/wiki/Logging-guidelines

jiivan (approved):
buakaw_throw

golem/task/taskserver.py
shadeofblue: okay, I'll try to remember that, although in this case, I'd say it's not worth a separate commit ...

PR #3809, opened 2019-01-24
Use subtasks_count in price estimation

jiivan:


golem/task/rpc.py
```diff
@@ -103,13 +103,6 @@ def prepare_and_validate_task_dict(client, task_dict):
         'concent_enabled',
         client.concent_service.enabled,
     )
-    # TODO #3474
```
mplebanski, line 106: Are we droping this backwards compatibility?

jiivan: This issue is closed. Electron doesn't use this. I would say it's about time.

mplebanski: `scripts/node_integration_tests/tasks/__init__.py` is still using `subtasks`

mplebanski (approved):

mdtanrikulu (approved):

PR #3815, opened 2019-01-24
fix an issue encountered when adding a task with `jpg` output

shadeofblue:


apps/core/task/coretask.py
```diff
@@ -528,12 +528,16 @@ def get_task_kwargs(self, **kwargs):
...
+        logger.debug(
+            "build_minimal_definition. task_type=%r, dictionary=%r",
+            task_type, dictionary
+        )
         definition = task_type.definition()
         definition.options = task_type.options()
         definition.task_type = task_type.name
         definition.compute_on = dictionary.get('compute_on', 'cpu')
         definition.resources = set(dictionary['resources'])
-        definition.subtasks_count = int(dictionary['subtasks_count'])
```
kmazurek, line 536: I don't think this should be renamed, afaik the name of this field is supposed to be `subtasks_count`.

kmazurek (changes requested):

shadeofblue:
I didn't expect the kind of Spanish Inquisition

PR #3830, opened 2019-01-30
B0.19 fix concent integration tests

shadeofblue:


golem/interface/formatters.py
```diff
@@ -63,6 +63,10 @@ def format(self, result):
 
             if self.prettify:
                 return yaml.safe_dump(result, allow_unicode=True, default_flow_style=False)
+
+            if not hasattr(result, 'keys'):
```
mfranciszkiewicz: Maybe `if not isinstance(result, dict)` and put it before `DictSerializer.dump`?

shadeofblue: hmm, `DictSerializer` does a bit more processing and I believe it handles other types of objects as well ... and - on the other hand - I'd say - to paraphrase ducktyping - if it gives keys like a dict, it can be treated as a dict... alternatively, I can put the subsequent line in a try/except so that any objects that do pass the iteration over `keys` are handled correctly and those that are not are just passed as `str(result)` ...

shadeofblue: @mfranciszkiewicz ^ updated

mfranciszkiewicz (approved):

scripts/node_integration_tests/playbooks/concent/concent_base.py
```diff
@@ -15,6 +15,66 @@ def step_clear_provider_output(self):
...
+            on_success=on_success, on_error=on_error,
+        )
+
+    def step_enable_provider_concent(self):
+        return self._step_enable_concent('provider')
+
+    def step_enable_requestor_concent(self):
+        return self._step_enable_concent('requestor')
+
+    def _step_ensure_concent_on(self, role):
```
kmazurek, line 55: Are there any cases where this could fail right after enabling was successful?

kmazurek (approved):

scripts/node_integration_tests/playbooks/concent/concent_base.py
shadeofblue: unsure, but putting it there just in case...

PR #3832, opened 2019-01-31
Skip updating HW presets when no preset name is provided

mfranciszkiewicz:
Prevents unintended hardware preset updates during startup that may cause VM to restart.

Observed on Windows 10 Pro (1803) + Hyper-V.


Wiezzel (approved):
👍

Issue #3817, opened 2019-01-24
golem_messages.exceptions.FieldError: Port not in range 1, 2**16-1 [port:0]


ederenn:
#### Encounered on:

Linux, mainnet
GOLEM Version: 0.18.3+dev739.gb7573ca
Protocol Version: 31
golem_messages Version: 2.23.0

around 10 minutes after starting the node, no upgrade, no additional changes, no new tasks started, just starting. No problems with network at this time.

```
2019-01-24 12:13:13 CRITICAL twisted                             Unhandled Error
Traceback (most recent call last):
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/python/log.py", line 103, in callWithLogger
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/python/log.py", line 86, in callWithContext
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/python/context.py", line 122, in callWithContext
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/python/context.py", line 85, in callWithContext
    
--- <exception caught here> ---
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/internet/posixbase.py", line 614, in _doReadOrWrite
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/internet/tcp.py", line 205, in doRead
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/internet/tcp.py", line 211, in _dataReceived
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/golem/network/transport/tcpnetwork.py", line 335, in dataReceived
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/golem/network/transport/tcpnetwork.py", line 361, in _interpret
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/golem/network/p2p/peersession.py", line 119, in interpret
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/golem/network/transport/session.py", line 75, in interpret
    
[...]

PR #3847, opened 2019-02-04
Add check that only peers with valid address are returned

dybi:
(probably) resolves: https://github.com/golemfactory/golem/issues/3817

I was not able to reproduce the issue, but it seems that this check will prevent the issue from happening again


kmazurek (approved):
LGTM

Issue #3851, opened 2019-02-04
Verification in task with many blender files (multi asset) fails for every subtask


ederenn:
####: Encountered on: Linux, Windows , GOLEM Version: 0.18.3+dev809.gfa38a40 and previous 19.0 rc

Verification of every subtask ends in result:REJECT for multiblend files.

```
(golem-env) ederenn@ederenn-ubnt1804:~/projects/golem$ golemcli --mainnet subtasks show 139b7e38-289d-11e9-97d0-6c15fa96c125
node_id: 0e9c5f4ab51d0cf3deecc25aecd31cc3efce1a6c2038355e93e9545cedf73a4c21de0a8936b73f45552e18359572b918a336934fcd8ea3a48246235b9cbc33fb
node_name: vm linux 18.3
progress: 1.0
results: ['/home/ederenn/.local/share/golem/default/mainnet/ComputerRes/f11691c0-289c-11e9-a014-6c15fa96c125/tmp/books_20001.png']
status: Failure
stderr: /home/ederenn/.local/share/golem/default/mainnet/ComputerRes/f11691c0-289c-11e9-a014-6c15fa96c125/tmp/139b7e38-289d-11e9-97d0-6c15fa96c125/stderr.log
stdout: /home/ederenn/.local/share/golem/default/mainnet/ComputerRes/f11691c0-289c-11e9-a014-6c15fa96c125/tmp/139b7e38-289d-11e9-97d0-6c15fa96c125/stdout.log
subtask_id: 139b7e38-289d-11e9-97d0-6c15fa96c125
time_remaining: 0.0
time_started: 1549299071.6115153

(golem-env) ederenn@ederenn-ubnt1804:~/projects/golem$ golemcli --mainnet subtasks show 13a619a6-289d-11e9-9469-6c15fa96c125
node_id: d1a98d16a83a7ec5cac43e2c87c7c62a28c14f7935cd559107366f0f6689ae2340a0356239b68f7cc3baefcd1b81c58929e3f6f137763fa76dc73155c8779036
node_name: zefla
progress: 1.0
results: ['/home/ederenn/.local/share/golem/default/mainnet/ComputerRes/f11691c0-289c-11e9-a014-6c15fa96c125/tmp/books_30001.png']
status: Failure
[...]

PR #3852, opened 2019-02-04
Fix resource package subdirectories

mfranciszkiewicz:
Resolves: #3851

### DIFF
--- a/golem/task/result/resultpackage.py
+++ b/golem/task/result/resultpackage.py
@@ -55,7 +55,11 @@ class Packager(object):
 
     @classmethod
     def _prepare_file_dict(cls, disk_files) -> Dict[str, str]:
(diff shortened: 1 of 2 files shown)


magdasta (approved):

kmazurek (approved):
LGTM, should resolve #3851

mplebanski:
It looks to me like a blender specific patch in generic code

mfranciszkiewicz:
@mplebanski How is it blender specific?

mfranciszkiewicz:
The flattened directory structure change was never supposed to be implemented. *That* approach was use-case specific. This preserves the original directory structure and is the expected behaviour.

mplebanski:
As discussed offline, OK to merge for me.

Issue #3785, opened 2019-01-18
output files with the same task name are overwritten


ederenn:
#### Encountered on:
Linux, Windows, golem-0.18.3+dev693.g6b46c7a
Steps to reproduce:
- Create a task, wait for task to finish, 
- create second task using the same input file, name task exactly the same as first one, wait for task to finish.
-check if there is new output file in output folder

shadeofblue:
@ederenn how did it work before?

shadeofblue:
iow, please provide missing details:

**Actual result**
_(What is the result of this issue)_

**Expected result**
_(What is the expected result)_

ederenn:
there was added suffix to the output file name to avoid overwriting files

actual result:
two tasks are finished, for example each task has two frames. In output folder there are only two frames created.

expected result:
in output folder there should be created 4 frames in total, with numeral suffixes added to task name to differentiate files

etam:
I don't remember such functionality. And I can't find anything supporting this in the history of our source code.
Anyway, how we should resolve this? Adding incrementing numbers to output filenames might be tricky and error prone with multithreading. Any ideas?

etam:
Or maybe it should be like this: When creating a new task, with the same name and output directory as another task, user should be warned that output files will get overwritten. It sounds like a lot of work with gui and stuff, but I tried to think about what user would expect.

shadeofblue:
@ederenn please confirm that this feature used to work in version b0.18.X

shadeofblue:
@etam we could try to do that on the back end when starting the task - may indeed by trick-ish if there is nothing created when the task is started and only when it's finished...

PR #3854, opened 2019-02-05
Bugfix output files with the same task name are overwritten

rwrzesien:
Resolves https://github.com/golemfactory/golem/issues/3785

### DIFF
--- a/apps/rendering/task/framerenderingtask.py
+++ b/apps/rendering/task/framerenderingtask.py
@@ -486,7 +486,7 @@ class FrameRenderingTask(RenderingTask):
             )
         )
 
-    def _add_suffix_to_existing_directory(self, directory):
+    def _add_suffix_to_existing_directory(self, directory):  # pylint: disable=no-self-use
         assert isinstance(directory, str)
 
         suffix = 1
--- a/apps/rendering/task/framerenderingtask.py
+++ b/apps/rendering/task/framerenderingtask.py
@@ -480,8 +480,8 @@ class FrameRenderingTask(RenderingTask):
         return self._add_suffix_to_existing_directory(
             os.path.normpath(
                 os.path.join(
-                    self.task_definition.task_id,
-                    os.path.dirname(self.output_file)
+                    os.path.dirname(self.output_file),
+                    self.task_definition.task_id
                 )
             )
         )
--- a/apps/rendering/task/framerenderingtask.py
+++ b/apps/rendering/task/framerenderingtask.py
@@ -154,16 +154,25 @@ class FrameRenderingTask(RenderingTask):
     def get_output_names(self):
         if self.use_frames:
             dir_ = os.path.dirname(self.output_file)
-            output_names = [os.path.normpath(os.path.join(dir_, self._get_output_name(frame))) for frame in self.frames]
+            output_names = [
+                os.path.normpath(
+                    os.path.join(dir_, self._get_output_name(frame))
+                ) for frame in self.frames
+            ]
         else:
             output_names = super(FrameRenderingTask, self).get_output_names()
 
         return self._suffix_existing_output_names(output_names)
 
     def _suffix_existing_output_names(self, output_names):
-        """ Checks if any of output names already exists, if yes, adds suffix to output name. """
+        """
+        Checks if any of output names already exists,
+        if yes, adds suffix to output name.
+        """
         assert isinstance(output_names, list)
-        assert all([isinstance(output_name, str) for output_name in output_names])
+        assert all(
+            [isinstance(output_name, str) for output_name in output_names]
+        )
 
         output_directory = os.path.dirname(self.output_file)
 
@@ -172,8 +181,11 @@ class FrameRenderingTask(RenderingTask):
             output_file_path = os.path.join(output_directory, output_name)
 
             while os.path.exists(output_file_path):
-                output_file_name, output_file_extension = os.path.splitext(output_file_path)
-                output_names[i] = f'{output_file_name}_{suffix}{output_file_extension}'
+                output_file_name, output_file_extension = os.path.splitext(
+                    output_file_path
+                )
+                output_names[i] = \
+                    f'{output_file_name}_{suffix}{output_file_extension}'
                 suffix += 1
 
         return output_names
--- a/apps/rendering/task/framerenderingtask.py
+++ b/apps/rendering/task/framerenderingtask.py
@@ -159,8 +159,7 @@ class FrameRenderingTask(RenderingTask):
                     os.path.join(self.output_dir, self._get_output_name(frame))
                 ) for frame in self.frames
             ]
-        else:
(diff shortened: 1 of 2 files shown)


etam (commented):
Question: Suppose the following scenario:

1. A task is created with 2 frames. Output files are "task0.png" and "task1.png" (or something like that).
2. A task is created with 4 frames. Output files are "task0_1.png", "task1_1.png", "task2.png", "task3.png"?

Am I correct? If so, then it's not a good solution.

rwrzesien:
@etam Yes, exactly. In issue description the expected behavior is described as:

> in output folder there should be created 4 frames in total, with numeral suffixes added to task name to differentiate files

Which part of this requirement is not filled in?

etam:
Probably requirements are not well specified. For me such output files suggest that "task?.png" files are one thing and "task?_?.png" another. But that's probably something for UX team to decide how this should work. I think that's confusing, but if UX agrees that it's how it should be, then I won't object anymore.

rwrzesien:
@etam 

> Probably requirements are not well specified.

I agree, so if you want to propose a better expected result, please let me know so I can implement a solution that fits. Saying that solution is not good without proposing a new one doesn't give me much area to improve.

etam:
Quick idea: If task has multiple frames, then put results in a subfolder that is named after task. If folder already exists, add suffix.

So for sequence of tasks with 2, 1, 3 and 1 frames the result would be:
```
task.png
task_1.png
task/task0.png
task/task1.png
task_1/task0.png
task_1/task1.png
task_1/task2.png
```

etam:
I asked @jamuszyn for comments.

jamuszyn:
> Quick idea: If task has multiple frames, then put results in a subfolder that is named after task. If folder already exists, add suffix.
> 
> So for sequence of tasks with 2, 1, 3 and 1 frames the result would be:
> 
> ```
> task.png
> task_1.png
> task/task0.png
> task/task1.png
> task_1/task0.png
> task_1/task1.png
> task_1/task2.png
> ```

This sounds reasonable. I would also include timestamps (task creation) to those folders (not the filenames).

tests/apps/rendering/task/test_framerenderingtask.py
```diff
@@ -130,6 +131,26 @@ def test_get_output_names(self):
...
         assert len(output_names) == 0
 
+    def test__get_output_dir(self):
+        frame_task = self._get_frame_task(True)
+        with mock.patch(
+            'apps.rendering.task.framerenderingtask.os.path.exists',
+            return_value=False
+        ):
+            output_dir = frame_task._get_output_dir()
+        assert output_dir == os.path.dirname(
```
dybi: two remarks:
1) please use `self.assertEquals()`
2) could you please put use string literal here, so it can be deduced from the test, how the path should look like?

You might also put comments at the beginning of the test explaining, i.e.:
```
"""
cwd: path_to_golem/tasks/whatever
task_name: some_task777
...
"""
```
etc, if you recon it would help understanding what is being concatenated ;)

tests/apps/rendering/task/test_framerenderingtask.py
```diff
@@ -130,6 +131,26 @@ def test_get_output_names(self):
...
+            frame_task.task_definition.output_file)
+
+    def test__get_output_dir_when_directory_already_exists(self):
+        frame_task = self._get_frame_task(True)
+        with mock.patch(
+            'apps.rendering.task.framerenderingtask.os.path.exists',
+            side_effect=[True, False]
+        ):
+            output_dir = frame_task._get_output_dir()
+        assert output_dir == \
```
dybi: same as above

dybi (approved):
Looks okay, but I would ask to make improvements in the tests (please see comments)
horse_approves

tests/apps/rendering/task/test_framerenderingtask.py
etam: We are using `pytest` to run tests, which handles `assert a == b` as nicely as `self.assertEquals()`. For me plain `assert` is more readable. Do you have any good reason to use `self.assertEquals()`?

dybi: @etam, in such case, the only reason would be personal preference. So, consider my remark as null and void ;)

rwrzesien: @dybi 

> 1. please use `self.assertEquals()`

I prefer `self.assertXXX` but I wanted to fit into existing style. 

> 2. could you please put use string literal here, so it can be deduced from the test, how the path should look like?

The case is that in each run the path is regenerated and is random (or build from something, I am not sure from what, but not from `subtask_id` nor `task_id`).

dybi: @rwrzesien, the formula that is used currently `assert output_dir == os.path.dirname(
            frame_task.task_definition.output_file)` works only because `frame_task.task_definition.task_id` is equal to `""`. Please include it calculations if using string literals is not feasible.

rwrzesien: @dybi done :+1:

tests/apps/rendering/task/test_framerenderingtask.py
rwrzesien: :+1: done

ederenn:
screenshot from 2019-02-08 16-57-16


- still not working for multiframe tasks, folders are not created, tasks are overwritten
- if two tasks have the same name but one task is single frame, second task multiframe it creates a mess in output directory
to be consistent  we should create separate folders for each multiframe task, starting from first output

apps/rendering/task/framerenderingtask.py
```diff
@@ -472,6 +476,26 @@ def _update_preview_task_file_path(self, preview_task_file_path):
...
+        return self._add_suffix_to_existing_directory(
+            os.path.normpath(
+                os.path.join(
+                    os.path.dirname(self.output_file),
+                    self.task_definition.task_id
+                )
+            )
+        )
+
+    def _add_suffix_to_existing_directory(self, directory):  # pylint: disable=no-self-use
```
jiivan, line 489: No self use lint is valid here. It should either be a function or classmethod

tests/apps/rendering/task/test_framerenderingtask.py
```diff
@@ -3,6 +3,7 @@
 import uuid
 from pathlib import Path
 
+import mock
```
jiivan, line 6:
```suggestion
from unittest import mock
```

apps/rendering/task/framerenderingtask.py
shadeofblue: actually, probably a `staticmethod`

etam:
hello?

etam:
anyone here?

kmazurek:
Closing since the underlying issue has been fixed in #4037.

PR #3856, opened 2019-02-06
Retry starting Hyper-V

mfranciszkiewicz:


golem/docker/hypervisor/hyperv.py
```diff
@@ -62,6 +62,8 @@ class events(Enum):
     },
 }
 
+START_VM_RETRIES = 2
```
Wiezzel: Maybe put it in `HyperVHypervisor` class with all the other constants?

Wiezzel (approved):
:+1:

PR #3857, opened 2019-02-06
Use native SSH in docker-machine

Wiezzel:
~The script would raise an error in case the VM's IP has changed. Try-catch block with certificate regeneration was added to mitigate against this issue.~

~Changed error class in try-except block in `HyperVHypervisor.start_vm()` since `run_powershell()` raises `RuntimeError` not `CalledProcessError`.~

## Use native SSH in docker-machine

SSH connections to VMs were very slow due to bug in OpenSSH. Actually, every time ssh was called it would wait for the amount of time specified by the ConnectTimeout option (which obviously **is not** the expected behaviour for this option).

Because the ConnectTimeout was hard-coded in docker-machine to 10 seconds we could not reconfigure it. Instead we decided to use the --native-ssh option for docker-machine which doesn't use OpenSSH at all but the native (Go-based) implementation.

Since this option significantly speeded up starting VM we don't need the custom script for doing it.


mfranciszkiewicz (dismissed):

mfranciszkiewicz (approved):
Even better

PR #3860, opened 2019-02-07
Create DockerWhitelist class

Krigpl:
Just per repository since tags are mutable so it's not secure in this sense. Alternatively, we can use image IDs which are not user-friendly.


mfranciszkiewicz (approved):

Wiezzel:
@Krigpl I believe you should change this PRs base to `epic/env_refactor`.

tests/golem/envs/docker/cpu/test_env.py
```diff
@@ -309,6 +310,12 @@ def test_env_disabled(self):
         with self.assertRaises(ValueError):
             self.env.install_prerequisites(prereqs)
 
+    def _patch_whitelist(self, return_value):
+        # Has to use twisted's patch because standard one doesn't work with
+        # Deferreds very well
```
Wiezzel, line 315: You could have used the same approach as I did with `local_client` in `test_pull_image_error` but this looks just as good.

Wiezzel (approved):

PR #3873, opened 2019-02-11
Add additional subtask info

### LINKED ISSUE #3850: Retain node info in finished subtasks on restart
Follow-up to: #3846.

When restarting a task with the option to retain finished subtasks (RPC call `comp.task.restart_subtasks`), finished subtask results are copied over to newly created subtasks. However, the node info from the original subtask (name, id, address) is not included in the state of the new subtask. This has no impact on Golem's operation, but would be a UX improvement.

The relevant call can be found [here](https://github.com/golemfactory/golem/blob/307022ea06a34bec682c4152e0b072b6686e46ef/golem/task/taskmanager.py#L549). The default empty strings should be replaced by relevant fields from the original subtask's state (`self.tasks_states[old_task_id]...`).

pawelkisielewicz:
Resolves https://github.com/golemfactory/golem/issues/3850

### DIFF
--- a/golem/task/rpc.py
+++ b/golem/task/rpc.py
@@ -173,7 +173,8 @@ def _restart_subtasks(
         client.task_manager.copy_results(
             old_task_id=old_task_id,
             new_task_id=new_task.header.task_id,
-            subtask_ids_to_copy=subtask_ids_to_copy
+            subtask_ids_to_copy=subtask_ids_to_copy,
+            address=client.node.pub_addr,
         )
     # Function passed to twisted.threads.deferToThread can't itself
     # return a deferred, that's why I defined inner deferred function
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -514,7 +514,8 @@ class TaskManager(TaskEventListener):
             self,
             old_task_id: str,
             new_task_id: str,
-            subtask_ids_to_copy: Iterable[str]) -> None:
+            subtask_ids_to_copy: Iterable[str],
+            address: Optional[str] = '') -> None:
 
         try:
             old_task = self.tasks[old_task_id]
@@ -540,11 +541,14 @@ class TaskManager(TaskEventListener):
             new_subtask_id = extra_data.ctd['subtask_id']
             self.subtask2task_mapping[new_subtask_id] = \
                 new_task_id
+
+            subtask_states: SubtaskState = next(iter(
+                self.tasks_states[old_task_id].subtask_states.values()))
             self.__add_subtask_to_tasks_states(
-                node_name=None,
-                node_id=None,
-                address=None,
-                price=0,
+                node_name=subtask_states.node_name,
+                node_id=subtask_states.node_id,
+                address=address,
+                price=subtask_states.price,
                 ctd=extra_data.ctd)
             new_subtasks_ids.append(new_subtask_id)
 
--- a/tests/golem/task/test_taskmanager.py
+++ b/tests/golem/task/test_taskmanager.py
@@ -1121,6 +1121,10 @@ class TestTaskManager(LogTestCase, TestDatabaseWithReactor,  # noqa # pylint: di
         self.tm.tasks['old_task_id'] = old_task
         self.tm.tasks['new_task_id'] = new_task
         self.tm.tasks_states['new_task_id'] = TaskState()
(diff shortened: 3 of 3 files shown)


tests/golem/task/test_taskmanager.py
```diff
@@ -1121,6 +1121,10 @@ def test_copy_results_subtasks_properly_generated(self):
         self.tm.tasks['old_task_id'] = old_task
         self.tm.tasks['new_task_id'] = new_task
         self.tm.tasks_states['new_task_id'] = TaskState()
+        self.tm.tasks_states['old_task_id'] = TaskState()
+        subtask_states = self.tm.tasks_states['old_task_id'].subtask_states
+        subtask_states['old_subtask_id1'] = SubtaskState()
+        subtask_states['old_subtask_id1'].node_name = 'abcabc'
```
shadeofblue, line 1127: @pawelkisielewicz could we have `'abcabc'` defined in one place instead of two?

golem/task/rpc.py
```diff
@@ -173,7 +173,8 @@ def deferred():
         client.task_manager.copy_results(
             old_task_id=old_task_id,
             new_task_id=new_task.header.task_id,
-            subtask_ids_to_copy=subtask_ids_to_copy
+            subtask_ids_to_copy=subtask_ids_to_copy,
+            address=client.node.pub_addr,
```
kmazurek, line 177: This is not the expected behaviour - the address here should not be the requestor's own address. Now that I think if it, this field should probably stay empty - the state will be updated anyway once the subtask is assigned to a provider and there's no added value in knowing the address of the provider who previously failed on the subtask.

golem/task/taskmanager.py
```diff
@@ -540,11 +541,14 @@ def copy_results(
             new_subtask_id = extra_data.ctd['subtask_id']
             self.subtask2task_mapping[new_subtask_id] = \
                 new_task_id
+
+            subtask_states: SubtaskState = next(iter(
```
kmazurek, line 545: There's a mapping between new and old subtasks available, you should use it here. please see `subtasks_to_copy` above.

kmazurek (changes requested):
#3850 was not merged into develop yet and it contains some changes to this code which might be relevant. Please update this once the merge is done.

tests/golem/task/test_taskmanager.py
pawelkisielewicz: I thought about it, but didn't want to create a new variable. Changed :+1:

golem/task/rpc.py
pawelkisielewicz: OK, removed. But please check the rest of flow in this function. `address` is used only in `self.__add_subtask_to_tasks_states` and **only in logger:**
```python
logger.debug('add_subtask_to_tasks_states(%r, %r, %r, %r)',
                     node_name, node_id, ctd, address)
```
Do we want to keep this and always log empty string? In this case I think that `address` should be simply removed.

golem/task/taskmanager.py
pawelkisielewicz: OK, I have only question about it. `subtasks_to_copy` is empty unless I pass `subtask_id` as `subtask_ids_to_copy`. And in this case subtask state if Failure. Please give me some more info how to do this. And how should it be called in test:
```python
self.tm.copy_results('old_task_id', 'new_task_id', ['some_subtask_id1 ?? '])
```
or 
```python
self.tm.copy_results('old_task_id', 'new_task_id', [])
```
as previous?

golem/task/rpc.py
kmazurek: Ok, I think that in this case it's fine to remove the `address` parameter.

golem/task/taskmanager.py
kmazurek: You can create a mapping analogous to `subtasks_to_copy` based on the `start_task` field, but for all other subtasks of the original task. Then, for each of the new subtasks created (after the call to `query_extra_data`, which assigns the value of `start_task`) you can try accessing the old subtask state. If that fails for some reason, just put in empty strings as defaults.

shadeofblue:
@zakaprov @pawelkisielewicz  do you have any idea on the status of this pull request?

kmazurek:
@shadeofblue There are still some changes left to be done in order for this to work (see the most recent comments from the discussions here).

PR #3875, opened 2019-02-11
Remove unnecessary attributes from TaskSession

jiivan:
Lints cleaning in TaskSession

### DIFF
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -33,7 +33,6 @@ from golem.ranking.manager.database_manager import (
 from golem.resource.resourcehandshake import ResourceHandshakeSessionMixin
 from golem.task import taskkeeper
 from golem.task.server import helpers as task_server_helpers
-from golem.task.taskstate import TaskState
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


golem/task/tasksession.py
```diff
@@ -142,7 +137,22 @@ def __init__(self, conn):
         self.err_msg = None  # Keep track of errors
         self.__set_msg_interpretations()
 
-        # self.threads = []
+    @property
```
mplebanski, line 140: Isn't using a @property here a bit of overkill? None of the class members are dynamically computed inside getter methods -> They could be simple fields

jiivan: Those attributes result in (justified IMO) linter warning: `too-many-instance-attributes`. They are just convenience variables to not i.e call `self.conn.server.client.concent_service` but `self.concent_service` instead. I believe that in this context properties are simpler than multiple fields. IMHO it's more straightforward to define them as RO properties. They're not meant to be modified. They're just convenience accessors.

mplebanski (approved):

PR #3876, opened 2019-02-11
Collect transcoding results

bidzyyys:


apps/transcoding/task.py
```diff
@@ -92,18 +89,22 @@ def initialize(self, dir_manager: DirManager):
...
             super(TranscodingTask, self).accept_results(subtask_id,
                                                         result_files)
+            # TODO remove these logs
+            for idx, result in enumerate(result_files):
+                logger.info("Result[{}]: {}".format(idx, result))
+
             self.num_tasks_received += 1
 
     def _get_next_subtask(self):
+        # with self.lock:
```
weaselix: uncomment it, please.

golem/verificator/ffmpeg_verifier.py
```diff
@@ -1,7 +1,55 @@
+import logging
+import os
+
 from golem.verificator import CoreVerifier
+from golem.verificator.verifier import SubtaskVerificationState
+
+logger = logging.getLogger("apps.ffmpeg")
```
weaselix: logger = logging.getLogger(\_\_name\_\_)

golem/verificator/ffmpeg_verifier.py
```diff
@@ -1,7 +1,55 @@
...
+            self.message = "Missing results in verification_data"
+        except RuntimeError:
+            self.message = "No results from provider"
+        except FileNotFoundError:
+            self.results.clear()
+            self.message = "No proper task result found"
+
+        return verdict
+
+    @staticmethod
```
rkondziolka: why do you decorate that method as static?

golem/verificator/ffmpeg_verifier.py
```diff
@@ -1,7 +1,55 @@
...
+        if not results:
+            raise RuntimeError
+
+        for result in results:
+            self._check_file_existence(result)
+            self.results.append(result)
+
+        return True
+
+    @staticmethod
```
rkondziolka: why do you decorate that method as static?

golem/verificator/ffmpeg_verifier.py
```diff
@@ -1,7 +1,55 @@
...
+    def _get_result_info(verification_data):
+        return verification_data["results"]
+
+    def _check_results(self, results):
+        if not results:
+            raise RuntimeError
+
+        for result in results:
+            self._check_file_existence(result)
```
weaselix: why do you check existence of result files? It is done by your supermethod (CoreVerifier.simple_verification). Use it and to avoid repeating.

golem/verificator/ffmpeg_verifier.py
```diff
@@ -1,7 +1,55 @@
...
+
+        return verdict
+
+    @staticmethod
+    def _get_result_info(verification_data):
+        return verification_data["results"]
+
+    def _check_results(self, results):
+        if not results:
+            raise RuntimeError
```
weaselix: It is done by your supermethod (CoreVerifier.simple_verification). Try to use it.
By the way:
1. Don't raise exception and catch it immediately 
2. don't raise RuntimeError in such situation.

tests/golem/docker/test_docker_manager.py
```diff
@@ -303,9 +303,9 @@ def command(key, *args, **kwargs):
 
         from apps.core import nvgpu
         if nvgpu.is_supported():
-            expected = 6
+            expected = 7
         else:
-            expected = 4
+            expected = 5
```
weaselix, line 308: why do you need edit test_docker_manager?

tests/golem/verificator/test_ffmpeg_verifier.py
```diff
@@ -0,0 +1,90 @@
+from os import remove
```
weaselix, line 1: Could you test accept_results in different scenarios? (like, two subtasks and there are valid and all expected files for the first one and there is something lacking for the second one)

apps/transcoding/task.py
bidzyyys: Ok

tests/golem/docker/test_docker_manager.py
bidzyyys: Because of failing tests - you've added FFmpeg docker image

weaselix: ok, I did not know that it is necessary.

golem/verificator/ffmpeg_verifier.py
bidzyyys: Ok

apps/transcoding/task.py
bidzyyys: It causes deadlock (`test_query_extra_data`), fix it by yourself @rkondziolka

weaselix: Actually, `query_extra_data` acquires that lock, so just remove that line.

Issue #3862, opened 2019-02-07
Delete outdated tasks before exposing tasks to clients


kmazurek:
Related to: #3807

When the field `clean_tasks_older_than_seconds` in client config is set to a value larger than 0, tasks older than their deadline + config value will be deleted.

In its current form, deleting tasks is done by calling [`clean_old_tasks`](https://github.com/golemfactory/golem/blob/85fea6dabb190102add7f2fb79aa0d70538e5f49/golem/client.py#L1162). This requires the tasks to be restored and loaded into task manager, making them available to clients and other parts of golem core.

This leads to a race condition - if a list of tasks is requested before cleaning old tasks is finished, the list will include tasks which are no longer available in golem core.

Fixing this will most likely require changing the task deletion mechanism to not require exposing tasks in task manager.

PR #3878, opened 2019-02-11
Prevent loading default clean interval from old config

kmazurek:
Related to: #3862

Adds a client config field which controls whether tasks and resources
should be cleaned periodically based on their age. This ensures that
deleting outdated tasks is disabled by default, regardless of the client
config from previous versions.


shadeofblue (approved):
I believe it's okay... though - shouldn't there be a test accompanying the change ?

jiivan (approved):

PR #3886, opened 2019-02-13
Add `sleep` in `RequestorPaysTest` tests

dybi:
`RequestorPaysTest` were ocasionally failing, probably due to the fact, that there might be not enough time to synchronize blockchains state after payment transactions

### DIFF
--- a/scripts/concent_acceptance_tests/base.py
+++ b/scripts/concent_acceptance_tests/base.py
@@ -367,8 +367,7 @@ class SCIBaseTest(ConcentBaseTest):
 
         if sci.get_deposit_value(sci.get_eth_address()) < amount:
             raise RuntimeError("Deposit failed")
-        sys.stderr.write('Long sleep. hrrrr\n')
-        time.sleep(120)
+        self._blockchain_sleep(120)
         dump_balance(sci)
 
     def requestor_put_deposit(self, price: int):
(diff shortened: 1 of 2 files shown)


scripts/concent_acceptance_tests/force_payment/test_payment.py
```diff
@@ -321,6 +321,9 @@ def _on_batch(receipt: golem_sci.structs.TransactionReceipt):
             'Batch transfer timeout',
         )
         sys.stderr.write('\n')
+        sleep_time = 60
+        sys.stderr.write(f'Going to sleep for: {sleep_time}\n')
+        time.sleep(sleep_time)
```
jiivan: Can we unify this with:
https://github.com/golemfactory/golem/blob/6965cc38a84da5bf1277a76419fcc6f6bde912a8/scripts/concent_acceptance_tests/base.py#L370-L371
Maybe we could add `.blockchain_sleep()` method to base test class?

shadeofblue (approved):

scripts/concent_acceptance_tests/force_payment/test_payment.py
dybi: OK

scripts/concent_acceptance_tests/base.py
```diff
@@ -381,3 +380,8 @@ def requestor_put_deposit(self, price: int):
     def provider_put_deposit(self, price: int):
         amount, _ = helpers.provider_deposit_amount(price)
         return self.put_deposit(self.provider_sci, amount)
+
+    @staticmethod
+    def _blockchain_sleep(sleep_time=60):
```
jiivan:
```suggestion
    def blockchain_sleep(sleep_time=60):
```

This method is not internal to base class

jiivan:
What about renaming calls from `_blockchain_sleep()` to `blockchain_sleep()`?

Issue #3858, opened 2019-02-06
Subtask dies with status 'Download'


ZmijaWA:
**Description**
_(Short description what is wrong, what area and where)_
Golem Version: golem-0.18.3+dev803.gfa38a40
Golem-Messages version:
Electron version (if used): v3
OS: Mac
Branch (if launched from source):
Mainnet/Testnet: Mainnet
Description of the issue:
Some subtask "dies" and the whole task is not finished.
Is usually happen when we have more than 20 subtask.
When one subtask failed on one node then it is game over for a whole task (subtask is ignored by the other nodes )

node_id: ac201342cfc39645202f2d84905ff9d757e60427be4d7681cdefd77b0f9febaf77e63434f6a2a27265e899f22614aa6427e1d0f5ce9ab5306082f7051483c8f1
node_name: vm linux
progress: 0.0
results: []
status: Failure
stderr: [GOLEM] Timeout
stdout: 
subtask_id: 7acb24e4-2a1e-11e9-b577-e593c8c41e5e
time_remaining: 0
time_started: 1549464600.98813

More info please take a look in logs 

**Repro steps**
_(short steps description how it happens)_

**Actual result**
_(What is the result of this issue)_

**Expected result**
_(What is the expected result)_

**Logs**
_(Any additional information that could help to resolve the issue, which systems were checked to reproduce the issue)_

```
2019-02-06 16:00:01 INFO     golem.task.taskmanager              Subtask '7acb24e4-2a1e-11e9-b577-e593c8c41e5e' dies with status 'Downloading'
2019-02-06 16:00:01 DEBUG    golem.task.taskkeeper               COMPTASK DUMP: /Users/lukaszzmijewski/Library/Application Support/golem/default/mainnet/tasks/comp_task_keeper.pickle
2019-02-06 16:00:01 DEBUG    apps.core                           _mark_subtask_failed. subtask_id='7acb24e4-2a1e-11e9-b577-e593c8c41e5e'
```

```2019-02-06 16:00:01 DEBUG    golem.task.taskmanager              Notice task updated. task_id=5351c4e2-2a1e-11e9-8e00-e593c8c41e5e, subtask_id=7acb24e4-2a1e-11e9-b577-e593c8c41e5e,op=SubtaskOp.TIMEOUT, persist=True```

**Proposed Solution**
_(What could be a solution for that issue)_

jiivan:
> task timeout: 1h
> subtask timeout: 15min
> 1 frame
[...]

PR #3888, opened 2019-02-14
Properly break task selection algorithm to ask for subtasks

jiivan:
fixes: #3858 

Providers' `TaskServer.requested_tasks` set used to grow potentially infinitely. It wasn't cleared on `CannotAssignTask`.

### DIFF
--- a/golem/task/taskserver.py
+++ b/golem/task/taskserver.py
@@ -211,8 +211,8 @@ class TaskServer(
         return self.task_keeper.environments_manager.get_environment_by_id(
             env_id)
 
-    # This method chooses random task from the network to compute on our machine
     def request_task(self) -> Optional[str]:
+        """Chooses random task from network to compute on our machine"""
         theader = self.task_keeper.get_task(self.requested_tasks)
         if theader is None:
             return None
@@ -281,7 +281,7 @@ class TaskServer(
                                                       supported)
         except Exception as err:
             logger.warning("Cannot send request for task: {}".format(err))
-            self.task_keeper.remove_task_header(theader.task_id)
+            self.remove_task_header(theader.task_id)
 
         return None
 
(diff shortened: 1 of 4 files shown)


golem/task/taskserver.py
```diff
@@ -211,8 +211,8 @@ def get_environment_by_id(self, env_id):
         return self.task_keeper.environments_manager.get_environment_by_id(
             env_id)
 
-    # This method chooses random task from the network to compute on our machine
     def request_task(self) -> Optional[str]:
+        """Chooses random task from network to compute on our machine"""
```
mfranciszkiewicz, line 215: It looks like `task_keeper.get_task` may now return tasks that have no subtasks left. AFAIR the headers are removed and ids are kept in `task_keeper.removed_tasks`. Maybe `get_task` should also exclude those?

jiivan: `removed_tasks` are only kept for 3 minutes
https://github.com/golemfactory/golem/blob/0df4d4241b28cf55f339d157e1fbd5e1031a1f05/golem/task/taskkeeper.py#L326

jiivan: We can't really tell (from the providers perspective) whether there are no more subtasks. Temporarily all subtasks may be delegated to providers, but if one subtask fails it becomes available again. I think this is core of #3858 I.e. that providers won't ask for subtasks once they get *CannotAssignTask*.

More elaborate solution would be to distinguish when:
 1. There are no more tasks for now, but keep asking (not all subtasks are confirmed)
 1. All subtasks are computed & verified. Don't ask me anymore (or you're banned, don't ask)

mfranciszkiewicz (approved):

golem/task/taskserver.py
shadeofblue: @jiivan 👍 

added:
https://github.com/golemfactory/golem-messages/issues/323
https://github.com/golemfactory/golem/issues/3898

Issue #3871, opened 2019-02-11
restarted task showing up as timed out


shadeofblue:

ederenn:
Steps to reproduce:
- run Golem
- start a task with at least two frames and wait for task to finish 
- restart whole Golemapp
- when Golem is launched again, in task finished previously try to restart one finished frame from second preview window
- whole task will end up in timeout immediately

![screenshot from 2019-02-12 15-47-04](https://user-images.githubusercontent.com/32061891/52649067-de18c880-2ee7-11e9-960b-430de749cca8.png)

![screenshot from 2019-02-12 15-47-34](https://user-images.githubusercontent.com/32061891/52649077-e53fd680-2ee7-11e9-9b67-6da3d52fa071.png)


if Golem is running without restarts of an application it is possible to restart a frame of a finished task without timeout

- run Golem
- start a task with at least two frames and wait for task to finish 
- go to the second preview window and restart one finished frame 
- task will be restarted

![screenshot from 2019-02-12 16-38-31](https://user-images.githubusercontent.com/32061891/52649190-233cfa80-2ee8-11e9-85c2-34fd47b006e0.png)

even though task has been restarted successfully no new nodes had started computations for restarted subtasks

logs on debug

Windows:
https://drive.google.com/open?id=1gu1e-pQvXhfAaZDW1jLBy033I_WHevol

Linux:
https://drive.google.com/open?id=1ykYtPC3qHlr2MEA4pm6O8-e1aTTrw6rA

PR #3889, opened 2019-02-14
 Enabled frame restarts on finished tasks

kmazurek:
Fixes: #3871

This changes the RPC call `comp.task.subtasks.frame.restart` to also work on tasks which are finished. Internally, this method uses the already existing logic for subtask restarts.

### DIFF
--- a/golem/task/rpc.py
+++ b/golem/task/rpc.py
@@ -20,14 +20,10 @@ from golem.core import common
 from golem.core import deferred as golem_deferred
 from golem.core import simpleserializer
 from golem.ethereum import exceptions as eth_exceptions
-from golem.ethereum.transactionsystem import TransactionSystem
 from golem.resource import resource
 from golem.rpc import utils as rpc_utils
-from golem.task import taskbase
-from golem.task import taskkeeper
-from golem.task import taskstate
-from golem.task import tasktester
-
+from golem.task import taskbase, taskkeeper, taskstate, tasktester
+from golem.task.taskstate import SubtaskState
 
 logger = logging.getLogger(__name__)
 TASK_NAME_RE = re.compile(r"(\w|[\-\. ])+$")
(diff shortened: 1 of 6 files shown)


golem/task/rpc.py
```diff
@@ -537,13 +533,47 @@ def restart_task(self, task_id: str, force: bool = False) \
...
+        logger.debug('restart_frame_subtasks. task_id=%r, frame=%r',
+                     task_id, frame)
+
+        frame_subtasks: typing.Dict[str, SubtaskState] =\
+            self.task_manager.get_frame_subtasks(task_id, frame)
+
+        if not frame_subtasks:
+            logger.error('Frame restart failed, frame has no subtasks.'
+                         'task_id=%r, frame=%r', task_id, frame)
+            return
```
mfranciszkiewicz, line 557: IMO we should raise an exception here, so that the other side (CLI or GUI) will know the cause of failure

Wiezzel: Actually, if look into `get_frame_subtasks` you can see that `None` is returned also when task with the given ID is not found. This relates to the next except clause. If `frame_subtasks` is not `None` and getting `task_state` raises a `KeyError` it means that something bad has happened and `taskmanager` is in an invalid state.

golem/task/rpc.py
```diff
@@ -537,13 +533,47 @@ def restart_task(self, task_id: str, force: bool = False) \
...
+
+        try:
+            task_state = self.client.task_manager.tasks_states[task_id]
+        except KeyError:
+            logger.error('Frame restart failed, unknown task.'
+                         'task_id=%r, frame=%r', task_id, frame)
+            return
```
Wiezzel: As stated above it's more like "invalid task_manager state" than "unknown task". But do we even need this try/except if we use `@run_safe` decorator?

Wiezzel (approved):
Looks fine, but see comments.

golem/task/rpc.py
kmazurek: Got it, in that case I'm going to drop the next try-except and keep the `@run_safe` decorator for catching any unexpected errors.

golem/task/rpc.py
kmazurek: Makes sense, I'm going to remove the try-except here.

PR #3894, opened 2019-02-15
Added integration test for frame restarts

kmazurek:


shadeofblue (approved):
looks good

scripts/node_integration_tests/playbooks/golem/restart_frame.py
```diff
@@ -0,0 +1,41 @@
+import typing
+
+from ..base import NodeTestPlaybook
+
+
+class RestartFrame(NodeTestPlaybook):
+    provider_node_script = 'provider/debug'
+    requestor_node_script = 'requestor/debug'
+    requestor_node_script_2 = 'requestor/always_accept_provider'
+    task_settings = 'default'
```
shadeofblue, line 10: though - we should probably make the task multi-frame and make sure only the specified frame is restarted and all of them...

kmazurek: Makes sense, but as discussed - let's make this as a separate change.

jiivan (approved):
lgtm

PR #3896, opened 2019-02-15
TaskSession almost stateless

jiivan:
depends on: https://github.com/golemfactory/golem-messages/pull/321
 1. Remove `task_id` from *TaskSession*
 1. Send *TaskToCompute* along with *WaitingForResults* (new GM)
 1. Verify owners of received *WaitingForResults*


golem/task/taskserver.py
```diff
@@ -995,17 +992,14 @@ def __remove_old_tasks(self):
...
-                sessions_to_remove.append(subtask_id)
-        for subtask_id in sessions_to_remove:
-            if sessions[subtask_id].task_computer is not None:
-                sessions[subtask_id].task_computer.session_timeout()
-            sessions[subtask_id].dropped()
+            if dt < self.last_message_time_threshold:
+                continue
+            if session.task_computer is not None:
+                session.task_computer.session_timeout()
+            session.dropped()
```
maaktweluit, line 1002: Will dropping the session here change the list of `all_sessions`? Think it was 2 loops for this reason

maaktweluit (changes requested):
Looks good to me!

One small question, will approve on reply or fix :)

golem/task/taskserver.py
jiivan: Actually `all_sessions` is a `frozenset()` thus it can't change. Also it's a dynamically generated property which is a copy of incoming and outgoing sessions sets combined.

maaktweluit: nice! ok, approving :D

maaktweluit (approved):

PR #3899, opened 2019-02-15
Add an integration/regression test for the output overwritting issue

### LINKED ISSUE #3869: Add an integration/regression test for the output overwritting issue
* add an integration test (node integration test) that reproduces https://github.com/golemfactory/golem/issues/3785

Suggested test:
* the test should run one task normally, verifying if the output is created in an expected location (this should already be the case in the `regular_task_run` scenario)
* then, the test should create another task with exactly the same properties and this time, verify if the output has been created in an expected way (so as not to overwrite the original output)

To achieve this, a new playbook needs to be added to `playbooks/golem` and a test needs to be added to `tests/test_golem.py`.

Possibly it will require adding another task definition to `tasks` as the original issue only manifested itself (afaik) in multi-frame tasks (if unable to reproduce, please consult with @ederenn ).

rwrzesien:
Resolves https://github.com/golemfactory/golem/issues/3869

### DIFF
--- a/scripts/node_integration_tests/playbooks/golem/duplicated_files_names.py
+++ b/scripts/node_integration_tests/playbooks/golem/duplicated_files_names.py
@@ -1,0 +1,42 @@
+import os
+import typing
+
+from ..base import NodeTestPlaybook
+
+
+class DuplicatedFilesNames(NodeTestPlaybook):
+    provider_node_script = 'provider/debug'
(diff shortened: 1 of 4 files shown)


dybi (approved):
thumsup

shadeofblue:
superseded by https://github.com/golemfactory/golem/pull/4046

PR #3901, opened 2019-02-18
Merge video

bidzyyys:


apps/transcoding/ffmpeg/resources/ffmpeg-scripts/m3u8_utils.py
```diff
@@ -17,24 +19,27 @@ def create_and_dump_m3u8(path, segment):
...
         playlist = m3u8.load(pl)
         for segment in playlist.segments:
             base.add_segment(segment)
     return base
 
-def get_playlists(playlists_dir):
-    playlists = [os.path.join(playlists_dir,f) for f in os.listdir(playlists_dir) if f.endswith('_TC.m3u8')]
+
+def get_playlists(chunks):
+    playlists = [os.path.join('/golem/resources', f) for f in chunks if f.endswith('_TC.m3u8')]
```
weaselix: '/golem/resources' -> use `DirManager`

apps/transcoding/ffmpeg/utils.py
```diff
@@ -66,8 +83,49 @@ def split_video(self, input_stream: str, parts: int,
...
+            work_dir = os.path.join(resources_dir, 'merge', 'work')
+            os.makedirs(work_dir)
+            files = collect_files(resources_dir, chunks)
+            return resources_dir, output_dir, work_dir, list(
+                map(lambda chunk: chunk.replace(resources_dir,
+                                                '/golem/resources'),
+                    files))
+        except FileNotFoundError:
+            raise ffmpegException('Result file does not exist')
+        except OSError:
```
weaselix, line 79: We have no specific handler for OSError here. Just catch Exception (instead of OSError)

apps/transcoding/ffmpeg/utils.py
```diff
@@ -66,8 +83,49 @@ def split_video(self, input_stream: str, parts: int,
...
+            os.makedirs(output_dir)
+            work_dir = os.path.join(resources_dir, 'merge', 'work')
+            os.makedirs(work_dir)
+            files = collect_files(resources_dir, chunks)
+            return resources_dir, output_dir, work_dir, list(
+                map(lambda chunk: chunk.replace(resources_dir,
+                                                '/golem/resources'),
+                    files))
+        except FileNotFoundError:
+            raise ffmpegException('Result file does not exist')
```
weaselix: We don't know that file does not exist. Try to get that information, please.

apps/transcoding/ffmpeg/utils.py
```diff
@@ -66,8 +83,49 @@ def split_video(self, input_stream: str, parts: int,
                         .format(input_stream, streams_list))
             return streams_list
 
-    def _do_job_in_container(self, dir_manager: DirManager, task_id: str,
-                             extra_data: dict, env: Environment = None,
+    @staticmethod
+    def prepare_merge_job(task_dir, chunks):
+        try:
+            # should it be task dir?
+            resources_dir = task_dir
```
weaselix, line 74: It appears that it uses `<task_id>/tmp` directory. For my eye it is ok.

apps/transcoding/ffmpeg/utils.py
```diff
@@ -66,8 +83,49 @@ def split_video(self, input_stream: str, parts: int,
...
+            # each chunk must be in the same directory
+            output_dir = os.path.join(resources_dir, 'merge', 'output')
+            os.makedirs(output_dir)
+            work_dir = os.path.join(resources_dir, 'merge', 'work')
+            os.makedirs(work_dir)
+            files = collect_files(resources_dir, chunks)
+            return resources_dir, output_dir, work_dir, list(
+                map(lambda chunk: chunk.replace(resources_dir,
+                                                '/golem/resources'),
```
weaselix: - Don't use the constant 'golem/resources'. Use DirManager instead. 
- I suggest not to replace prefix by "string replace method". Construct that path by: `os.path.join(_get_golem_resources(), filename)`. For my eye it is more clear. However, it is my personal opinion.

apps/transcoding/ffmpeg/utils.py
```diff
@@ -25,6 +25,23 @@
...
     TRANSCODE = ('transcode', '')
+    MERGE = ('merge', '')
+
+
+def collect_files(dir, files):
+    results = []
+    os.makedirs(dir, exist_ok=True)
+    for file in files:
+        if not os.path.isfile(file):
+            raise FileNotFoundError
```
weaselix: We don't know what file is missing here. We want to know it.

apps/transcoding/ffmpeg/utils.py
```diff
@@ -25,6 +25,23 @@
...
+        if os.path.dirname(file) != dir:
+            new_path = file.replace(os.path.dirname(file), dir)
+            copy2(file, new_path)
```
weaselix: We just cannot copy files. Please note that chunk video files can be large.

apps/transcoding/ffmpeg/utils.py
```diff
@@ -25,6 +25,23 @@
 class Commands(enum.Enum):
     SPLIT = ('split', 'split-results.json')
     TRANSCODE = ('transcode', '')
+    MERGE = ('merge', '')
+
+
+def collect_files(dir, files):
```
weaselix: Why it is not private method in StreamOperator?

apps/transcoding/ffmpeg/utils.py
```diff
@@ -66,8 +83,49 @@ def split_video(self, input_stream: str, parts: int,
                         .format(input_stream, streams_list))
             return streams_list
 
-    def _do_job_in_container(self, dir_manager: DirManager, task_id: str,
-                             extra_data: dict, env: Environment = None,
+    @staticmethod
+    def prepare_merge_job(task_dir, chunks):
```
weaselix: private method?

apps/transcoding/ffmpeg/utils.py
```diff
@@ -66,8 +83,49 @@ def split_video(self, input_stream: str, parts: int,
...
+            raise ffmpegException("Failed to prepare merge job")
+
+    def merge_video(self, filename, task_dir, chunks):
+        resources_dir, output_dir, work_dir, chunks = \
+            self.prepare_merge_job(task_dir, chunks)
+
+        extra_data = {
+            'script_filepath': FFMPEG_BASE_SCRIPT,
+            'command': Commands.MERGE.value[0],
+            'output_stream': os.path.join('/golem/output', filename),
```
weaselix: '/golem/output' <- probably we can replace it with DirManager.

apps/transcoding/ffmpeg/utils.py
```diff
@@ -89,6 +147,14 @@ def _get_dir_mapping(self, dir_manager: DirManager, task_id: str):
...
-        return DockerTaskThread.specify_dir_mapping(
-            output=task_output_dir, temporary=tmp_task_dir,
-            resources=resources_task_dir, logs=tmp_task_dir, work=tmp_task_dir)
+        return self._specify_dir_mapping(output=task_output_dir,
+                                         temporary=tmp_task_dir,
+                                         resources=resources_task_dir,
+                                         logs=tmp_task_dir, work=tmp_task_dir)
+
+    @staticmethod
+    def _specify_dir_mapping(output, temporary, resources, logs, work):
```
weaselix, line 161: What is the purpose for that you wrapped `DockerTaskThread.specify_dir_mapping`?

tests/apps/ffmpeg/task/test_fmmpegtranscoding.py
```diff
@@ -30,21 +33,78 @@ def setUp(self):
...
+                                         tc_playlists)
+        assert os.path.isfile(os.path.join(playlist_dir, 'merge',
+                                           'output', output_name))
+
+    def test_merge_video_empty_dir(self):
+        with self.assertRaises(ffmpegException):
+            self.stream_operator.merge_video('output.mp4', self.tempdir, [])
+
+    def test_collect_nonexistent_results(self):
+        with self.assertRaises(OSError):
```
weaselix: Perhaps should we wrap catch exception and re-raise it as (for example) `TranscondingException`?

apps/transcoding/task.py
```diff
@@ -97,12 +102,24 @@ def accept_results(self, subtask_id, result_files):
...
+
+    def _collect_results(self, results):
+        self.collected_files.extend(results)
+
+    def _merge_video(self):
+        stream_operator = StreamOperator()
+        path = stream_operator.merge_video(os.path.basename(self.task_definition.output_file),
+                                           self.task_dir, self.collected_files)
+        os.makedirs(os.path.dirname(self.task_definition.output_file), exist_ok=True)
+        copy2(path, self.task_definition.output_file)
```
weaselix: We cannot copy it. Can we mount specific directories to the container instead?

weaselix (changes requested):
- please add some logs (on info and debug level) about merge process
- lines are longer than 80 characters (please make it shorter)
- please add test for the case where one of the chunks is invalid.

apps/transcoding/ffmpeg/utils.py
bidzyyys: It is suggested by PyCharm. Have you thought about benefits from rewriting it to private method? IMO Python is not strictly object-oriented language.

apps/transcoding/ffmpeg/utils.py
bidzyyys: `/golem/output` is a directory inside docker container, docker does not know anything about the realpath

apps/transcoding/ffmpeg/utils.py
bidzyyys: We should have all video chunks in a single directory that is mounted as `/golem/resources/`. If you are sure that they are collected in the same place, I'll change it.

apps/transcoding/ffmpeg/utils.py
bidzyyys: Here I prepare list of video chunks for docker container - to make them visible the real directory should be replaced by `/golem/resources/`

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/m3u8_utils.py
bidzyyys: `DirManger` does not know anything about `/golem/resources/` (it is the path inside docker container with video chunks).

apps/transcoding/ffmpeg/utils.py
bidzyyys: It is raised by `collect_files` method.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/m3u8_utils.py
bidzyyys: I will define constant docker's paths (output, resources and work).

apps/transcoding/ffmpeg/utils.py
bidzyyys: I can change it, but `OSError` can be raised by `os.makedis`, no other exception is expected here.

apps/transcoding/ffmpeg/utils.py
weaselix: > Have you thought about benefits from rewriting it to private method?

Yes. In that way you hide implementation details. Your function `collect_files` is a specific  function and should not be available in the interface. You put it in an interface. What is the purpose?

> IMO Python is not strictly object-oriented language

Yes, it is not. And what about it in our specific case?

apps/transcoding/ffmpeg/utils.py
weaselix: I am not sure, but I think you can check it.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/m3u8_utils.py
weaselix: You can use `DockerJob.RESOURCES_DIR` (instead of `DirManager`- I was wrong here)

apps/transcoding/ffmpeg/utils.py
weaselix: Yes, but we would like to know what file is missing (like 'Not found /tmp/test_file.mp4')

apps/transcoding/ffmpeg/utils.py
weaselix: You can use `DockerJob.RESOURCES_DIR` (instead of `DirManager`- I was wrong here)

apps/transcoding/ffmpeg/utils.py
weaselix: You can use `DockerJob.RESOURCES_DIR` (instead of `DirManager`- I was wrong here)

apps/transcoding/task.py
```diff
@@ -97,12 +102,24 @@ def accept_results(self, subtask_id, result_files):
...
             super(TranscodingTask, self).accept_results(subtask_id,
                                                         result_files)
-            # TODO remove these logs
-            for idx, result in enumerate(result_files):
-                logger.info("Result[{}]: {}".format(idx, result))
+            self._collect_results(result_files)
 
             self.num_tasks_received += 1
 
+        if self.num_tasks_received == self.total_tasks:
```
weaselix: You need the acquired lock here.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/m3u8_utils.py
bidzyyys: I haven't seen this :+1:

apps/transcoding/task.py
bidzyyys: We mount one directory as `/golem/resources/`, so I will check if each file has proper path (and raise `ffmpegException`)

apps/transcoding/ffmpeg/utils.py
bidzyyys: To enable specifying custom directories, see `merge_video`

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/m3u8_utils.py
bidzyyys: Btw this is executed inside docker

apps/transcoding/ffmpeg/utils.py
weaselix: Yes, but I can not understand why do you create wrapping method instead using `DockerTaskThread.specify_dir_mapping`. I suggest to remove that method (it is redundant).

apps/transcoding/task.py
weaselix: If I understand correctly `path` points to the result file (a merged video). Can we move it to the destination instead of copying?

weaselix (changes requested):
* Some lines are still longer than 80 characters

PR #3911, opened 2019-02-20
Update install.sh

mdtanrikulu:


shadeofblue (approved):

Installer/Installer_Linux/install.sh
```diff
@@ -339,7 +339,7 @@ function install_golem()
         error_msg "ERROR) Cannot extract ${ELECTRON_PACKAGE}. Exiting..."
         return 1
     fi
-    ELECTRON_DIR=$(find . -maxdepth 1 -name "linux-unpacked" -type d -print | head -n1)
+    ELECTRON_DIR=$(find . -maxdepth 1 -name "golem-electron-beta-linux" -type d -print | head -n1)
```
mbenke: Maybe
```suggestion
    ELECTRON_DIR=$(find . -maxdepth 1 -name "golem-electron-beta-linux*" -type d -print | head -n1)
```

this might fail though if there is more than one directory matching

PR #3914, opened 2019-02-20
Differentiate CannotAssignTask responses 

### LINKED ISSUE #3902: Differentiate the `CannotAssignTask` responses depending on the task status
Depends on:
* [x] https://github.com/golemfactory/golem-messages/issues/323

As a requestor, differentiate the `CannotAssignTask` message:
* when task is temporarily out of subtasks, send CAT with `NoMoreSubtasks` reason to tell the provider node that the requestor doesn't have subtasks for this task at this moment but tasks can still reappear on the market if any of the currently assigned nodes fail so they can still ask about this task in near future,
* when  all subtasks had their results completed, delivered and verified, the task is definitely finished and there won't ever be any more subtasks on the market that pertain to this task - the requestor should then send a CAT with `TaskFinished` reason sent to let the providers know they need to stop asking about this task.

jiivan:
fixes: #3902


shadeofblue (approved):
lgtm

PR #3917, opened 2019-02-21
Remove resource size from task header

badb:
If we want to make possible to add different resources for different subtasks we cannot send just one size in the TaskHeader. This PR:
* Removes resource_size from TaskHeader
* Removes providers check of resource size at the task selection state
* Provider may refuse to compute tasks with too big resources after getting TaskToCompute with specific resource size. 
* Taskkeeper keeping time is updated after TaskToCompute with proper resource size is sent. 

Golem-messages PR: https://github.com/golemfactory/golem-messages/pull/319


golem/task/server/resources.py
```diff
@@ -125,10 +124,9 @@ def _filter_options(_options):
             options = received_options
 
         options = _filter_options(options)
-        task_header = task_keeper.task_headers.get(task_id)
 
-        if task_header:
-            options.set(size=task_header.resource_size)
+        if size and options:
+            options.set(size=size)
```
shadeofblue, line 129: @badb there seems to be something not quite right here... unless I'm misreading...

what I mean is that previously, the size here had always been set to the resource size specified in the task header... 

right now, the only call to this method that specifies the `size`, uses the `ReportComputedTask.size` which is not the resource size but rather the _result_ size ... is that indeed what we need here? am I wrong?

badb: Size option can be also already specified in options (that should be the case for TaskToCompute (check line 233 in taskkeeper). 
This option is used to estimate how long may it take to download resources. With ReportComputedTask we want to dowload results so their size should be used. I guess that usage of resource_size in this place was just a heuristic (we assumed that results size won't be significantly different  that size of resources, which of course doesn't have to be true).

shadeofblue: :+1:

shadeofblue (approved):
girl a okay stockphoto

PR #3918, opened 2019-02-21
Many improvements spread all over

jiivan:


golem/appconfig.py
```diff
@@ -83,6 +83,11 @@
 # Updating by 1 bit increases number of workers 2x
 MASK_UPDATE_NUM_BITS = 1
 
+DISALLOW_NODE_TIMEOUT_SECONDS = 5*60
```
kmazurek: The logic for temporarily blocking providers should be made optional by default. This means the disallow timeouts here should be set to some special value (e.g. 0) and treated appropriately.

golem/clientconfigdescriptor.py
```diff
@@ -69,6 +69,12 @@ def __init__(self):
         self.mask_update_interval = 0
         self.mask_update_num_bits = 0
 
+        self.disallow_node_timeout_seconds = 5*60
```
kmazurek: Same as above, the default timeouts here should be set to a special value (disabled by default).

golem/task/acl.py
```diff
@@ -21,55 +28,95 @@ def disallow(self, node_id: str, timeout_seconds: int, persist: bool) \
...
-    def _deadline(timeout: int = DEFAULT_TIMEOUT):
+    def _deadline(timeout: int) -> float:
         return time.time() + timeout
 
 
 class _AllowAcl(Acl):
-    def __init__(self, allow_set: Set[str], list_path: Path) -> None:
+    def __init__(self, allow_set: Optional[Set[str]] = None,
+                 list_path: Optional[Path] = None,
+                 max_times: int = 1) -> None:
```
kmazurek: Since this is not implemented, I think the `max_times` parameter here should be removed. Also, I don't really see a good use case for having this feature on allow ACL anyway.

golem/task/taskserver.py
```diff
@@ -417,6 +424,7 @@ def verify_header_sig(cls, header: dt_tasks.TaskHeader):
             return False
         return True
 
+    @rpc_utils.expose('comp.tasks.known.delete')
```
kmazurek, line 427: Should this one be exposed?

golem/task/taskserver.py
```diff
@@ -714,46 +736,107 @@ def should_accept_provider(  # noqa pylint: disable=too-many-arguments,too-many-
...
+            reason=reason.value,
+            details=details,
+        )
+
     def should_accept_requestor(self, node_id):
         allowed, reason = self.acl.is_allowed(node_id)
         if not allowed:
-            short_id = short_node_id(node_id)
-            logger.info('requestor %s. node=%s', reason, short_id)
+            short_id = node_id  # for AR
```
kmazurek: This should stay in short form.

golem/task/tasksession.py
```diff
@@ -228,6 +228,13 @@ def verification_finished():
             if not task_to_compute.sig:
                 task_to_compute.sign_message(self.my_private_key)
 
+            config_desc = self.task_server.config_desc
```
kmazurek, line 231: Please make these temporary disallows optional (add conditions to check against the config's disabled value).

golem/task/taskserver.py
jiivan: Exposure is moved from client https://github.com/golemfactory/golem/pull/3918/files#diff-9f351727974c5655127c6a4dc0671995L1157

kmazurek: Right, missed that one.

golem/appconfig.py
```diff
@@ -83,6 +83,12 @@
 # Updating by 1 bit increases number of workers 2x
 MASK_UPDATE_NUM_BITS = 1
 
+# Experimental temporary banning options
+DISALLOW_NODE_TIMEOUT_SECONDS = None
+DISALLOW_IP_TIMEOUT_SECONDS = None
+ALLOW_ID_MAX_TIMES = 1
```
kmazurek: Actually, the names of these `ALLOW` options are counter-intuitive, maybe let's change the prefix to `DISALLOW`? I think that would better describe their function, since they control the number of times `disallow` needs to be called for a given provider to result in a ban.

kmazurek (approved):
tldr

Issue #3890, opened 2019-02-14
Identify unit tests that leave temporary data behind and add clean up


shadeofblue:
The problem is that right now, each test run leaves cruft after execution and each run adds to the cruft deposited in the system, causing the buildbot instances to get clogged after just a few days.

We need to identify those tests that leave garbage behind and add adequate cleanup methods to remove the cruft after each run of the unit test suite.

## Garbage list:
- docker
  - docker images
  - docker volumes
- /tmp
  - To Be Checked...
- ~/.local/share/golem
  - To Be Checked...

rwrzesien:
@shadeofblue Could you give me a hint what exactly is left over? Or in which directories those things are?

Krigpl:
Isn't this the issue basically? https://github.com/golemfactory/golem/blob/develop/golem/testutils.py#L43-L47

rwrzesien:
@Krigpl This is just one of many. My question is not valid anymore because I was able to identify most of the cases where something is left out, and I am currently working on fixing it,

shadeofblue:
@maaktweluit is this still a thing? @Krigpl went into some discussion with @rwrzesien in the pull request and it was never merged ... but I have a feeling that either you've done some reconfiguration to buildbot or you're actively maintaing the buildbot so that it's not an issue anymore ...  ?

maaktweluit:
@shadeofblue yes this is still a thing. 
there are some cron-jobs for cleanup but it still fills up every now and then

This issue can be used to list the files that are left by the tests and create new issues per case, for each case we can see if its the production code or the test code that should clean up ( as per discussion in the closed PR above )

kmazurek:
A possible solution here is to use the [`tmpdir` fixture](https://docs.pytest.org/en/latest/tmpdir.html) in all tests which require creating directories.

PR #3920, opened 2019-02-21
Identify unit tests that leave temporary data behind and add clean up

rwrzesien:
Resolves https://github.com/golemfactory/golem/issues/3890.

### DIFF
--- a/tests/apps/blender/benchmark/test_blenderbenchmark.py
+++ b/tests/apps/blender/benchmark/test_blenderbenchmark.py
@@ -26,6 +26,11 @@ class TestBlenderBenchmark(unittest.TestCase, testutils.PEP8MixIn):
 
     def setUp(self):
         self.bb = BlenderBenchmark()
+        self.addCleanup(self.__clean_files)
+
+    def __clean_files(self):
+        if os.path.isfile(self.bb.task_definition.output_file):
+            os.remove(self.bb.task_definition.output_file)
 
     def test_is_instance(self):
         self.assertIsInstance(self.bb, BlenderBenchmark)
--- a/tests/apps/blender/task/test_blenderrendertask.py
+++ b/tests/apps/blender/task/test_blenderrendertask.py
@@ -634,6 +634,15 @@ class TestPreviewUpdater(TempDirFixture, LogTestCase):
 
 class TestBlenderRenderTaskBuilder(TempDirFixture):
 
+    def setUp(self):
+        super().setUp()
+        _, self.blender_file_path = tempfile.mkstemp('.blend')
+        self.addCleanup(self.__clean_files)
+
+    def __clean_files(self):
+        if os.path.isfile(self.blender_file_path):
+            os.remove(self.blender_file_path)
+
     @property
     def _task_dictionary(self):
         return {
@@ -643,7 +652,7 @@ class TestBlenderRenderTaskBuilder(TempDirFixture):
             "subtask_timeout": "0:09:50",
             "subtasks_count": 1,
             "bid": 1.0,
-            "resources": [tempfile.mkstemp('.blend')[1]],
+            "resources": [self.blender_file_path],
             "options": {
                 "output_path": '',
                 "format": "PNG",
--- a/tests/apps/dummy/benchmark/test_dummybenchmark.py
+++ b/tests/apps/dummy/benchmark/test_dummybenchmark.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 from unittest.mock import patch
 
 from apps.dummy.benchmark.benchmark import DummyTaskBenchmark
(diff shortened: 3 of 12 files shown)


tests/apps/blender/task/test_blenderrendertask.py
```diff
@@ -634,6 +634,15 @@ def test_error_in_preview_update(self):
 
 class TestBlenderRenderTaskBuilder(TempDirFixture):
 
+    def setUp(self):
+        super().setUp()
+        _, self.blender_file_path = tempfile.mkstemp('.blend')
```
Krigpl, line 639: that could just use `self.tmp_dir` and get cleaned automatically

tests/apps/blender/benchmark/test_blenderbenchmark.py
```diff
@@ -26,6 +26,11 @@ class TestBlenderBenchmark(unittest.TestCase, testutils.PEP8MixIn):
 
     def setUp(self):
         self.bb = BlenderBenchmark()
+        self.addCleanup(self.__clean_files)
+
+    def __clean_files(self):
+        if os.path.isfile(self.bb.task_definition.output_file):
+            os.remove(self.bb.task_definition.output_file)
```
Krigpl, line 33: that doesn't feel like the right solution here - in production code we still leave this temp file.  imo benchmark classes should clean up after themselves

tests/apps/dummy/benchmark/test_dummybenchmark.py
```diff
@@ -11,6 +12,11 @@ class TestDummyBenchmark(TempDirFixture):
     def setUp(self):
         super().setUp()
         self.db = DummyTaskBenchmark()
+        self.addCleanup(self.__clean_files)
+
+    def __clean_files(self):
+        if os.path.isdir(self.db._task_definition.tmp_dir):
+            shutil.rmtree(self.db._task_definition.tmp_dir)
```
Krigpl, line 19: same as above

tests/golem/docker/test_docker_job.py
```diff
@@ -37,11 +37,11 @@ def _get_test_tag(self):
     TEST_SCRIPT = "print 'Adventure Time!'\n"
 
     def setUp(self):
-        main_dir = get_local_datadir('tests-' + str(uuid.uuid4()))
-        if not os.path.exists(main_dir):
-            os.makedirs(main_dir)
+        self.main_dir = get_local_datadir('tests-' + str(uuid.uuid4()))
```
Krigpl, line 40: Why not just use `self.tmp_dir` here and below ?

Krigpl (commented):
That looks like the majority of the leftover files comes from the production code that doesn't clean them up rather than anything specific to tests - so it'd be better to look into how to improve production code.

rwrzesien:
@Krigpl 
> That looks like the majority of the leftover files comes from the production code that doesn't clean them up rather than anything specific to tests - so it'd be better to look into how to improve production code.

Probably yes, but a separate issue should be created for that. This one is specifically about cleaning after what unit tests run leaves.

tests/apps/dummy/benchmark/test_dummybenchmark.py
rwrzesien: See https://github.com/golemfactory/golem/pull/3920#issuecomment-466386794.

tests/apps/blender/benchmark/test_blenderbenchmark.py
rwrzesien: See https://github.com/golemfactory/golem/pull/3920#issuecomment-466386794.

tests/apps/blender/task/test_blenderrendertask.py
rwrzesien: I don't want to modify any existing tests that I am not aware what they are doing and what they suppose to really test.

tests/golem/docker/test_docker_job.py
rwrzesien: I don't want to modify any existing tests that I am not aware what they are doing and what they suppose to really test.

Krigpl (changes requested):
These tests expose an issue with production code and I'm not fine with just applying some bandaid solution to improve tests while sweeping the real problem under the rug.
Please modify `test_docker_job` and `test_blenderrendertask` as suggested, remove all the benchmark related hacks and create an issue to fix leftover benchmarks files.

Krigpl:
Any progress on that?

mplebanski:
Can we close this?

lukasz-tarka:
If the solution for the problem explicitly described as test related in the issue is not acceptable, it should be closed and a new issue opened for fixing production code instead of tests itself.

Krigpl:
rip [*]

PR #3923, opened 2019-02-22
Fix GNT conversion when there's GNT top up in the meantime

Krigpl:
Previously when one topped up the account twice in quick succession the second amount of GNT would not convert.
That is fixed now.

### DIFF
--- a/golem/ethereum/transactionsystem.py
+++ b/golem/ethereum/transactionsystem.py
@@ -15,6 +15,7 @@ from typing import (
     Iterable,
     List,
     Optional,
+    Tuple,
 )
 
 from ethereum.utils import denoms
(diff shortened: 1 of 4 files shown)


golem/ethereum/transactionsystem.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, datadir: Path, config) -> None:
         self._payment_processor: Optional[PaymentProcessor] = None
 
         self._gnt_faucet_requested = False
-        self._gnt_conversion_status = ConversionStatus.NONE
+        self._gnt_conversion_status = (ConversionStatus.NONE, None)
```
shadeofblue: @Krigpl it would be good to either add a type hint for the second value or maybe just add a comment what it is when not `None` ...

shadeofblue (approved):
incredible_contribution

golem/ethereum/transactionsystem.py
Krigpl: image

Issue #3874, opened 2019-02-11
Reuse the keystore files in the `node_integration_tests`


shadeofblue:
Right now, the `node_integration_tests` start golem instances in empty data directories. This is done to eliminate any artifacts from previous runs and tasks that could influence execution of subsequent tests and thus, cause false negatives and generally make it harder to track down issues.

The downside is that by starting afresh, each test has to wait for the nodes' wallets to get seeded with test ETH and test GNT from faucet and for GNT to be converted to GNTB, plus, for GNT deposit to get created when starting each task that requires the Concent. Not only it causes the integration tests to be terribly slow (each run of a single test takes at least ~15 minutes of such waiting) but it also causes the ETH faucet to be drained unacceptably fast.

Thus, we'd like to add some means (possibly using `setUpClass`/`tearDownClass`) to reuse the ethereum keys between the test runs - although by strictly reusing the `rinkeby/keys/keystore.json` file and no other files from the datadirs.

Additionally, at least one test in a run should be executed with completely fresh directories (we need to ensure we're including the retrieving test ETH/GNT, GNTB conversions and creating the deposits into the integration testing). So, the first test to run should always use a completely fresh datadir and only the subsequent tests should reuse the keystores.

Apart from that - if a specific PROVIDER/REQUESTOR datadirs to use are provided using environment variables directly, the reuse should not be performed.

There should also be a way to disable key reuse for certain tests - e.g. Concent UC5 test (using a decorator?).

PR #3930, opened 2019-02-26
Feature reuse keys and run faster node integration tests

pawelkisielewicz:
Resolves https://github.com/golemfactory/golem/issues/3874


scripts/node_integration_tests/README.md
```diff
@@ -171,6 +171,29 @@ environment variable, e.g.:
 GOLEM_INTEGRATION_TEST_DIR=/some/location pytest scripts/node_integration_tests
 ```
 
+There is possibility to reuse provider's and requestor's node keys. It is done 
+to make this tests faster. Only first test creates new ETH accounts, all other 
+use the same account, so there is no need to wait for ETH, tGNT and GNTB any 
+more. If you want to change this behaviour set --reuse_keys.
+
+To enable reusing keys fo all tests add --reuse_keys=no
```
shadeofblue: and not the other way round? ...

scripts/node_integration_tests/run_test.py
```diff
@@ -89,5 +108,6 @@
         ]
            and v
     }
+    _add_to_kwargs_reuse_node_keys_parsed_from_string_to_boolean()
```
shadeofblue: @pawelkisielewicz modifying an external value inside a function doesn't sound like a good engineering practice, does it? ...

scripts/node_integration_tests/README.md
pawelkisielewicz: I just fixed readme, please check it now. There is also decorator setting default behavior, if there is no flag, tests run with default, decorator setting. I set it to True, so by default it will be True

scripts/node_integration_tests/tests/base.py
```diff
@@ -1,22 +1,110 @@
...
+    def decorator(test_function: Callable)-> Callable:
+        @wraps(test_function)
+        def wrap(*args, **kwargs) -> None:
+            kwargs.update({'reuse_keys_default': default_option})
+            test_function(*args, **kwargs)
+        return wrap
+    return decorator
+
+
+class ReuseNodeKeys(object):
```
shadeofblue: `(object)` is not needed in python 3...

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -475,6 +476,25 @@ def call_provider(self, method, *args,
             **kwargs,
         )
 
+    @staticmethod
+    def _replace_keystore_and_wallet(destination: str) -> None:
+        src = '/tmp/provider_reuse_keystore/keystore.json' if 'provider' in \
+            destination else '/tmp/requestor_reuse_keystore/keystore.json'
+        dst = destination + '/rinkeby/keys/keystore.json'
+        os.mkdir(destination+'/rinkeby')
```
shadeofblue: @pawelkisielewicz `os.makedirs`?

pawelkisielewicz: ok, done

scripts/node_integration_tests/tests/base.py
pawelkisielewicz: :+1:

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -475,6 +476,25 @@ def call_provider(self, method, *args,
             **kwargs,
         )
 
+    @staticmethod
+    def _replace_keystore_and_wallet(destination: str) -> None:
+        src = '/tmp/provider_reuse_keystore/keystore.json' if 'provider' in \
```
Krigpl: These filepath operations are not gonna work on Windows, are they? Use `pathlib` module instead. Also use `tempfile` for creating temporary files

scripts/node_integration_tests/run_test.py
```diff
@@ -47,6 +60,12 @@
         help="use the mainnet environment to run the test "
              "(the playbook must also use mainnet)",
     )
+    parser.add_argument(
+        '--reuse_node_keys',
```
mfranciszkiewicz: Please replace underscores with hyphens, that way it will follow the naming convention of other arguments.

Why do we need a new argument name here?

scripts/node_integration_tests/playbooks/base.py
pawelkisielewicz: fixed :+1:

scripts/node_integration_tests/playbooks/base.py
pawelkisielewicz: done, but I'd prefer to leave two reusable 'keystore.json' in folders as it is now and replace this folders each time. Less mess, it will be always replaced

Krigpl: fwiw I find it noisy/confusing to see comments like "done"/"fixed" without corresponding changes in the published code, the standalone comment like that has little value until there's code to look at

scripts/node_integration_tests/run_test.py
pawelkisielewicz: renamed. It is to disable reusing node keys. With this argument all tests will start with new node keys.

scripts/node_integration_tests/run_test.py
pawelkisielewicz: It was modifying only test_kwargs, mkaybe the name of function was misleading. But now it is removed anyway

scripts/node_integration_tests/playbooks/base.py
pawelkisielewicz: sorry, I made changes with list and wrote here, but didn't push because of tests. Sorry one more time, it is pushed now

scripts/node_integration_tests/tests/base.py
```diff
@@ -1,22 +1,103 @@
...
+            cls.instance.first_test_in_set = False
+        return ReuseNodeKeys.instance
+
+
 class NodeTestBase:
     def setUp(self):
         test_dir = pathlib.Path(get_testdir()) / self._relative_id()
-        self.provider_datadir = test_dir / 'provider'
-        self.requestor_datadir = test_dir / 'requestor'
+        self.provider_datadir = test_dir/'provider'
```
shadeofblue: @pawelkisielewicz why are we removing the separators here?

scripts/node_integration_tests/tests/base.py
```diff
@@ -30,4 +111,21 @@ def _run_test(self, playbook_class_path: str, *args, **kwargs):
...
             test_args.append(v)
 
+        if self._should_node_keys_be_reused():
+            self._replace_files_to_used_before()
+
         return subprocess.call(args=test_args)
+
+    @staticmethod
+    def _replace_keystore(destination: pathlib.PosixPath) -> None:
+        src = pathlib.Path(get_testdir()).parent/'provider_reuse_keystore/' \
```
shadeofblue: and same here -> please add spaces for readability here

scripts/node_integration_tests/tests/test_concent.py
```diff
@@ -1,6 +1,5 @@
 import unittest
-
```
shadeofblue: and pls, don't remove the separating line here

scripts/node_integration_tests/tests/base.py
```diff
@@ -1,22 +1,103 @@
 import os
 import pathlib
+import shutil
 import subprocess
+import time
+from functools import wraps
+from shutil import copyfile
```
shadeofblue: if we already have `shutil` imported, wouldn't it be better to just call `shutil.copyfile` rather than import `copyfile` again? especially that for someone reading the code, it may not be perfectly obvious that `copyfile` comes from `shutil` ... and actually, same with `functools.wraps`... although, maybe indeed in the latter case it's a bit more obvious...

conftest.py
```diff
@@ -2,14 +2,26 @@
 import pytest
 import _pytest
 
+DISABLE_KEY_REUSE_COMMAND_LINE = False
```
shadeofblue: @pawelkisielewicz please move the key reuse code to a conftest inside `scripts` or actually `scripts/node_integration_tests` so that this setting doesn't interfere with regular unittest runs

scripts/node_integration_tests/run_test.py
shadeofblue: @pawelkisielewicz as mentioned above, please move this code to a location more specific to these tests here

scripts/node_integration_tests/tests/base.py
pawelkisielewicz: done

scripts/node_integration_tests/tests/base.py
pawelkisielewicz: done

scripts/node_integration_tests/tests/test_concent.py
pawelkisielewicz: ok, sorry for this

scripts/node_integration_tests/tests/base.py
pawelkisielewicz: right, done

conftest.py
pawelkisielewicz: done, moved to `node_integration_tests/tests/base.py`

conftest.py
```diff
@@ -2,7 +2,8 @@
 import pytest
 import _pytest
 
-DISABLE_KEY_REUSE_COMMAND_LINE = False
+from scripts.node_integration_tests.tests.base import \
+    disable_reuse_keys_command_line
```
shadeofblue: @pawelkisielewicz the point was not to pollute the global `conftest.py` with something that's used only in the `node_integration_tests`...

conftest.py
shadeofblue: @pawelkisielewicz not quite what I meant ... :/

pawelkisielewicz: sorry, fixed

conftest.py
pawelkisielewicz: fixed

scripts/node_integration_tests/README.md
```diff
@@ -181,3 +183,26 @@ pytest scripts/node_integration_tests/tests/test_golem.py::GolemNodeTest::test_r
 Suggestion: when you _don't_ provide the `GOLEM_INTEGRATION_TEST_DIR` variable
 to pytest, run `pytest -s -v [...]` so that you can see the paths generated
 automatically during the test run.
+
+#### Node key reuse
+
+Normally, each test starts with empty provider and requestor datadirs. That
+also means that the nodes need to initialize they keystores, request GNT and
```
jiivan:
```suggestion
also means that the nodes need to initialize their keystores, request GNT and
```

jiivan (approved):

scripts/node_integration_tests/README.md
```diff
@@ -181,3 +183,26 @@ pytest scripts/node_integration_tests/tests/test_golem.py::GolemNodeTest::test_r
...
+also means that the nodes need to initialize their keystores, request GNT and
+ETH from the faucets and finally convert GNT into GNTB which adds several
+minutes to each run.
+
+To optimize that, we're now only initializing the keystores on the first test
+run by default and all subsequest tests reuse the same node key pairs.
+Thus, nodes don't need to wait for ETH, GNT and GNTB anymore.
+
+In some tests, we need to ensure there are no side effects on the blockchain.
+For that, we can disable key reuse for this particular test by adding the
```
shadeofblue:
```suggestion
To achieve that, we can disable key reuse for this particular test by adding the
```

scripts/node_integration_tests/run_test.py
pawelkisielewicz: After all fixes this argument is not passed to tests, so this code is not here any more

PR #3932, opened 2019-02-27
Replace script_filepath with a custom command

Krigpl:
Makes DockerJob non-Python specific


maaktweluit (approved):
LGTM!

Lets see if the slow tests think the same ;)

mfranciszkiewicz (approved):

PR #3933, opened 2019-02-27
Don't try to build Docker images

Krigpl:
Dockerfiles and their context directories will be moved outside Golem repo so attempting to build them won't make sense anymore

### DIFF
--- a/apps/images.ini
+++ b/apps/images.ini
@@ -1,6 +1,6 @@
-golemfactory/base core/resources/images/base.Dockerfile 1.4 .
-golemfactory/nvgpu core/resources/images/nvgpu.Dockerfile 1.2 . apps.core.nvgpu.is_supported
-golemfactory/blender blender/resources/images/blender.Dockerfile 1.8 blender/resources/images/
-golemfactory/blender_verifier blender/resources/images/blender_verifier.Dockerfile 1.0 blender/resources/images/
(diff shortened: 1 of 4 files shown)


maaktweluit (approved):
LGTM!

Nice cleanup!

It hurts a bit that the .ini file is using space separated data instead of properties, but that is out of scope to fix here.

mfranciszkiewicz:
LGTM, though I wouldn't merge this yet, unless no one uses Golem to build their Docker images.

Krigpl:
@mfranciszkiewicz And what if they do? This kind of thing won't be supported so what's the reason for delaying?

mfranciszkiewicz:
@Krigpl The reason is that not every image can be uploaded to Docker Hub. E.g. https://github.com/golemfactory/golem/compare/amdgpu?expand=1

Krigpl:
@mfranciszkiewicz Why can't it be uploaded?

mfranciszkiewicz:
Locally installed driver libraries and configuration files need to be added to the image and they must always match what's installed in the host os.

Krigpl:
That sounds like something that should be configured in the runtime rather than built into the Docker image, though. Otherwise images tagged under the same name will be essentially completely different on different hosts which contradicts the idea of consistent Docker images, doesn't it?

shadeofblue:
@Krigpl how did it go? a single repo for everything ? ... and now what?

Krigpl:
@blue Apps were never meant to live inside the same repo

PR #3938, opened 2019-02-28
Fix docker scripts [b0.19]

Wiezzel:
### Fixed missing backslash in `prepare-docker-for-windows.ps1`

Added missing backslash to `$golemDataDir` path (without the script only worked if `$appDataDir` was provided with a trailing backslash)

 ---

### Fixed missing quotes in `create-share.ps1`

Added missing quotation marks around paths. Without them the script would fail if the username contained whitespace.

---

### Increased timeout for all powershell commands to ~10~ 60 seconds

### DIFF
--- a/golem/core/windows.py
+++ b/golem/core/windows.py
@@ -1,7 +1,7 @@
 import subprocess
 from typing import List, Optional
 
-DEFAULT_SCRIPT_TIMEOUT = 5  # seconds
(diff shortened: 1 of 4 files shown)


Wiezzel:
There is a separate PR for develop (which also includes some additional changes): #3939

mfranciszkiewicz (approved):

maaktweluit (approved):

PR #3939, opened 2019-02-28
Fix docker scripts

Wiezzel:
### Revert 3870850

Changed PowerShell.exe invocation back to simple `&` because it was not needed after all and it made the terminal close after running the script, which was quite inconvenient.

---

### Fixed missing backslash in `prepare-docker-for-windows.ps1`

Added missing backslash to `$golemDataDir` path (without the script only worked if `$appDataDir` was provided with a trailing backslash)

---

###  Fixed missing quotes in `create-share.ps1`

Added missing quotation marks around paths. Without them the script would fail if the username contained whitespace.

### DIFF
--- a/scripts/docker/create-share.ps1
+++ b/scripts/docker/create-share.ps1
@@ -56,11 +56,11 @@ if (Get-SmbShare | Where-Object -Property Name -EQ $SmbShareName) {
 
 "Sharing directory..."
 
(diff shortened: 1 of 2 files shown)


mfranciszkiewicz (approved):

maaktweluit (approved):
LGTM

curious how the golem-data-dir worked without slash..

As long as we tests both installer and manually running the script it should be fine.

PR #3942, opened 2019-03-01
Use res_id instead task_id in ResourceManager

badb:
This one will be needed if we want to introduce resources per subtask and not just per task.
res_id is used instead of task_id in Resource, ResourceStorage, ResourceManager and ResourceServer.
That way function names and arguments are more universal and we will be able to use them for subtasks resources, additional resources etc.


shadeofblue (approved):
looks good as far as I could tell :)

PR #3943, opened 2019-03-04
Integration tests of transcoding

bidzyyys:


golem/testutils.py
```diff
@@ -167,3 +240,152 @@ def test_conformance(self):
         result = style.check_files(absolute_files)
         self.assertEqual(result.total_errors, 0,
                          "Found code style errors (and warnings).")
+
+
+class TestTaskIntegration(TempDirFixture):
+    TEST_FAILED = False
```
nieznanysprawiciel, line 246: This is static variable. You probably should place this variable in setUp function. Otherwise all tests using TestTaskIntegration will share this variable.

nieznanysprawiciel (approved):

PR #3947, opened 2019-03-04
Fix `comp.task.state` RPC outputs for blender task

### LINKED ISSUE #3604: TaskState for BlenderRenderTask does not define 'outputs'
RPC call to `comp.task.state` returns `state` object with `outputs` string array containing information about where outputs are available for given `task_id`. This is taken from `get_output_names` method for each task.

mplebanski:
Fixes #3604


badb (changes requested):
This will not work for mulitfram tasks

Issue #3937, opened 2019-02-28
Skip virtualization check when Docker is supported natively


kmazurek:
Currently, when checking whether Docker can be installed on a given system, a hardware virtualisation check (RPC call `env.hw.virtualization`) is always performed. This check should be ignored for Linux systems, where Docker does not require virtualisation.

PR #3949, opened 2019-03-04
Disabled virtualization check on Linux platforms

kmazurek:
Resolves: #3937

This change skips the virtualization check on Linux systems, always returning positive support status.

### DIFF
--- a/golem/core/virtualization.py
+++ b/golem/core/virtualization.py
@@ -15,14 +15,15 @@ WIN_SCRIPT_PATH = os.path.join(
 
 
 @rpc_utils.expose('env.hw.virtualization')
-def is_virtualization_enabled() -> bool:
+def is_virtualization_satisfied() -> bool:
     """ Checks if hardware virtualization is available on this machine.
-    Currently, this check is limited to Intel CPUs (VT and VT-x support).
-    :return bool: True if virtualization is available or the OS is Linux.
+    If hardware virtualization is not required to run Golem (e.g. on Linux),
+    the actual check is skipped. Currently, this function is limited to Intel
+    CPUs (VT and VT-x support).
+    :return bool: True if virtualization is available or is not required.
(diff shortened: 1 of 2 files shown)


golem/core/virtualization.py
```diff
@@ -18,10 +18,13 @@
...
     Currently, this check is limited to Intel CPUs (VT and VT-x support).
-    :return bool: True if virtualization is available. On Windows, we also check
-    if the feature is enabled in firmware.
+    :return bool: True if virtualization is available or the OS is Linux.
+    On Windows, we additionally check if the feature is enabled in firmware.
     """
-    if is_windows():
+    if is_linux():
+        # Virtualization support is not required to run Golem on Linux.
+        return True
```
mfranciszkiewicz, line 27: Virtualization is not enabled per se, it's simply not required. I'd treat this fix as temporary. Maybe this call should be simply skipped by the UI?

kmazurek: I agree, hence the comment I left there. Another option would be to change this RPC / create a new one specific to the use case (i.e. checking for Docker support).

etam: I agree with mfranciszkiewicz. This change makes a weird exception to the simple question "is virtualization enabled?". Instead of complicating the answer, we should fix GUI who asks the bogus question.

etam (changes requested):
I agree with mfranciszkiewicz. This change makes a weird exception to the simple question "is virtualization enabled?". Instead of complicating the answer, we should fix GUI which asks the wrong question.

kmazurek:
@shadeofblue: what's your take on this?

shadeofblue:
sigh... okay @zakaprov please change the name of the function to `is_virtualization_satisfied` or `is_virtualization_requirement_satisfied` ...

etam (approved):
ok

PR #3964, opened 2019-03-06
Fix free providers

### LINKED ISSUE #3936: Requestor cannot accept a `WantToComputeTask` message with price set to 0
* [x] add an integration test that reproduces this issue
* [x] addresss the issue 

When receiving the `WantToComputeTask` message, the processing crashes when the price specified by the provider is set to 0

etam:
fixes #3936


scripts/node_integration_tests/playbooks/golem/zero_price.py
```diff
@@ -0,0 +1,17 @@
+from ..base import NodeTestPlaybook
+
+
+class RegularRun(NodeTestPlaybook):
```
kmazurek: Besides the file name, I think the test name should also reflect the logic being tested.

kmazurek (approved):

scripts/node_integration_tests/playbooks/golem/zero_price.py
etam: done

PR #3975, opened 2019-03-08
Fix and add logs in blender verifier

pawelkisielewicz:
Add and fix logs in blender verifier. Resolves part of https://codepoets.atlassian.net/browse/GF-47

### DIFF
--- a/golem/verificator/blender_verifier.py
+++ b/golem/verificator/blender_verifier.py
@@ -129,7 +129,7 @@ class BlenderVerifier(FrameRenderingVerifier):
                     Exception('Verification result negative'))
 
         logger.info(
-            f'Data for verification prepared. '
+            f' Starting docker thread for:  '
             f'Subtask_id: {subtask_id}. Extra data:{json.dumps(extra_data)}.')
         d = self.docker_task.start()
         d.addErrback(error)
--- a/golem/verificator/blender_verifier.py
+++ b/golem/verificator/blender_verifier.py
@@ -40,7 +40,9 @@ class BlenderVerifier(FrameRenderingVerifier):
     def start_verification(self, verification_data):
         self.time_started = datetime.utcnow()
         self.verification_data = verification_data
-
+        logger.info(
+            f'Start verification in BlenderVerifier. '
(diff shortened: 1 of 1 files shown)


Jakub89 (approved):

golem/verificator/blender_verifier.py
```diff
@@ -106,25 +113,24 @@ def failure(exception):
...
+            logger.warning(
+                f'Verification process exception. '
+                f'Subtask_id: {subtask_id}. Exception: {exception}')
             self.finished.errback(exception)
 
         def callback(*_):
             with open(os.path.join(dir_mapping.output, 'verdict.json'), 'r') \
                     as f:
                 verdict = json.load(f)
-
```
dybi, line 116: Why removing this?

golem/verificator/blender_verifier.py
```diff
@@ -106,25 +113,24 @@ def failure(exception):
...
-            )
             if verdict['verdict']:
                 self.finished.callback(True)
             else:
                 self.finished.errback(
-                    Exception('Verification result negative', verdict))
+                    Exception('Verification result negative'))
 
+        logger.info(
+            f'Data for verification prepared. '
```
dybi: if I get it right, it's more like:
```suggestion
            f' Starting docker thread for:  '
```

dybi (commented):
Please clarify.

golem/verificator/blender_verifier.py
pawelkisielewicz: Because in next step there is callback or errback called. And result is logged there. Leaving this message here would make logs duplicated.

dybi: In such case, the exception logging in `error()` is unnecessary as well..

pawelkisielewicz: Not necessary... `failure` (with log Verification verdict: negative.) is called in two cases: 
1. in callback, when result is negative. This is normal behavior, verification finished without errors, but verdict is negative.
2. When some unexpected error  during verification occurred. In this case I think it should be logged, that verification raised exception. This is not expected behavior.

dybi (approved):
thumbs_up_lady

PR #3479, opened 2018-10-19
Make marketplace use the provider selection algorithm

Krigpl:

PR #3976, opened 2019-03-08
fix math of calculating scaled price for marketplace

etam:
1. revert from `msg.price / task.header.max_price` back to `task.header.max_price / msg.price`
2. put this calculation, with handling of 0, into a function in marketplace module.

Related issues: #3453, #3479, #3964

### DIFF
--- a/golem/marketplace/offerpool.py
+++ b/golem/marketplace/offerpool.py
@@ -1,3 +1,4 @@
+import sys
 import logging
 from typing import List, Dict, ClassVar, Tuple
 
@@ -11,7 +12,8 @@ logger = logging.getLogger(__name__)
 
 def scale_price(task_price: float, offered_price: float) -> float:
     if offered_price == 0:
-        return float('inf')
+        # using float('inf') breaks math in order_providers, when alpha < 1
+        return sys.float_info.max
     return task_price / offered_price
 
 
--- a/golem/marketplace/offerpool.py
+++ b/golem/marketplace/offerpool.py
@@ -9,6 +9,12 @@ from .rust import order_providers
 logger = logging.getLogger(__name__)
 
 
+def scale_price(task_price: float, offered_price: float) -> float:
+    if offered_price == 0:
+        return float('inf')
+    return task_price / offered_price
+
+
 class Offer:
     def __init__(
             self,
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -20,7 +20,7 @@ from golem.core.keysauth import KeysAuth
 from golem.core import variables
 from golem.docker.environment import DockerEnvironment
 from golem.docker.image import DockerImage
-from golem.marketplace import Offer, OfferPool
+from golem.marketplace import scale_price, Offer, OfferPool
 from golem.model import Actor
 from golem.network import history
 from golem.network.concent import helpers as concent_helpers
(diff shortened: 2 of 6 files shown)


golem/task/tasksession.py
```diff
@@ -623,11 +623,7 @@ def _offer_chosen(is_chosen: bool) -> None:
 
         task = self.task_manager.tasks[msg.task_id]
         offer = Offer(
-            scaled_price=(
-                msg.price / task.header.max_price
```
Krigpl, line 627: Why did you change it without any confirmation in the first place?

golem/marketplace/offerpool.py
```diff
@@ -9,6 +9,12 @@
 logger = logging.getLogger(__name__)
 
 
+def scale_price(task_price: float, offered_price: float) -> float:
+    if offered_price == 0:
+        return float('inf')
```
Krigpl: Did you verify this is correctly handled by the Rust module?

golem/task/tasksession.py
etam: Because you made a confusing comment in another pull request (which now I know was wrong) and the design document is written in such way that it makes very hard to find related formula (I found it now).

golem/marketplace/offerpool.py
etam: It is now.

golem/marketplace/offerpool.py
```diff
@@ -9,6 +10,12 @@
 logger = logging.getLogger(__name__)
 
 
+def scale_price(task_price: float, offered_price: float) -> float:
+    if offered_price == 0:
+        return sys.float_info.max  # using float('inf') complicates calculations in rust part
```
Krigpl: What kind of complications are we talking about? It fails with an error? There's `test_rust.py` which is supposed to test proper integration with Rust and I'd expect a test there that guarantees no "complications in rust part".

tests/golem/marketplace/test_offerpool.py
```diff
@@ -1,7 +1,15 @@
...
-from golem.marketplace import Offer, OfferPool
+from golem.marketplace import scale_price, Offer, OfferPool
+
+
+class TestScalePrice(TestCase):
+    def test_basic(self):
+        assert scale_price(5, 2) == 2.5
+
+    def test_zero(self):
+        assert scale_price(5, 0) == float('inf')
```
Krigpl: is this still valid then?

Krigpl (changes requested):

tests/golem/marketplace/test_offerpool.py
etam: of course no

golem/marketplace/offerpool.py
etam: The complication was not because of python-rust communication of `float('inf')` value, but because of calculations done in `order_providers_impl` (for which I expanded tests in `marketplace.py` file).
The problem is when `alpha = 0`, because `inf * 0 = nan`. And using `inf` would totally break cases when `0 < alpha < 1`, because the score would be always `inf`.
By using float max it's at least giving a good impression of behaving correctly. I tried to talk with @mbenke about it, but he's not responding right now.

tests/golem/marketplace/test_rust.py
```diff
@@ -6,6 +7,7 @@ def test_order_providers():
     offer0 = Offer(scaled_price=2., reputation=1., quality=(1., 1., 6., 1.))
     offer1 = Offer(scaled_price=3., reputation=5., quality=(1., 3., 1., 4.))
     offer2 = Offer(scaled_price=4., reputation=2., quality=(2., 1., 1., 3.))
-    res = order_providers([offer0, offer1, offer2])
+    offer3 = Offer(scaled_price=sys.float_info.max, reputation=3., quality=(1., 2., 3., 4.))
```
Krigpl: I'd be inclined to use `scale_price` function here rather than a raw value - this value is an implementation detail which may change (and it already has done it once; from inf)

Krigpl (approved):

tests/golem/marketplace/test_rust.py
etam: done

PR #3983, opened 2019-03-11
Restructure scripts in Blender Docker so PYTHONPATH is not required

Krigpl:
Use relative imports instead which will also work when running the scripts locally in unit tests.


jiivan (approved):

Krigpl:
@jiivan 
Futurama-Fry

PR #3991, opened 2019-03-13
Keep Golem running after Docker issues at startup

### LINKED ISSUE #3966: Stop golem from dying when docker issues are encountered
we need that to ensure that the front-end can still communicate with the back-end

* [ ] ensure that `task.create` fails and doesn't try to create a task when docker is not available

kmazurek:
Fixes #3966

This change adds handling of Docker issues during client startup, preventing Golem from exiting.
Golem will continue running without connecting to the network and with basic RPCs registered to support communication with its frontend.

### DIFF
--- a/golem/node.py
+++ b/golem/node.py
@@ -14,7 +14,7 @@ from typing import (
 
 from pathlib import Path
 from twisted.internet import threads
-from twisted.internet.defer import gatherResults, Deferred, succeed, fail
+from twisted.internet.defer import gatherResults, Deferred, succeed, fail, FirstError
 from twisted.python.failure import Failure
 
 from apps.appsmanager import AppsManager
@@ -161,9 +161,19 @@ class Node(HardwarePresetsMixin):
                 docker = self._start_docker()
                 return gatherResults([terms_, keys, docker], consumeErrors=True)
 
+            def on_start_error(failure: FirstError):
+                failure.value.subFailure.trap(EnvironmentError)
+                logger.warning(
+                    """
+                    There was a problem setting up the environment. Golem will
+                    run with limited functionality to support communication with
+                    local clients.
+                    """
+                )
+
             chain_function(rpc, on_rpc_ready).addCallbacks(
                 self._setup_client,
-                self._error('keys or docker'),
+                on_start_error,
             ).addErrback(self._error('setup client'))
             self._reactor.run()
         except Exception:  # pylint: disable=broad-except
--- a/golem/node.py
+++ b/golem/node.py
@@ -164,7 +164,7 @@ class Node(HardwarePresetsMixin):
 
             def on_start_error(failure: FirstError):
                 failure.value.subFailure.trap(EnvironmentError)
-                logger.warning(
+                logger.error(
                     """
                     There was a problem setting up the environment. Golem will
                     run with limited functionality to support communication with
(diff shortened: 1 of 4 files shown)


golem/node.py
```diff
@@ -161,9 +162,19 @@ def on_rpc_ready() -> Deferred:
                 docker = self._start_docker()
                 return gatherResults([terms_, keys, docker], consumeErrors=True)
 
+            def on_start_error(failure: FirstError):
+                failure.value.subFailure.trap(EnvironmentError)
+                logger.warning(
```
shadeofblue: I believe it should be an `ERROR` not `WARNING`

golem/task/rpc.py
```diff
@@ -99,7 +99,7 @@ def validate_client(client: Client):
     if client.config_desc.in_shutdown:
         raise CreateTaskError(
             'Can not enqueue task: shutdown is in progress, '
-            'toggle shutdown mode off to create a new tasks.')
+            'toggle shutdown mode off to create new tasks.')
```
shadeofblue, line 102: I'd update this error message to be more meaningful, e.g.:

```
Cannot enqueue task: golem is running in a limited-functionality state,
either because of an initiated shutdown or because of an environment misconfiguration.
To be able to create new tasks, toggle shutdown mode off or consult golem log to resolve the configuration issue.
```

or something like that...

golem/node.py
```diff
@@ -161,9 +162,19 @@ def on_rpc_ready() -> Deferred:
...
+                    """
+                    There was a problem setting up the environment. Golem will
```
shadeofblue: + maybe mention that the error pertains to either Docker or the the keystore ... ? (unsure what the previous error message`keys or docker` meant ... )

shadeofblue: @zakaprov or maybe the message in the `EnvironmentError` is meaningful enough to include it in this logged message?

golem/task/rpc.py
```diff
@@ -99,7 +99,7 @@ def validate_client(client: Client):
     if client.config_desc.in_shutdown:
```
Wiezzel, line 99: Where is this flag being set? I don't see it in the error handler.

kmazurek: In the end I'm not making use of the shutdown flag. The error handler prevents client from being initialised which results in the relevant RPCs (creating/restarting tasks, running benchmarks, updating configs) not being registered. Golem remains in a state analogous to when terms of use are not yet accepted.

golem/node.py
kmazurek: The `EnvironmentError` doesn't include a message itself, but there's an error message logged before raising it (e.g. [here](https://github.com/golemfactory/golem/blob/8e123f99e2fe02a63a9fed0aec9b9c9285620dfd/golem/docker/manager.py#L50)). I'll update the message logged here in the error handler.

golem/task/rpc.py
kmazurek: Ah snap, this is a leftover nit from when I was trying out the shutdown flag approach. In the current approach this error won't be raised. Instead, a generic "RPC not registered" error will be raised when trying to create a task.

golem/task/rpc.py
Wiezzel: OK, fine.

Wiezzel (approved):

Issue #3630, opened 2018-12-04
fix the logic of VerificationTask


etam:
I tried to create a verifier that always reports `SubtaskVerificationState.WRONG_ANSWER`. While doing that I discovered some bugs:

https://github.com/golemfactory/golem/blob/79a858069cd0aff533ee4ca35644e452e15ac28a/apps/core/verification_task.py#L14-L21

https://github.com/golemfactory/golem-verificator/blob/661e84ddb8490c4a7da49e207e463e27163dbf0f/golem_verificator/core_verifier.py#L17-L24

1)
  - If:
    - task is not timed out
    - `simple_verification` returns `True`
    - `CoreVerifier._verify_result` returns `False`
  - then:
    - `VerificationTask.start` returns None, even though it's return type annotation is `Deferred`.
    - `self.state` is not set to `SubtaskVerificationState.WRONG_ANSWER`.

2) 
  - If `simple_verification` returns `False`
  - then `self.subtask_info` is not filled with data, because `start_verification` is not called.

3) When running tests from `tests/apps/blender/task/test_blenderrendertask.py`, when I change the code so that `entry.start(verifier_cls)` in `VerificationQueue._run` returns something other than `None`, then neither of `callback` nor `errback` is called.

PR #4004, opened 2019-03-14
per task disable or lenient verification

### LINKED ISSUE #3940: Add options to task creation dictionary to make the requestor more lenient towards the verification results
* add an `x-disable-verification` option to the task creation parameters - when specified and `True` - it should completely disable verification for the given task and treat all results as valid without performing any verification - default `False`
* add an `x-lenient-verification` option to the task creation parameters - when specified, the verification should still be performed and then in the event of negative verification result - the subtask itself should be marked as failed as usual and rescheduled and the provider should also be banned from performing any additional subtasks in this task and a failure should still be reported to the golem monitor - BUT - when this flag is present in the task definition - the provider should get a `SubtaskResultsAccepted` response and be issued a payment just as it would had the result been correct. In other words - this flag should cause the verification result to seem positive from the provider's POV.

Those options should only be known to the requestor itself. The options should NOT be exposed in the `TaskHeader` and should be clearly marked in the accompanying comments within the code as experimental.

etam:
Resolves #3940 

Blocked by #3630 

Testing plan: I think the best way to test this is to make proper integration tests. For this I need a verificator that always fails (I don't think that we could have a task that always and reliably fails verification) and an option to force using it.


etam:
Comment by @shadeofblue: for integration test, instead of failing verificator, we could have a provider that always returns bogus files (for example provided 1 black pixel).

apps/core/task/coretaskstate.py
```diff
@@ -13,6 +13,28 @@
 DEFAULT_SUBTASK_TIMEOUT = 20 * 60
 
 
+class RunVerification(enum.Enum):
+    """
+    Disabled:
```
shadeofblue: @etam I'd also add a docstring for `Enabled` and say it's a default to make it perfectly clear what's going on here

etam: done

apps/core/task/coretask.py
```diff
@@ -208,10 +210,19 @@ def computation_finished(self, subtask_id, task_result,
...
 
-        def verification_finished_(subtask_id, verdict, result):
+        def verification_finished_(subtask_id,
+                                   verdict: SubtaskVerificationState, result):
             self.verification_finished(subtask_id, verdict, result)
             verification_finished()
 
+        if self.task_definition.run_verification == RunVerification.disabled:
+            logger.debug("subtask %s: verification disabled; "
+                         "calling verification_finished", subtask_id)
```
jiivan: It's only debug, but I would follow the convention from https://github.com/golemfactory/golem-internal/wiki/Logging-guidelines#log-message-structure

golem/appconfig.py
```diff
@@ -201,6 +201,8 @@ def load_config(cls, datadir, cfg_file_name=CONFIG_FILENAME):
             hyperdrive_address=DEFAULT_HYPERDRIVE_ADDRESS,
             hyperdrive_rpc_port=DEFAULT_HYPERDRIVE_RPC_PORT,
             hyperdrive_rpc_address=DEFAULT_HYPERDRIVE_RPC_ADDRESS,
+            # testing
```
jiivan, line 204: ?

jiivan (approved):
Looks good :+1:

golem/appconfig.py
etam: What's the problem?

jiivan: I don't understand this comment. Could you ELIF? :)

etam: You can see above, there are comments, that are some sort of section headers.

apps/core/task/coretask.py
etam: 1) What exactly you have in mind? Which point?
2) Why is this guideline in private repo?

jiivan: ad1.
> Always start with a message, print data as "kwargs". For example Task deadline passed. id=%r, state=%r

ad2. dunno

etam: done

etam:
I rebased changes onto latest develop and squashed some commits.

PR #4012, opened 2019-03-15
Fixed Hyper-V VM remaining in invalid state

Wiezzel:
If creating the machine failed during virtual disk creation then the VM would remain in a an invalid state (actually not the VM but some config files used by docker-machine). This in turn would make it impossible to re-create the VM.
The error had been unsuccessfully mitigated against by adding a remove command in `HyperVHypervisor._failed_to_create()` method. This was not effective because removing the VM sometimes fails if it done shortly after the unsuccessful creation.
To make the cleanup more robust a retry loop was added. It required changing the logic of `DockerMachineHypervisor.vms()` because `docker-machine remove -q` doesn't print out VMs in invalid states.

### DIFF
--- a/golem/docker/hypervisor/hyperv.py
+++ b/golem/docker/hypervisor/hyperv.py
@@ -3,6 +3,7 @@ import logging
 import os
 from pathlib import Path
 import subprocess
+import time
 from typing import Any, ClassVar, Dict, Iterable, List, Optional
 
 from os_win.constants import HOST_SHUTDOWN_ACTION_SAVE, \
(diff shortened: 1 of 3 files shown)


maaktweluit (approved):
Nice find on the `-q`!

Sad the retry is needed but good its more robust now

golem/docker/hypervisor/hyperv.py
```diff
@@ -232,13 +234,25 @@ def _parse_create_params(
...
+                break
+
+            try:
+                self.command('rm', name, args=['-f'])
+            except subprocess.CalledProcessError:
+                logger.warning(f'{self.DRIVER_NAME}: Attempt to remove '
+                               f'machine "{name}" failed')
+            time.sleep(0.1)
+
+        else:
```
mfranciszkiewicz, line 252: Doesn't this always trigger on the last iteration?

Wiezzel: That was the idea. If all the iterations passed then we assume that removing the VM failed. If it succeeded then `break` would be triggered and `else` clause wouldn't be executed.

mfranciszkiewicz (approved):

PR #4016, opened 2019-03-18
Message Queue Storage

jiivan:
Complete API for message queue storage.

part of #2223

### DIFF
--- a/golem/database/schemas/024_create_message_queue.py
+++ b/golem/database/schemas/024_create_message_queue.py
@@ -1,0 +1,25 @@
+# pylint: disable=no-member
+# pylint: disable=unused-argument
+import datetime
+
+import peewee as pw
+
+SCHEMA_VERSION = 21
+
+
+def migrate(migrator, database, fake=False, **kwargs):
+    @migrator.create_model  # pylint: disable=unused-variable
+    class QueuedMessage(pw.Model):
+        node = pw.CharField()
+        msg_version = pw.CharField()
+        msg_cls = pw.CharField()
+        msg_data = pw.BlobField()
+        created_date = pw.DateTimeField(default=datetime.datetime.now)
+        modified_date = pw.DateTimeField(default=datetime.datetime.now)
+
+        class Meta:
+            db_table = "queuedmessage"
+
+
+def rollback(migrator, database, fake=False, **kwargs):
+    migrator.remove_model("queuedmessage")
--- a/golem/model.py
+++ b/golem/model.py
@@ -4,10 +4,13 @@ import inspect
 import json
 import pickle
 import sys
+import time
 from typing import Optional
 
 from eth_utils import decode_hex, encode_hex
 from ethereum.utils import denoms
+import golem_messages
+from golem_messages import datastructures as msg_dt
 from golem_messages import message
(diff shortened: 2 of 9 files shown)


golem/database/schemas/024_create_message_queue.py
```diff
@@ -0,0 +1,25 @@
...
+def migrate(migrator, database, fake=False, **kwargs):
+    @migrator.create_model  # pylint: disable=unused-variable
+    class QueuedMessage(pw.Model):
+        node = pw.CharField()
+        msg_version = pw.CharField()
+        msg_cls = pw.CharField()
+        msg_data = pw.BlobField()
+        created_date = pw.DateTimeField(default=datetime.datetime.now)
+        modified_date = pw.DateTimeField(default=datetime.datetime.now)
+
```
shadeofblue, line 19: will messages live forever if they cannot be sent? maybe they should have some lifetime/timeout set? possibly depending on a type maybe? ... (unsure)

golem/model.py
```diff
@@ -515,6 +535,51 @@ def as_message(self) -> message.base.Message:
...
+        message.base.verify_version(str(self.msg_version))
+        msg = golem_messages.load(
+            self.msg_data,
+            None,
+            None,
+            check_time=False,
+        )
+        msg.header = msg_dt.MessageHeader(
+            msg.header[0],
+            int(time.time()),
```
shadeofblue, line 567: are we sure that this always produces the correct unix epoch?

golem/network/transport/msg_queue.py
```diff
@@ -0,0 +1,77 @@
...
+            except msg_exceptions.MessageError:
+                logger.info(
+                    'Invalid message in queue.'
+                    ' db_model=%s',
+                    db_model,
+                    exc_info=True,
+                )
+                continue
+            finally:
+                db_model.delete_instance()
```
shadeofblue, line 57: hmm... what if sending of a message fails at the time it's performed? will message be returned to the queue? ... shouldn't a message be removed from the queue only after it has been successfully sent? ...

golem/model.py
jiivan: Only case it didn't produce expected results was on AppVeyor. We don't use AppVeyor anymore.

golem/database/schemas/024_create_message_queue.py
kmazurek: I think these are removed after `MESSAGE_QUEUE_MAX_AGE`, see method `sweep` in `msg_queue.py`.

shadeofblue: yeah, I noticed later

golem/network/transport/msg_queue.py
jiivan: IMHO the cost of implementing this would be higher than the benefits.

kmazurek (approved):
LGTM

golem/network/transport/msg_queue.py
shadeofblue: okay, we may leave that as a future improvement if it deems necessary...

shadeofblue (approved):
okay, as mentioned above, we can leave it as a future improvement

PR #4017, opened 2019-03-18
Allow to push resources after query_extra_data

badb:
We want to make sure that user is able to push additional resource for new subtasks. That can be useful if there is a task that may have dynamic resources that cannot be all defined and pushed during task creation (for example they are based on previous subtasks results).

Previously resources were only send during task creation and it was not possible to add them later.


golem/task/rpc.py
```diff
@@ -95,7 +94,7 @@ def _validate_task_dict(client, task_dict) -> None:
             )
 
 
-def validate_client(client: Client):
```
mplebanski, line 99: Why has this type hint removed here along with the import?

golem/task/rpc.py
```diff
@@ -277,10 +276,35 @@ def _get_mask_for_task(client, task: coretask.CoreTask) -> masking.Mask:
 
 
 @defer.inlineCallbacks
-def _inform_subsystems(client, task, packager_result):
-    task_id = task.header.task_id
+def add_resources(client, resources, res_id, timeout):
```
mplebanski, line 280: Consider passing a resource_server directly, since you are never using client, but only it's field.

golem/task/rpc.py
```diff
@@ -277,10 +276,35 @@ def _get_mask_for_task(client, task: coretask.CoreTask) -> masking.Mask:
...
+    files = copy.copy(list(resources))
+
+    packager_result = yield client.resource_server.create_resource_package(
+        files,
+        res_id
+    )
     package_path, package_sha1 = packager_result
     resource_size = os.path.getsize(package_path)
+    client_options = client.task_server.get_share_options(res_id, None)
```
mplebanski, line 289: Comment unrelated to this PR. Do you have any idea why get_share_options has a unused first parameter?

golem/task/rpc.py
```diff
@@ -277,10 +276,35 @@ def _get_mask_for_task(client, task: coretask.CoreTask) -> masking.Mask:
...
+    client_options.timeout = timeout
+    resource_server_result = yield client.resource_server.add_resources(
+        package_path,
+        package_sha1,
+        res_id,
+        resource_size,
+        client_options=client_options,
+    )
+
+    logger.info(
```
mplebanski: I don't understand this log message. What subsystems are we informing here and where do we do that?

golem/task/rpc.py
```diff
@@ -343,20 +362,9 @@ def enqueue_new_task(client, task, force=False) \
...
-        client=client,
-        task=task,
-    )
-
-    logger.info(
-        "Resource package created. Informing subsystems. task_id=%r",
-        task_id,
-    )
-
     resource_server_result = yield _inform_subsystems(
```
mplebanski, line 363: We should consider refactoring this `_inform_subsystems` function name. A name like this implies that a result is something like "we managed to inform" or "we didn't manage to inform subsystems", not a `resource_server_result`. It's hard to reason about this code based on identifiers.

golem/task/tasksession.py
```diff
@@ -639,19 +652,24 @@ def _offer_chosen(is_chosen: bool) -> None:
...
                 provider_id=self.key_id,
-                package_hash='sha1:' + task_state.package_hash,
+                package_hash='sha1:' + package_hash,
                 concent_enabled=msg.concent_enabled,
                 price=price,
-                size=task_state.package_size,
+                size=package_size,
                 resources_options=self.task_server.get_share_options(
-                    ctd['task_id'], self.address).__dict__
+                    ctd['subtask_id'], self.address).__dict__
```
mplebanski: OK now my comment about `get_share_options` is related, because there you are changing `task_id` to `subtask_id`.

golem/task/tasksession.py
```diff
@@ -627,6 +629,17 @@ def _offer_chosen(is_chosen: bool) -> None:
...
+            resources_result = None
+            if ctd["resources"]:
+                resources_result = yield add_resources(
+                    self.task_server.client,
+                    ctd["resources"],
+                    ctd["subtask_id"],
+                    common.deadline_to_timeout(ctd["deadline"])
+                )
+                resources = self.task_server.get_resources(ctd['subtask_id'])
+                ctd["resources"] = resources
```
mplebanski: Why resources from `task_server` are reassigned to CTD?

golem/task/tasksession.py
```diff
@@ -639,19 +652,24 @@ def _offer_chosen(is_chosen: bool) -> None:
                 msg.price,
                 task.header.subtask_timeout,
             )
+            if resources_result:
+                _, _, package_hash, package_size = resources_result
+            else:
+                package_hash = task_state.package_hash
```
mplebanski: What is the `package_hash` held by `task_state`? Is this some sort of resource that is always sent to the subtask if there are no other resources defined in CTD?

golem/task/rpc.py
mplebanski: OK actually when I think about it I expect a function named `_inform_subsystems` to throw an exception on error and not return anything on success.

mplebanski:
I rebased on my `glambda0.3` branch to run my integration test with thin client. They work so it's rock solid change. Except from what I mentioned in the comments it looks good to me.

mplebanski:
Your PR will need to align with https://github.com/golemfactory/golem/issues/2223 `Use MQ in offer_chosen callback` pretty soon.

golem/task/rpc.py
badb: Because I've added rpc import in tasksession and created circular dependency and Client module was important here just to do this single check.

golem/task/rpc.py
badb: From history of Golem: It was used when we have another method for adding resources (ie. third golem network, resource server) and we pushed there information about peers holding this type of resource. 
But you're right it definietly should be refactored.

badb: Issue: https://github.com/golemfactory/golem/issues/4121

golem/task/rpc.py
badb: It's not a bad idea, but it should be part of separete PR.

golem/task/tasksession.py
badb: Because previously they were in the form that the app may give (just paths to files) and we want to change them to the one that were serialized properly by resource_manage (to_wire function). Provider expects to get resources in the serialized form (so from_wire function may be applied).

golem/task/tasksession.py
badb: It's hash of the resources that were pushed to the network during task creation.

badb:
**For QA team:** 

I need you to focus on testing typical computation scenarios. Especially the one with:
* multiple structure resource (ie. classroom) 
* many subtasks 
* some simple golem restarting scenarios (turning off and on during computation). 

No specific and modified version of Electron is needed

@ZmijaWA @ederenn

golem/task/tasksession.py
mfranciszkiewicz: Could you please add a brief comment?

golem/task/rpc.py
mplebanski: #4126

badb:
@ZmijaWA @ederenn Branch updated.

ZmijaWA:
@badb the task is under "testing" status ;)

golem/task/rpc.py
shadeofblue: @badb fyi, it can be done this way:

https://github.com/golemfactory/golem/blob/develop/golem/task/taskcomputer.py#L27
https://github.com/golemfactory/golem/blob/develop/golem/task/taskcomputer.py#L54

shadeofblue (approved):
looks okay as far as I could tell :)

PR #4018, opened 2019-03-19
Completed the rename from private to public

maaktweluit:
Running the `concent_acceptance_test` failed on b0.19.


shadeofblue (approved):
strawberry leaf

PR #4020, opened 2019-03-19
optionally dump node output on fail of a node integration test

shadeofblue:


scripts/node_integration_tests/conftest.py
```diff
@@ -36,17 +38,48 @@ def enabled(self):
         return REUSE_KEYS
 
 
+class DumpOutput:
+    @staticmethod
+    def on_crash():
```
kmazurek: This method name sounds like a handler, perhaps add a type hint here?

scripts/node_integration_tests/conftest.py
```diff
@@ -36,17 +38,48 @@ def enabled(self):
...
+        return DUMP_OUTPUT_ON_CRASH
+
+    @staticmethod
+    def enable_on_crash():
+        global DUMP_OUTPUT_ON_CRASH
+        DUMP_OUTPUT_ON_CRASH = True
+
+    @staticmethod
+    def on_fail():
```
kmazurek: Same as above.

kmazurek (approved):
lgtm

scripts/node_integration_tests/conftest.py
shadeofblue: right, I'll rather just change the name to something else, thanks :)

Issue #3015, opened 2018-07-11
Improve command feedback in golemcli


shadeofblue:
Currently, cli only reports "Completed in ... " with no actual feedback
* [x] Analyze output of all golemcli commands https://github.com/golemfactory/golem/issues/3968

Changes required:

1. 	[x] account withdraw [address] [Amount GNT] [gas fee WEI]
		is: ['Tx Hash']

	account withdraw  [address] [Amount GNT] [gas fee WEI]
		Sent `Amount` GNT to: [`address`]. With gas fee: `Amount` WEI
		Transaction hash (check on etherscan.io): ['Tx Hash']

---

2. 	[x] concent switch turn on/off
		is: Completed in 0.01 s

	concent switch turn on/off
		Concent switch turned on. Completed in 0.01 s

---

3.	[x] envs disable 'env_name' eg. BLENDER
		is: Completed in 0.01 s

	envs disable `env_name`
		env `env_name` disabled. Completed in 0.01 s

---

4.	[x] envs enable 'env_name' eg. BLENDER
		is: Completed in 0.01 s

	envs enable `env_name`
		env `env_name` enabled. Completed in 0.01 s

---

5.	[x] envs perf_mult_set `value`
		is: Completed in 0.01 s

	envs perf_mult_set `value`
		envs perf_mult_set to `value`. Completed in 0.01 s

---

6. 	[x] network block `node_id`
		is: Completed in 0.01 s

	network block `node_id`
		`node_id` blocked in 0.01 s

---

7. 	[x] network connect 
		The following arguments are required: ip, port_

---

8. 	[x] network connect `ip`
		The following arguments are required: port_

	network connect `ip`
		Required arguments are `ip` and `port`. Port is still missing

---

9. 	[ ] network connect `ip` `port`
		is: Completed in 0.01 s

	network connect `ip` `port`
		Connected with: Ip: `ip` Port: `port` in 0.01 s

---

10.	[x] settings set node_name `node name`
		is: Completed in 0.01 s

	settings set node_name `node name`
		Node name changed to: `node name` in 0.01 s. To confirm run `golemcli settings show`

---

11. [x] settings set accept_tasks `0/1`
		is: Completed in 0.01 s

	settings set accept_tasks `0/1`
		Your node will accept tasks. To confirm run `golemcli settings show`
		Your node will not accept tasks (acting as requestor only). To confirm run `golemcli settings show`

---

12. [ ] settings set getting_tasks_interval `value`
[...]

PR #4023, opened 2019-03-20
[WIP] Add required return messages

pawelkisielewicz:
Resolves https://github.com/golemfactory/golem/issues/3015

### DIFF
--- a/golem/interface/cli.py
+++ b/golem/interface/cli.py
@@ -2,13 +2,17 @@ import argparse
 import shlex
 import sys
 import time
+from collections import namedtuple
 from copy import deepcopy
-from typing import Text, Dict, Callable
+from typing import Text, Dict, Callable, List, Optional
 
-from golem.interface.command import CommandHelper, CommandStorage, command, Argument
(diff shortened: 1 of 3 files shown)


golem/interface/cli.py
```diff
@@ -56,6 +60,11 @@ def disable_withdraw(
     return new_children
 
 
+def _ask_for_confirmation(question: str) -> bool:
+    text = input(f'{question} (y/n) ')
+    return True if text == 'y' else False
```
Jakub89: This function should be change in a way that all unix questions works. Empty value should be `True` just as it works when installing some packages. You could also make a `while` loop that would protect against typos, so that only available inputs are Y,y,N,n and empty space as `True`.

pawelkisielewicz: fixed

Jakub89 (approved):
Looks good to me, just please remember about `pylint` fixes.

tworec:
golemcli has been rewritten to Rust
see https://github.com/golemfactory/golem-client

Issue #4024, opened 2019-03-20
Windows Hyper-V check passes when the module is disabled manually


kmazurek:
The script used for checking virtualization support on Windows (`get-virtualization-state.ps1`) gives a false positive when the Hyper-V module is installed, but manually disabled by the user.

PR #4025, opened 2019-03-20
Fixed virtualization detection on Windows

kmazurek:
Resolves: #4024

Adds handling of a scenario where Hyper-V is installed, but has been
manually disabled by the user. Additionally, this removes the call to
WMI, since when Hyper-V is on it reports hardware virtualization as disabled.

### DIFF
--- a/golem/docker/hypervisor/hyperv.py
+++ b/golem/docker/hypervisor/hyperv.py
@@ -91,9 +91,11 @@ class HyperVHypervisor(DockerMachineHypervisor):
     VOLUME_DRIVER = "cifs"
     SMB_PORT = "445"
 
-    SCRIPTS_PATH = os.path.join(get_golem_path(), 'scripts', 'docker')
+    SCRIPTS_PATH = os.path.join(get_golem_path(), 'scripts')
     GET_VSWITCH_SCRIPT_PATH = \
-        os.path.join(SCRIPTS_PATH, 'get-default-vswitch.ps1')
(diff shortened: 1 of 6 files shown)


Wiezzel (commented):
Is the new version intended to work for Hyper-V only? What about DockerToolbox support?

kmazurek:
@Wiezzel I did not take Docker Toolbox into account, but that's also true for the previous version of this script. My understanding is that the script should return true if either Hyper-V or Docker Toolbox is available. Is that correct?

Wiezzel:
According to my knowledge the new version of the script will return `False` even if DockerToolbox is installed and virtualization enabled in BIOS.

kmazurek:
Ok, understood - let me bring back that check.

Wiezzel (approved):
Seems OK now.

golem/docker/hypervisor/hyperv.py
```diff
@@ -185,10 +185,16 @@ def start_vm(self, name: Optional[str] = None) -> None:
 
     @classmethod
     def is_available(cls) -> bool:
-        command = "@(Get-Module -ListAvailable hyper-v).Name | Get-Unique"
         try:
-            output = run_powershell(command=command)
-            return output == "Hyper-V"
+            output = run_powershell(
+                script=os.path.join(
```
Wiezzel: Please, extract this path into a constant like `GET_VSWITCH_SCRIPT_PATH `.

Wiezzel (changes requested):

Wiezzel (approved):
Alright now.

PR #4029, opened 2019-03-21
Random hacks of kindness

etam:
Some dead code removed, some typing improved, stuff like that.


apps/core/task/coretaskstate.py
```diff
@@ -50,7 +51,7 @@ def __init__(self):
 
         self.max_price = 0
 
-        self.verification_options = None
```
jiivan, line 53: This value is modified
`apps/dummy/benchmark/benchmark.py:        self.verification_options["subtask_id"] = "DummyBenchmark"`

And generally it's best to avoid modifiable default values. So it should be `Optional`

golem/task/taskserver.py
```diff
@@ -319,28 +319,26 @@ def send_results(self, subtask_id, task_id, result):
         if 'data' not in result:
             raise AttributeError("Wrong result format")
 
-        header = self.task_keeper.task_headers[task_id]
+        if subtask_id in self.results_to_send:
+            raise RuntimeError("Incorrect subtask_id: {}".format(subtask_id))
```
jiivan, line 323: :+1:

jiivan (commented):
I like this cleanup! Only one issue found.

mplebanski (approved):
LGTM!

apps/core/task/coretaskstate.py
etam: `TaskDefinition.verification_options` is unrelated to `DummyTaskBenchmark.verification_options`. Is `TaskDefinition.verification_options` used anywhere at all?

etam: It looks like it's not used at all. I removed it, run unit tests and `regular_run` integration test and everything passed fine.

jiivan (approved):

PR #4034, opened 2019-03-21
Docker CPU environment

Wiezzel:


golem/envs/__init__.py
```diff
@@ -18,8 +16,41 @@
...
+
+class EnvConfig(Serializable, ABC):
+    """ Environment-wide configuration. Specifies e.g. available resources. """
+
+
+class Prerequisites(Serializable, ABC):
+    """
+    Environment-specific requirements for computing a task. Distributed with the
+    task header. Providers are expected to prepare (download, install, etc.)
+    prerequisites in advance no to waste computation time.
```
mfranciszkiewicz:
```suggestion
    prerequisites in advance not to waste computation time.
```

golem/envs/__init__.py
```diff
@@ -18,8 +16,41 @@
...
+
+
+class Payload(Serializable, ABC):
+    """
+    A definition for Runtime. Environment-specific description of computation to
+    be run. Received when provider is assigned a subtask.
+    """
+
+
+class EnvSupportStatus(NamedTuple):
```
mfranciszkiewicz, line 40: This already exists, see `golem.environments.environment.SupportStatus`

golem/envs/docker/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+    """ This exists because NamedTuple must be single superclass """
+
+    def to_dict(self) -> Dict[str, Any]:
+        dict_ = self._asdict()
+        dict_['source'] = str(dict_['source'])
+        return dict_
+
+    @classmethod
+    def from_dict(cls, dict_: Dict[str, Any]) -> 'DockerBind':
+        source = Path(dict_.pop('source'))
```
mfranciszkiewicz: Do we really need to modify `dict_`?

golem/envs/docker/cpu.py
```diff
@@ -0,0 +1,290 @@
...
+        logger.info('Environment supported.')
+        return EnvSupportStatus(True)
+
+    @classmethod
+    def _check_docker_version(cls) -> bool:
+        logger.info('Checking docker version...')
+        try:
+            version_string = DockerCommandHandler.run("version")
+        except SubprocessError:
+            logger.info('Checking docker version failed.')
```
mfranciszkiewicz: `logger.[error|exception]`?

golem/docker/hypervisor/__init__.py
```diff
@@ -32,6 +32,7 @@ def __init__(self,
 
     @classmethod
     def is_available(cls) -> bool:
+        # FIXME: Implement an actual check
```
mfranciszkiewicz: Maybe `raise NotImplementedError` instead?

golem/envs/__init__.py
```diff
@@ -75,78 +118,93 @@ class EnvStatus(Enum):
...
     @abstractmethod
-    def metadata(self) -> EnvMetadata:
+    def parse_prerequisites(cls, prerequisites_dict: Dict[str, Any]) \
+            -> Prerequisites:
+        """ Build Prerequisites struct from supplied dictionary. Returned value
+            is of appropriate type for calling prepare_prerequisites(). """
+        raise NotImplementedError
+
+    @abstractmethod
+    def prepare_prerequisites(self, prerequisites: Prerequisites) -> Deferred:
```
mfranciszkiewicz: IMO the name `prepare_prerequisites` doesn't suggest that we may modify some global state here

golem/envs/__init__.py
```diff
@@ -75,78 +118,93 @@ class EnvStatus(Enum):
...
+    @classmethod
+    @abstractmethod
+    def metadata(cls) -> EnvMetadata:
+        """ Get Environment metadata. """
+        raise NotImplementedError
+
+    @classmethod
     @abstractmethod
-    def metadata(self) -> EnvMetadata:
+    def parse_prerequisites(cls, prerequisites_dict: Dict[str, Any]) \
```
mfranciszkiewicz, line 150: Could this be a part of `Prerequisites`?

golem/envs/__init__.py
```diff
@@ -18,8 +16,41 @@
 RuntimeEvent = Any  # TODO: Define runtime events
 
 
-class Payload:
-    pass
+class Serializable(ABC):
```
mfranciszkiewicz: Already exists: `golem.core.simpleserializer.DictSerializable`

golem/envs/docker/cpu.py
Wiezzel: Yep, this should be `exception()`

golem/envs/docker/__init__.py
Wiezzel: I could make a copy.

golem/envs/__init__.py
Wiezzel: Well, there are two reasons why I'm reluctant to use it:
1. The `_desc` field of `SupportStatus` is kind of oddly coupled with `UnsupportReason` which doesn't fit the new environments picture (e.g. `CONCENT_REQUIRED` or `MASK_MISMATCH` have nothing to do with environments).
2. The current `environments` package is going away when we're done with refactoring, isn't it?

golem/docker/hypervisor/__init__.py
Wiezzel: OK, why not.

golem/envs/__init__.py
Wiezzel: I'd gladly welcome some better name suggestion.
`download_prerequisites`?
`install_prerequisites`?

golem/envs/__init__.py
Wiezzel: Not really. The assumption is that caller of this method doesn't have to know the exact type of `Prerequisites` returned. I imagine something like this:

```python
...
env_id = task_header.env_id
prereqs_dict = task_header.prerequisites
env = env_manager.environment(env_id)
prereqs = env.parse_prerequisites(prereqs_dict)
env.prepare_prerequisites(prereqs)
```

golem/docker/hypervisor/__init__.py
Wiezzel: Actually, this comment is referring to VirtualBox. We need an availability check for `VirtualBoxHypervisor`.

Wiezzel: And for Xhyve as well, as it seems.

golem/envs/__init__.py
mfranciszkiewicz: Perhaps `install`?

Wiezzel: OK, will change.

golem/envs/docker/cpu.py
```diff
@@ -0,0 +1,290 @@
...
+        return DockerPayload.from_dict(payload_dict)
+
+    def runtime(self, payload: Payload, config: Optional[EnvConfig] = None) \
+            -> DockerCPURuntime:
+        assert isinstance(payload, DockerPayload)
+        if config is not None:
+            assert isinstance(config, DockerCPUConfig)
+        else:
+            config = self.config()
+        logger.info("Creating runtime...")
```
maaktweluit, line 288: 3 dots at the end of a log messages makes me assume it also reports when its done

maaktweluit (approved):
LGTM!

Nice and clear comments!

golem/envs/__init__.py
mfranciszkiewicz: Yes, we're going to remove the old `environments` package

mfranciszkiewicz (approved):

PR #4037, opened 2019-03-22
Create a separate directory for every task's output

kmazurek:
Fixes: #3785

This changes the way output paths for tasks are handled. For each new task, a directory gets created in the output path from the task definition. The new directories' names are based on the task's name and their time of creation, e.g. `some_task_2019-01-01_12-00-00`.

### DIFF
--- a/apps/core/task/coretask.py
+++ b/apps/core/task/coretask.py
@@ -591,19 +591,19 @@ class CoreTaskBuilder(TaskBuilder):
     @classmethod
     def get_output_path(cls, dictionary, definition):
         options = dictionary['options']
+
         base_path = options['output_path']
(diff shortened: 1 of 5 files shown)


etam:
How this relates to #3854 ?

kmazurek:
@etam It fixes the same issue, but instead of focusing solely on the specific reported case (i.e. multi-frame tasks) this changes the way task output is handled in general.
Each task will get its own directory in the output path, with the directory names based on the task's name and creation timestamp (precision rounded up to seconds). This ensures there are no collisions between subsequent runs of the same task on one machine.

jiivan (approved):
greencanoe

etam (approved):
I like this approach.

ok

PR #4044, opened 2019-03-26
Improved checking Hyper-V status on Windows

kmazurek:
Resolves: #4024

Updated the relevant Powershell script to use the field
HypervisorPresent from Win32_ComputerSystem WMI object. This provides a
more reliable way of determining hypervisor availability on the system.

### DIFF
--- a/scripts/virtualization/get-hyperv-state.ps1
+++ b/scripts/virtualization/get-hyperv-state.ps1
@@ -1,4 +1,20 @@
 $ErrorActionPreference = "Stop"
 [Console]::OutputEncoding = [Text.UTF8Encoding]::UTF8
 
-return (gwmi Win32_ComputerSystem).HypervisorPresent
+$IsHypervisorPresent = (gwmi Win32_ComputerSystem).HypervisorPresent
+
+if ($IsHypervisorPresent) {
+    $HyperVServices = @("vmms", "vmcompute")
+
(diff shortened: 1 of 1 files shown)


Wiezzel (changes requested):
`HypervisorPresent` is true even if hypervisor service is not running.

kmazurek:
Ok, I'll keep the check for `vmms` process running. However, from my testing `HypervisorPresent` covers more cases than the call to `Get-Module -ListAvailable hyper-v`. I'm going to replace the latter with the former in the script.

Wiezzel:
@zakaprov Maybe instead of this try-catch do this:
```powershell
(Get-Service -Name vmms).Status -Eq "Running"
```

Wiezzel:
And you could also check `vmcompute` service.

kmazurek:
@Wiezzel Good one, thanks. I might still need to wrap the call to `Get-Service` in a try-catch since these services may not be present on a fresh system. Will try to verify that.

Wiezzel (commented):
If `$IsHypervisorPresent` is `False` the script should return `False`.

kmazurek:
@Wiezzel It does return `False`, the last line of the script got conveniently truncated in the GitHub changes viewer.

Wiezzel (approved):

PR #4046, opened 2019-03-26
Created integration test for task output directories

kmazurek:


scripts/node_integration_tests/playbooks/golem/task_output.py
```diff
@@ -0,0 +1,60 @@
...
+        task.
+    * Checking whether the names of the output directories match the expected
+        format (e.g. 'some_task_2019-01-01_12-00-00').
+    """
+    OUTPUT_DIR_TIME_FORMAT = '%Y-%m-%d_%H-%M-%S'
+
+    provider_node_script = 'provider/debug'
+    requestor_node_script = 'requestor/debug'
+
+    def step_set_task_in_creation_to_false(self):
```
shadeofblue: @zakaprov why is this needed?

kmazurek: Otherwise the test gets stuck creating the task for the second time. This step was copied over from #3899, but now that you mentioned it I think `step_create_task` should be updated to include resetting this flag. Let me change that.

etam (approved):
ok

Issue #4039, opened 2019-03-25
Golem installation for account with space in name ends in failure


ederenn:
## Description

**Golem Version**: 0.19.1
bulid 3966

**Golem-Messages version** (leave empty if unsure): 2.24.3

**Electron version** (if used): 0.1.19

**OS** [e.g. Windows 10 Pro]: Windows

**Branch** (if launched from source): -

**Mainnet/Testnet**: Mainnet

**Description of the issue**:

When installing Golem app on account with space in user name whole installation ends in failure "installation has failed"

**Actual result**:

It is not possible to end installation process

**Screenshots**:

<img width="568" alt="Screen Shot 2019-03-22 at 15 55 04" src="https://user-images.githubusercontent.com/32061891/54927839-40b3aa00-4f13-11e9-8f56-47fb117c245a.png">

## Steps To Reproduce
- On Windows os create an account with space in name for example: Bad Wolf
- on this account run installation with golemapp.exe
- installation wil end in failure with error log included
- save error log

## Expected behavior
_(What is the expected behavior and/or result in this scenario)_

## Logs and any additional context

[MSI66AA.LOG](https://github.com/golemfactory/golem/files/3002921/MSI66AA.LOG)

## Proposed Solution?
_(Optional: What could be a solution for that issue)_

Wiezzel:
I was unable to reproduce the issue. The cause of failure must something else than whitespace in username.

Wiezzel:
Investigated further today. It appears to be an issue with the user not being an administrator, unrelated to username.

ederenn:
One of the users had a problem with installation. Installation ended in failure. Adding admin rights to the account did not help for that user. Installation on account with admin rights ended in failure anyway.

![Zrzut ekranu 2019-03-28 o 12 30 12](https://user-images.githubusercontent.com/32061891/55154587-77790280-5155-11e9-817d-04fb9544232a.png)

installation logs below

before adding admin rights:
[MSIEA51.LOG](https://github.com/golemfactory/golem/files/3017676/MSIEA51.LOG)

after adding admin rights:
[MSI727E.LOG](https://github.com/golemfactory/golem/files/3017829/MSI727E.LOG)

Wiezzel:
@ederenn I have created an account with the extact same name as user's (`Jack Daniel's`) and I'm unable to reproduce this issue. I suspect it has already been fixed by #3995 .

PR #4052, opened 2019-03-28
Require administrator privilege to run the Windows installer

Wiezzel:
Powershell scripts included in the installer didn't escalate privilege so that installation would for non-admin users. In order to remove this issue and potential future permission issues the installer now requires to be run with administrator privilege right from the start.


Wiezzel:
Fixes #4039

shadeofblue (approved):

mfranciszkiewicz (approved):

PR #4054, opened 2019-03-28
Enable HyperG to work using arbitrary ports

shadeofblue:


golem/client.py
```diff
@@ -408,7 +408,19 @@ def start_network(self):
 
         logger.info("Starting resource server ...")
 
-        self.daemon_manager = HyperdriveDaemonManager(self.datadir)
+        self.daemon_manager = HyperdriveDaemonManager(
+            self.datadir,
+            daemon_config={
+                'host': self.config_desc.hyperdrive_address,
+                'port': self.config_desc.hyperdrive_port,
+                'rpc_host': self.config_desc.hyperdrive_address,
```
mfranciszkiewicz: The RPC and daemon listening addresses should not be the same; i.e. RPC is a local server bound to `127.0.0.1` and the other is a public one, bound to `0.0.0.0` by default

mfranciszkiewicz (dismissed):

golem/client.py
shadeofblue: ah, haven't thought of that... okay, I'll update the code further

shadeofblue: @mfranciszkiewicz updated

golem/appconfig.py
```diff
@@ -191,10 +196,21 @@ def load_config(cls, datadir, cfg_file_name=CONFIG_FILENAME):
...
+            hyperdrive_rpc_port=DEFAULT_HYPERDRIVE_RPC_PORT,
+            hyperdrive_rpc_address=DEFAULT_HYPERDRIVE_RPC_ADDRESS,
         )
 
         cfg = SimpleConfig(node_config, cfg_file, keep_old=False)
-        return AppConfig(cfg, cfg_file)
+        return cls(cfg, cfg_file)
+
+    def __repr__(self):
+        return '{}: {}'.format(self.__class__, {
```
jiivan: Let repr be repr

> [...]otherwise the representation is a string enclosed in angle brackets that contains the name of the type of the object together with additional information[...]
https://docs.python.org/3/library/functions.html#repr

```suggestion
        return '<{}: {}>'.format(self.__class__, {
```

golem/network/hyperdrive/client.py
```diff
@@ -53,6 +51,9 @@ def __init__(self, port=DEFAULT_HYPERDRIVE_RPC_PORT,
         self._url = 'http://{}:{}/api'.format(self.host, self.port)
         self._headers = {'content-type': 'application/json'}
 
+    def __repr__(self):
+        return '{} at {}'.format(self.CLIENT_ID, self._url)
```
jiivan:
```suggestion
        return f'<{self.__class__.__name__} {self.CLIENT_ID} at {self._url}>'
```

jiivan (approved):
LGTM, besides repr contents

mfranciszkiewicz (approved):
Looks good but the tests need more love.

PR #4058, opened 2019-03-29
Multinode integration tests

etam:
This refactoring allows having more nodes in integration tests.

### DIFF
--- a/scripts/node_integration_tests/README.md
+++ b/scripts/node_integration_tests/README.md
@@ -25,8 +25,9 @@ behavior - e.g. a requestor node that always fails the submitted results.
 This allows us to simulate various scenarios of two nodes interacting with
 each other and verifying they do indeed react the way we intend them to.
 
-We could theoretically create tests that employ more than two golem instances
-but we don't do that yet and it's not yet supported by the test suite. 
+In most cases there are two nodes (provider and requestor), and this is a
+default configuration for tests. If needed, it is possible to define any number of
+nodes with custom names.
 
 ### Components
 
@@ -41,8 +42,12 @@ introduce modifications to the regular node behavior, implemented using
 #### playbooks
 
 The scripts for the tests themselves, each of which is an instance of
-`playbooks.base.NodeTestPlaybook` and is comprised of discrete steps
-through which the execution of the test progresses.
+`playbooks.base.NodeTestPlaybook`, is comprised of discrete steps through which
+the execution of the test progresses.
+
+A configuration for each test is a class that derives from
+`playbooks.test_config_base.TestConfigBase`. This allows to use it outside of
+test itself, like setting up key reuse by pytest (see below).
 
 After both golem instances are spawned, most of the steps consist of RPC calls
 to either the requestor or to the provider node and checking for expected
@@ -127,16 +132,13 @@ example:
 full usage:
 
 ```
-run_test.py [-h] [--task-package TASK_PACKAGE]
-            [--task-settings TASK_SETTINGS]
-            [--provider-datadir PROVIDER_DATADIR]
-            [--requestor-datadir REQUESTOR_DATADIR] [--mainnet]
-            test_path
+run_test.py [-h] [--task-package TASK_PACKAGE] [--task-settings TASK_SETTINGS]
+            [--datadir NODE PATH] [--mainnet] test_path
 
 Runs a single test playbook.
 
 positional arguments:
-  test_path             a dot-separated path to the playbook class within
+  test_path             a dot-separated path to the test module within
                         `playbooks`, e.g. golem.regular_run
 
 optional arguments:
@@ -145,10 +147,8 @@ optional arguments:
                         a directory within `tasks` containing the task package
   --task-settings TASK_SETTINGS
                         the task settings set to use, see `tasks.__init__.py`
-  --provider-datadir PROVIDER_DATADIR
-                        the provider node's datadir
-  --requestor-datadir REQUESTOR_DATADIR
-                        the requestor node's datadir
+  --datadir NODE PATH   override datadir path for given node. standard node
+                        names are 'requestor' and 'provider'
   --mainnet             use the mainnet environment to run the test (the
                         playbook must also use mainnet)
   --dump-output-on-crash
@@ -160,6 +160,9 @@ optional arguments:
                         (may result in very large test reports on failures)
 
 ```
+
+Argument `--datadir` may be used multiple times to override paths for multiple
+nodes. Example: `run_test.py --datadir requestor /tmp/foo --datadir provider /tmp/bar`.
 
 ### Running with pytest
 
--- a/scripts/node_integration_tests/playbooks/base.py
+++ b/scripts/node_integration_tests/playbooks/base.py
@@ -1,3 +1,4 @@
+from functools import partial
 from pathlib import Path
 import re
 import sys
@@ -15,32 +16,36 @@ from scripts.node_integration_tests import helpers, tasks
 
 from golem.rpc.cert import CertificateError
 
+from .test_config_base import NodeId
+
 if typing.TYPE_CHECKING:
+    from queue import Queue
+    from subprocess import Popen
     from .test_config_base import TestConfigBase, NodeConfig
 
 
 _sslverify.platformTrust = lambda: None
 
 
+def print_result(result):
+    print(f"Result: {result}")
+
+
+def print_error(error):
+    print(f"Error: {error}")
+
+
 class NodeTestPlaybook:
     INTERVAL = 1
 
-    start_time = None
+    start_time: float
 
-    _loop = None
+    _loop: task.LoopingCall
 
     nodes_root: typing.Optional[Path] = None
-    provider_node = None
-    requestor_node = None
-    provider_output_queue = None
-    requestor_output_queue = None
-
-    provider_port = None
 
     exit_code = None
     current_step = 0
-    provider_key = None
-    requestor_key = None
     known_tasks = None
     task_id = None
     started = False
@@ -53,10 +58,6 @@ class NodeTestPlaybook:
     reconnect_countdown_initial = 10
     reconnect_countdown = None
 
-    playbook_description = 'Runs a golem node integration test'
-
-    node_restart_count = 0
-
     @property
     def task_settings_dict(self) -> dict:
         return tasks.get_settings(self.config.task_settings)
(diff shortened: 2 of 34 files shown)


scripts/node_integration_tests/params.py
```diff
@@ -1,11 +1,12 @@
+from copy import copy
```
etam: Don't pay much attention to this file. It will be deleted.

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -1,645 +0,0 @@
-from pathlib import Path
```
etam: This file was renamed to `playbook_base.py`. It was so much changed, that github doesn't count this as a rename and edit, but deletion and addition of files.

scripts/node_integration_tests/playbooks/golem/restart_frame.py
```diff
@@ -1,41 +1,69 @@
...
+    from ..playbook_base import NodeTestPlaybook
+    from ..test_base import Config
+
+
+class RestartFrame(DebugTest):
+    @staticmethod
+    def get_config() -> 'Config':
+        config = DebugTest.get_config()
+        requestor_config = config.nodes[NodeId.requestor]
+        requestor_config_2 = deepcopy(requestor_config)
```
etam: I'm not quite happy with how it turned out. See `test_base.py` for more comments.

scripts/node_integration_tests/playbooks/test_base.py
```diff
@@ -0,0 +1,153 @@
...
+    def get_config() -> Config:
+        concent = os.environ.get('GOLEM_CONCENT_VARIANT', 'staging')
+        return Config(
+            nodes={
+                NodeId.requestor: NodeConfig(
+                    concent=concent,
+                    password=os.environ.get(
+                        'GOLEM_REQUESTOR_PASSWORD', 'dupa.8'),
+                    rpc_port=int(os.environ.get(
+                        'GOLEM_REQUESTOR_RPC_PORT', '61000')),
```
etam: Where are those env variables used?

I'm not quite happy with this, because it discourages creating `NodeConfig` instances in tests and encourages editing those created here.

scripts/node_integration_tests/helpers.py
```diff
@@ -50,19 +51,25 @@ def gracefully_shutdown(process: subprocess.Popen, node_type: str):
     print("%s shut down correctly." % node_type)
 
 
-def run_golem_node(node_type: str, *args,
-                   nodes_root: typing.Optional[pathlib.Path] = None):
+def _params_from_dict(d: typing.Dict[str, typing.Any]) -> typing.List[str]:
```
shadeofblue: @etam couldn't we have used `params_from_dict` from `node_integration_tests.params`?

scripts/node_integration_tests/helpers.py
```diff
@@ -50,19 +51,25 @@ def gracefully_shutdown(process: subprocess.Popen, node_type: str):
...
+    return list(
+        itertools.chain.from_iterable(
+            [k, str(v)] if v is not None else [k] for k, v in d.items()
+        )
+    )
+
+
+def run_golem_node(node_type: str, args: typing.Dict[str, typing.Any]) \
```
shadeofblue: @etam why exactly have we gotten rid of the `nodes_root` parameter?

scripts/node_integration_tests/params.py
shadeofblue: @etam could you elaborate on why?

scripts/node_integration_tests/playbooks/base.py
shadeofblue: @etam still, why rename the file for the sake of renaming? even if github treats that the same?

scripts/node_integration_tests/playbooks/golem/regular_run_stop_on_reject.py
```diff
@@ -1,43 +1,45 @@
...
-        NodeTestPlaybook.step_get_subtasks,
-        NodeTestPlaybook.step_verify_provider_income,
-    )
+from ..test_base import DebugTest, NodeId
+
+if TYPE_CHECKING:
+    from ..playbook_base import NodeTestPlaybook
+
+
+class RegularRun(DebugTest):
```
shadeofblue: @etam could you elaborate on exact reason for such a deep rewrite? plus, haven't you considered discussing any rewrites of similar magnitude before actually emarking on such a quest? ... I know sometimes almost complete rewrites are needed but it's not okay if they're done with no prior discussion et al .. especially that this change WILL cause conficts with branches containing test not migrated to the new structure

scripts/node_integration_tests/playbooks/test_base.py
shadeofblue: @etam I'm unsure what `NodeConfig` is but those tests are designed so they can work with existing golem installations and those env variables allow that

shadeofblue:
I'm generally unconvinced that such a deep rewrite was indeed _required_ to satisfy the original premise... and especially that a complete, conflict-prone and backwards-incompatible refactoring of each and every playbook is justified for the sake of a single test...

shadeofblue:
plus @etam if you really think that adding type hints for the sake of adding type hints is indeed needed, please do so in a _separate_ pull request ...

etam:
we agreed to make as separate PRs:
- excessive type annotations
- separate config from internal state of NodeTestPlaybook and convert `params.py` into config.

and restore `nodes_root`

etam:
Ideas to explore:
- Rename `Test` to `TestConfig` and merge it with `Config`.
- Keep playbook and config in separate files. If test has only config, then it could be a single file. Otherwise they should be in a subfolder.

etam:
Separation of config from playbook: #4088

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -108,158 +106,105 @@ def previous(self):
...
-        if requestor_config is None:
-            self.fail("requestor node config is not defined")
-            return
+        return self.call(node_id, 'net.ident.key',
+                         on_success=on_success, on_error=on_error)
 
-        if not requestor_config.opts:
+    def step_configure_node(self, node_id: NodeId):
+        opts = self.config.current_nodes[node_id].opts
+        if not opts:
```
kmazurek, line 146: I think it could be worth printing a message about the lack of options in this case.

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -108,158 +106,105 @@ def previous(self):
...
-                              ("localhost", self.provider_port, ),
-                              on_success=on_success)
+        return self.call(node_id, 'net.peer.connect',
+                         ("localhost", self.nodes_ports[target_node]),
+                         on_success=on_success)
 
-    def step_verify_peer_connection(self):
+    def step_verify_peer_connection(self, node_id: NodeId, target_node: NodeId):
         def on_success(result):
             if len(result) > 1:
```
kmazurek: Does this still apply in the case of multinode tests?

scripts/node_integration_tests/playbooks/base.py
etam: Maybe? The lack of options is a default value, so it would be an unnecessary noise. If you have options, you get a success or failure message. You know what your `opts` are, so you know what to expect.

scripts/node_integration_tests/playbooks/base.py
etam: No. Apparently I need to create a simple multinode test and make sure it works.

kmazurek: Got it.

scripts/node_integration_tests/playbooks/base.py
kmazurek: Right, I missed the fact it's the default. Never mind then, let's leave that as it is.

scripts/node_integration_tests/README.md
```diff
@@ -25,8 +25,9 @@ behavior - e.g. a requestor node that always fails the submitted results.
 This allows us to simulate various scenarios of two nodes interacting with
 each other and verifying they do indeed react the way we intend them to.
 
-We could theoretically create tests that employ more than two golem instances
-but we don't do that yet and it's not yet supported by the test suite. 
+In most cases there are two nodes (provider and requestor), and this is a
+default configuration for tests. But it is possible to define any number of
```
shadeofblue:
```suggestion
default configuration for tests. If needed, it is possible to define any number of
```

scripts/node_integration_tests/run_test.py
```diff
@@ -1,22 +1,27 @@
...
-from scripts.node_integration_tests.playbooks.base import NodeTestPlaybook
+from scripts.node_integration_tests.playbooks.test_config_base import NodeId
 
 if TYPE_CHECKING:
     from scripts.node_integration_tests.playbooks.test_config_base \
-        import NodeConfig, TestConfigBase
+        import TestConfigBase
+
+
+class DictAction(argparse.Action):
```
shadeofblue, line 17: comment required on how this works as it's not obvious at first glance

scripts/node_integration_tests/run_test.py
```diff
@@ -35,12 +40,11 @@ def parse_args() -> argparse.Namespace:
         help='the task settings set to use, see `tasks.__init__.py`'
     )
     parser.add_argument(
-        '--provider-datadir',
-        help="the provider node's datadir",
-    )
-    parser.add_argument(
-        '--requestor-datadir',
-        help="the requestor node's datadir",
+        '--datadir',
```
shadeofblue, line 63: what's the `--help` output here?

etam:
```
  --datadir NODE PATH   override datadir path for given node
```

scripts/node_integration_tests/run_test.py
etam: done

scripts/node_integration_tests/run_test.py
etam: I've also updated README

scripts/node_integration_tests/playbooks/golem/multinode_regular_run/playbook.py
```diff
@@ -0,0 +1,79 @@
...
+
+        if unpaid:
+            print("Found subtasks with no matching payments: %s" % unpaid)
+            self.fail()
+            return
+
+        print("All subtasks accounted for.")
+        self.next()
+
+    initial_steps: typing.Tuple = (
```
kmazurek, line 37: Idea: change playbook steps to a list instead of a tuple to allow for generating repetitive steps (e.g. common initial steps for tests of a certain type).

kmazurek (approved):
nod

scripts/node_integration_tests/playbooks/golem/multinode_regular_run/playbook.py
etam: I like the idea. I'll make a separate PR, once this is merged.

kmazurek (approved):

scripts/node_integration_tests/README.md
```diff
@@ -127,11 +132,8 @@ example:
 full usage:
 
 ```
-run_test.py [-h] [--task-package TASK_PACKAGE]
-            [--task-settings TASK_SETTINGS]
-            [--provider-datadir PROVIDER_DATADIR]
-            [--requestor-datadir REQUESTOR_DATADIR] [--mainnet]
-            test_path
+run_test.py [-h] [--task-package TASK_PACKAGE] [--task-settings TASK_SETTINGS]
+            [--datadir NODE PATH] [--mainnet] test_path
```
shadeofblue, line 136: @etam please provide a usage example for a test that uses more than one node so that it's more clear of how exactly the roles and their datadirs should be specified

scripts/node_integration_tests/README.md
```diff
@@ -145,10 +147,7 @@ optional arguments:
                         a directory within `tasks` containing the task package
   --task-settings TASK_SETTINGS
                         the task settings set to use, see `tasks.__init__.py`
-  --provider-datadir PROVIDER_DATADIR
-                        the provider node's datadir
-  --requestor-datadir REQUESTOR_DATADIR
-                        the requestor node's datadir
+  --datadir NODE PATH   override datadir path for given node
```
shadeofblue: maybe we should mention the "standard" node names here?

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -74,192 +75,137 @@ def current_step_method(self):
 
     @property
     def current_step_name(self) -> str:
-        method = self.current_step_method
-        return method.__name__ if method else ''
+        return repr(self.current_step_method)
```
shadeofblue: hmm - won't it result in:

```
<bound method Playbook.some_method of <__main__.Playbook object at 0x7f4ef7ac8b38>>
```

?

shadeofblue: or alternatively - what's the motivation behind this change?

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -74,192 +75,137 @@ def current_step_method(self):
...
+        for node_id, output_queue in self.output_queues.items():
+            if self.config.dump_output_on_fail or (
+                    self.config.dump_output_on_crash
+                    and self.nodes_exit_codes[node_id] is not None):
+                helpers.print_output(output_queue, node_id.value + ' ')
 
         self.stop(1)
 
-    def success(self):
+    def _success(self):
```
shadeofblue, line 97: why suddenly hidden?

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -74,192 +75,137 @@ def current_step_method(self):
...
+    def step_wait_node_gnt(self, node_id: NodeId):
         def on_success(result):
-            return self._wait_gnt_eth('requestor', result)
-
-        return self.call_requestor('pay.balance', on_success=on_success)
+            return self._wait_gnt_eth(node_id, result)
+        return self.call(node_id, 'pay.balance', on_success=on_success)
 
-    def step_get_provider_key(self):
+    def step_get_node_key(self, node_id: NodeId):
```
shadeofblue: @etam I guess, generally, we could just remove `_node` from those step names as, usually, they all refer to some activity against one node or another...

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -74,192 +75,137 @@ def current_step_method(self):
...
+    def step_get_node_network_info(self, node_id: NodeId):
         def on_success(result):
             if result.get('listening') and result.get('port_statuses'):
-                provider_ports = list(result.get('port_statuses').keys())
-                self.provider_port = provider_ports[0]
-                print("Provider's port: {} (all: {})".format(
-                    self.provider_port, provider_ports))
+                port = list(result.get('port_statuses').keys())[0]
+                self.nodes_ports[node_id] = port
+                print(f"{node_id.value}'s port: {port}")
```
shadeofblue: @etam why have we removed all the other ports?

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -74,192 +75,137 @@ def current_step_method(self):
...
-                print("Waiting for Provider's network info...")
+                print(f"Waiting for {node_id.value}'s network info...")
                 time.sleep(3)
 
-        return self.call_provider('net.status',
-                             on_success=on_success, on_error=self.print_error)
+        return self.call(node_id, 'net.status', on_success=on_success)
 
-    def step_ensure_requestor_network(self):
+    def step_ensure_node_network(self, node_id: NodeId):
```
shadeofblue: hmm... @etam how is this method different than `step_get_node_network_info` after refactoring?

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -74,192 +75,137 @@ def current_step_method(self):
...
                 return
-            elif len(result) == 1:
-                peer = result[0]
-                if peer['key_id'] != self.provider_key:
-                    print("Connected peer: {} != provider peer: {}",
-                          peer.key, self.provider_key)
+            elif len(result) > 0:
+                expected_peer_key = self.nodes_keys[target_node]
+                if expected_peer_key not in [peer['key_id'] for peer in result]:
+                    print(f"Expected peer: {expected_peer_key} not in connected"
```
shadeofblue: I think the check (and the message) should rather verify if there aren't any _unexpected_ nodes connected ...

I can imagine a situation when we're waiting for several nodes to connect and they are unable to do that all immediately resulting in some joining right away and some of them after a short delay ... it wouldn't be an issue for the test as a whole if someone was still not connected after the first one succeeded ... but it _would_ be an issue if a completely unexpected node from outside the test network connected here...

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -276,19 +222,17 @@ def on_success(result):
                     print("Waiting for nodes to sync...")
                     time.sleep(10)
 
-        return self.call_requestor('net.peers.connected',
-                              on_success=on_success, on_error=self.print_error)
+        return self.call(node_id, 'net.peers.connected', on_success=on_success)
 
-    def step_get_known_tasks(self):
+    def step_get_known_tasks(self, node_id: NodeId):
```
shadeofblue: actually, here, when in almost _all_ cases this will be used against a requestor, I think the default for `node_id` should be `NodeId.requestor` and then, the step could be defined without `partial` in the list below and only ever overridden if indeed it needs to be called for a different node

scripts/node_integration_tests/playbooks/base.py
etam: The motivation is that `partial` object has no `__name__`

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -276,19 +222,17 @@ def on_success(result):
...
-            print("Got current tasks list from the requestor.")
+            print(f"Got current tasks list from the {node_id.value}.")
             self.next()
 
-        return self.call_requestor('comp.tasks',
-                              on_success=on_success, on_error=self.print_error)
+        return self.call(node_id, 'comp.tasks', on_success=on_success)
 
-    def step_create_task(self):
+    def step_create_task(self, node_id: NodeId):
```
shadeofblue: same here:

```suggestion
    def step_create_task(self, node_id: NodeId = NodeId.requestor):
```

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -309,12 +253,10 @@ def on_success(result):
...
             self.task_in_creation = True
-            return self.call_requestor('comp.task.create',
-                                       self.config.task_dict,
-                                  on_success=on_success,
-                                  on_error=self.print_error)
+            return self.call(node_id, 'comp.task.create', self.config.task_dict,
+                             on_success=on_success)
 
-    def step_get_task_id(self):
+    def step_get_task_id(self, node_id: NodeId):
```
shadeofblue: and here

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -327,18 +269,17 @@ def on_success(result):
                 print("Task id: {}".format(self.task_id))
                 self.next()
 
-        return self.call_requestor('comp.tasks',
-                              on_success=on_success, on_error=self.print_error)
+        return self.call(node_id, 'comp.tasks', on_success=on_success)
 
-    def step_get_task_status(self):
+    def step_get_task_status(self, node_id: NodeId):
```
shadeofblue: ditto

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -327,18 +269,17 @@ def on_success(result):
...
             print("Task status: {}".format(result['status']))
             self.next()
 
-        return self.call_requestor('comp.task', self.task_id,
-                              on_success=on_success, on_error=self.print_error)
+        return self.call(node_id, 'comp.task', self.task_id,
+                         on_success=on_success)
 
-    def step_wait_task_finished(self):
+    def step_wait_task_finished(self, node_id: NodeId):
```
shadeofblue: +

scripts/node_integration_tests/playbooks/base.py
etam: for non-partial method the output is:
```
<function Foo.foo at 0x7fa5b44aaa60>
```
and for partial:
```
functools.partial(<function Foo.foo at 0x7fa5b44aaa60>, i=5)
```

scripts/node_integration_tests/playbooks/concent/concent_base.py
```diff
@@ -1,79 +1,55 @@
...
-    def step_enable_provider_concent(self):
-        return self._step_enable_concent('provider')
-
-    def step_enable_requestor_concent(self):
-        return self._step_enable_concent('requestor')
-
-    def _step_ensure_concent_on(self, role):
-        call_method = getattr(self, 'call_' + role)
-
+    def step_ensure_node_concent_on(self, node_id: NodeId):
```
shadeofblue: again, I think `_node` is superfluous

scripts/node_integration_tests/playbooks/golem/multinode_regular_run/playbook.py
shadeofblue: pls, don't do that - I believe it is a good part of those playbooks readability that those steps are rather explicit

scripts/node_integration_tests/playbooks/base.py
etam: I made a separate commit for this: 02052aafa86f881ce00e159e9801045216db1700

Observation 1: All tests are composed of steps. If all steps are successful, then the whole test is successful.
Observation 2: When writing tests it caught me by surprise, that the tests was ending at one step, that wasn't the last one. It turned out that it was calling `success`, what was terminating the test.

Conclusion: I changed the logic, that every step calls `next`. If a step was the last one, `_success` is called from inside `next`.

shadeofblue (changes requested):
comments added

scripts/node_integration_tests/playbooks/base.py
shadeofblue: @etam ugh... and no better way to present that so that it outputs just `foo` ?

scripts/node_integration_tests/playbooks/base.py
shadeofblue: :+1: 

ah, so yeah, we need to remove the "ran out of steps" check in `run`

etam: done

scripts/node_integration_tests/playbooks/base.py
etam: Which other ports?

scripts/node_integration_tests/playbooks/base.py
etam: The same as before refactoring. `step_get_node_network_info` saves port number in `self.nodes_ports`. `step_ensure_node_network` just checks if any port is returned.

So yes, they are very similar and `step_ensure_node_network` could be deleted.

scripts/node_integration_tests/playbooks/base.py
etam: :+1:

scripts/node_integration_tests/README.md
etam: done

scripts/node_integration_tests/README.md
etam: done

scripts/node_integration_tests/playbooks/base.py
shadeofblue: okay, maybe ... 

```
method = self.current_step_method
if hasattr(method, 'func'):
    method = getattr(method, 'func')
return getattr(method, '__name__', '')
```

@etam ^

scripts/node_integration_tests/playbooks/base.py
shadeofblue: @etam ??? ... those that you removed from the previous output

scripts/node_integration_tests/playbooks/base.py
shadeofblue: @etam so yeah, please unify them

scripts/node_integration_tests/playbooks/base.py
shadeofblue: @etam ^ ?

scripts/node_integration_tests/playbooks/base.py
shadeofblue: @etam ^

scripts/node_integration_tests/README.md
```diff
@@ -161,6 +161,9 @@ optional arguments:
 
 ```
 
+Argument `--datadir` may be used multiple times to override paths for multiple
+nodes. Example: `run_test.py --datadir requestor /tmp/foo --datadir provider /tmp/bar`.
```
shadeofblue, line 165: :+1:

scripts/node_integration_tests/playbooks/concent/concent_base.py
shadeofblue: @etam ^

scripts/node_integration_tests/playbooks/base.py
etam: https://www.youtube.com/embed/f5d8pVg3Qtg?start=215&end=232

shadeofblue: @etam I really have no way to better ask than to point to that line before the change: https://github.com/golemfactory/golem/pull/4058/files#diff-cb2db76beee60d6f1a9a0c95e70f717cL216

scripts/node_integration_tests/playbooks/base.py
etam: done

scripts/node_integration_tests/playbooks/base.py
etam: done

scripts/node_integration_tests/playbooks/base.py
etam: done

scripts/node_integration_tests/playbooks/concent/concent_base.py
etam: done

scripts/node_integration_tests/playbooks/base.py
etam: how about now?
For my examples above the outputs now are:
```
foo
foo(i=5)
```

scripts/node_integration_tests/playbooks/base.py
etam: Good catch. Actually I totally borked the logic in this step.

etam: done

scripts/node_integration_tests/playbooks/golem/multinode_regular_run/playbook.py
etam: I did this and it looks like this:
```python
    initial_steps: typing.Tuple = \
        tuple(helpers.map_partial(step_get_node_key,
                                  node_id=[NodeId.provider, NodeId.requestor]))
    + tuple(helpers.map_partial(step_configure_node,
                                node_id=[NodeId.provider, NodeId.requestor]))
    + (
        partial(step_get_node_network_info, node_id=NodeId.provider),
        partial(step_ensure_node_network, node_id=NodeId.requestor),
        partial(step_connect_nodes, node_id=NodeId.requestor,
                target_node=NodeId.provider),
        partial(step_verify_peer_connection, node_id=NodeId.requestor,
                target_node=NodeId.provider),
        partial(step_wait_node_gnt, node_id=NodeId.requestor),
        partial(step_get_known_tasks, node_id=NodeId.requestor),
    )
```
so I think I agree with blue that this is not better.

shadeofblue (approved):

kmazurek (approved):

etam:
Yay! Now I'm gonna cleanup git history and rebase onto develop after #4160 is merged.

PR #4067, opened 2019-04-01
Use the correct option for passing audio bitrate to ffmpeg

cameel:
This trivial change fixes the name of the option used to pass audio bitrate fo ffmpeg. The right option is `-b:a` or `-ab`. `-c:a` is for selecting the audio codec.

There is currently no mechanism in Golem for unit testing the code running in Docker containers so I could not add a failing test case for this. I tried to do this indirectly - by adding audio bitrate value to one of the test cases in `test_ffmpegintegration.py` but they use video files without audio tracks and it seems that ffmpeg silently ignores invalid `-c:a` values and does not fail in that case.

So I can only show that using this option manually in a ffmpeg command produces an error:
``` bash
ffmpeg -i ForBiggerBlazes-\[codec\=h264\].mp4 -c:v libx265 -r 25 -c:a aac -c:a 128k -vf scale=320:240 out.mp4
```
```
Unknown encoder '128k'
```

The test file is [ForBiggerBlazes-[codec=h264].mp4](https://github.com/golemfactory/transcoding-experiments/blob/master/tests/videos/different-codecs/ForBiggerBlazes-[codec%3Dh264].mp4).

### DIFF
--- a/apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_commands.py
+++ b/apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_commands.py
@@ -127,7 +127,7 @@ def transcode_video_command(track, output_playlist_name, targs, use_playlist):
         pass
     try:
         abitrate = targs['audio']['bitrate']
-        cmd.append("-c:a")
+        cmd.append("-b:a")
         cmd.append(abitrate)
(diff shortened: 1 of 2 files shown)


nieznanysprawiciel (approved):

bidzyyys (approved):
:+1:

PR #4068, opened 2019-04-01
Fix ffmpeg integration tests splitting file twice in setup

cameel:
`TestffmpegTask.setUp()` prepares an ffmpeg task for use in individual tests. This involves splitting the input video into segments. Not all tests in `TestffmpegTask` use it though. Some of them just ignore it and build another task for themselves - all in the same temporary directory. This means that video is being split into segments again and segments and/or playlists may be overwritten.

This does not fail on `CGI/transcoding/master` because the split command used in the ffmpeg image right now seems happy to overwrite existing segments without asking when it's being told to use a M3U8 playlist as output. It does fail on my new branch where I split the video with a different method.

Even if it did not fail, it's unnecessary an makes the tests slower. My change extracts the part that builds the task into a function and makes only tests that need it call it. It also removes one of the tests which does not do anything beyond building a task.

### DIFF
--- a/tests/apps/ffmpeg/task/test_ffmpegtask.py
+++ b/tests/apps/ffmpeg/task/test_ffmpegtask.py
@@ -152,13 +152,6 @@ class TestffmpegTask(TempDirFixture):
 
         self.assertEqual(td.output_file, '/tmp/test task.mp4')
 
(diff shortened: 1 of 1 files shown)


bidzyyys (approved):
:+1:

PR #4069, opened 2019-04-01
Miscellaneous tweaks and enhancements in ffmpeg task

cameel:
This pull request contains several small tweaks that aren't big or important enough to warrant putting in a pull request of their own. The changes include:
- The ffmpeg image had parts of code wrapped in bare `except:`, catching all exceptions (even ones like `SyntaxError` or `SystemExit`). I think the purpose was to ignore `KeyError`s due to missing values in the `params` dictionary. I've already run into problems with this while trying to debug the parameters. I've replaced it with simple `if`s so that legitimate errors are not suppressed.
- Adding `-nostdin` to all ffmpeg commands to prevent it from processing input from TTY. This may happen in multiple situations:
    - ffmpeg asking the user if its OK to overwrite a file.
    - ffmpeg runs in an interactive mode by default and can be controlled by pressing various keys. I can't see any reason why it would be good to have it enabled when running in Docker. `-nostdin` disables it.
- Minor wording/wrapping changes or removing duplicate lines.

### DIFF
--- a/apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_commands.py
+++ b/apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_commands.py
@@ -102,49 +102,41 @@ def transcode_video_command(track, output_playlist_name, targs, use_playlist):
         cmd.extend(playlist_cmd)
 
     # video settings
-    try:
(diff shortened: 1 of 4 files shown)


bidzyyys (commented):
:+1:

PR #4078, opened 2019-04-03
Always use long Windows paths for shared directories

Wiezzel:
Name of the share is a hash of the shared directory's path. Any non-canonical forms of the path would break file sharing because they would have a different hash. Therefore we make sure to standardize the
path before computing the hash function.
This commit improves the path standardization by introducing expansion of shortened paths (a.k.a. 8.3 paths, DOS paths). Such paths have been observed during Golem installation as "Jack Daniel's" user.

### DIFF
--- a/golem/docker/smbshare.py
+++ b/golem/docker/smbshare.py
@@ -2,8 +2,6 @@ import binascii
 import hashlib
 from os import path, makedirs
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


shadeofblue (approved):
ba90134f42960c6092b0d8f97198fcf3

PR #4081, opened 2019-04-03
Fix handling task output directories

### LINKED ISSUE #4074: Running unit tests causes output directories to be created in Golem source dir
## Description

**Golem Version**: 0.19.1

**OS**: Ubuntu 18.04

**Branch**: develop (commit 6197cdd0)

**Description of the issue**:
After successfully running Golem unit tests using `pytest` a number of directories are created in Golem's source directory. These directories are created when building task definitions in the tests (see: `coretask#get_output_path`). Their names follow the convention used in the given source file, e.g.: `name_2019-04-02_17-15-50`.
Here's a list of tests which currently result in output directories being created:
```
tests/apps/blender/task/test_blenderrendertask.py
tests/apps/dummy/task/test_dummytask.py
tests/apps/rendering/resources/test_imgrepr.py
tests/apps/rendering/resources/test_utils.py
tests/apps/rendering/task/test_framerenderingtask.py
tests/apps/rendering/task/test_renderingtask.py
tests/apps/rendering/task/test_renderingtaskstate.py
tests/apps/wasm/test_job.py
tests/apps/wasm/test_task.py
```

## Steps To Reproduce
1. From the Golem source directory, run unit tests for Golem apps, e.g.:
```
pytest tests/apps
```
2. After the tests have successfully run, check the Golem source for newly created directories with times in their names, e.g. `name_2019-04-02_17-15-50`.
3. If one or more such directories are present, you have reproduced the issue.

## Expected behavior
Running unit tests should not create any new directories outside of the the temp directory.

kmazurek:
Fixes #4074

### DIFF
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -267,8 +267,9 @@ class TaskManager(TaskEventListener):
         `/some/output/dir`.
         """
         output_dir = self._get_task_output_dir(task_def)
-        if output_dir:
-            output_dir.mkdir(parents=True, exist_ok=True)
+        if not output_dir:
+            return
+        output_dir.mkdir(parents=True, exist_ok=True)
 
     def _try_remove_task_output_dir(self, task_def: TaskDefinition):
         """
@@ -276,18 +277,20 @@ class TaskManager(TaskEventListener):
         This will only succeed if the directory is empty.
         """
         output_dir = self._get_task_output_dir(task_def)
-        if output_dir:
-            try:
-                output_dir.rmdir()
-            except OSError:
-                pass
+        if not output_dir:
+            return
+
+        try:
+            output_dir.rmdir()
+        except OSError:
+            pass
 
     @staticmethod
     def _get_task_output_dir(task_def: TaskDefinition) -> Optional[Path]:
-        if task_def.output_file:
-            return Path(task_def.output_file).resolve().parent
+        if not task_def.output_file:
+            return None
 
-        return None
+        return Path(task_def.output_file).resolve().parent
 
     @staticmethod
     def _migrate_status_to_enum(state: TaskState) -> None:
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -16,6 +16,7 @@ from twisted.internet.threads import deferToThread
 
 from apps.appsmanager import AppsManager
 from apps.core.task.coretask import CoreTask
+from apps.core.task.coretaskstate import TaskDefinition
 from golem.core.common import get_timestamp_utc, HandleForwardedError, \
     HandleKeyError, node_info_str, short_node_id, to_unicode, update_dict
 from golem.manager.nodestatesnapshot import LocalTaskStateSnapshot
@@ -185,6 +186,8 @@ class TaskManager(TaskEventListener):
         self.tasks_states[task_id] = ts
         logger.info("Task %s added", task_id)
 
+        self._create_task_output_dir(task.task_definition)
+
         self.notice_task_updated(task_id,
                                  op=TaskOp.CREATED,
                                  persist=False)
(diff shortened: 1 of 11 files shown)


etam:
After putting some thought into the problem I came up with this: This change is fine on it's own, but it doesn't solve the problem.

Imagine that you have the sources, but no write permission (for example a buildbot might do it, to ensure that tests are not screwing up anything). Then the tests will fail, because of lack of permission to create those directories.

I think they shouldn't create directories in source code directory.

kmazurek:
After discussing the issue further I've decided this is not a good approach, namely because this could introduce memory leaks due to tasks being kept in memory by `atexit`. I'm going to move the directory creation and cleanup to `taskmanager`.

golem/task/taskmanager.py
```diff
@@ -254,6 +257,38 @@ def remove_dump(self, task_id: str):
...
+        """
+        output_dir = self._get_task_output_dir(task_def)
+        if output_dir:
+            try:
+                output_dir.rmdir()
+            except OSError:
+                pass
+
+    @staticmethod
+    def _get_task_output_dir(task_def: TaskDefinition) -> Optional[Path]:
```
etam, line 298: [Bikeshedding](https://en.wiktionary.org/wiki/bikeshedding):

I prefer
```
if something unusual:
    handle this

do the thing
```

golem/task/taskmanager.py
```diff
@@ -254,6 +257,38 @@ def remove_dump(self, task_id: str):
...
+        """
+        Creates the output directory for a task along with any parents,
+        if necessary. The path is obtained from `output_file` field in the
+        task's definition.
+        For example, for an output file with the following path:
+        `/some/output/dir/result.png` the created directory will be:
+        `/some/output/dir`.
+        """
+        output_dir = self._get_task_output_dir(task_def)
+        if output_dir:
```
etam: continuing my comment from _get_task_output_dir: I'd write `if not output_dir: return`

tests/golem/docker/docker-blender-cycles-task.json
```diff
@@ -60,5 +59,6 @@
       "id": null
     }
   ],
-  "concent_enabled": false
+  "concent_enabled": false,
+  "output_file": "/tmp/output.png"
```
etam: move it back

tests/golem/docker/docker-blender-cycles-task.json
```diff
@@ -43,7 +43,7 @@
     900
   ],
   "output_format":"PNG",
-  "output_file":"hel.png",
+  "output_file": "/tmp/output.png",
```
jiivan: Will it work on windows?

jiivan (approved):

tests/golem/docker/docker-blender-cycles-task.json
kmazurek: This directory is not actually created in any of the tests, but I think you're right, I'll change it back to a single file name.

golem/testutils.py
```diff
@@ -155,7 +155,7 @@ class MyTestCase(unittest.TestCase, testutils.PEP8MixIn):
     in this attribute.
     """
 
-    def test_conformance(self):
+    def test_conformance(self, *_):
```
etam, line 158: why?

jiivan: Probably to handle `@path` on a whole TestCase.

kmazurek: Exactly, that's for handling the patch on `TestTaskManager`.

etam: Well, this shows that this shouldn't be a mix-in, but rather just a class used separately, like in tests/apps/rendering/resources/test_utils.py or tests/golem/test_client.py, because it causes unnecessary dependencies. But that's a topic for another discussion or PR.

tests/golem/task/test_taskmanager.py
```diff
@@ -1284,6 +1284,55 @@ def test_copy_results_error_in_copying(self):
             logger.error.assert_called_once()
             restart.assert_called_once_with('new_subtask_id')
 
+    def test_add_new_task_creates_output_directory(self, *_):
+        task_definition = Mock()
+        task_definition.output_file = '/some/output/file.png'
+        task_mock = self._get_task_mock(task_definition=task_definition)
+
+        output_dir_mock = Mock()
+        with patch('golem.task.taskmanager.TaskManager._get_task_output_dir',
```
etam: You already patch this on whole class. You could get this mock as an method argument and set return value to it.

etam (approved):
ok

PR #4088, opened 2019-04-04
separate config class for integration tests

etam:
This puts integration config into a separate class. This will allow to read the config without loading playbook and it's dependencies like twisted.

This also changes the expected test structure. It's either a single file containing a `TestConfig` class, which must derive from `TestConfigBase`, or a module containing `test_config.TestConfig` class and `playbook.Playbook` class.


yang980 (approved):

scripts/node_integration_tests/playbooks/test_config_base.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}: {self.__dict__}"
+
+
+def make_node_config_from_env(role: str, counter: int) -> NodeConfig:
+    node_config = NodeConfig()
+    node_config.concent = os.environ.get('GOLEM_CONCENT_VARIANT', 'staging')
+    node_config.log_level = 'DEBUG'
+    node_config.password = os.environ.get(f'GOLEM_{role}_PASSWORD', 'dupa.8')
```
shadeofblue: 'dupa.8' is already defined as a default in `NodeConfig` so no need to specify that here ... 
```suggestion
    node_config.password = os.environ.get(f'GOLEM_{role}_PASSWORD', node_config.password)
```

scripts/node_integration_tests/playbooks/test_config_base.py
```diff
@@ -0,0 +1,88 @@
...
+    node_config.rpc_port = \
+        int(os.environ.get(f'GOLEM_{role}_RPC_PORT', 61000 + counter))
```
shadeofblue, line 53: if we're allowing the ports to be potentially different from test to test, we need to output them alongside the data directories currently in use (to e.g. facilitate connections through e.g. `golemcli`)

scripts/node_integration_tests/playbooks/test_config_base.py
```diff
@@ -0,0 +1,88 @@
...
+            args['--mainnet'] = None
+        return args
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}: {self.__dict__}"
+
+
+def make_node_config_from_env(role: str, counter: int) -> NodeConfig:
+    node_config = NodeConfig()
+    node_config.concent = os.environ.get('GOLEM_CONCENT_VARIANT', 'staging')
```
shadeofblue:
```suggestion
    node_config.concent = os.environ.get('GOLEM_CONCENT_VARIANT', node_config.concent)
```

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -185,7 +167,10 @@ def on_error(result):
                               on_success=on_success, on_error=on_error)
 
     def step_configure_provider(self):
-        if not self.provider_opts:
+        provider_config = self.config.current_provider
+        assert provider_config is not None
```
shadeofblue: I'd rather not have naked `assert`s here but rather appropriate `.fail()` calls with informative failure messages ... 

*) where `informative failure messages` meaning that some suggestion is made on how to address the issue

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -197,11 +182,14 @@ def on_error(_):
...
             self.fail()
 
-        return self.call_provider('env.opts.update', self.provider_opts,
+        return self.call_provider('env.opts.update', provider_config.opts,
                                   on_success=on_success, on_error=on_error)
 
     def step_configure_requestor(self):
-        if not self.requestor_opts:
+        requestor_config = self.config.current_requestor
+        assert requestor_config is not None
```
shadeofblue: same here (re: `.fail()` + informative msg)

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -429,26 +418,7 @@ def step_stop_nodes(self):
...
-
-        if self.requestor_enabled:
-            requestor_replacement_script = getattr(
-                self,
-                'requestor_node_script_%s' % (self.node_restart_count + 1),
-                None,
-            )
-            if requestor_replacement_script:
-                self.requestor_node_script = requestor_replacement_script
+        self.config.next_nodes()
```
shadeofblue: `next_nodes` sounds awkward-ish ... but my ideas are not much better ... `replace_nodes`, `bump_nodes`... maybe @zakaprov would have something better? ...

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -502,10 +472,12 @@ def call_requestor(self, method, *args,
                        on_success=lambda x: print(x),
                        on_error=lambda: None,
                        **kwargs):
+        requestor_config = self.config.current_requestor
+        assert requestor_config is not None
```
shadeofblue: same here -> please add an appropriate `.fail()` call with an informative message

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -516,33 +488,36 @@ def call_provider(self, method, *args,
                       on_success=lambda x: print(x),
                       on_error=None,
                       **kwargs):
+        provider_config = self.config.current_provider
+        assert provider_config is not None
```
shadeofblue: same here -> please add an appropriate .fail() call with an informative message

scripts/node_integration_tests/run_test.py
```diff
@@ -1,100 +1,137 @@
...
+             " e.g. golem.regular_run",
     )
     parser.add_argument(
         '--task-package',
-        default='test_task_1',
         help='a directory within `tasks` containing the task package'
     )
     parser.add_argument(
         '--task-settings',
-        required=False,
```
shadeofblue, line 24: "explicit is better than implicit"

shadeofblue (changes requested):

scripts/node_integration_tests/run_test.py
etam: Should I also specify all other arguments, even if they have default value?

scripts/node_integration_tests/run_test.py
```diff
@@ -1,100 +1,137 @@
...
-    playbook_kwargs = {
-        k: v
-        for k, v in vars(args).items()
+
+def override_config(config: 'TestConfigBase', args: argparse.Namespace) -> None:
+    for k, v in vars(args).items():
+        if v is None:
+            continue
+
+        def override_datadir(
```
shadeofblue: why introduce an inner function when it doesn't make use of the outside context? separate functions are usually more readable in such a case...

shadeofblue: or, "simple is better than complex"

shadeofblue: and "readability counts"

shadeofblue: or actually "flat is better than nested" ... choose yours ;)

shadeofblue: bottom line is - please don't create inner functions when just plain functions or methods can do the same job just as well

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -185,7 +167,10 @@ def on_error(result):
                               on_success=on_success, on_error=on_error)
 
     def step_configure_provider(self):
-        if not self.provider_opts:
+        provider_config = self.config.current_provider
+        assert provider_config is not None
+
+        if not provider_config.opts:
```
kmazurek, line 175: Bikeshedding: I know this comes from the name of the RPC (`env.opts.update`), but I still find the field name `opts` unnecessarily cryptic. IMO even `options` or `settings` (while still generic) would be more readable. Your thoughts on this?

scripts/node_integration_tests/run_test.py
etam: https://docs.python.org/3.6/library/argparse.html#required says:
> In general, the argparse module assumes that flags like -f and --bar indicate optional arguments, which can always be omitted at the command line. To make an option required, True can be specified for the required= keyword argument to add_argument()

> Required options are generally considered bad form because users expect options to be optional, and thus they should be avoided when possible.

My conclusion: adding `required=False` is stating the obvious and should be avoided.

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -516,33 +488,36 @@ def call_provider(self, method, *args,
...
         )
 
     def start_nodes(self):
-        print("Provider data directory: %s" % self.provider_datadir)
-        print("Requestor data directory: %s" % self.requestor_datadir)
-
-        if self.provider_enabled:
+        provider_config = self.config.current_provider
+        if provider_config is not None:
+            print("Provider data directory: %s" % provider_config.datadir)
```
kmazurek: Idea: perhaps we should pretty-print all the args (or the config) when starting a node?

scripts/node_integration_tests/playbooks/test_config_base.py
kmazurek: I bet repeating it is about making a statement.

scripts/node_integration_tests/playbooks/base.py
kmazurek: `move_to_next_nodes`? Verbose, but it would be good for the name to imply the potential change.

scripts/node_integration_tests/run_test.py
shadeofblue: okay...

scripts/node_integration_tests/playbooks/base.py
etam: thoughts

scripts/node_integration_tests/playbooks/test_config_base.py
etam: done

scripts/node_integration_tests/playbooks/base.py
etam: done

scripts/node_integration_tests/run_test.py
etam: done

scripts/node_integration_tests/playbooks/base.py
etam: done

scripts/node_integration_tests/playbooks/base.py
etam: My thoughts are: I see no big difference between `opts`, `options` or `settings`. As you pointed out, `opts` refers to rpc method name, which distinguishes it from other forms of options, like those passed as arguments to `golemapp`.

scripts/node_integration_tests/playbooks/base.py
etam: `use_next_nodes`?

scripts/node_integration_tests/playbooks/base.py
kmazurek: Okay, fair enough. Let's keep this as `opts` then.

scripts/node_integration_tests/playbooks/base.py
kmazurek: Ok, sounds good to me, @shadeofblue?

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -185,7 +167,12 @@ def on_error(result):
                               on_success=on_success, on_error=on_error)
 
     def step_configure_provider(self):
-        if not self.provider_opts:
+        provider_config = self.config.current_provider
+        if provider_config is None:
+            self.fail("provider is not defined")
```
shadeofblue:
```suggestion
            self.fail("provider node config is not defined")
```

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -197,11 +184,16 @@ def on_error(_):
...
 
-        return self.call_provider('env.opts.update', self.provider_opts,
+        return self.call_provider('env.opts.update', provider_config.opts,
                                   on_success=on_success, on_error=on_error)
 
     def step_configure_requestor(self):
-        if not self.requestor_opts:
+        requestor_config = self.config.current_requestor
+        if requestor_config is None:
+            self.fail("requestor is not defined")
```
shadeofblue:
```suggestion
            self.fail("requestor node config is not defined")
```

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -502,10 +476,15 @@ def call_requestor(self, method, *args,
                        on_success=lambda x: print(x),
                        on_error=lambda: None,
                        **kwargs):
+        requestor_config = self.config.current_requestor
+        if requestor_config is None:
+            self.fail("requestor is not defined")
```
shadeofblue:
```suggestion
            self.fail("requestor node config is not defined")
```

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -516,33 +495,39 @@ def call_provider(self, method, *args,
                       on_success=lambda x: print(x),
                       on_error=None,
                       **kwargs):
+        provider_config = self.config.current_provider
+        if provider_config is None:
+            self.fail("provider is not defined")
```
shadeofblue:
```suggestion
            self.fail("provider node config is not defined")
```

shadeofblue (approved):

kmazurek (approved):
lgtm

PR #4089, opened 2019-04-04
Fix linter breakage and warning filtering

cameel:
The current implementation of `lintdiff.sh` is broken in several ways:
1) If fails if there are too many differences between your branch and `develop`.
2) It silently ignores failures. E.g. when `grep` returns an error because it got too many patterns as input, it does not exit with a non-zero status. This way `lint.sh` assumes that there are no new warnings even if there are.
3) It does not filter the warnings correctly. It can sometimes include warnings from lines that were not modified in your branch.
4) It may parse the diff incorrectly if your code contains `---`, `+++` or `@@` sequences.
5) It does not report any warnings if the whole file is moved.

This pull request fixes problems 1-4. I'm not sure if 5 is a real issue in practice so I left it as is. See commit descriptions for details of each individual fix (I strongly recommend reviewing them individually rather than looking at the combined result). I have also added comments explaining the regexes in `diff-lines()` to make them easier to understand and decrease the likelihood of future bugs.

### Reproducing the problems
If you want to see the problem, just check out `CGI/transcoding/master` and run the linter:
``` bash
git checkout CGI/transcoding/master
./lint.sh`
```
Linting will end with nothing found even though there are tons of warning if you run `pylint` or `mypy` manually. That's because the `grep` call used to select only warnings on modified lines gets too many patterns and fails. You can even see the error in the output if you run `./lint.sh -v` and try to rerun the `lintdiff.sh` command it prints by yourself.

Rebase my branch locally on top of `CGI/transcoding/master` and run the linter again:
``` bash
git checkout fix-linter-breakage-and-warning-filtering
git checkout -b tmp
git rebase --onto CGI/transcoding/master origin/develop
./lint.sh
```

Now you can see all the warnings that used to be suppressed.

### DIFF
--- a/lintdiff.sh
+++ b/lintdiff.sh
@@ -60,7 +60,7 @@ cleanup_artifacts() {
     git checkout "$CURRENT_BRANCH" -- || exit 1
 }
 
-diff-lines() {
+list-added-lines-from-diff() {
     # Function taken from: https://stackoverflow.com/questions/8259851/using-git-diff-how-can-i-get-added-and-modified-lines-numbers
     # Adjusted the 'echo' line to not print the line content.
     local path=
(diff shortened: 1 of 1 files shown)


dybi (approved):
LGTM

maaktweluit (approved):
LGTM!

jiivan (approved):

PR #4094, opened 2019-04-08
Call p2pservice.connect_network() once on start

maaktweluit:
Golem always started with some duplicate peer warnings.

When looking into this i found that `p2pservice.connect_network()` is called twice, causing duplicate peer connections.

Fixed this bug and improved the logging surrounding it.


Wiezzel (approved):
LGTM
By the way `start_network()` method is too damn long.

mfranciszkiewicz (approved):

Issue #4102, opened 2019-04-10
render_entrypoint.py no such file or directory; error 127


ederenn:
## Description

**Golem Version**: GOLEM Version: 0.19.1+dev272.g0f90171

**Golem-Messages version** (leave empty if unsure):  golem_messages Version: 3.3.0

**Electron version** (if used): 0.1.19

**OS** [e.g. Windows 10 Pro]: Windows

**Branch** (if launched from source): develop and any develop based

**Mainnet/Testnet**: Mainnet

**Description of the issue**:

local render test ends in 127 error after uploading a file 

**Actual result**:

it is not possible to run a task

**Screenshots**:

<img width="518" alt="Screen Shot 2019-04-10 at 15 30 15" src="https://user-images.githubusercontent.com/32061891/55882983-bb342900-5ba5-11e9-9aa8-08172de175ba.png">

## Steps To Reproduce
_Short description of steps to reproduce the behavior:_
e.g.
1. Run Golem from source on Windows, develop branch
2. Upload a file by drag and drop
3. Error 127 appears in Gui
4. Check in logs for error

## Expected behavior
It is possible to upload a file and start a task

## Logs and any additional context

```
2019-04-10 15:30:10 INFO     golem.task.rpc                      Running test task "{'name': 'scene-Helicopter-27-cycl', 'resources': ['C:\\Users\\ederenn\\Desktop\\scene-Helicopter-27-cycles.blend'], 'compute_on': 'cpu', 'type': 'Blender', 'subtasks_count': 1}" ...
2019-04-10 15:30:11 WARNING  golem.docker.task_thread            Task error - exit_code=127
stderr:
/usr/local/bin/entrypoint.sh: line 13: python3 /golem/entrypoints/render_entrypoint.py: No such file or directory

tail of stdout:


2019-04-10 15:30:11 WARNING  golem.task.taskthread               Task computing error Subtask computation failed with exit code 127
2019-04-10 15:30:11 WARNING  golem.task                          Task not tested properly Subtask computation failed with exit code 127
2019-04-10 15:30:11 WARNING  golem.task.rpc                      Test task error "{'name': 'scene-Helicopter-27-cycl', 'resources': ['C:\\Users\\ederenn\\Desktop\\scene-Helicopter-27-cycles.blend'], 'compute_on': 'cpu', 'type': 'Blender', 'subtasks_count': 1, 'concent_enabled': False}": ('Subtask computation failed with exit code 127',)
[...]

PR #4113, opened 2019-04-11
DockerJob: Fix container command for Windows

mfranciszkiewicz:
`entrypoint.sh` behaviour is different between Windows and other OSes. The command invocation line `$@` exclusive for Windows behaves differently than `exec su-exec sh -c "$@"` (shortened here) in case of Linux and macOS. Without this patch, starting a container results in an error described in https://github.com/golemfactory/golem/issues/4102.

Because the entrypoint change would require updating all docker images, this solution should work as a temporary fix until an entrypoint change is required.

### DIFF
--- a/golem/docker/job.py
+++ b/golem/docker/job.py
@@ -126,6 +126,8 @@ class DockerJob:
 
         host_cfg = client.create_host_config(**self.host_config)
 
+        # FIXME: Make the entrypoint.sh behaviour consistent between Windows
+        #  and other OSes. See issue #4102
         if is_windows():
             command = self.entrypoint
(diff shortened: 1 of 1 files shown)


maaktweluit (approved):
LGTM!

Wiezzel:
I don't quite understand what the bug is. Also, it should be made clear whether we mean `entrypoint` or `command` here. These are two different things in docker and the code suggests some confusion in that matter.

Wiezzel (approved):

mfranciszkiewicz:
Updated the description to include more information about the bug

PR #4115, opened 2019-04-12
Retrieve samples parameter on 'comp.task' RPC

mdtanrikulu:


shadeofblue (approved):
wolfofwallstreet

shadeofblue:
though @mdtanrikulu please add or modify a test to make sure that this value is covered

tests/apps/blender/task/test_blenderrendertask.py
```diff
@@ -650,7 +650,8 @@ def _task_dictionary(self):
                 "resolution": [
                     320,
                     240
-                ]
+                ],
+                "samples": 0
```
shadeofblue: could I suggest using a value different than 0? @mdtanrikulu

mdtanrikulu: normally there was no samples parameter at all, but sure I updated it.

PR #4119, opened 2019-04-15
Set DNS servers for the Docker daemon on macOS

mfranciszkiewicz:


golem/docker/hypervisor/docker_for_mac.py
```diff
@@ -24,23 +24,33 @@ class DockerForMac(Hypervisor):
         cpu='cpus',
         mem='memoryMiB',
     )
+    DAEMON_CONFIG_FILE = os.path.expanduser(
+        "~/.docker/daemon.json"
+    )
```
Wiezzel, line 29: I encourage using `pathlib.Path`. It supports all methods you need: `expanduser()`, `exists()`, `open()`, `parent`

tests/golem/docker/test_hypervisor.py
```diff
@@ -525,3 +525,30 @@ def test_constrain(self):
...
+
+        with mock.patch.object(hypervisor, 'DAEMON_CONFIG_FILE', config_file):
+            hypervisor._configure_daemon()
+            assert_dns_configured(config_file)
+
+            extra_settings = {'option': ['value1', 'value2']}
+            with open(config_file, 'w') as f:
+                json.dump(extra_settings, f)
+
+            hypervisor._configure_daemon()
```
Wiezzel, line 546: Could you split this into two tests? It's better to have only one call to the tested method per unit test

golem/docker/hypervisor/docker_for_mac.py
```diff
@@ -24,23 +24,33 @@ class DockerForMac(Hypervisor):
...
         else:
+            self._configure_daemon()
             self.start_vm()
 
     @classmethod
     def is_available(cls) -> bool:
         return os.path.exists(cls.COMMAND_HANDLER.APP)
 
-    def create(self, name: Optional[str] = None, **params) -> bool:
+    def create(self, _name: Optional[str] = None, **_params) -> bool:
```
Wiezzel: Maybe it would be better to keep the same argument names as in `Hypervisor` abstract class to have a consistent interface?

Wiezzel (commented):
We don't care about Xhyve, right?

golem/docker/hypervisor/docker_for_mac.py
mfranciszkiewicz: One of the linters complained. I guess they're broken after recent changes.

golem/docker/hypervisor/docker_for_mac.py
mfranciszkiewicz: I can either write in the same style as the existing code or refactor the class to use `Path`. I went with the first option.

mfranciszkiewicz:
@Wiezzel Xhyve should be removed

Wiezzel (approved):

golem/docker/hypervisor/docker_for_mac.py
mfranciszkiewicz: Parent signatures restored

PR #4120, opened 2019-04-15
Created task fragments RPC

### LINKED ISSUE #4041: Task fragments RPC
Add a `comp.task.fragments` RPC

## Parameters
`task_id`

### Output
* List of task fragments - that is the pieces of the task which are the basis for the subtasks (jobs) with meta information describing the piece (e.g. crop coordinates, frame range, etc)
* Each fragment entry should contain the list of subtasks (jobs) that have been scheduled with the same information as provided in the `comp.task.subtasks` RPC

kmazurek:
Resolves: #4041

Adds an RPC endpoint which returns fragments for a given rendering task.
A fragment is a collection of subtasks which all refer to a common part
of the whole task (currently it is the start_task index from subtask
extra_data).


kmazurek:
Sample JSON response from the RPC in its current form (subtask_count is 5, only one subtask has been assigned):
```
{
    "1": [
        {
            "node_id": "da0853e0ec7477edbbf5fdd036093066616c5763f5aa6a670742617dc11f0e325846c53dbccba9f3aada8c6dceecf2acbbe0c44f0bfe2ec87db7cf53553a19a8",
            "node_name": "",
            "progress": 1.0,
            "results": [
                "/home/.../ComputerRes/31d06954-5c6f-11e9-8455-03db94344a05/tmp/small_shark100_10001.png",
                "/home/.../ComputerRes/31d06954-5c6f-11e9-8455-03db94344a05/tmp/small_shark100_10002.png"
            ],
            "status": "Finished",
            "stderr": "/home/.../ComputerRes/31d06954-5c6f-11e9-8455-03db94344a05/tmp/47cb889c-5c6f-11e9-8d2d-03db94344a05/stderr.log",
            "stdout": "/home/.../ComputerRes/31d06954-5c6f-11e9-8455-03db94344a05/tmp/47cb889c-5c6f-11e9-8d2d-03db94344a05/stdout.log",
            "subtask_id": "47cb889c-5c6f-11e9-8d2d-03db94344a05",
            "time_remaining": 0.0,
            "time_started": 1554996863.0686917
        }
    ],
    "2": [],
    "3": [],
    "4": [],
    "5": []
}
```

golem/client.py
```diff
@@ -18,6 +18,7 @@
     Deferred)
 
 from apps.appsmanager import AppsManager
+from apps.rendering.task.renderingtask import RenderingTask
```
jiivan: Instead of importing to already overbloated `client.py` I would like to see this implemented and exposed in `apps.rendering.task.renderingtask`

golem/task/rpc.py
```diff
@@ -5,7 +5,7 @@
 import logging
 import os.path
 import re
-import typing
+from typing import Any, Dict, Generator, Iterable, List, Optional, Tuple
```
shadeofblue: why?... maybe it's just me but I find `typing.X` immediately more readable than just `X` ...

golem/task/rpc.py
```diff
@@ -668,3 +669,21 @@ def get_estimated_cost(self, _task_type: str, options: dict) -> dict:
         }
         logger.info('Estimated task cost. result=%r', result)
         return result
+
+    @rpc_utils.expose('comp.task.rendering.task_fragments')
+    def get_fragments(self, task_id: str) -> Optional[Dict[int, List[Dict]]]:
+        task = self.task_manager.tasks.get(task_id)
+        if task is None or not isinstance(task, RenderingTask):
+            return None
```
shadeofblue: unsure -> maybe it would be better to have it report an error in such a case... although, unsure how errors are expressed through our RPC in general now ...

shadeofblue (approved):
abstract-paper-fracture-detail-lasercut-paper-by-robin-hanhart

golem/task/rpc.py
kmazurek: I find prefixing overly verbose, especially for collection type signatures (when the prefix gets repeated multiple times). I suppose it's mostly subjective, but `from typing` seems to be more widely used in our codebase (113 vs 29 files).

golem/task/rpc.py
kmazurek: Sure, the convention here is to return a tuple with an optional error message, I'll add this to the endpoint.

golem/task/rpc.py
jiivan: I second @shadeofblue in that. IMO this is a case of namespace pollution.

jiivan (commented):
Added remote procedure looks good :+1:

I don't like changing clean namespace into this polluted... thing. :/

```python
>>> import this
The Zen of Python, by Tim Peters
[...]
Namespaces are one honking great idea -- let's do more of those!
```

golem/task/rpc.py
kmazurek: surrender

jiivan (approved):
buakaw_throw

PR #4122, opened 2019-04-18
Cleaned up typing for linter errors in taskkeeper

maaktweluit:

PR #4123, opened 2019-04-18
Task computer "stuck" on subtask

maaktweluit:
My golem node got stuck on a subtask:
```
provider_state: {'status': 'Computing', 'subtask': {'subtask_id': 'SUB-ABC', 'progress': 0.0, 'seconds_to_timeout': -416547.8222870827, 'running_time_seconds': 419350.25729084015, 'outfilebasename': 'test_task', 'output_format': 'PNG', 'scene_file': 'bmw.blend', 'frames': [0], 'start_task': 5, 'total_tasks': 50}, 'environment': None}
```

When investigating i think the header was already deleted by the `check_max_tasks_per_owner()` cleanup, while the task was computing:
```
2019-04-11 12:25:21 INFO     golem.task.taskcomputer             Starting computation of subtask 'SUB-ABC' (task: 'TASK-DEF', deadline: 1554988323, docker images: [{'repository': 'golemfactory/blender', 'image_id': None, 'tag': '1.4'}])
2019-04-11 12:27:31 WARNING  golem.task.taskkeeper               Too many tasks from 839b...9cba, dropping 1 tasks
2019-04-11 12:30:20 ERROR    golem.task.taskcomputer             No subtask with id 'SUB-ABC'
```

This PR solves it 2x:
- Better errors in `task_computer`, no longer return on error but let it clean up like other failures.
- Knowledge in `task_keeper` of what tasks are running, so they are not removed during cleanup.

~Blocked by: #4122 ( lint errors )~

### DIFF
--- a/golem/task/taskkeeper.py
+++ b/golem/task/taskkeeper.py
@@ -538,8 +538,10 @@ class TaskHeaderKeeper:
         # headers, remove the rest
         to_remove = by_age[self.max_tasks_per_requestor:]
 
-        logger.warning("Too many tasks from %s, dropping %d tasks",
-                       owner_key_id, len(to_remove))
+        logger.warning(
+            "Too many tasks, dropping %d tasks. owner=%s, ids_to_remove=%r",
+            len(to_remove), common.short_node_id(owner_key_id), to_remove
+        )
 
         for tid in to_remove:
             self.remove_task_header(tid)
@@ -549,6 +551,11 @@ class TaskHeaderKeeper:
         return: False if task was already removed
         """
         if task_id in self.removed_tasks:
+            return False
+
+        if task_id in self.running_tasks:
+            logger.warning("Can not remove task header, task is running. "
+                           "task_id=%s", task_id)
             return False
 
         try:
--- a/golem/task/taskkeeper.py
+++ b/golem/task/taskkeeper.py
@@ -334,6 +334,8 @@ class TaskHeaderKeeper:
         self.task_headers: typing.Dict[str, dt_tasks.TaskHeader] = {}
         # ids of tasks that this node may try to compute
         self.supported_tasks: typing.List[str] = []
+        # ids of tasks that are computing on this node
+        self.running_tasks = []
         # results of tasks' support checks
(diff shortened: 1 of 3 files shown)


golem/task/taskkeeper.py
```diff
@@ -527,7 +529,9 @@ def check_max_tasks_per_owner(self, owner_key_id):
         if len(owner_task_set) <= self.max_tasks_per_requestor:
             return
 
-        by_age = sorted(owner_task_set,
+        not_running = [x for x in owner_task_set if x not in self.running_tasks]
```
maaktweluit: maybe this needs to be before then `len(..` check..

since now it will not clean up any when there are exactly `max_tasks_per_requestor`

Wiezzel: If you made `self.running_tasks` a set then you could write this nicer:
```python
not_running = set(owner_task_set) - self.running_tasks
```
Also, using set is more appropriate for a collection that doesn't need to be ordered. And gives constant-time remove operation.

Wiezzel (commented):
Apart from the remark about using set instead of list, I would suggest unit-testing this change somehow.

golem/task/taskkeeper.py
maaktweluit: nice one! updated it to be a set

maaktweluit:
Rebased to develop now #4122 is solved

maaktweluit:
@Wiezzel thanks for the review!

Changed list to set and extended the tests, please check again :)

tests/golem/task/test_taskkeeper.py
```diff
@@ -276,6 +276,7 @@ def test_task_limit(frozen_time, self):  # pylint: disable=no-self-argument
         self.assertIn(tb_id, tk.task_headers)
 
     def test_check_max_tasks_per_owner(self):
```
mfranciszkiewicz: This test func got really long. Is it possible to split it?

Wiezzel (approved):
Looks fine

maaktweluit:
Duplicate of #4159

PR #4124, opened 2019-04-18
'TaskServer' object has no attribute 'key_id'

maaktweluit:
self.key_id did not exist, so passed it as parameter


Wiezzel (approved):
Looks fine.

PR #4127, opened 2019-04-23
Issue template for new features

shadeofblue:
it should allow us to have a better insight into what the implemented features are and how they should be tested. Later on, they should alllow us to track the progress of the development and the QA.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -3,7 +3,7 @@ name: Bug report
 about: Please tell us if you encountered an issue with Golem
 title: ''
 labels: brass, bug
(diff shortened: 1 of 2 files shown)


ZmijaWA:
`## Issues encountered during QA . ` - as we spoke... can we change that to `## Issues encountered during QA that are not a blockers`

shadeofblue:
> `## Issues encountered during QA . ` - as we spoke... can we change that to `## Issues encountered during QA that are not a blockers`

@ZmijaWA hmm... that could be even more confusing ... I assume some of those issues _will_ be blockers - but the question is whether they are _introduced_ by this feature or not

badb (approved):

ZmijaWA (approved):

PR #4128, opened 2019-04-24
Task is finished when it's not active

jiivan:


shadeofblue (commented):
add/update the unit test?

shadeofblue (approved):
unit tests

PR #4130, opened 2019-04-24
Fix ffmpeg.task._get_extra_data() truncating chunk names at the first dot

cameel:
If the input file name contains any dots other than the one indicating the extension (e.g. `test.video.mp4`) the names of the video segments used for subtasks are truncated (`test[num=0]_TC.m3u8` instead of `test.video[0]_TC.m3u8`).

This is because `pathlib.Path` expects to get a full file name, not just a fragment. If we give it a fragment without extension, any other dots are interpreted as the start of the extension.

The names that are generated this way don't match expectations in other parts of the code (e.g. in `join_playlists()`) and the transcoding fails.

### DIFF
--- a/apps/transcoding/ffmpeg/task.py
+++ b/apps/transcoding/ffmpeg/task.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import pathlib
 
 from ffmpeg_tools.codecs import VideoCodec, AudioCodec
 from ffmpeg_tools.formats import Container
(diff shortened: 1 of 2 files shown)


cameel:
Rebased on the latest `CGI/transcoding/master`. I see that the test `setUp()` has changed so that `RESOURCE_STREAM` no longer is an instance variable (it's now a class variable) and my test relied on being able to change it. Fortunately it's now possible to pass your own file name to `_build_ffmpeg_task()` so I have updated my test to use that.

Note that one of the tests (`TestffmpegDockerJob.test_ffmpeg_trancoding_job`) fails on this branch but that's because it's broken in `CGI/transcoding/master`. This is not caused by my fix.

nieznanysprawiciel (approved):
This I don't know why, but this test doesn't fail on both transcoding master and your branch.

cameel:
@nieznanysprawiciel 
So it's not failing for you? In that case I took a closer look at it and it's either another bug or the expected workflow has changed.

The failure I'm getting is caused by an exception when the test tries to run a docker container:
```
docker.errors.ImageNotFound: 404 Client Error: Not Found ("No such image: golemfactory/ffmpeg:1.0")
```

This means it can't find `golemfactory/ffmpeg:1.0`.  Since the branches I'm working with often have different versions of the code in the image, I always delete the image and let `setup.py` rebuild it before I run tests:

``` bash
docker rmi golemfactory/ffmpeg:1.0 golemfactory/ffmpeg:latest
./setup.py develop
pytest --tb=short
```

So the problem is that `setup.py` no longer builds the image even though your tests still try to use it. Is that intentional or is it a bug? I see that you have uploaded an image called `ffmpeg-experimental` to DockerHub and now `setup.py` is using that. Maybe the test cases should be updated to use that image too?

PR #4133, opened 2019-04-25
New subtask status "Cancelled"

### LINKED ISSUE #2562: SubtaskState.subtask_rem_time is always 0
It is set to zero:
https://github.com/golemfactory/golem/blob/develop/golem/task/taskstate.py#L52
And updated to zero:
https://github.com/golemfactory/golem/blob/develop/golem/task/taskmanager.py#L465
https://github.com/golemfactory/golem/blob/develop/golem/task/taskmanager.py#L522

But never really updated.

### LINKED ISSUE #3908: New subtask status "Cancelled"
We need a new status for a subtask - `cancelled` - which - for the purpose of task completion and re-assignment should be treated exactly as `restarted` or `failed` with the exception that the node that cancelled the subtask should never be banned.

This status should be assigned to a subtask when the provider node sends an "Offer Cancelled" message within the allowed cancellation window.

jiivan:
fixes: #2562
fixes: #3908 

Also make `SubtaskState` a `Container` and standardized field names with it's _dictionary_ representation.

### DIFF
--- a/golem/task/taskstate.py
+++ b/golem/task/taskstate.py
@@ -1,8 +1,10 @@
 from enum import Enum, auto
+import functools
 import time
 from typing import Dict
 
-from golem.core.common import to_unicode
+from golem_messages import datastructures
+from golem_messages import validators
 
 
 class TaskState(object):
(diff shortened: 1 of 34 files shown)


golem/task/taskstate.py
```diff
@@ -1,8 +1,10 @@
 from enum import Enum, auto
+import functools
 import time
 from typing import Dict
 
-from golem.core.common import to_unicode
+from golem_messages import datastructures
```
shadeofblue, line 6: hmm... general question - should `golem_messages` be used for stuff that's not really part of the protocol?

shadeofblue (approved):
I guess it's mostly okay... though, I have reservations against using `golem_messages` for structures that are not part of Golem protocol...

PR #4135, opened 2019-04-25
Only ask for new tETH donation when half of donated funds are used

maaktweluit:
While working on the node_integration_tests i noticed there are always 2 donations done on each new key. 

The donation amount (`0.01`) is equal to the threshold to request a new donation ( `0.01` ). Causing a new donation after a single transaction has been made.


Krigpl (approved):
what-could-possibly-go-wrong_o_5802169

PR #4138, opened 2019-04-25
Added more detailed error info to task creation RPC

### LINKED ISSUE #3881: Make the return information for `comp.task.create` RPC more meaningful
* update the `comp.task.create` RPC to return the exact amount of missing ETH/GNT
* if `concent_enabled`, verify if the held ETH/GNT would be sufficient to create a task without Concent and if so, provide the estimated missing difference of both ETH and GNT

kmazurek:
Resolves: #3881

The exceptions related to insufficient ETH and GNT funds have been
reworked. They can now include info about multiple missing currencies and have a corresponding dict representation.

### DIFF
--- a/golem/ethereum/exceptions.py
+++ b/golem/ethereum/exceptions.py
@@ -1,23 +1,70 @@
+import typing
+
 from ethereum.utils import denoms
 import golem_sci.structs
 
 
 class EthereumError(Exception):
-    pass
+    def to_dict(self) -> dict:
+        return {
+            'error_type': self.__class__.__name__,
+            'error_msg': self.__str__(),
+            'error_details': {}
+        }
+
+
+class MissingFunds(typing.NamedTuple):
+    """
+    Represents a single entry for missing funds in a given currency.
+    """
+    required: int
+    available: int
+    extension: str
 
 
 class NotEnoughFunds(EthereumError):
-    def __init__(self, required: int, available: int, extension="GNT") -> None:
+    def __init__(
+            self,
+            missing_funds: typing.Optional[typing.List[MissingFunds]] = None,
+            required: int = -1,
+            available: int = -1,
+            extension: str = '') -> None:
         super().__init__()
-        self.required = required
-        self.available = available
-        self.extension = extension
+        if not missing_funds:
+            self.missing_funds = [
+                MissingFunds(
+                    required=required,
+                    available=available,
+                    extension=extension
+                )
+            ]
+        else:
+            self.missing_funds = missing_funds
 
-    def __str__(self):
-        return "Not enough %s available. Required: %f, available: %f" % \
-               (self.extension,
-                self.required / denoms.ether,
-                self.available / denoms.ether)
+    def __str__(self) -> str:
+        return "Not enough funds available." + self._missing_funds_to_str()
+
+    def to_dict(self) -> dict:
+        err_dict = super().to_dict()
+        err_dict['error_details']['missing_funds'] = \
+            [entry._asdict() for entry in self.missing_funds]
+
+        return err_dict
+
+    def _missing_funds_to_str(self) -> str:
+        res = ""
+        for required, available, extension in self.missing_funds:
+            res += f"\nRequired {extension}: {required / denoms.ether:f}, " \
+                f"available: {available / denoms.ether:f}"
+
+        return res
+
+
+class NotEnoughDepositFunds(NotEnoughFunds):
+    def __str__(self) -> str:
+        return "Not enough funds for Concent deposit." \
+               "Top up your account or create the task with Concent disabled." \
+               + self._missing_funds_to_str()
 
 
 class TransactionError(EthereumError):
--- a/golem/ethereum/exceptions.py
+++ b/golem/ethereum/exceptions.py
@@ -19,7 +19,7 @@ class MissingFunds(typing.NamedTuple):
     """
     required: int
     available: int
-    extension: str
+    currency: str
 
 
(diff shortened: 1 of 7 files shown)


kmazurek:
With these changes applied the RPC `comp.task.create` will return objects for errors related to insufficient funds. Here's a sample response:
```
{
  "error_type": "NotEnoughFunds",
  "error_msg": "Not enough funds available.\nRequired GNT: 2.0, available: 0.001\nRequired ETH: 3.0, available: 0.001",
  "error_details": {
    "missing_funds": [
      {
        "required": 2e+18,
        "available": 1000000000000000,
        "extension": "GNT"
      },
      {
        "required": 3e+18,
        "available": 1000000000000000,
        "extension": "ETH"
      }
    ]
  }
}
```
For other errors the behaviour is unchanged (a plain string with the error message will be returned).
The CLI interface `tasks` has been updated to handle dicts and display the value under `error_msg` in these cases.

The proposed RPC error structure could potentially be applied to all Golem endpoints, enabling differentiation between various error types (instead of relying solely on error messages). However, this is outside of the scope of this PR.

golem/ethereum/exceptions.py
```diff
@@ -1,23 +1,70 @@
...
 
 
 class NotEnoughFunds(EthereumError):
-    def __init__(self, required: int, available: int, extension="GNT") -> None:
+    def __init__(
+            self,
+            missing_funds: typing.Optional[typing.List[MissingFunds]] = None,
+            required: int = -1,
+            available: int = -1,
+            extension: str = '') -> None:
```
jiivan: I know it's been called extension before your changes, but... Let's call it currency. :)

jiivan (commented):
LGTM, just old odd naming issue

golem/ethereum/exceptions.py
kmazurek: Agreed, will do

jiivan (approved):
heavyrain_nod

golem/ethereum/transactionsystem.py
```diff
@@ -603,24 +613,37 @@ def validate_concent_deposit_possibility(
...
                 'Gas price is high. It can take some time to mine deposit.',
             )
 
+        required_deposit_difference = required - self.concent_balance()
+        gntb_balance = self.get_available_gnt()
+        if gntb_balance < required_deposit_difference:
+            missing_funds.append(exceptions.MissingFunds(
+                required=required,
+                available=gntb_balance,
+                currency='GNTB'
```
Krigpl: Just go with `GNT` as in other places

golem/ethereum/exceptions.py
```diff
@@ -1,23 +1,70 @@
...
+    currency: str
 
 
 class NotEnoughFunds(EthereumError):
-    def __init__(self, required: int, available: int, extension="GNT") -> None:
+    def __init__(
+            self,
+            missing_funds: typing.Optional[typing.List[MissingFunds]] = None,
+            required: int = -1,
+            available: int = -1,
```
Krigpl: Ugh, this combo constructor sucks. Make it take just the list of MissingFunds and create a static helper method that takes required, available, currency and returns an instance.

golem/ethereum/exceptions.py
```diff
@@ -1,23 +1,70 @@
...
+        err_dict = super().to_dict()
+        err_dict['error_details']['missing_funds'] = \
+            [entry._asdict() for entry in self.missing_funds]
+
+        return err_dict
+
+    def _missing_funds_to_str(self) -> str:
+        res = ""
+        for required, available, currency in self.missing_funds:
+            res += f"\nRequired {currency}: {required / denoms.ether:f}, " \
```
Krigpl: Nitpicking, but I wouldn't put `\n` here but in other places.

golem/task/rpc.py
```diff
@@ -392,8 +392,13 @@ def enqueue_new_task(client, task, force=False) \
     return task
 
 
-def _create_task_error(e, _self, task_dict, **_kwargs):
+def _create_task_error(e, _self, task_dict, **_kwargs) \
```
Krigpl, line 395: Why does it always return `None` as the first element in the tuple? What is that for? Is there some magic here?

golem/ethereum/exceptions.py
kmazurek: I agree, constructor overloading is a pretty neat idea that hasn't made its way to Python somehow. Still, I thought of this as the "pythonic" way of things. Still, let me change it.

golem/ethereum/transactionsystem.py
kmazurek: ACK

golem/ethereum/exceptions.py
kmazurek: What do you mean by other places? The line break here is for the sake of readability when the error message is displayed directly (e.g. when the RPC is called through CLI).

golem/task/rpc.py
kmazurek: It's there to be compatible with what `create_task` returns, i.e. a tuple with optional task ID as its first element in case of success. There is nothing successful here.

golem/ethereum/exceptions.py
Krigpl: It's just that I find it weird that calling this function gets you a string that starts with a new line. Imo callers should care about formatting and prepend new line when necessary.
But as I said, it's nitpicking.

kmazurek: I see, okay - I suppose I can move the new line to be a suffix.

PR #4140, opened 2019-04-25
Docker input & output

Wiezzel:


golem/envs/__init__.py
```diff
@@ -62,6 +62,56 @@ def __str__(self) -> str:
         return self.name
 
 
+class RuntimeInput(ContextManager['RuntimeInput'], ABC):
+    """ A handle for writing to standard input stream of a running Runtime.
+        Input could be either raw (bytes) or encoded (str). Could be used as a
+        context manager to call .close() automatically. """
+
+    def __init__(self, encoding: Optional[str] = None) -> None:
```
mfranciszkiewicz, line 70: Since `self._encoding` is required for `str`s and `_encode` accepts both `str` and `bytes`, `encoding` shouldn't be optional.

golem/envs/docker/cpu.py
```diff
@@ -54,6 +56,49 @@ def from_dict(dict_: Dict[str, Any]) -> 'DockerCPUConfig':
...
+        super().__init__(encoding=encoding)
+        self._raw_output = raw_output
+
+    def __iter__(self) -> Iterator[Union[str, bytes]]:
+        buffer = b""
+
+        for chunk in self._raw_output:
+            buffer += chunk
+            lines = buffer.split(b"\n")
+            buffer = lines.pop()
```
mfranciszkiewicz, line 74: Wouldn't this make non-fully flushed `_raw_output` contain only a part of the last (intended) line?

golem/envs/docker/cpu.py
```diff
@@ -149,9 +197,20 @@ def _prepare():
...
 
-            for warning in result.get("Warnings", []):
+            for warning in result.get("Warnings") or []:
                 logger.warning("Container creation warning: %s", warning)
 
+            sock = client.attach_socket(
+                container_id, params={'stdin': True, 'stream': True}
+            )
+            # Due to terrible design of attach_socket() we have to do this
+            if isinstance(sock, WrappedSocket):
```
mfranciszkiewicz: Maybe create a custom socket wrapper? We need to branch multiple times in this class.

golem/envs/__init__.py
Wiezzel: I don't quite get your comment. `Optional[str]` is equivalent to `Union[str, None]` and accepting `None` is needed because that's how we get a non-encoded stream. Or did you mean that there should be no default value?

golem/envs/docker/cpu.py
Wiezzel: Could you provide some example? I don't see what could go wrong.

golem/envs/docker/cpu.py
Wiezzel: Good idea. 👍

golem/envs/__init__.py
```diff
@@ -62,6 +62,56 @@ def __str__(self) -> str:
...
+        self._encoding = encoding
+
+    def _encode(self, line: Union[str, bytes]) -> bytes:
+        if self._encoding:
```
mfranciszkiewicz, line 76: IMO this should be
```
if isinstance(line, str):
    assert self._encoding
```

golem/envs/__init__.py
mfranciszkiewicz: What I mean is that we need `self._encoding` for `_encode` to work properly with strings. IMO `self._encoding` should always be set. Please also check my comment for `_encode`

Wiezzel: The idea behind `_encoding` is "is this input encoded and if yes then with which encoding". Sometimes it is set to `None` which means "this input is not encoded", and sometimes it is set to a `str` which means "this input is encoded with encoding X".
Further, input which is non-encoded expects `_encode()` to be called with `bytes`, while encoded input expects `_encode()` to be called with `str`.

mfranciszkiewicz: From the interface point of view I can pass both `bytes` and `str` to `_encode`. Passing a `str` when encoding is not set will raise an `AssertionError`. I'll learn that later on, in runtime.

golem/envs/docker/cpu.py
mfranciszkiewicz: E.g. `_raw_output` is something like
```
['first line\n', 'second line\n', 'last li']
```

and is later filled with
```
['ne\n']
```

golem/envs/__init__.py
Wiezzel: Yep, that's the only way I see for having support for both raw and encoded input without making two separate classes. If you find this kind of 'hidden' type constraints misleading then I can split Input into two classes:
  * `RawInput` which doesn't have `_encoding` at all and accepts only `bytes`
  * `EncodedInput` which has mandatory `_encoding` and accepts only `str`

But then, to keep things completely clear we would probably have to add two separate methods for `Runtime` as well:
  * `get_raw_input() -> RawInput`
  * `get_encoded_input(encoding: str) -> EncodedInput`

golem/envs/docker/cpu.py
Wiezzel: This will work correctly.

After reading `'last li'`:
```python
buffer += chunk              # buffer is 'last li'
lines = buffer.split(b"\n")  # lines is ['last li']
buffer = lines.pop()         # buffer is 'last li', lines is []
# nothing is yielded
```
After reading `'ne\n'`:
```python
buffer += chunk              # buffer is 'last line\n'
lines = buffer.split(b"\n")  # lines is ['last line', '']
buffer = lines.pop()         # buffer is '', lines is ['last line']
# 'last line\n' is yielded
```

golem/envs/docker/cpu.py
Wiezzel: Created `InputSocket` wrapper class: https://github.com/golemfactory/golem/pull/4140/commits/820de6d3d6be4aabf1d9da14831963aca1982ef9

PR #4147, opened 2019-04-30
Make incomes and payments CLI calls print short-id by default

maaktweluit:
Found this branch while cleaning up my local git, still looks like a nice improvement to me

### DIFF
--- a/golem/interface/client/payments.py
+++ b/golem/interface/client/payments.py
@@ -1,6 +1,6 @@
 from ethereum.utils import denoms
 
-from golem.core.common import to_unicode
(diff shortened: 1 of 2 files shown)


golem/interface/client/payments.py
```diff
@@ -10,6 +10,13 @@
 
 filterable_statuses = ['awaiting', 'confirmed']
 
+full_table = Argument(
+    '--full',
+    optional=True,
+    default=False,
+    help="Show full table contents"
```
Krigpl: Maybe a little moar descriptive like "Expand shortened columns"

Krigpl (approved):

PR #4150, opened 2019-05-02
Moved responsibility for web3 retry to the ProxyProvider

maaktweluit:
The integration tests are failing frequently with [GETH errors](https://buildbot.golem.network/buildbot/#/builders/15/builds/646).

When investigating i found the re-try mechanism is not properly working:
- When making all URL's invalid it does not try each 3x and then fail
- When connection is lost the code triggers, but still fails to early

This PR solves both issues, by moving the responsibility of retry to ProviderProxy, cleaning up ethereum/node and ethereum/middleware


tests/golem/ethereum/test_ethereum_node.py
```diff
@@ -10,28 +10,7 @@ class TestPublicNodeList(TestCase):
...
-        assert node.initial_addr_list
-        assert node.addr_list is None
-
-        node.provider_proxy.provider = Mock()
-        node.addr_list = []
-        node._handle_remote_rpc_provider_failure()
-
-        assert node.provider_proxy.provider is None
-        assert node.start.called
+        assert node.web3.isConnected
```
Krigpl: This assert doesn't really do anything considering you set it to `Mock` above

Krigpl (approved):
This is fine I think

tests/golem/ethereum/test_ethereum_node.py
maaktweluit: Good one! made it a `assert_called` :)

mfranciszkiewicz (approved):

PR #4152, opened 2019-05-06
Brass Marketplace paper linked in docs + naming fix for thirst

tworec:
resolves golemfactory/golem-model#74


.gitignore
```diff
@@ -81,3 +81,6 @@ apps/dummy/test_tmp
 .python-version
 luxrender-output.png
 .testmondata
+
+# rust
```
Krigpl: There's a `Rust` section already
Also see https://github.com/golemfactory/golem/pull/3647#pullrequestreview-182303798, we keep `Cargo.lock` in the repo

golem/task/timer.py
```diff
@@ -60,7 +64,7 @@ def __init__(self) -> None:
         self._thirst = 0.2
 
     @property
-    def thirst(self):
+    def profit_factor(self):
```
Krigpl, line 68: Since you link the doc everywhere and the doc mentions `thirst` and doesn't mention `profit_factor`, what is this change for?

Krigpl (changes requested):

.gitignore
tworec: > There's a `Rust` section already

ok, didn't noticed bc of case sensitive search ;)

> Also see [#3647 (review)](https://github.com/golemfactory/golem/pull/3647#pullrequestreview-182303798), we keep `Cargo.lock` in the repo

I'm restoring. It keeps bothering me to be changed after each `setup.py develop` invocation. Why do you want to keep it in repo? I'm just curious.

golem/task/timer.py
tworec: read [the paper](https://docs.golem.network/About/img/Brass_Golem_Marketplace.pdf) carefully and you'll notice that `thrist` is `t_2 - t_1` and here is calculated profit factor. This name is proposed by @mbenke, see https://github.com/golemfactory/golem-model/issues/74#issuecomment-488568824

golem/task/timer.py
```diff
@@ -51,24 +51,30 @@ def _finish(self) -> None:
...
 
 class ThirstTimer(ActionTimer):
+    """
+    Enables calculation of profit factor that adjusts expected profit as
+    proposed in https://docs.golem.network/About/img/Brass_Golem_Marketplace.pdf
+    """
 
     _ALPHA: ClassVar[float] = 0.00001
     _BETA: ClassVar[float] = 4 * _ALPHA
+    _INITIAL_PROFIT_FACTOR = 0.2
```
Krigpl: typehint

Krigpl (approved):

.gitignore
Krigpl: Does it change really? Just tried and it didn't.

tworec: it was adding two lines, maybe it's because I have newer rust version
I've committed them https://github.com/golemfactory/golem/pull/4152/commits/56e6e008bd90f4e91d18b604fa775513dd2ab050#diff-f1f51e603003432b4874d91a6299019c

Krigpl: Right, so not every time, just once because of the newer compiler version. This isn't something to be bothered by.

tworec: ok, I've reverted, because buildbot complained about not stashed changes after cargo changed lock.

I'm asking by pure educational purposes: why do you want `Cargo.lock` in repo?

golem/task/timer.py
```diff
@@ -51,24 +51,30 @@ def _finish(self) -> None:
 
 
 class ThirstTimer(ActionTimer):
```
mfranciszkiewicz, line 53: Should probably be renamed

.gitignore
Krigpl: The reason is included in the link I posted - Golem is a binary, not a library, thus Cargo.lock is meant to be in the version control.

tworec: oops, I didn't see hidden/resolved comment, thx

golem/task/timer.py
tworec: nope, timer is measuring thirst (in seconds) but we are transforming thirst into profit factor. 

see  https://github.com/golemfactory/golem-model/issues/74#issuecomment-488568824

mfranciszkiewicz: But it effectively measures the profit factor. Since `thirst` has been renamed, it does much more than measuring idle time.

tworec: it does not measure the `profit factor`, it *calculates* it based on `thirst`, which is the measured value. But feel free to rename it. I was also willing to name it as you're proposing but rethought and for me now it is ok.

PR #4153, opened 2019-05-06
Environment 'ERROR' status

Wiezzel:


mfranciszkiewicz:
It would be helpful to know the reason why the error occurred

Wiezzel:
@mfranciszkiewicz The exception is always logged and re-raised. So the caller will have their reason.

mfranciszkiewicz (approved):

mfranciszkiewicz:
I guess this will suffice. To be clear: the point was to store the reason with the status so it can be queried later on.

Wiezzel:
@mfranciszkiewicz I see. If this is needed we might think of adding it later on. But AFAIK Python enums don't support storing any additional data so it would require some more substantial changes to the way statuses are represented.

Krigpl (approved):

PR #4163, opened 2019-05-09
Integration tests minor stuff

etam:
Things I've gathered along the way.

### DIFF
--- a/scripts/node_integration_tests/playbooks/concent/force_accept/playbook.py
+++ b/scripts/node_integration_tests/playbooks/concent/force_accept/playbook.py
@@ -8,7 +8,7 @@ from golem_messages.factories.tasks import ReportComputedTaskFactory
 
 from scripts.node_integration_tests import helpers
 
-from ..concent_base  import ConcentTestPlaybook
+from ..concent_base import ConcentTestPlaybook
 from ...test_config_base import NodeId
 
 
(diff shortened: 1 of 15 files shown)


scripts/node_integration_tests/playbooks/concent/force_accept/playbook.py
```diff
@@ -86,10 +86,9 @@ def on_success(result):
             self.next()
             return
 
-        if ((not self.sra_received) and
-                self.sra_deadline and
-                datetime.datetime.now() > self.sra_deadline
-        ):
+        if (not self.sra_received) and \
```
shadeofblue: @etam actually, I'm unsure how this change improves stuff ... ? 

I think that outer braces are more readable and/or less error-prone than the backtick...

shadeofblue (approved):
fine by me

scripts/node_integration_tests/playbooks/concent/force_accept/playbook.py
etam: I was touching this in the first place, because "closing bracket does not match visual indentation [E124]". For me it makes no difference if it's a bracket or backslash ([backtick](https://en.wikipedia.org/wiki/Grave_accent#Use_in_programming) is something different).

I can change it back to [parentheses](https://en.wikipedia.org/wiki/Bracket) if you want.

etam: done

kmazurek (approved):
okok

Issue #4165, opened 2019-05-09
test_router.py breaks b0.19 because it leaves zombies


maaktweluit:
https://buildbot.golem.network/buildbot/#/builders/9/builds/11624

Above build is completed but still timed out, when looking at the build-agent i see a process is stuck:
![Screenshot 2019-05-09 at 10 10 12](https://user-images.githubusercontent.com/10008353/57452663-aca06680-7264-11e9-95d6-a58c3947727b.png)

maaktweluit:
`TestDockerDummyTask` is also disabled, it seems to have the same problems

PR #4166, opened 2019-05-09
Reported issue and disabled breaking tests

maaktweluit:
Disables tests that break due to #4165


maaktweluit:
No longer needed, other solves are linked in the issue

PR #4169, opened 2019-05-09
Transaction history update

### LINKED ISSUE #3880: Transaction history update
* Add support for more payment statuses in the history for both incoming and outgoing payments: 
** (blockchain transaction) rejected, payment overdue (awaiting beyond expected payment time), confirming (on blockchain) @Krigpl (is that possible already - or would it require changes to ETS?)

```
Incoming payments:
1.0 awaiting
2.0 payment overdue (try to use Concent), reasons may include:
Requestor made a transaction that didn’t cover this payment (can be detected earlier)
Requestor didn’t make a transaction at all (actual overdue payment)
4.0 confirmed

Outgoing payments:
1.0 awaiting (/queued)
2.0 payment overdue, reasons may include:
insufficient ETH/GNT
Golem bug
3.0 confirming (on blockchain)
4.0 confirmed
5.2137 operational costs ETH

```

* Outgoing: Add information about the transaction fee (only on confirmed transactions)

* (possibly?) replace the subtask id in the history with a node name

jiivan:
Fixes: #3880


golem/ethereum/transactionsystem.py
```diff
@@ -385,6 +386,10 @@ def get_subtasks_payments(
     def get_incomes_list(self) -> List[Dict[str, Any]]:
         incomes = self._incomes_keeper.get_list_of_all_incomes()
 
+        # Our version of peewee (2.10.2) doesn't support
+        # .join(attr='XXX'). So we'll have to join manually
+        lru_node = functools.lru_cache()(functools.partial(nodeskeeper.get))
```
etam: unneeded partial

golem/ethereum/paymentprocessor.py
```diff
@@ -248,3 +253,22 @@ def sendout(self, acceptable_delay: int = PAYMENT_MAX_DELAY):
         )
 
         return True
+
+    def update_overdue(self) -> None:
```
etam, line 257: Maybe it should be `@staticmethod` and operate only on database? Or would doing this break records kept in `_awaiting`? (I don't know how peewee would handle this)

etam:
Generally :+1: , but added 2 comments.

golem/ethereum/paymentprocessor.py
jiivan: I'm afraid that because instances in `_awaiting` would became desynchronized with db, it can cause problems.

etam (approved):
ok

PR #4177, opened 2019-05-13
send multiple TTC in response to multi-WTCT

tworec:
Multiple TTC are sent as a response to `num_subtasks` filed in WTCT.
Resolves (partly) golemfactory/golem-unlimited#216

The general idea is to transit from a __one subtask__ per single Provider to **one WTCT** per single Provider. It is done by remembering WTCT msg hash, then counting started and accepted subtasks.

Limitation for just one WTCT is hardcoded and not parametrized intentionally. Current Provider code is not amended, so WTCT will have `num_subtasks == 1` for standard Golem Brass Providers.  Golem Unlimited is supposed to use this functionality solely.

The Provider is allowed to ask for multiple TTC expressing it via `num_subtasks` parameter in WTCT. If Provider's offer will be selected he will get at most `num_subtasks` TTC's. It can be less when there is no more work to perform. 

Until all assigned work (TTC's) are accepted by Requestor, Provider is not able to ask for more. This way we are enforcing Providers to plan their resources consumption beforehand sending WTCT. 

_Technical note_:
This change rewrites `TaskClient` class and changes API of the base `Task` class by adding WTCT hash and `num_subtasks` attribute to client acceptance methods.


golem/task/taskclient.py
```diff
@@ -23,51 +24,57 @@ def __getstate__(self):
         return state
 
     @staticmethod
-    def assert_exists(node_id, node_dict):
+    def assert_exists(node_id: str,
```
kmazurek: While we're at it, I think this method name is misguiding - `assert` implies that it will raise an exception if the value is not in the dict being checked. I think something like `try_create` would be more accurate.

tests/golem/task/test_taskclient.py
```diff
@@ -10,19 +10,116 @@ def test(self):
         node_id = str(uuid.uuid4())
         node_dict = {}
 
+        # when
```
jiivan, line 13: Is it possible to test all those scenarios separately (in separate test methods)?

tworec: it can be split for sure, but I can do it in a week or two from now on. I think it is not a blocker here

golem/task/taskmanager.py
```diff
@@ -386,20 +387,18 @@ def task_needs_computation(self, task_id: str) -> bool:
             return False
         return True
 
-    def get_next_subtask(
-            self, node_id, task_id, estimated_performance, price,
-            max_resource_size, max_memory_size):
+    # noqa pylint: disable=too-many-arguments,too-many-return-statements
+    def get_next_subtask(self,
```
kmazurek, line 392: Nitpicking: since you're changing the signature here you could add type annotations to all parameters.

tests/golem/task/test_taskclient.py
kmazurek: I agree with @jiivan here, the test should be split into several separate methods.

tworec:
@jiivan @zakaprov since your review is kinda stylistic, can you approve for now to enable this to be included in release tests? I will address your comments in separate PR next week, ok?

kmazurek (commented):
The logic looks okay to me, I left comments about some potential improvements. With that said, I'd recommend running the integration test suite on this branch before merging. Also, it's a large enough change to call for an end-to-end test of its own.

kmazurek (approved):

golem/task/taskclient.py
tworec: I've come up with `get_or_initialize`, which is most descriptive of what I can imagine

tests/golem/task/test_taskclient.py
tworec: done

tworec: @zakaprov @jiivan can you review once again after changes?

tests/golem/task/test_taskmanager.py
```diff
@@ -330,54 +328,35 @@ def test_get_next_subtask(self, *_):
...
+        assert self.tm.is_my_task("xyz")
+        assert self.tm.get_next_subtask("DEF", "xyz", 1000, 10, None) is None
 
-        wrong_task = not self.tm.is_my_task("xyz")
-        subtask = self.tm.get_next_subtask(
-            "DEF", "xyz", 1000, 10, 5, 10, None)
-        assert subtask is None
-        assert not wrong_task
+        assert self.tm.is_my_task("xyz")
+        assert self.tm.get_next_subtask("DEF", "xyz", 1000, 10, None) is None
```
tworec: I'm not sure if those tree checks are adding any value here. Wouldn't be better to replace them with one?

apps/core/task/coretask.py
```diff
@@ -384,8 +379,6 @@ def interpret_task_results(self, subtask_id, task_results, sort=True):
 
     @handle_key_error
     def result_incoming(self, subtask_id):
-        self.counting_nodes[self.subtasks_given[
```
kmazurek, line 387: Why is this removed?

apps/core/task/coretask.py
```diff
@@ -476,25 +469,27 @@ def _get_resources_root_dir(self):
...
+                      node_id: str,
+                      wtct_hash: Union[None, bytes] = None,
+                      num_subtasks: int = 1
+                     ) -> AcceptClientVerdict:
+        verdict = self.should_accept_client(node_id, wtct_hash)
 
-        if verdict == AcceptClientVerdict.ACCEPTED:
-            client = TaskClient.assert_exists(node_id, self.counting_nodes)
-            client.start()
+        client = TaskClient.get_or_initialize(node_id, self.counting_nodes)
```
kmazurek: This part is always called, regardless of the verdict obtained above. Is that intentional?

apps/core/task/coretask.py
```diff
@@ -476,25 +469,27 @@ def _get_resources_root_dir(self):
...
         return os.path.dirname(prefix)
 
-    def should_accept_client(self, node_id):
-        client = TaskClient.assert_exists(node_id, self.counting_nodes)
-        finishing = client.finishing()
-        max_finishing = self.max_pending_client_results
-
+    def should_accept_client(self,
+                             node_id: str,
+                             wtct_hash: Union[None, bytes] = None
```
kmazurek: I think it's more readable to use `Optional[bytes]` instead of `Union[None, bytes]` here.

apps/core/task/coretask.py
```diff
@@ -476,25 +469,27 @@ def _get_resources_root_dir(self):
...
+        elif client.should_wait(wtct_hash):
             return AcceptClientVerdict.SHOULD_WAIT
 
         return AcceptClientVerdict.ACCEPTED
 
-    def accept_client(self, node_id):
-        verdict = self.should_accept_client(node_id)
+    def accept_client(self,
+                      node_id: str,
+                      wtct_hash: Union[None, bytes] = None,
```
kmazurek: ditto

golem/task/taskbase.py
```diff
@@ -337,11 +337,18 @@ def to_dictionary(self):
         pass
 
     @abc.abstractmethod
-    def should_accept_client(self, node_id):
+    def should_accept_client(self,
+                             node_id: str,
+                             wtct_hash: Union[None, bytes] = None
```
kmazurek: Like above, please use `Optional[bytes]` here.

golem/task/taskbase.py
```diff
@@ -337,11 +337,18 @@ def to_dictionary(self):
...
+                             node_id: str,
+                             wtct_hash: Union[None, bytes] = None
+                            ) -> AcceptClientVerdict:
         pass
 
     @abc.abstractmethod
-    def accept_client(self, node_id):
+    def accept_client(self,
+                      node_id: str,
+                      wtct_hash: Union[None, bytes] = None,
```
kmazurek: ditto

golem/task/taskclient.py
```diff
@@ -23,51 +24,67 @@ def __getstate__(self):
         return state
 
     @staticmethod
-    def assert_exists(node_id, node_dict):
+    def get_or_initialize(node_id: str,
+                          node_dict: Dict[str, 'TaskClient']) -> 'TaskClient':
+        """ If given `node_id` is already in `node_dict` it's corresponding
```
kmazurek: nitpicking: it's -> its

golem/task/taskmanager.py
```diff
@@ -387,27 +388,22 @@ def task_needs_computation(self, task_id: str) -> bool:
...
-    def get_next_subtask(
-            self, node_id, task_id, estimated_performance, price,
-            max_resource_size, max_memory_size):
+    # noqa pylint: disable=too-many-arguments,too-many-return-statements
+    def get_next_subtask(self,
+                         node_id: str,
+                         task_id: str,
+                         estimated_performance: float,
+                         price: int,
+                         wtct_hash,) -> Optional[ComputeTaskDef]:
```
kmazurek: nit: `wtct_hash` has no type hint.

golem/task/tasksession.py
```diff
@@ -408,92 +398,96 @@ def _offer_chosen(
...
-            price=price,
-            size=package_size,
-            resources_options=self.task_server.get_share_options(
-                ctd['subtask_id'], self.address).__dict__
-        )
-        ttc.generate_ethsig(self.my_private_key)
-        if ttc.concent_enabled:
+        wtct_hash = msg.get_short_hash()
+        for _i in range(msg.num_subtasks):
+            ctd = self.task_manager.get_next_subtask(
```
kmazurek, line 415: As discussed offline - the entire body of this loop could be moved to a separate method of `TaskSession`, e.g. `_assign_new_subtask(self, msg: message.tasks.WantToComputeTask, wtct_hash: bytes)`.

tests/golem/task/test_taskmanager.py
kmazurek: I agree, no point in keeping it this way.

kmazurek (commented):
Btw, are there any plans for adding integration tests of this feature?

apps/core/task/coretask.py
tworec: good catch, thx.

apps/core/task/coretask.py
tworec: bc there is no `finish` anymore in taskclient. I've added description for this class since it was rewritten in large part here.

tests/golem/task/test_taskmanager.py
tworec: removed

apps/core/task/coretask.py
tworec: done

golem/task/taskclient.py
tworec: done

golem/task/taskmanager.py
tworec: done

golem/task/taskmanager.py
tworec: done

tworec (commented):
@zakaprov pls look again

golem/task/tasksession.py
tworec: this is not so straightforward when there are `return` and `yield` statements. Lets keep things simple.

tworec:
> Btw, are there any plans for adding integration tests of this feature?

would love to do, but don't have knowledge... who can help?

kmazurek:
> > Btw, are there any plans for adding integration tests of this feature?
> 
> would love to do, but don't have knowledge... who can help?

As discussed offline - @tworec is going to check whether it's possible to cover this use-case using Golem node integration tests (without the involvement of a Golem Unlimited gateway). Let me know should you need more support with writing these tests.

kmazurek (approved):
LGTM
lgtm

PR #4160, opened 2019-05-08
Cargo.lock with comment for rust on BB upgraded

tworec:
also added small enhancement to `lint.sh` to communicate changed file names. sample

```
⚡️  You must commit or stash changes:

  M golem/task/taskbase.py
D  golem/task/taskmanager.py
 M lint.sh

```

PR #4178, opened 2019-05-14
Green control_test for b0.19

maaktweluit:
Contains:
- Cherry-picked version of #4160
- Fix to kill all zombie hyperg's before starting a new test
- Skipping tests that break due to Twisted making zombies #4165


scripts/test-daemon-stop.sh
```diff
@@ -5,4 +5,6 @@ rm $H_FILE || echo "Error, not able to delete '$H_FILE'"
 
 kill $H_PID || echo "Error, not able to stop hyperg"
 
+pkill -f hyperg
```
mfranciszkiewicz: This basically replaces the PID-file mechanism. Maybe something like this instead?
`kill $H_PID || pkill -f hyperg || echo "Error, not able to stop hyperg"`

shadeofblue (approved):

scripts/test-daemon-stop.sh
maaktweluit: It is supposed to catch zombies when the PID mechanism leaves some `hyperg-workers`

Also at this point there should never be any `hyperg(-worker)` running so it ensures a clean test slate

mfranciszkiewicz: That's true, but the PID file's role was to target a specific hyperg instance, as there may be more of them more running.

With `pkill -f` we're eliminating the need for a PID file.

mfranciszkiewicz (approved):

PR #4182, opened 2019-05-14
Add resource usage to GLambda App

mplebanski:


kubkon (approved):
LGTM! It would be great if you could log the usage after parsing it as well :+1:

mbenke (approved):
LGTM

PR #4183, opened 2019-05-14
Eliminate hanging processes left after `rpc/test_router.py` tests

### LINKED ISSUE #4165: test_router.py breaks b0.19 because it leaves zombies

shadeofblue:
closes #4165

### DIFF
--- a/tests/golem/rpc/test_router.py
+++ b/tests/golem/rpc/test_router.py
@@ -211,10 +211,7 @@ class _TestRouter(TestDirFixtureWithReactor):
         self.reactor_thread.reactor.stop()
 
         if self.process.is_alive():
-            raise Exception('alive!')
-
-
-            #self.process.terminate()
+            self.process.terminate()
 
     def _run_test(self, expect_error, *args, **kwargs):
         self.process = Process(
--- a/tests/golem/rpc/test_router.py
+++ b/tests/golem/rpc/test_router.py
@@ -1,4 +1,4 @@
-    # pylint: disable=protected-access
+# pylint: disable=protected-access
 # The code below is organised in classes, each class running one test only.
 # This is because closing reactor and router and running them again ends with
 # a timeout error during session.connect(). This proved to be challenging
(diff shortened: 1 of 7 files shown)


tests/golem/rpc/test_router.py
```diff
@@ -120,9 +123,27 @@ def format_errors(self):
...
+        p = psutil.Process()
+        return set([c.pid for c in p.children(recursive=True)])
+
     def setUp(self):
         super().setUp()
         self.state = _TestRouter.State()
+        self._children_on_start = self._get_process_children()
+
+    def tearDown(self):
+        super().tearDown()
```
maaktweluit: i think its better to teardown at the end

tests/golem/rpc/test_router.py
```diff
@@ -1,4 +1,4 @@
-# pylint: disable=protected-access
+    # pylint: disable=protected-access
```
maaktweluit: why are these spaces here?

maaktweluit (approved):
LGTM!

Nice fixes! some small comments, nothing blocking

tests/golem/rpc/test_router.py
shadeofblue: ugh, that's probably some artifact...

tests/golem/rpc/test_router.py
shadeofblue: hmm... it shouldn't matter - but particularly, in this case, I think it should be called first -> as there is stuff that touches the reactor et al in the original teardown

I wanted to be absolutely sure that we're doing our own cleanup as the very last thing before the test finishes

mfranciszkiewicz:
What do you think on creating a test fixture from this cleanup code?

tests/golem/rpc/test_router.py
maaktweluit: ok, sounds good to me!

mfranciszkiewicz (approved):

PR #4185, opened 2019-05-15
Environment events & runtime events

Wiezzel:
Environment events:
  * ENV_ENABLED
  * ENV_DISABLED
  * PREREQUISITES_INSTALLED
  * CONFIG_UPDATED
  * ERROR_OCCURRED

Runtime events:
  * PREPARED
  * STARTED
  * STOPPED
  * TORN_DOWN
  * ERROR_OCCURRED


Wiezzel:
@Krigpl @mfranciszkiewicz Plx, review

Krigpl:
Is it somewhere described what will be listening on these events and how it will react?

Wiezzel:
@Krigpl I believe no. But I can image some scenarios, e.g. forwarding these events to UI to provide feedback for the user.

Krigpl:
I have mixed feelings about adding ~700 lines of code for a feature we're not sure how it'll be used at the moment.

golem/envs/__init__.py
```diff
@@ -16,11 +18,41 @@
...
+class RuntimeEvent(NamedTuple):
+    type: RuntimeEventType
+    details: Optional[Dict[str, Any]] = None
+
+
+RuntimeEventListener = Callable[[RuntimeEvent], Any]
+
+
+class EnvEventType(Enum):
+    ENV_ENABLED = 1
```
mfranciszkiewicz:
```suggestion
    ENABLED = 1
```

golem/envs/__init__.py
```diff
@@ -16,11 +18,41 @@
...
+    ENV_DISABLED = 2
```
mfranciszkiewicz:
```suggestion
    DISABLED = 2
```

golem/envs/__init__.py
```diff
@@ -149,38 +183,78 @@ def _change_status(
...
+            details=details
+        )
+        self._logger.debug("Emit event: %r", event)
+        for listener in self._event_listeners.get(event_type, ()):
+            listener(deepcopy(event))
+
+    def _prepared(self, *_) -> None:
+        """ Acknowledge that Runtime has been prepared. Log message, set status
+            and emit event. Arguments are ignored (for callback use). """
+        self._logger.info("Runtime prepared.")
```
mfranciszkiewicz, line 216: Maybe include class name for more information?

mfranciszkiewicz:
We **do** need these events to give feedback to the user.

golem/envs/__init__.py
```diff
@@ -275,16 +348,83 @@ class Environment(ABC):
...
+            every listener registered for this type of events. """
+
+        event = EnvEvent(
+            env_id=self.metadata().id,
+            type=event_type,
+            details=details
+        )
+        self._logger.debug("Emit event: %r", event)
+        for listener in self._event_listeners.get(event_type, ()):
+            listener(deepcopy(event))
```
mfranciszkiewicz: Sending events should be synchronous but I'm not entirely sure that we need to wait for the listener to finish executing.

Wiezzel: What would you suggest then? Pushing events to a queue and having a delegated event-handling thread? Or maybe spawning a new thread for each event/handler?

golem/envs/__init__.py
Wiezzel: I think logger name is enough (e.g. Docker CPU will use `golem.envs.docker.cpu`).

golem/envs/__init__.py
Wiezzel: Done.

golem/envs/__init__.py
Wiezzel: Done.

golem/envs/__init__.py
mfranciszkiewicz: We can either leave it for the developer to decide how to implement the listener or delegate the call to a thread pool (e.g. via `golem.core.golem_async.async_run`)

Wiezzel: OK, resolved by calling `deferToThread`.

mfranciszkiewicz (approved):

PR #4191, opened 2019-05-16
Cgi/transcoding/resource per subtask

weaselix:


golem/task/rpc.py
```diff
@@ -5,6 +5,7 @@
 import logging
 import os.path
 import re
+import threading
```
nieznanysprawiciel: Why is this necessary?

golem/task/tasksession.py
```diff
@@ -4,6 +4,7 @@
 import enum
 import functools
 import logging
+import threading
```
nieznanysprawiciel: Why is this necessary?

golem/task/rpc.py
weaselix: Thanks that you notice that. It is not necessary. It was added for test purposes by me. It was removed.

golem/task/tasksession.py
weaselix: Same as above.

nieznanysprawiciel (approved):

PR #4195, opened 2019-05-16
New unit tests for MemoryChecker

### LINKED ISSUE #4090: TestMemoryChecker.test_memory is flaky
Test randomly fails with a ridiculously long stacktrace.
Example: https://buildbot.golem.network/buildbot/#/builders/9/builds/10866/steps/5/logs/stdio

Wiezzel:
* Not flaky
* No sleeps
* Cover more cases

Fixes #4090

### DIFF
--- a/tests/golem/vm/test_memorychecker.py
+++ b/tests/golem/vm/test_memorychecker.py
@@ -1,4 +1,4 @@
-from time import sleep as originalsleep
+from typing import Iterable
 from unittest import TestCase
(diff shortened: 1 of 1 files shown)


mfranciszkiewicz (approved):
One thing - IMO using the original `patch` method would make the code easier to understand.

maaktweluit (approved):
LGTM!

An alternative would be to use `freezetime`, but i like your fix too!

PR #4200, opened 2019-05-17
fix: clean-docker-vm assumes Docker Toolbox is installed 

maaktweluit:
- Moved CleanDocker action to the end of the installation
- Updated the script to use the bundled docker version
- Added logs to the script

### DIFF
--- a/scripts/docker/clean-docker-vm.ps1
+++ b/scripts/docker/clean-docker-vm.ps1
@@ -40,10 +40,18 @@ try {
 "Done"
 #>
 "Clean Docker Machine..."
+
+"hasInstalled: " + $hasInstalled
+"pfSFFolder: " + $pfSFFolder
 # TODO: Change to docker or docker-machine updated, always execute this time from TB -> binaries.
 #if ( -Not ( $hasInstalled.indexof("Docker") = -1 ))
 #{
-$dockerRmCmd = """" + $pfSFFolder + "Docker Toolbox\docker-machine.exe"" rm -f golem"
+$dockerBin = """" + $pfSFFolder + "Golem\docker-machine.exe"""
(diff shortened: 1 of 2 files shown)


scripts/docker/clean-docker-vm.ps1
```diff
@@ -40,10 +40,18 @@ try {
...
+"hasInstalled: " + $hasInstalled
+"pfSFFolder: " + $pfSFFolder
 # TODO: Change to docker or docker-machine updated, always execute this time from TB -> binaries.
 #if ( -Not ( $hasInstalled.indexof("Docker") = -1 ))
 #{
-$dockerRmCmd = """" + $pfSFFolder + "Docker Toolbox\docker-machine.exe"" rm -f golem"
+$dockerBin = """" + $pfSFFolder + "Golem\docker-machine.exe"""
+$dockerVersionCmd = $dockerBin + " --version"
+"Executing: " + $dockerVersionCmd
+cmd.exe /c $dockerVersionCmd
```
Wiezzel, line 52: Why do we need to call `cmd.exe` instead of executing `docker-machine.exe` directly?

maaktweluit: Good one!
There where some issues with the PATH when writing this, but should not be a problem when calling it with absolute path..

Will give it a shot :)

maaktweluit: hmm, now i'm testing it it comes back, executing a string is good old windows shit..

We could also use `& ` or `Invoke-Expression ` but both mess up with the quotes.

Propose to leave as is for now

Wiezzel: Meh... okay, let's leave it.

Wiezzel (approved):

PR #4202, opened 2019-05-17
Specify gVisor secure runtime version

mplebanski:


kubkon:
LGTM, but could you add some description in the PR as to why we're hard-coding the version of gvisor?

EDIT: such a description would then be for posterity ;-)

kubkon (approved):

mplebanski:
@kubkon Stability concerns. I don't want users to download `latest` which is not tested by us. Also this is harded to track down as we will have to ask users for logs/information about `gvisor` revision.

PR #4209, opened 2019-05-19
StatusPublisher's events with message codes

mfranciszkiewicz:
Events emitted by the StatusPublisher now contain message labels (codes), so that each error / warning can be more easily identified in the UI. This way the user may be presented with a solution to the occurring issue.

Data field carried by events is now equal to:
- `None`
- `{"status": _1, "value": _2}`, where:
    - when `_1 == "message"`, `_2` is a message provided by core
    - when `_1 != "message"`, `_1` is the message "label" with certain payload `_2`, understood by the UI

### DIFF
--- a/golem/docker/hypervisor/hyperv.py
+++ b/golem/docker/hypervisor/hyperv.py
@@ -5,7 +5,7 @@ import os
 from pathlib import Path
 import subprocess
 import time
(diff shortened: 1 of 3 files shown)


shadeofblue (approved):

mdtanrikulu (commented):
Approved

PR #4214, opened 2019-05-20
Configure gVisor to use host direct filesystem

mplebanski:
I can see two ways of fixing this:

- set +x permission on user's `.local` in the installation script (remember that GLambda app is enabled only on Linux providers, other OSes are not an issue). It's hard to tell what is the impact of loosening the permissions here, depends on what apps put in there, 
- configure gVisor to access host filesystem directly (this is what this PR introduces),
- move datadir to somewhere else e.g. `.cache`

### DIFF
--- a/Installer/Installer_Linux/install.sh
+++ b/Installer/Installer_Linux/install.sh
@@ -356,7 +356,10 @@ function install_dependencies()
 import json
 
 runtime_config = {
-    'path': '/usr/local/bin/runsc'
+    'path': '/usr/local/bin/runsc',
+    'runtimeArgs': [
+        '-file-access shared'
+    ]
 }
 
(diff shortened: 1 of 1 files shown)


kubkon (approved):

kubkon:
LGTM!

kubkon:
OK, after further discussion and a bit of thinking, we might want to rethink this change. It sounds like it will essentially annihilate any benefit of `gvisor` on direct FS access by the Docker app, and I guess we don't want that.

mplebanski:
@kubkon exactly.

mplebanski:
@prekucki What do you think?

mplebanski:
Invalid. This is only the case on some old gVisor revision. The one we are installing in our script is fine.

PR #4219, opened 2019-05-22
node_integration_tests - Fix timeout test on buildbot

maaktweluit:
- Increased timeout for slower machines like buildbot
- `task_timeout` test was hanging on `Finished` status, so made the check more simple

### DIFF
--- a/scripts/node_integration_tests/playbooks/golem/task_timeout/playbook.py
+++ b/scripts/node_integration_tests/playbooks/golem/task_timeout/playbook.py
@@ -29,12 +29,15 @@ class Playbook(NodeTestPlaybook):
     def step_wait_subtask_completed(self):
         def on_success(result):
             if result:
-                statuses = map(lambda s: s.get('status'), result)
-                if any(map(lambda s: s == 'Finished', statuses)):
+                statuses = [s.get('status') for s in result]
+                if 'Finished' in statuses:
                     print("First subtask finished")
                     self.next()
-                    return
-                print("Subtasks status: {}".format(list(statuses)))
+                elif 'Failure' in statuses:
+                    print('Subtask failed :(')
+                    self.fail("Got status 'Failure', expected 'Finished'.")
+                else:
+                    print("Subtasks status: {}".format(statuses))
 
             time.sleep(10)
 
--- a/scripts/node_integration_tests/tasks/__init__.py
+++ b/scripts/node_integration_tests/tasks/__init__.py
@@ -23,8 +23,8 @@ _TASK_SETTINGS = {
     '2_short': {
         'type': "Blender",
         'name': 'test task',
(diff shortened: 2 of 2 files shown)


scripts/node_integration_tests/playbooks/golem/task_timeout/playbook.py
```diff
@@ -29,12 +29,15 @@ class Playbook(NodeTestPlaybook):
     def step_wait_subtask_completed(self):
         def on_success(result):
             if result:
-                statuses = map(lambda s: s.get('status'), result)
-                if any(map(lambda s: s == 'Finished', statuses)):
+                statuses = list(map(lambda s: s.get('status'), result))
```
Wiezzel:
```suggestion
                statuses = [s.get('status') for s in result]
```
This is what list comprehensions are for. Much clearer.

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -99,11 +99,14 @@ def _wait_gnt_eth(self, node_id: NodeId, result):
         if gnt_balance > 0 and eth_balance > 0 and gntb_balance > 0:
             print("{} has {} total GNT ({} GNTB) and {} ETH.".format(
                 node_id.value, gnt_balance, gntb_balance, eth_balance))
+            if self.has_requested_eth:
+                time.sleep(30)
```
Wiezzel: Why sleep if we already got the ETH (since `eth_balance > 0`)?

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -311,7 +314,7 @@ def on_success(result):
             unpaid = self.subtasks - payments
             if unpaid:
                 print("Found subtasks with no matching payments: %s" % unpaid)
-                self.fail()
+                time.sleep(3)
```
Wiezzel: I don't quite get this change. Isn't this function supposed to assert that all tasks have been paid? What purpose does this `sleep()` call serve?

maaktweluit: This PR has been split up into #4227

The sleep will make the fn be called again in 3 seconds, instead of failing the first time it checked

scripts/node_integration_tests/playbooks/base.py
maaktweluit: `self.has_requested_eth = False` by default, so if the eth is enough the first time it will not sleep, if it has_requested it should sleep before progressing to the next step

Also this PR is split up into #4228 to keep it all clean

maaktweluit:
@Wiezzel @shadeofblue Tests are passing :D https://buildbot.golem.network/buildbot/#/builders/15/builds/691

Please review again

Wiezzel (approved):

PR #4226, opened 2019-05-23
node_integration_tests - Added failure after 100 tries on the same step

maaktweluit:
When this test was hanging it would timeout from buildbot and not generate any logs, made the test timeout on its own.

Added 100 retry maximum for each step, this way the test can decide on the timeout
for example 15 seconds of sleep gives a 25 minute timeout while 3 seconds of sleep makes it 5 minutes.

Split out from #4219


Wiezzel (approved):
Looks fine.

PR #4235, opened 2019-05-24
Non-hypervised Docker CPU Environment

Wiezzel:
A temporary subclass of `DockerCPUEnvironment` that always uses `DummyHypervisor` and therefore performs no operations on Docker virtual machine. It is meant to enable usage of the new Environment API without removing `DockerManager` yet. Using standard `DockerCPUEnvironment` would cause potential conflicts during VM reconfiguration.


golem/task/taskcomputer.py
```diff
@@ -67,9 +70,14 @@ def __init__(self, task_server: 'TaskServer', use_docker_manager=True,
 
         self.docker_manager: DockerManager = DockerManager.install()
         if use_docker_manager:
-            self.docker_manager.check_environment()
-
+            self.docker_manager.check_environment()  # pylint: disable=no-member
         self.use_docker_manager = use_docker_manager
+
+        os.makedirs(self.dir_manager.root_path, exist_ok=True)
```
maaktweluit: feels like this belongs in `DirManager.__init__()`

maaktweluit (approved):
LGTM!

So many test patches.. at least it has good coverage :P

One small comment, still approving!

golem/task/taskcomputer.py
Wiezzel: Generally this is a nice idea, but it would require fixing some tests that use `DirManager` and **I really don't want to fix any more tests in this PR**.

maaktweluit: Haha

ok, fully agree!

Wiezzel:
@mfranciszkiewicz @maaktweluit The code is ready for review. The only thing missing is Golem-Messages update. I'll update the requirements as soon as I get to release Golem-Messages with this change: https://github.com/golemfactory/golem-messages/pull/349

maaktweluit (approved):
LGTM!

tests/golem/task/dummy/test_runner_script.py
```diff
@@ -5,6 +5,7 @@
 from tests.golem.task.dummy import runner, task
 
 
+@mock.patch('golem.task.taskserver.NonHypervisedDockerCPUEnvironment')
```
Krigpl, line 8: fyi, if you use `mock.Mock()` as the second argument you won't need to put the extra `_` argument in the test methods.

mfranciszkiewicz (approved):

PR #4236, opened 2019-05-24
Wamp serialization error fix: huge unsigned int

mdtanrikulu:


golem/ethereum/exceptions.py
```diff
@@ -17,8 +17,8 @@ class MissingFunds(typing.NamedTuple):
     """
     Represents a single entry for missing funds in a given currency.
     """
-    required: int
-    available: int
+    required: any
```
Krigpl: Don't change that. This is supposed to hold ints, period. See comment below.

golem/ethereum/exceptions.py
```diff
@@ -47,7 +47,8 @@ def __str__(self) -> str:
     def to_dict(self) -> dict:
         err_dict = super().to_dict()
         err_dict['error_details']['missing_funds'] = \
-            [entry._asdict() for entry in self.missing_funds]
+            [MissingFunds._make([str(item) for item in entry])._asdict()
```
Krigpl: Don't create a new `MissingFunds` object, instead call `entry._asdict()` and then convert the values to string. And even better do that conversion here: https://github.com/golemfactory/golem/pull/4138/files#diff-cfa6ad63fdbdae4250cf225821638922R400, since there's no reason `to_dict` method should convert any ints to strings.

Krigpl (changes requested):

golem/ethereum/exceptions.py
mdtanrikulu: Reverted

golem/task/rpc.py
```diff
@@ -401,7 +401,8 @@ def _create_task_error(e, _self, task_dict, **_kwargs) \
     logger.error("Cannot create task %r: %s", task_dict, e)
 
     if hasattr(e, 'to_dict'):
-        return None, e.to_dict()
+        tempDict = rpc_utils.big_int_to_string(e.to_dict())
```
Krigpl: Let convert all `int`s to strings, it's awkward that the same field might be sometimes an integer and sometimes a string.
Also no camelCase in Python.

golem/rpc/utils.py
```diff
@@ -40,3 +40,23 @@ def predicate(member):
             continue
         mapping[uri] = method
     return mapping
+
+
+def big_int_to_string(collection):
+    # Deeply convert all big integer elements of collections to string
+    def _big_int_to_string(v):
```
Krigpl: This helper function is not needed. Just add another `elif` condition on line 61 to check against `int` and that's it. (and change argument name to `v`)

Krigpl (changes requested):

golem/rpc/utils.py
mdtanrikulu: Make sense. 👍

golem/task/rpc.py
mdtanrikulu: Fixed

Krigpl (approved):
obama_approves

PR #4240, opened 2019-05-27
Migrate Payments to TaskPayment

### LINKED ISSUE #2457: Remove dead code from PaymentsKeeper & PaymentsDatabase
Most of the code is probably not longer used. Check which parts are still used and remove all the rest.

### LINKED ISSUE #4197: Migrate Payments to Transaction
## Rationale
To achieve _Epic Show top-ups and withdraws in the transaction history_ #4172 .

This issue does not cover `Income` nor `DepositPayment`.

### Functional specification
* Implement DB storage as specified in #4172
* Migrate `Payments` to `WalletOperation` and `TaskPayment`. Remove `Payments` table. (use schema migrations)

### User Interface

All rpc that depend on `Payments` should remain the same API (no changes of CLI nor Electron)

### Technical specification
* DB
```
WalletOperation:
    hash: BlockchainTransactionField
    direction: enum(Incoming, Outgoing)
    type: enum(Transfer, DepositTransfer, TaskPayment, DepositPayment)
    status: enum(Awaiting, Sent, Confirmed, Overdue)
    sender_address: [eth address]
    recipient_address: [eth address]
    amount: HexIntegerField
    currency: enum(GNT, ETH)
    gas_cost: HexIntegerField

TaskPayment:
    wallet_operation: ->[WalletOperation 1to1]
    node: str
    task: str
    subtask: str
    paid_amount: HexIntegerField
    expected_amount: HexIntegerField
    accepted_ts: int
    settled_ts: int
```
* Refactor payment RPCs

## Dependencies
None

### Sub-components
None

### Blockers
None

### Additional tests
* Update `Payments` unit tests
* No additional integration tests needed

## QA
_Description of the QA process for the feature_

#### Base test process
_specify one of: ( if unsure, consult @ZmijaWA or @ederenn )_
** QA limited to the single feature (in case of very trivial changes)
** Smoke Test (the limited set of tests designed to ensure that most critical functionality doesn't experience regressions)
** Full Test (full set of tests for the whole application - in case of more extensive application updates, refactoring etc)

### New feature tests
_Add test scenarios for the new feature_

#### Positive Scenario 1
* Step 1
* Step 2
* ...
* Expected results

#### Positive Scenario 2
* ...

#### Negative Scenario 1
* Step 1
* Step 2
* ...
* Some expected problem
* Expected problem handling

### Possible regressions
* is there any functionality that is likely to experience regressions?
* are there any existing application components that should be tested more thoroughly?

### Development
* [ ] Feature implemented
* [ ] Feature unit/integration tests implemented

### Dev QA
* [ ] Basic tests by the developer pass
* [ ] Unit tests pass
* [ ] Golem integration tests pass
* [ ] Concent integration tests pass
* [ ] Concent acceptance tests pass

### QA team
* [ ] Base scenario passes
* [ ] Additional test 1 passes... 
* [ ] ... 

## Issues encountered during QA
_When adding an issue here, please update testing scenarios and QA progress above_
* [ ] link to issue... 
* [ ] ...

jiivan:
fixes: #4197 
additionally fixes: #2457


golem/client.py
```diff
@@ -833,17 +833,22 @@ def get_task(self, task_id: str) -> Optional[dict]:
...
+            p.wallet_operation.status in statuses_of_interest
             for p in subtasks_payments)
         if not subtasks_payments or not all_sent:
             task_dict['cost'] = None
             task_dict['fee'] = None
         else:
-            # Because details are JSON field
-            task_dict['cost'] = sum(p.value or 0 for p in subtasks_payments)
+            task_dict['cost'] = sum(
+                p.wallet_oepration.amount for p in subtasks_payments
```
kmazurek: typo: `wallet_operation`

golem/model.py
```diff
@@ -292,34 +261,55 @@ def python_value(self, value):
...
+    sender_address = CharField()
+    recipient_address = CharField()
+    amount = HexIntegerField(null=True)
+    currency = StringEnumField(enum_type=CURRENCY)
+    gas_cost = HexIntegerField(null=True)
+
+
+class TaskPayment(BaseModel):
+    wallet_operation = ForeignKeyField(WalletOperation, unique=True, null=False)
+    node = CharField(null=False)
```
kmazurek, line 311: Opinion: suffixing these three field names (`node`, `task`, `subtask`) with `_id` would remove ambiguity on the values they store. What do you think?

golem/model.py
```diff
@@ -292,34 +261,55 @@ def python_value(self, value):
...
+    accepted_ts = IntegerField(null=True)
+    settled_ts = IntegerField(null=True)  # set if settled by the Concent
 
-    def __init__(self, *args, **kwargs):
-        super(Payment, self).__init__(*args, **kwargs)
-        # For convenience always have .details as a dictionary
-        if self.details is None:
-            self.details = PaymentDetails()
-
-    def __repr__(self) -> str:
```
kmazurek, line 311: `__repr__` is missing on both of the newly introduced classes, is that intentional?

golem/model.py
```diff
@@ -222,56 +236,11 @@ def __self__(self):
...
-    def __init__(self,
-                 node_info: Optional[dt_p2p.Node] = None,
-                 fee: Optional[int] = None,
-                 block_hash: Optional[str] = None,
-                 block_number: Optional[int] = None,
-                 check: Optional[bool] = None,
-                 tx: Optional[str] = None) -> None:
-        self.node_info = node_info
-        self.fee = fee
-        self.block_hash = block_hash
```
kmazurek, line 235: Shouldn't we keep these (i.e. `block_hash` and `block_number`) for the sake of bookkeeping? Or are they considered redundant?

golem/ethereum/paymentprocessor.py
```diff
@@ -134,24 +139,39 @@ def _payment_confirmed(payment: Payment, timestamp: int) -> None:
...
             subtask_id=payment.subtask,
-            payee=payment.payee,
+            payee=payment.wallet_operation.recipient_address,
             delay=delay,
         )
 
-    def add(self, subtask_id: str, eth_addr: str, value: int) -> Payment:
+    def add(
+            self,
+            task_header: dt_tasks.TaskHeader,
```
Krigpl: Don't make this module depend on golem-messages just because you need two values from this struct.
Just provide node_id and task_id explicitly.

golem/ethereum/transactionsystem.py
```diff
@@ -327,12 +343,18 @@ def stop(self):
 
     def add_payment_info(
             self,
+            task_header: dt_tasks.TaskHeader,
```
Krigpl: ditto

golem/model.py
jiivan: Yes, I like suffixing in this case, but all other models have fields without them. For the sake of consistency I would keep it as is.

golem/model.py
jiivan: It's not needed anywhere that's why I removed them.

golem/model.py
```diff
@@ -292,34 +261,55 @@ def python_value(self, value):
...
-            .format(
-                self.subtask,
-                float(self.value) / denoms.ether,
-                self.status,
-                tx,
-                bn,
-                self.processed_ts
-            )
+    @property
+    def processed_ts(self) -> int:
```
Krigpl: I'd really prefer to have this property set explicitly using the same logic as the batch transfer uses to have a guarantee this works as expected. and afaict `time.time()` is not defined to return the UTC timestamp. Please revert back to the previous approach of using common function is those two places, you're not gaining much by making this a property anyway.

Krigpl (changes requested):

golem/model.py
jiivan: Please look here:
https://github.com/golemfactory/golem/blob/402a22a2daa76cdd250cf9637df383310ea47c75/golem/model.py#L46-L61

This guarantees that `created_date` is in UTC. Only function of this property is to translate datetime to integer timestamp. Do you think there is an incentive in duplicating this value?

Krigpl: But during batch transfer you use `time.time()` which doesn't have to generate the same values as `created_date`. I think it's reasonable to make this very explicit that `processed_ts` and `closure_time` originate from the same place considering this is payments related, so we'd rather not risk introducing any bugs or implicit assumption here.

golem/ethereum/paymentprocessor.py
jiivan: `[too-many-arguments] Too many arguments (6/5)`
I'll stay with pylint in this case. Moved the dependency into `TYPE_CHECKING` branch.

golem/client.py
jiivan: Thanks, fixed

golem/model.py
jiivan: Good point. I'll update batch transfer to use plain utc datetimes.

golem/ethereum/paymentprocessor.py
```diff
@@ -134,24 +145,39 @@ def _payment_confirmed(payment: Payment, timestamp: int) -> None:
...
             subtask_id=payment.subtask,
-            payee=payment.payee,
+            payee=payment.wallet_operation.recipient_address,
             delay=delay,
         )
 
-    def add(self, subtask_id: str, eth_addr: str, value: int) -> Payment:
+    def add(
+            self,
+            task_header: 'dt_tasks.TaskHeader',
```
Krigpl: That's really not what I meant, it doesn't change dependencies at all. There are exactly zero reasons transaction system should be aware of the existence of the `TaskHeader`. Please do what I suggested earlier.

golem/ethereum/paymentprocessor.py
```diff
@@ -213,7 +240,7 @@ def sendout(self, acceptable_delay: int = PAYMENT_MAX_DELAY):
         if not self._awaiting:
             return False
 
-        now = get_timestamp()
+        now = int(time.time())
```
Krigpl: Still not a consistent way with the `processed_ts` values.

Krigpl (changes requested):

golem/model.py
jiivan: Maybe we'll add them, when they're needed?

golem/ethereum/transactionsystem.py
```diff
@@ -16,11 +16,13 @@
     List,
     Optional,
     Tuple,
+    TYPE_CHECKING,
 )
 
 from ethereum.utils import denoms
 from eth_keyfile import create_keyfile_json, extract_key_from_keyfile
 from eth_utils import is_address
+from golem_messages.datastructures import p2p as dt_p2p
```
Krigpl: The same applies to the transaction system and the whole `ethereum` directory in general - don't make it depend on message-related structures from `golem_messages`

tests/golem/ethereum/test_paymentprocessor.py
```diff
@@ -6,20 +6,22 @@
 import unittest.mock as mock
 from os import urandom
 
+from golem_messages.factories.datastructures import tasks as dt_tasks_factory
```
Krigpl: The same applies to the tests. Since the module doesn't use structures from `golem-messages`, the tests should not use factories from it as well.

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -10,6 +10,7 @@
 import faker
 from freezegun import freeze_time
 from golem_messages.factories import p2p as p2p_factory
+from golem_messages.factories.datastructures import tasks as dt_tasks_factory
```
Krigpl: ditto, obviously

Krigpl (changes requested):
Please, no unnecessary `golem-messages` dependencies in the `ethereum` module.

golem/ethereum/transactionsystem.py
jiivan: Sure, but in this case is it better to manually construct a dict? IMHO it's better to use `node.to_dict()`.

https://github.com/golemfactory/golem/blob/5914e62e02ce9251f1328bbe8fff8ae06b28b873/golem/ethereum/transactionsystem.py#L86-L94

Krigpl: That's the thing - it wouldn't be a problem if I was made aware of the previous set of changes, the ones that moved all the rpc calls directly to this module. This was not the right move. This module is not supposed to do random formatting and returning arbitrary dicts as in e.g. `get_incomes_list`. It also is not supposed to be aware of what `node_id` is connected to (the Node structure), because it doesn't care about that. This kind of things should happen outside of this module and then we wouldn't have this kind of unnecessary dependencies.

This kind of mess happens when you don't include module maintainers in your PRs. I'm really not happy about this.

jiivan: Moved to separate module.

golem/model.py
```diff
@@ -292,34 +274,46 @@ def python_value(self, value):
...
+        deposit_payment = enum.auto()
+
+    class CURRENCY(msg_dt.StringEnum):
+        ETH = enum.auto()
+        GNT = enum.auto()
+
+    tx_hash = BlockchainTransactionField(null=True)
+    direction = StringEnumField(enum_type=DIRECTION)
+    operation_type = StringEnumField(enum_type=TYPE)
+    status = StringEnumField(enum_type=STATUS, default=STATUS.awaiting.value)
```
Krigpl: Do we need the default value here? It's not clear to me whether `awaiting` is still a good default value

golem/model.py
```diff
@@ -292,34 +274,46 @@ def python_value(self, value):
...
+    sender_address = CharField()
+    recipient_address = CharField()
+    amount = HexIntegerField(null=True)
```
Krigpl, line 304: When can it be `null`?

Krigpl (commented):
Thanks, that looks muuuch better. Mostly lgtm but have a couple of questions inline.

golem/model.py
jiivan: When it's an income (direction incoming). It's null until we receive confirmation from SCI.

golem/model.py
jiivan: We could set status explicitly every time. But I'm not sure whether `awaiting` is a bad default.

tests/golem/ethereum/test_paymentprocessor.py
```diff
@@ -188,12 +203,19 @@ def test_failed_transaction(self):
         self.sci.get_gntb_balance.return_value = balance_gntb
 
         gnt_value = 10**17
-        self.pp.add("test_subtask_id", encode_hex(urandom(20)), gnt_value)
+        self.pp.add(
+            subtask_id="test_subtask_id",
+            eth_addr=encode_hex(urandom(20)),
+            value=gnt_value,
+            node_id='00adbeef' + 'deadbeef' * 15,
```
Krigpl: Did you mean `0x` prefix?

golem/model.py
Krigpl: Because there's only one case where we set `awaiting` during object creation.
Withdrawals will set `sent`
Top ups will set `confirmed`
Deposit operations will also not use `awaiting` anywhere

`awaiting` is only used for incoming task payments, that's why I don't see why it is a good enough reason to make it a default.

golem/ethereum/transactionsystem.py
```diff
@@ -39,13 +40,16 @@
...
-from golem.network import nodeskeeper
 from golem.rpc import utils as rpc_utils
 from golem.utils import privkeytoaddr
 
 from . import exceptions
 from .faucet import tETH_faucet_donate
 
+if TYPE_CHECKING:
+    # pylint: disable=unused-import,ungrouped-imports
+    from golem_messages.datastructures import tasks as dt_tasks
```
Krigpl: That's not used anymore

golem/ethereum/paymentprocessor.py
```diff
@@ -257,17 +293,20 @@ def sendout(self, acceptable_delay: int = PAYMENT_MAX_DELAY):
     def update_overdue(self) -> None:
         """Sets overdue status for awaiting payments"""
 
-        processed_ts_deadline = int(time.time()) - PAYMENT_DEADLINE
+        processed_ts_deadline = datetime.datetime.now(
+            tz=datetime.timezone.utc
```
Krigpl, line 297: There's `get_timestamp_utc` in `golem.core.common`, might be cleaner to use it

golem/ethereum/paymentprocessor.py
```diff
@@ -257,17 +293,20 @@ def sendout(self, acceptable_delay: int = PAYMENT_MAX_DELAY):
     def update_overdue(self) -> None:
         """Sets overdue status for awaiting payments"""
 
-        processed_ts_deadline = int(time.time()) - PAYMENT_DEADLINE
+        processed_ts_deadline = datetime.datetime.now(
```
Krigpl: rename to `created_ts_deadline` for consistency

Krigpl (approved):

golem/ethereum/paymentprocessor.py
jiivan: Unfortunately it returns float, not datetime.

PR #4242, opened 2019-05-28
Remove entrypoint.sh from Docker images

Krigpl:
The same can be now achieved by using the `--user` flag in `docker run` command.


mfranciszkiewicz (approved):
Approving for slow tests

Krigpl:
I ran unittests on macos as well as a manual e2e test and everything works fine.

mfranciszkiewicz (approved):
Re-approving

PR #4247, opened 2019-05-28
Add funds validation to task/subtask restart endpoints

### LINKED ISSUE #3895: Change task restart RPCs to return results/errors
The RPC calls `comp.task.subtasks.frame.restart` and `comp.task.restart_subtasks` do not return any results. As a consequence, uncaught exceptions will not be visible to the client. To enable reporting errors, these calls should return tuples of optional values - a return value and an error. This way they will be compatible with other task RPC calls (e.g. `comp.task.restart`) and can be properly handled by the `safe_run` decorator.

### LINKED ISSUE #4180: Cost information and insufficient balance check for `comp.task.subtask.restart`
## Functional description
`comp.task.restart_subtasks`

* For UI/UX purposes it's good to have estimated cost information for the given subtask (with another rpc provided), and if user balance is not enough to restart the subtask, the RPC should return proper information about insufficient balance.
* Optionally, the RPC should provide a way to optionally disable the Concent service when restarting an already finished task.

### User Interface
It will need changes on UI: 
Proper restart modal to show cost information of selected subtask(s).
Insufficient balance error modal in error case.

kmazurek:
Resolves: #4180
Also fixes: #3895

Replaced the RPC endpoint `comp.task.restart_subtasks` with `comp.task.subtasks.restart`. Updated the endpoint to handle both finished and active tasks. Added error reporting via the method's return value.
Added funds validation for both subtask and whole task restarts.

### DIFF
--- a/golem/task/rpc.py
+++ b/golem/task/rpc.py
@@ -545,6 +545,14 @@ class ClientProvider:
         # Create new task that is a copy of the definition of the old one.
         # It has a new deadline and a new task id.
         try:
+            task = self.task_manager.tasks[task_id]
+            self._validate_enough_funds_to_pay_for_task(
+                task.subtask_price,
+                task.get_total_tasks(),
+                task.header.concent_enabled,
+                force
+            )
+
             task_dict = copy.deepcopy(
                 self.task_manager.get_task_definition_dict(
                     self.task_manager.tasks[task_id],
--- a/golem/task/rpc.py
+++ b/golem/task/rpc.py
@@ -19,7 +19,6 @@ from golem.core import golem_async
 from golem.core import common
 from golem.core import deferred as golem_deferred
 from golem.core import simpleserializer
-from golem.client import Client
 from golem.ethereum import exceptions as eth_exceptions
 from golem.resource import resource
 from golem.rpc import utils as rpc_utils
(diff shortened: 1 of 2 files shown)


golem/task/rpc.py
```diff
@@ -411,6 +414,17 @@ def _restart_task_error(e, _self, task_id, **_kwargs):
     return None, str(e)
 
 
+def _restart_subtasks_error(e, _self, task_id, subtask_ids, **_kwargs) \
+        -> typing.Union[str, typing.Dict]:
+    logger.error("Failed to restart subtasks from task: %r, subtask_ids: %r,"
+                 "%s", task_id, subtask_ids, e)
```
jiivan: > Always start with a message, print data as "kwargs" 
https://github.com/golemfactory/golem-internal/wiki/Logging-guidelines#log-message-structure

golem/task/rpc.py
```diff
@@ -552,6 +581,62 @@ def restart_task(self, task_id: str, force: bool = False) \
...
+        """
+        try:
+            task = self.task_manager.tasks[task_id]
+
+            self._validate_enough_funds_to_pay_for_task(
+                task.subtask_price,
+                len(subtask_ids),
+                False if disable_concent else task.header.concent_enabled,
+                ignore_gas_price
+            )
```
jiivan: I believe `_validate_enough_funds_to_pay_for_task` should be outside of this `try..except`

jiivan (approved):

PR #4249, opened 2019-05-29
Enhanced transcoding integration tests with ffprobe report

cameel:
This pull request contains major part of my and @pawelkisielewicz's work on adding detailed ffprobe metadata comparisons to transcoding integration tests. It has been rebased on the latest `CGI/transcoding/master` and updated to take into account changes from the master branch and to work with the current `ffmpeg_tools` `master`.

This is the first part. It's meant to work with the current transcoding method - without changing the split&merge method or adding support for new formats. It contains the bulk of the implementation but due to the limitations of the current method it does not test everything it could. That will come in subsequent, smaller pull requests.

### What's included
1) `ffprobe_report.py` (+unit tests): classes that wrap JSON data obtained from ffprobe and define an interface for accessing the attributes. This is an intermediate layer that abstracts out low level details like: deciding which field to use (sometimes there's more than one with similar information), how to parse and compare it it, which fields are actually of interest to us, etc.
2) `ffprobe_report_set.py` (+unit tests): a helper class that gathers a set of ffprobe report diffs from the executed test cases and can print a summary in a table at the end. It's hard to get an overview of what works and what does not for a big set of video files otherwise. The test output is too verbose to be useful when 1000 tests fail.
3) `simulated_transcoding_operation.py` (+unit tests): a class that started out as a set of private helper methods in `test_ffmpegintegration.py` but grew enough to warrant creating a separate class. Its main purpose is to keep stuff out of the tests themselves to make them shorter and more readable. It allows us to concisely define how the file should be transcoded, define what to ignore when comparing results and run the transcoding operation.
4) New integration tests for the following cases:
    - codec/container change
    - resolution change
    - frame rate change
    - splitting into various numbers of segments

### What is **not** included
1) Testing on the big, varied set of video files that we prepared. The current tests run only on `test_video.mp4` and `test_video2.mp4` included in the repo. My work on getting the tests to run on the whole set is nearing completion but there are still issues I need to deal with and I did not want to add failing tests to the pull request. I'm going to be adding files from that set in the subsequent pull requests.
2) Ignoring/handling the attribute differences in the failing tests that we agreed upon. Some of the smaller ones are already here but the rest was on top of the rest of my branches and getting them to work with the new master is what I'm currently working on. They'll be coming as separate pull requests for ease of review.
3) The new split&merge method with audio separation. I still have to update that code to work on the current master. And the changes have to go to two separate repositories (`golem` and `ffmpeg_tools`). I've decided that it will be easier to first get the tests in. This way you'll be able to see that it does not break stuff that used to work and I'll be able to submit the rest in smaller, more manageable pieces.
4) Support for a wider range of codecs/contrainers. That will come with the new split&merge implementation.

### Stuff you may want to take a closer look at when reviewing
1) The range of currently supported codec/container combinations is very limited so only conversion to H.264/AVI is being tested. I have more but I'm going to add them along with the new split&merge implementation.
2) The current resolution validations don't allow any conversion if the **input** file does not have one of the predefined sizes (even if the conversion would retain the aspect ratio). So for example `test_video.mp4` is 320x240 and you can't convert it to 640x480. I'm not sure it that's something that's going to be fixed or a deliberate decision. Anyway, that's the reason why I'm testing conversion from 320x240 to 320x240 :) No other sizes would work with that file. Even though it does not really change the size, it does run ffmpeg with the scaling filter so it is testing something. I'll be able to add more resolutions when I add more test files or when this validation rule is relaxed.
3) The tests are ignoring video bitrate when comparing results. Turns out that even if you give ffmpeg your desired bitrate with `-b:v`, the bitrate of the resulting file can vary wildly. I'll expand on that in a separate pull request in `ffmpeg_tools` for a feature that implements preserving bitrate. Here I just wanted to let you know that I disabled these comparisons. It's done explicitly with `operation.exclude_from_diff({'video': {'bitrate'}})` in the test and is even in a separate commit to make it hard to miss.
4) ffprobe reports wrong frame count in `test_video.mp4` transcoded to H.264/AVI using your current `ffmpeg_tools`. The input file has 50 frames and so does the output file (I checked with `-show_frames`) but the `nb_frames` field reported by ffprobe says 100.  Also `r_frame_rate` changes from 25 to 50 in the output file even though the conversion does not specify frame rate.  `avg_frame_rate`, on the other hand, remains unchanged at 25. This is something I did not expect and we'll have to decide what to do about it but the bottom line is that the frame count read from `nb_frames` is not reliable. I'm ignoring the `FfprobeReport.frame_count` in the tests.
5) We have implemented `StreamOperator.get_metadata()` because there used to be no other way to run the `compute-metrics` operation from the ffmpeg image. I see that `split` operation now returns the metadata but that only works for the input file. For the transcoded file we have to get the metadata separately.
    - You may want to take a look at intermediate files/directories produced by `StreamOperator.get_metadata()` and `FfprobeReport.build()`. I'm not sure it this fits well into the directory structure in the image.
6) `FfprobeReportSet` is not enabled because it prints a big markdown table that is only relevant to you if you're working on transcoding. And the table only looks remotely readable if you render it as markdown (too much information to make it readable in the terminal unfortunately). Currently the lines that enable it are simply commented out  in `test_ffmpegintegration` test case `setUp()` and you can uncomment it if you want to see the result. Is this acceptable? Or should we have some switch/setting for it? Or maybe just print it to a file and add it to `.gitignore` just like it's done with the code coverage report?
7) I have kept existing integration test cases mostly intact (apart from extracting the task definition into a single place). Do you want it left that way or should they also be rewritten using the new classes?


cameel:
Looks like CI fails - but only on Mac OS and Windows:

```
______ ERROR collecting tests/apps/ffmpeg/task/test_ffmpegintegration.py _______
tests/apps/ffmpeg/task/test_ffmpegintegration.py:68: in <module>
    class TestffmpegIntegration(FfmpegIntegrationTestCase):
tests/apps/ffmpeg/task/test_ffmpegintegration.py:87: in TestffmpegIntegration
    for video_file in FfmpegIntegrationTestCase.VIDEO_FILES
E   AttributeError: 'function' object has no attribute 'VIDEO_FILES'
```

But why would code like this not fail on Linux if `VIDEO_FILES` was really missing? The curious thing is that it says that `FfmpegIntegrationTestCase` is a `'function' object` (it should be a class).

After looking at it closely, I think that the `@ci_skip` decorator is the culprit. The failing code is a part of the `@parameterized.expand` decorator call in `TestffmpegIntegration`. The code tries to access a class variable on `FfmpegIntegrationTestCase` called `VIDEO_FILES`:

``` python
    @parameterized.expand(
        (video_file, video_codec, container)
        for video_file in FfmpegIntegrationTestCase.VIDEO_FILES
        for video_codec, container in [
            (VideoCodec.H_264, Container.c_AVI),
        ]
    )
    @pytest.mark.slow
    def test_split_and_merge_with_codec_change(self,
```

The problem stems from the way `@ci_skip` was implemented. It doesn't really just "skip" the class it's applied to. It actually prevents it from being defined:
``` python
def ci_skip(obj):
    import unittest
    if in_appveyor() or in_travis():
        return unittest.skip('Unsupported CI environment')
    return obj
```

So in CI `FfmpegIntegrationTestCase` doesn't actually exist and the code that inherits from it or uses it in any way won't work properly. So this is really a non-obvious and undocumented limitation of `@ci_skip`.

Fortunately in my case I was able to refactor the code to remove the inheritance. I've just pushed a patch. We'll see if it passes through CI.

tests/apps/ffmpeg/task/ffprobe_report.py
```diff
@@ -0,0 +1,641 @@
...
+        for attribute in attributes_to_compare:
+            original_value = getattr(original_stream_report, attribute)
+            modified_value = overrides.get(
+                attribute,
+                getattr(modified_stream_report, attribute),
+            )
+
+            if (
+                    attribute not in excludes and
+                    modified_value != original_value
```
nieznanysprawiciel: What happens in case when we override attribute? 
As far as I understand we compare value from original stream to expected value after transcoding.
And what we want, is to compare expected value from overrides to value from modified stream (after transcoding).

cameel: It works like in your second sentence but the names are wrong.  When we run `diff()` in tests the report from the input file is passed into the "modified" parameter and the transcoded report is the "original". "Modified" should really be called "expected" and "original" should be called something else. Yeah, I know that's confusing and it's something I wanted to change. It ended up like this because I did not want to interfere with the stuff @pawelkisielewicz was working on at the time and it was not critical - this change is still somewhere in my TODO. But now that the pull request is ready, I can go back to it.

Do you want it fixed in this pull request or is it OK to submit a separate one later? I'd prefer a separate one because I'm now working to get my other changes out and I probably won't be able to do this quickly. I think the other stuff has higher priority.

nieznanysprawiciel: This can wait

tests/apps/ffmpeg/task/test_ffmpegintegration.py
```diff
@@ -1,49 +1,189 @@
...
+        # Uncomment this to enable report generation:
+        #from tests.apps.ffmpeg.task.ffprobe_report_set import FfprobeReportSet
+        #cls._ffprobe_report_set = FfprobeReportSet()
+
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+
+        if cls._ffprobe_report_set is not None:
+            print(cls._ffprobe_report_set.to_markdown())
```
nieznanysprawiciel: Writing results to file would be good idea. But probably to tests directory with all other tests temporary stuff.

tests/apps/ffmpeg/task/test_ffprobe_report.py
```diff
@@ -0,0 +1,728 @@
...
+                'location': 'subtitle',
+                'original_stream_index': 6,
+                'modified_stream_index': None,
+                'reason': 'No matching stream'
+            },
+        ]
+        self.assertCountEqual(diff, expected_diff)
+
+    def test_report_should_have_video_fields_with_proper_values(self):
+        assert RAW_REPORT_WITH_MPEG4['streams'][0]['width'] == 560
```
nieznanysprawiciel: This solution introduces dependency between tests. If you change RAW_REPORT_WITH_MPEG4, this test won't work either, despite it could still work.
Better solution is to use copy.deepcopy(RAW_REPORT_ORIGINAL) like in other tests and set required fields to expected values.
The same applies to all other tests that asserts on RAW_REPORT_WITH_MPEG4

nieznanysprawiciel:
Files structure:
- it would be good to move all files related to reports and tests utils you wrote to separate directory (for example test/apps/ffmpeg/task/utils or something like this.
- tests testing these utils should be placed in the same directory as utils

It's not the end. I'm still reviewing this PR. There's a lot of changes here...

tests/apps/ffmpeg/task/ffprobe_report_set.py
```diff
@@ -0,0 +1,119 @@
...
+        table[coordinates[0]][coordinates[1]][coordinates[2]] = value
+
+    @classmethod
+    def _describe_ffprobe_report_diff(cls, diff: Diff) -> str:
+        output_lines = []
+
+        stream_mismatch_found = False
+        stream_types_different = False
+        for diff_dict in diff:
+            if diff_dict['reason'] == "Different attribute values":
```
nieznanysprawiciel: I'm a little bit afraid of this comparison. If error message changes even slightly this condition will fail. Is there any other way to do this ?

nieznanysprawiciel (commented):
Ok, I finished review and left some comments.

There was question, if our integration tests should be rewritten to new model. I think, most of them shouldn't, because they check other things, before task is created and ran. Probably test_simple_case could be rewritten, because it runs whole task.

Generally tests should be in 2 groups. One group will be ran on CI and consists probably of all tests that are now.
Second group test extended video set and we will run it only sometimes.

tests/apps/ffmpeg/task/ffprobe_report_set.py
cameel: Good point. We should change that.

The original reason for adding this field was that there are various kinds of differences - "different attribute values" and "no matching stream in the other file". When you see the first one in the diff dict its meaning is pretty clear. Not so with the other one. So I added a special value that would be useful both as a clarification and would allow us handle different types of differences differently. It's not really a free-form error message.

But I see your point. It looks like a human-readable error message. I should have defined an enum or constants.

Actually, with hindsight, I see that I could have done even better. Each type of difference should be a `namedtuple` rather than a dict.

We'll change that, but please say which variant you prefer:
1) **Enum**:
    ```python
    class DiffReason(Enum):
        DifferentAttributeValues = "Different attribute values"
        NoMatchingStream = "No matching stream"
    ```
2) **Constants**:
    ```python
    REASON_DIFFERENT_ATTRIBUTE_VALUES = "Different attribute values"
    REASON_NO_MATCHING_STREAM = "No matching stream"
    ```
3) **Proper types for different types of differences**
    ``` python
    class StreamType(Enum):
        Video = "video"
        Audio = "audio"
        Subtitle = "subtitle"
        Data = "data"

    FormatAttributeDifference(typing.namedtuple):
        attribute: str
        original_value: Any
        modified_value: Any

    StreamAttributeDifference(typing.namedtuple):
        stream_type: StreamType
        attribute: str
        original_value: Any
        modified_value: Any
        original_stream_index: Optional[int]
        modified_stream_index: Optional[int]

    StreamDifference(typing.namedtuple):
        original_stream_index: Optional[int]
        modified_stream_index: Optional[int]
    ```

(1) or (2) would be the quickest and we could add it as simple change in this pull request.

(3) is nicer in my opinion but it's a refactor. Changes would be still simple but there would be more of them so if you want this we'd do it in a separate pull request.

nieznanysprawiciel: Hmmm, first version looks best.
But if you store enum directly, what happens if you print this dict. Will it print enum string value or something human unreadable ?

tests/apps/ffmpeg/task/test_ffprobe_report.py
cameel: For simple assumptions like the value of an individual field - yeah. There's not much difference between comparing width and setting your own width value. But what about more complex cases like `assert len(RAW_REPORT_WITH_MPEG4['streams']) == 2`? Checking this assumption with a single assert is clean and readable. Hard-coding the whole 'streams' list in the test - not so in my opinion. It defeats the purpose of having a shared fixture. Sharing fixtures does make multiple tests dependent on their content but in my opinion that's a reasonable trade-off.

That's my opinion but the decision here is up to you. If you insist, we'll change it. My suggestion would be a compromise - setting required values in the test in cases like the one you pointed out - where that does not require hard-coding big dicts. Keeping `assert` in cases where this would make tests overly long.

tests/apps/ffmpeg/task/ffprobe_report.py
cameel: Actually, we agreed with @pawelkisielewicz that he'll take care of review fixes while I'm preparing the final pull requests. So we can take care of this one along with the stuff you pointed out in other comments.

cameel:
> There was question, if our integration tests should be rewritten to new model. I think, most of them shouldn't, because they check other things, before task is created and ran. Probably test_simple_case could be rewritten, because it runs whole task.

Sure. We'll rewrite `test_case_simple` and no other tests in that case.

> Generally tests should be in 2 groups. One group will be ran on CI and consists probably of all tests that are now.
> Second group test extended video set and we will run it only sometimes.

In that case I think we'll split it onto two separate test classes (in the same file - so just a matter of adding a `class` header between the current methods; no refactor) to make the separation clear.

tests/apps/ffmpeg/task/test_ffprobe_report.py
nieznanysprawiciel: You are right, this length check can ramain.
But in case where this is only substitution of value it is worth changing, I think

nieznanysprawiciel:
I would separate these tests for extended video testset to other file.
But in this PR don't move anything. These tests you wrote until now, could be in this basic test set on CI.

Probably test_case_simple checks things, that your tests check too. Maybe it can be removed entirely.

tests/apps/ffmpeg/task/ffprobe_report_set.py
cameel: It will print something like this:
``` python
<DiffReason.DifferentAttributeValues: 'Different attribute values'>
```
so a bit long but still human-readable.

Alternatively, we could put `DiffReason.DifferentAttributeValues.value` in diffs and decode it back from string to an enum in the rare situations where we have to do something based on its value (rather than just print it).

nieznanysprawiciel: Hmmm, then I would prefere this second option.

tests/apps/ffmpeg/task/test_ffprobe_report.py
cameel: OK.

tests/apps/ffmpeg/task/ffprobe_report_set.py
cameel: OK

tests/apps/ffmpeg/task/test_ffmpegintegration.py
cameel: OK

cameel:
> I would separate these tests for extended video testset to other file.
> But in this PR don't move anything. These tests you wrote until now, could be in this basic test set on CI.

OK. So just so that I'm sure I understood this correctly:
- In this pull request:
    - Don't move any tests outside of `test_ffmpegintegration.py` or even inside the file.
    - Creating a separate class without moving things around is OK.
- In subsequent pull requests:
    - Leave the current tests for codec/resolution/frame rate change in `test_ffmpegintegration.py` and have them test only `test_video.mp4` and `test_video2.mp4`. Keep them there so that they can run on CI.
    - Create a separate module and copy the current tests for codec/resolution/frame rate change tests there. The files from the extended video set should be added to this module.

Is that correct?

> Probably test_case_simple checks things, that your tests check too. Maybe it can be removed entirely.

Well, the new tests don't cover what it does yet (conversion to H.265; they convert to H.264) but will when I extend them.

Actually, I think I'd keep this test as is. Since it's so simple it might be much easier to debug than the other ones when things break.

cameel:
@pawelkisielewicz To summarize, here's a final list of fixes to do. Look at the individual threads for details.

1) [Saving `FfprobeReportSet` to a file in the test directory](https://github.com/golemfactory/golem/pull/4249#discussion_r289033147)
2) [Hard-coding important fixture values in tests instead of verifying them with `assert`](https://github.com/golemfactory/golem/pull/4249#discussion_r289423750)
3) [Constants for `reason` field in the diff](https://github.com/golemfactory/golem/pull/4249#discussion_r289791824)
4) [Move `FfprobeReport`, `FfprobeReportSet` and `SimulatedTranscodingOperation` and their tests to a separate module](https://github.com/golemfactory/golem/pull/4249#pullrequestreview-244856669)
5) [Better names for modified/original in `FfprobeReport` diff methods](https://github.com/golemfactory/golem/pull/4249#discussion_r288981901)
6) [Copy the new integration tests to a separate directory but keep them also in `test_ffmpegintegration` for CI](https://github.com/golemfactory/golem/pull/4249#pullrequestreview-244856669) (later, in a separate pull request)

nieznanysprawiciel:
>     * Creating a separate class without moving things around is OK.
I don't see reason to separate class.
Rest is correct.

>    * Move FfprobeReport, FfprobeReportSet and SimulatedTranscodingOperation and their tests to a separate module

I meant directory.

cameel:
> I don't see reason to separate class.

OK. I thought it might be useful to separate the tests into two groups. But I guess just copying them to another module achieves the same thing.

> I meant directory.

Right. My bad. Fixed.

tests/apps/ffmpeg/task/ffprobe_report.py
cameel: Done.

tests/apps/ffmpeg/task/test_ffprobe_report.py
cameel: Done.

tests/apps/ffmpeg/task/ffprobe_report_set.py
cameel: Done.

Actually @pawelkisielewicz has pointed out that we can still have an enum and just always use it with `.value` using the values a string constants. This gives us best of (1) and (2). So that's what we went with.

cameel: Done.

Actually @pawelkisielewicz has pointed out that we can still have an enum and just always use it with `.value` using the values a string constants. This gives us best of (1) and (2). So that's what we went with.

cameel:
I've just pushed 5 commits with review fixes from @pawelkisielewicz that address points 1-5 from the above.

In addition to that there are 7 new commits with a few extra tweaks that make more sense here than in a separate pull request. I strongly recommend just looking at the individual commits. They're small and have descriptions explaining what they do.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
cameel: Done. It's now enabled by default and the output is stored in a file called `ffmpeg-integration-test-transcoding-diffs.md`.

tests/apps/ffmpeg/task/utils/ffprobe_report.py
```diff
@@ -0,0 +1,653 @@
...
+                getattr(expected_report, attribute),
+            )
+
+            if attribute not in excludes and expected_value != actual_value:
+                diff_dict = {
+                    'location': 'format',
+                    'attribute': attribute,
+                    'actual_value': actual_value,
+                    'expected_value': expected_value,
+                    'reason': "Different attribute values",
```
nieznanysprawiciel: Here should be used DiffReason.DifferentAttributeValues.value

tests/apps/ffmpeg/task/utils/ffprobe_report.py
```diff
@@ -0,0 +1,653 @@
...
+            if (
+                    attribute not in excludes and
+                    expected_value != actual_value
+            ):
+                diff_dict = {
+                    'location': actual_stream_report.codec_type,
+                    'attribute': attribute,
+                    'actual_value': actual_value,
+                    'expected_value': expected_value,
+                    'reason': "Different attribute values",
```
nieznanysprawiciel: DiffReason.DifferentAttributeValues.value

tests/apps/ffmpeg/task/utils/ffprobe_report.py
```diff
@@ -0,0 +1,653 @@
...
+                    diff_dict['expected_stream_index'] = expected_stream_index
+
+                diffs += shortest_diff
+                unmatched_reports.remove(matched_expected_stream_id)
+            else:
+                diffs.append({
+                    'location': actual_stream_reports[0].codec_type,
+                    'actual_stream_index': actual_idx,
+                    'expected_stream_index': None,
+                    'reason': "No matching stream",
```
nieznanysprawiciel: DiffReason.NoMatchingStream.value

tests/apps/ffmpeg/task/utils/ffprobe_report.py
```diff
@@ -0,0 +1,653 @@
...
+                })
+
+        for expected_idx in unmatched_reports:
+            diffs.append({
+                'location': expected_stream_reports[0].codec_type,
+                'actual_stream_index': None,
+                'expected_stream_index':
+                    expected_stream_reports[expected_idx].index,
+                'reason': "No matching stream",
```
nieznanysprawiciel: DiffReason.NoMatchingStream.value

tests/apps/ffmpeg/task/test_ffprobe_report.py
nieznanysprawiciel: But there are places where still string are used instead of enum.value.
I placed comments in code

tests/apps/ffmpeg/task/ffprobe_report_set.py
nieznanysprawiciel: But there are places where still string are used instead of enum.value.
I placed comments in code

tests/apps/ffmpeg/task/utils/ffprobe_report.py
cameel: Yeah, looks like @pawelkisielewicz missed those. I just pushed a fix. Two commits:
- One moves `DiffReason` to `ffprobe_report.py` (it should have been there to begin with).
- The other replaces are remaining strings with values from the enum.

tests/apps/ffmpeg/task/utils/ffprobe_report.py
cameel: Done. Already answered in: https://github.com/golemfactory/golem/pull/4249#discussion_r292504506

tests/apps/ffmpeg/task/utils/ffprobe_report.py
cameel: Done. Already answered in: https://github.com/golemfactory/golem/pull/4249#discussion_r292504506

tests/apps/ffmpeg/task/utils/ffprobe_report.py
cameel: Done. Already answered in: https://github.com/golemfactory/golem/pull/4249#discussion_r292504506

PR #4251, opened 2019-05-30
fix incorrect volumes

weaselix:
It looks like someone was trying to introduce the following optimization: 
> Let's avoid copying resources before sending them to provider and just pack them to the ZIP _in place_. 

However, it does not work when it comes to the verification process. A verificator assumes that resources are copied to `ComputerRes`. 

Issue was resolved by copying needed resources to `ComputerRes` but, in the future, we need to resolve that in another way. From the other hand, mounting user directories directly to virtualmachine / docker filesystem seems to be not reasonable.

### DIFF
--- a/golem/verificator/blender_verifier.py
+++ b/golem/verificator/blender_verifier.py
@@ -1,4 +1,6 @@
+import shutil
 from datetime import datetime
+from pathlib import Path
 from typing import Type
 
 import logging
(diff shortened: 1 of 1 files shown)


magdasta:
Needs to be manually tested on windows

Krigpl:
> It looks like someone was trying to introduce the following optimization:

Not sure if passive aggressive or...

Could you point to the specific code in the commit which is supposedly trying to introduce this optimization?

weaselix:
@Krigpl,
I have written it basing on the effects/observations:

1) We expect that the requestor has copied resources to the `ComputerRes`.

2) However, in the `develop` revision there is no copied resources in `ComputerRes`. However, there is a `zip` that contains resources. So, this is why (basing on those observations) I assumed that it was optimization.

Because we (me) introduce copying resources again I intentionally pointed it out in my pull request to highlight that fact. Copying resources is, in some sense, a regression (if it was an optimization actually).  And this is why I mentioned it. 

I cannot point specific code because:

1) I am not sure that it is optimization
2) I guess that there were a lot of commits from the last release to that day so I would rather to avoid reading them all. 

So, to sum up, I wrote that note to point that I am making (possibly) a regression.

I cannot understand why do you treat it as a `passive aggressive` and why do you treat it personally?

Krigpl:
This explanation is much more descriptive and provides more context than the initial summary of the commit and it'd be great to have it from the beginning, I wouldn't have had any doubts.

I'm glad it wasn't passive aggressive and I didn't take it personally, that's why I put "not sure" :) but in general in written text sentences like "looks like somebody did something [bad]" can be interpreted as such, considering you cannot infer the context from the tone of the voice. That's why I consider the second explanation much better than the initial one.

Krigpl (approved):

ederenn:
causes: https://github.com/golemfactory/golem/issues/4259

Failure docker.errors.APIError: 500 Server Error: Internal Server Error ("VolumeDriver.Mount: exit status 255")

PR #4255, opened 2019-05-31
Transcoding: segment split and concat demuxer merge

cameel:
This pull request updates the `ffmpeg-experimental` Docker image to correctly use the new merge method introduced in https://github.com/golemfactory/ffmpeg-tools/pull/3 and Golem code to pass video file names rather than playlist names to the split and merge commands.

A few things of note:
1) I have disabled one of the target FPS values in the FPS change test. The new method seems to have problems when transcoding `test_video.mp4` to 12.5 FPS. The video should be converted from 500 frames at 25 FPS and 20 seconds to 250 frames at 12.5 FPS and 20 seconds. Instead the result has 254 frames and 20.32 seconds. This is odd and will probably need to be investigated. But since won't support weird values like 12.5 FPS anyway I went ahead and just disabled this failing test. One of the upcoming pull requests to `ffmpeg_tools` will contain validation that will prevent transcoding to odd frame rates like this and then I'll add tests for the supported values.
2) It might seem that the patch changes the naming of the segments produced by split command from `segment[num=<n>]` to `segment_<n>`. Note however that it's not the case. Until now the split command was producing a set of `segment[num=<n>].m3u` playlists, each containing a single `segment[num=<n>].mp4` file. I simply removed the playlists and the file names are now visible in the code.

### Dependencies
This pull request is based on #4249. Technically, it could work directly on `CGI/transcoding/master` but I think it's better if it's merged after that branch.

For the code to work you also need a `golemfactory/ffmpeg-experimental` image built with https://github.com/golemfactory/ffmpeg-tools/pull/3.

**NOTE**: I have set the base branch of this pull request to `enhanced-transcoding-integration-tests-with-ffprobe-report` (#4249) rather than `CGI/transcoding/master`. Otherwise you would see changes from that branch here. The downside is that github now expects it to be merged into that base branch rather than into `CGI/transcoding/master` and will simply close the pull request (instead or marking it as "merged") if you merge it into master. To prevent this please remember to change the base branch back to `CGI/transcoding/master` before you merge.


cameel:
Note that unit tests on buildbot won't pass until a new version of the `ffmpeg-experimental` image is released.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
```diff
@@ -52,16 +56,50 @@ def do_transcode(track, targs, output, use_playlist):
...
-        os.path.basename(outputfilename))
-    merged = join_playlists(chunks)
-    merged_filename = os.path.join(RESOURCES_DIR,
-                                   output_playlist + ".m3u8")
-    file = open(merged_filename, 'w')
-    file.write(merged.dumps())
-    file.close()
-
-    commands.merge_videos(merged_filename, outputfilename)
+    def select_transcoded_video_paths(output_file_paths, output_extension):
```
nieznanysprawiciel: These functions don't capture any value from merge function as I see. Can you move it outside of do_merge function?

cameel: Sure. I defined them inside because this function was getting big and I just wanted to make it more readable. I did not think they were generic enough to make sense on their own. But if you prefer them outside I can move them.

nieznanysprawiciel:
I pushed replication of bug. All tests use test_video2 instead of test_video2.mp4.
We can't assume user provides video with valid extension :(

The problem is with building paths. They build path like this path_to_file%d and this causes docker fail in splitting phase.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
nieznanysprawiciel: Maybe you could split this file into separate files for each command or something like that.

cameel: Yeah, that's what I'd normally do but I did not think it was worth it just for these three small functions. But I'll split it if it turns out later that I need to add more. For now I think just getting them out of `do_merge()` is fine.

EDIT: I misread your comment. I meant that I'd normally put utility functions in a separate module if they were bigger. As for putting each command in a separate module - I have no strong opinion. I don't think they're big enough for this yet but it's not a bad idea either. Personally I'd leave it as is but if you want me to, I can split them.

cameel:
> I pushed replication of bug. All tests use test_video2 instead of test_video2.mp4.
We can't assume user provides video with valid extension :(
>
> The problem is with building paths. They build path like this path_to_file%d and this causes docker fail in splitting phase.

OK. I'll take a look at this tomorrow and prepare a fix.

cameel:
This is a design problem and requires a design decision.

The issue here is that my split method tries to preserve the original container until the transcoding step and does so based on the input file extension. `ffmpeg -f segment` simply fails if you do not specify the output container format - either by giving it an extension (which is missing) or giving it with the `-segment_format` option (which I'm not doing here).

Unfortunately simply switching to `-segment_format` (or `-f`) has drawbacks too. `format_name` reported by ffprobe cannot be directly passed to ffmpeg with that option because it contains demuxer name while the option expects a muxer name. In some cases (e.g. `mpegts`, `flv`, `asf`) the muxer and demuxer have the same name. But there are very significant exceptions, e.g. `matroska,webm` or `mov,mp4,m4a,3gp,3g2,mj2` - in those cases you have to choose one of several possible muxers and ffprobe does not give you any hint to decide which one is correct. If you choose wrong, the container might not be able to store the streams from the input file. For example `webm` is a very limited subset of `matroska` and by design does not support all the codecs the latter does. Relying on the extension has other drawbacks but at least does not introduce this ambiguity.

Another option is to always use the same intermediate container type. That's what your split method was doing with the `mpegts` container but this severely limited the range of supported codecs. The conversion to an intermediate container (even with `-codec copy`) might also strip some metadata we would otherwise be able to preserve. This might still be a good option if we can find a versatile container. I think that `matroska` might be a viable choice because it's designed to support basically anything. So doing it this way is another possible solution but it might significantly affect the output. I'd rather do that as a separate task, not as a fixup in a pull request that several other pull requests depend on. Also, if we want this, it's probably best to introduce it after the pull request that adds extract and replace steps because then only the video streams will need to go through this intermediate container.

So, to summarize the options for fixing this:
1) Short-term fix (in this pull request):
    1) Patch this problem temporarily by requiring the input to have an extension. 
    2) Just leave it as is until the long-term fix.
2) Long-term fix (in a separate pull request):
    1) Keep preserving the container but switch to the `-segment_format` option.
        - This will never work perfectly for `matroska`, `webm`, `mov`, `mp4`, `m4a`, `3gp`, `3g2`, `mj2`, `svcd` and possibly some other containers because we need to guess the muxer. But it might work well enough in many cases.
    2) Always use `matroska` as the intermediate container.
        - The risk is that while `matroska` supports lots of codecs, there might still be some we need that it does not.
    3) Something between (i) and (ii): instead of trying to preserve the original container, for every demuxer find an intermediate container type that works. This would give us more flexibility than simply assuming that `matroska` can handle everything.

My suggestion would be to go with (1.ii) and (2.ii) because that's simplest to implement and switch to (2.iii) if `matroska` turns out to be problematic in some cases.

cameel:
One more thing. There's an interesting quirk with `-segment_format`. It allows you to specify the format and ignores the extension **but** it still requires the extension to be present and it must be one of the supported extensions.

This has even been reported in ffmpeg bugtracker as [#4483 -segment_format not works with some filenames](https://trac.ffmpeg.org/ticket/4483). I'm experiencing the exact same thing but the issue was closed as invalid. I wonder why. You can work around it by giving the file an extension but it still seems like bug to me.

Try this:
``` bash
ffmpeg                          \
    -v              error       \
    -i              test_video2 \
    -codec          copy        \
    -f              segment     \
    -segment_format matroska    \
    -segment_time   10          \
    test_video2-segment%d
```

Result:
```
Output file #0 does not contain any stream
```

Or give it a dummy extension:
``` bash
ffmpeg                          \
    -v              error       \
    -i              test_video2 \
    -codec          copy        \
    -f              segment     \
    -segment_format matroska    \
    -segment_time   10          \
    test_video2-segment%d.someext
```

Again:
```
Output file #0 does not contain any stream
```

But this works and produces a valid mkv container (not mp4, even though the extension is mp4):
``` bash
ffmpeg                          \
    -v              error       \
    -i              test_video2 \
    -codec          copy        \
    -f              segment     \
    -segment_format matroska    \
    -segment_time   10          \
    test_video2-segment%d.mp4

ffprobe                      \
    test_video2-segment0.mp4 \
    -v            error      \
    -show_entries format=format_name,format_long_name,filename
```
```
[FORMAT]
filename=test_video2-segment0.mp4
format_name=matroska,webm
format_long_name=Matroska / WebM
[/FORMAT]
```

So if we use `-segment_format` we'll have to give files spurious extensions just to prevent ffmpeg from failing. At least for the intermediate files.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
nieznanysprawiciel: We could group commands and leave ffmpeg_task.py only as commands dispatcher. But it doesn't matter so much actually.

nieznanysprawiciel:
As I understand, problem with leaving the same container is only with matroska,webm or mov,mp4,m4a,3gp,3g2,mj2.
I think we should preserve container for most cases and choose intermediate container for these groups (each group could have different intermediate container). This would require another map containing only these exceptional cases.

Edit:
I can't merge this pull request before this fix will be ready. It breaks CGI/transcoding/master. Our master branch must work with our transcoding web service, that we are creating at this moment.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
```diff
@@ -52,16 +56,50 @@ def do_transcode(track, targs, output, use_playlist):
...
+
+    def sorted_transcoded_video_paths(transcoded_video_paths):
+        path_index = {int(re.findall(TRANSCODED_VIDEO_REGEX, path)[0]): path
+                      for path in transcoded_video_paths}
+        return [value for key, value in sorted(path_index.items())]
+
+
+    def build_and_store_ffconcat_list(chunks, output_filename, list_basename):
+        assert len(chunks) >= 1
+        assert len(set(os.path.dirname(chunk) for chunk in chunks)) == 1, \
```
nieznanysprawiciel: Check if these asserts shouldn't be replaced with exceptions.

cameel:
> As I understand, problem with leaving the same container is only with matroska,webm or mov,mp4,m4a,3gp,3g2,mj2.

Not only but those are the most prominent ones. I suspect that these formats are important enough for you that any solution that does not handle them properly would not be acceptable.

Another, less prominent example is `svcd` (the container format of `.vob` files). Its demuxer is `mpeg` but if you use it as a muxer, you won't get a SVCD file back. You'll get a MPEG file. You have to specifically use `-f svcd` to get a SVCD container. I'm not sure how much of a problem this would be since these formats are pretty similar (they are handled by a single demuxer after all) but the fact is that the container does change and that this case is different from the cases where the demuxer name is just a combination of several muxer names.

> I think we should preserve container for most cases and choose intermediate container for these groups (each group could have different intermediate container). This would require another map containing only these exceptional cases.

So it's basically a variant of option 2.iii. Seems good to me.

> I can't merge this pull request before this fix will be ready. It breaks CGI/transcoding/master. Our master branch must work with our transcoding web service, that we are creating at this moment.

By breaking master do you mean the test modifications that removed the extension from `test_video2.mp4`? It that's the case then I think that just adding a temporary validation that rejects files without extension would be an acceptable stopgap. Or do you mean that the service is actually getting valid files without extensions from users and is expected to handle them? That would require an actual fix.

Anyway, the biggest issue with doing a bigger fix here are the two pull requests sitting on top of this one. I'd prefer not to introduce any big, conflicting changes between them. I've split them mostly for ease of review anyway. So how about I implement the fix on top of those three and when it's done we merge all of them at once? I see that there were no major change requests in them so this should not hold back this one too much. I have a change that adds support for the `-f` option ready in another branch and I'll need it for the fix. Rebasing it on top of those three should be no problem but going deeper, down to this one would unnecessarily complicate things in my opinion.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
cameel: - The one that requires at least one chunk - maybe. Like I explained in another comment, `ffmpeg-scripts` does not validate its input. I'm not validating it either. ffmpeg will (probably) fail if you try to use it without giving it any input files so I'm just making it fail earlier, louder and in a way that makes it apparent that this is not expected to work.

    I can make it an exception if you want but I think the empty chunk list should be rejected long before we get here. Every command should have a function that performs full validations for its input.
- The one that ensures that chunks are in the same directory - maybe.  I mean, it's entirely possible to make it work with chunks in different directories. I just thought that it would be wasted effort because the current implementation of the task does store them in a single directory anyway and getting the ffmpeg command just right requires some experimentation (it did not work when I tried it the most obvious way; but I did not try too hard). Instead I left this assertion here so that it does not go unnoticed if the task implementation changes.

    I can make it an exception but, again, in my opinion this should be caught at a different layer. Or simply improved to work with arbitrary paths.

nieznanysprawiciel: Ok, than change it to exceptions and handle these errors on other level too.
The second change only to exception.

nieznanysprawiciel:
Service is renaming files for security reasons and doesn't preserve extension. This way we avoid validating paths. But even if we preserved extension, we can't count on it, that extension will be the same as real container. We would have to check it, so it's better to just ignore extension.

Ok you can implement it on the top of these PRs. but please implement this before other PRs you are working on, I want to merge these changes (all three PRs) as soon as possible.

Today I will review third pair of PRs

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
cameel: Done. No more assertions. The code now raises `commands.InvalidArgument` instead.

I think that `do_merge()` (or a separate validating function called from it) is the place where the input should be validated so I added those checks there too.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
cameel: Done. The helpers are now outside of `do_merge()`.

As for grouping commands or creating separate modules for them - I can do it that way if you want me to but that would be a pretty invasive change given the pull requests on top of this one it so I'd prefer to put it off until this series of pull requests get merged.

cameel:
> Ok you can implement it on the top of these PRs. but please implement this before other PRs you are working on, I want to merge these changes (all three PRs) as soon as possible.

Sure. It's going to be like this:

1) #4255 (this PR)
2) #4256 (extract and replace)
3) #4257 (remove m3u)
4) #4315 (support for `-f` option)
5) **PR with the fix**

And when we're done with those, the code will be ready for the final PR that adds all files, codecs and the workarounds I've been working on. The code is ready in `transcoding-all-files-codecs-framerates-resolutions` branch and is pretty much ready for review but I think I'll wait for the PRs above to be done before creating another one on top of them.

nieznanysprawiciel (commented):
Approved.
I will merge changes and update images after you resolve conflicts and prepare PR with extensions fix.

cameel:
Great. Is it OK for me to solve conflicts by rebasing all those pull requests to be again on top of one another? That's the cleanest way to do this in my opinion. When it's all in one line I can easily keep track of which branch has which commits. Merging 4 pull requests all into one another is going to make it a total mess.

nieznanysprawiciel:
Ok you can rebase it as you want this will help to read code for sure. But on the end I will merge them into one, because I need fix fo videos without extensions to merge everything and test

cameel:
You mean squashing the whole branch into a single commit? Why would you do that? Actually, I see that this is how #4249 got merged. That doesn't seem good.

I'm heavily using git branches in development, often having multiple with different variants of the feature, and I very much rely on being able to see whether a particular commit went into particular branch. I also take care to keep only related changes in a single commit which adds important context (e.g. in `git blame`) to see why I did each change. It's like a single, atomic step. I can easily move, delete or copy it because it contains only relevant changes an nothing more or less. Squashing them into one big commit erases all that information and makes it much harder to work with the code. Could we please reverse that?

If it's the matter of merging being problematic (conflicts and all), I can do that for you, that's no problem at all. It just seems a bit extreme to throw out all the history if that's the reason.

nieznanysprawiciel:
You can reverse squash and merge it back without squash if you want.

I meant rather, that I will merge all commit as one pull request. Otherwise I will have 4 not working commits on CGI/transcoding/master without this fix to video extensions. Moreover i want to test better final pull request with all changes.

PR #4256, opened 2019-05-31
Transcoding: extract and replace commands

cameel:
This pull request updates the `ffmpeg-experimental` Docker image with new commands that allow extracting video streams from the input file before split and inserting them back after merge. It then changes the transcoding task implementation to use those new commands.

### Dependencies
This pull request depends on #4255, which enables the new split and merge methods.

For the code to work you also need a `golemfactory/ffmpeg-experimental` image built with https://github.com/golemfactory/ffmpeg-tools/pull/4 (extract and replace command implementation).

**NOTE**: I have set the base branch of this pull request to `transcoding-segment-split-and-concat-demuxer-merge` (#4255) rather than `CGI/transcoding/master`. Otherwise you would see changes from that branch here. The downside is that github now expects it to be merged into that base branch rather than into `CGI/transcoding/master` and will simply close the pull request (instead or marking it as "merged") if you merge it into master. To prevent this please remember to change the base branch back to `CGI/transcoding/master` before you merge.

### DIFF
--- a/apps/transcoding/task.py
+++ b/apps/transcoding/task.py
@@ -98,12 +98,8 @@ class TranscodingTask(CoreTask):  # pylint: disable=too-many-instance-attributes
         input_file = self.task_resources[0]
 
         stream_operator = StreamOperator()
-        video_only_file = stream_operator.extract_video_streams(
+        chunks, video_metadata = stream_operator.extract_video_streams_and_split(
             input_file,
-            dir_manager,
-            task_id)
-        chunks, video_metadata = stream_operator.split_video(
-            video_only_file,
             self.task_definition.subtasks_count,
             dir_manager,
             task_id)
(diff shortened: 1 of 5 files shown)


cameel:
Note that unit tests on buildbot won't pass until a new version of the `ffmpeg-experimental` image is released.

apps/transcoding/task.py
```diff
@@ -161,15 +162,23 @@ def _merge_video(self):
         logger.info('Merging video [task_id = %s]',
                     self.task_definition.task_id)
 
+        output_basename = os.path.basename(self.task_definition.output_file)
+
+        assert len(self.task_definition.resources) == 1, \
+            "Assumption: input file is the only resource in a transcoding task"
+        input_file = next(iter(self.task_definition.resources))
```
nieznanysprawiciel, line 169: If we use input file to replace video stream with merged videos, how does changing container format works ?

cameel: The original input file and the merged file with the video stream are used as inputs and the output goes to a new file. I'm using the `-map` option to tell it which streams to take from which input. ffmpeg seems to have no problem when the inputs are of two different container types. It just demuxes each input separately and then muxes the selected streams into the output.

nieznanysprawiciel (approved):
I will merge all PRs at once when fix for videos extensions will be ready

nieznanysprawiciel:
Merged in PR:
https://github.com/golemfactory/golem/pull/4407

PR #4264, opened 2019-06-03
Finish rename of rpc command comp.task.subtasks.restart

maaktweluit:
Broken by #4247 

Found by [integration tests](https://buildbot.golem.network/buildbot/#/builders/15/builds/707)

@zakaprov:
- Please use a global find in the project when you rename a rpc command
- Maybe keep the old command with a deprecation warning
- Please also check other projects that use this rpc like golem-electron and golemcli-rust


kmazurek:
@maaktweluit This was a change required by Electron, I missed updating the CLI side, sorry about that!

PR #4270, opened 2019-06-04
Less verbose faucet error messages

Krigpl:
Golem needs 6 confirmations until it sees the funds, while faucet doesn't. Make golem not repeatedly request eth after the first requests. It'll result in fewer confusing messages in the logs.


maaktweluit (approved):
<3

PR #4276, opened 2019-06-05
node_integration_tests - use granary for key reuse

maaktweluit:
- Separated out the key reuse feature in its own file ( commit 1 )
- Extended it to pick between local files and the remote granary by specifying the granary hostname ( commit 2).

The granary is disabled by default.

### DIFF
--- a/scripts/node_integration_tests/key_reuse.py
+++ b/scripts/node_integration_tests/key_reuse.py
@@ -41,46 +41,62 @@ class NodeKeyReuseConfig:
     instance: 'Optional[NodeKeyReuseConfig]' = None
     provider: Optional[NodeKeyReuseBase]  = None
     enabled: bool = True
+
+    # Provider specific variables
     granary_hostname: Optional[str] = None
+    local_reuse_dir : Optional[Path] = None
 
     @classmethod
-    def get(cls, test_dir: Path):
+    def get(cls):
+        _log("NodeKeyReuseConfig.get() called.")
         if not cls.instance:
             cls.instance = cls()
             if cls.granary_hostname:
                 print("key_reuse - granary selected:", cls.granary_hostname)
                 cls.provider = NodeKeyReuseGranary(cls.granary_hostname)
             else:
-                print("key_reuse - local folder selected:", test_dir)
-                cls.provider = NodeKeyReuseLocalFolder(test_dir)
+                print("key_reuse - local folder selected:", cls.local_reuse_dir)
+                assert cls.local_reuse_dir is not None, \
+                    "ERROR: No folder for reuse, call set_dir() first"
+                cls.provider = NodeKeyReuseLocalFolder(cls.local_reuse_dir)
         return cls.instance
 
     @classmethod
+    def set_dir(cls, dir: Path):
+        _log("NodeKeyReuseConfig.set_dir() called. dir=", dir)
+        assert cls.provider is None, "ERROR: Can not set_dir() after get()"
+        cls.local_reuse_dir = dir
+
+    @classmethod
     def begin_test(cls, datadirs: Dict[NodeId, Path]):
+        _log("NodeKeyReuseConfig.begin_test() called. dirs= ", datadirs)
         if cls.enabled and cls.provider:
-            _log("NodeKeyReuseConfig.begin_test() called. dirs= ", datadirs)
             cls.provider.begin_test(datadirs)
 
     @classmethod
     def end_test(cls):
+        _log("NodeKeyReuseConfig.end_test() called.")
         if cls.enabled and cls.provider:
-            _log("NodeKeyReuseConfig.end_test() called.")
             cls.provider.end_test()
 
     @classmethod
     def disable(cls):
+        _log("NodeKeyReuseConfig.disable() called.")
         cls.enabled = False
 
     @classmethod
     def enable(cls):
+        _log("NodeKeyReuseConfig.enable() called.")
         cls.enabled = True
 
     @classmethod
     def reset(cls):
+        _log("NodeKeyReuseConfig.reset() called.")
         cls.instance = None
 
     @classmethod
     def set_granary(cls, hostname):
+        _log("NodeKeyReuseConfig.set_granary() called. host=", hostname)
         cls.granary_hostname = hostname
 
 
@@ -92,8 +108,8 @@ class NodeKeyReuseLocalFolder(NodeKeyReuseBase):
         self._first_test = True
 
     def begin_test(self, datadirs: Dict[NodeId, Path]) -> None:
-        self.datadirs = datadirs
         _log("NodeKeyReuseLocalFolder.begin_test() called.")
+        self.datadirs = datadirs
 
         if not self._first_test:
             _log("Moving keys from reuse-dirs to data-dirs")
@@ -113,7 +129,7 @@ class NodeKeyReuseLocalFolder(NodeKeyReuseBase):
     def _recycle_keys(self) -> None:
         # this is run before running second and later tests
         for i, (node_id, datadir) in enumerate(self.datadirs.items()):
-            _log("NodeKeyReuseLocalFolder._copy_keystores() loop. "
+            _log("NodeKeyReuseLocalFolder._recycle_keys() loop. "
                  "i", i, 'node_id', node_id, 'datadir', datadir)
             reuse_dir = self.dir / str(i)
             if not reuse_dir.exists():
@@ -180,7 +196,6 @@ class NodeKeyReuseGranary(NodeKeyReuseBase):
             return
 
     def _recycle_keys(self) -> None:
-        _log("Recycle keys")
         # this is run before tests
         for i, (node_id, datadir) in enumerate(self.datadirs.items()):
             account = self.granary.request_account()
@@ -239,7 +254,6 @@ class NodeKeyReuseGranary(NodeKeyReuseBase):
 
     @staticmethod
     def _save_private_key(key, key_path: Path, password: str) -> None:
-        _log("_save_private_key() password=", password)
         keystore = create_keyfile_json(
             key,
             password.encode('utf-8'),
--- a/scripts/node_integration_tests/key_reuse.py
+++ b/scripts/node_integration_tests/key_reuse.py
@@ -14,12 +14,19 @@ from scripts.node_integration_tests.playbooks.test_config_base import NodeId
 KEYSTORE_DIR = 'rinkeby/keys'
 KEYSTORE_FILE = 'keystore.json'
 TRANSACTION_FILE = 'tx.json'
-PASSWORD = 'dupa.8'
+PASSWORD = 'goleM.8'
 
 _logging = False
 
 
-class NodeKeyReuseBase():
+def _log(*args):
+    # Private log function since pytest.tearDown() does not print logger
+    if _logging:
+        print(*args)
+
+
+class NodeKeyReuseBase:
+    ### Base class for key reuse providers
     def begin_test(self, datadirs: Dict[NodeId, Path]) -> None:
         raise NotImplementedError()
 
@@ -27,8 +34,11 @@ class NodeKeyReuseBase():
         raise NotImplementedError()
 
 
-class NodeKeyReuse:
-    instance: 'Optional[NodeKeyReuse]' = None
+class NodeKeyReuseConfig:
+    # Configuration singleton class for reusing keys
+    # Selects the right provider ( subclass of NodeKeyReuseBase )
+    # Based on command line arguments ( TBA: and enviromnet vars )
+    instance: 'Optional[NodeKeyReuseConfig]' = None
     provider: Optional[NodeKeyReuseBase]  = None
     enabled: bool = True
     granary_hostname: Optional[str] = None
@@ -48,15 +58,13 @@ class NodeKeyReuse:
     @classmethod
     def begin_test(cls, datadirs: Dict[NodeId, Path]):
         if cls.enabled and cls.provider:
-            if _logging:
-                print("NodeKeyReuse.begin_test() called. dirs= ", datadirs)
+            _log("NodeKeyReuseConfig.begin_test() called. dirs= ", datadirs)
             cls.provider.begin_test(datadirs)
 
     @classmethod
     def end_test(cls):
         if cls.enabled and cls.provider:
-            if _logging:
-                print("NodeKeyReuse.end_test() called.")
+            _log("NodeKeyReuseConfig.end_test() called.")
             cls.provider.end_test()
 
     @classmethod
@@ -77,6 +85,7 @@ class NodeKeyReuse:
 
 
 class NodeKeyReuseLocalFolder(NodeKeyReuseBase):
+    # Key reuse provider implementation that uses a local folder to store keys
     def __init__(self, test_dir: Path):
         self.dir: Path = test_dir / 'key_reuse'
         self.datadirs: Dict[NodeId, Path] = {}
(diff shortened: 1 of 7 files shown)


scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,126 @@
...
+from scripts.node_integration_tests.playbooks.test_config_base import NodeId
+KEYSTORE_DIR = 'rinkeby/keys'
+KEYSTORE_FILE = 'keystore.json'
+
+_logging = False
+
+
+class NodeKeyReuse:
+    instance = None
+    provider = None
```
etam: Type annotations plz

scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,126 @@
...
+        self.datadirs = datadirs
+        if _logging:
+            print("NodeKeyReuseLocalFolder.begin_test() called.")
+        if not self._first_test:
+            if _logging:
+                print("Moving keys from reuse-dirs to data-dirs")
+            self._recycle_keys()
+
+    def end_test(self) -> None:
+        print("NodeKeyReuseLocalFolder.end_test() called.")
```
etam: `if logging:`?

scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,126 @@
...
+_logging = False
+
+
+class NodeKeyReuse:
+    instance = None
+    provider = None
+    enabled = True
+
+    @classmethod
+    def get(cls, test_dir: Path):
```
etam: Not critical: If you call this second time, with different `test_dir` you might get a bit surprising results.

scripts/node_integration_tests/key_reuse.py
```diff
@@ -124,3 +141,96 @@ def _copy_keystore(datadir: Path, reuse_dir: Path) -> None:
...
+class NodeKeyReuseGranary():
+    def __init__(self, hostname: str):
+        self.datadirs: Dict[NodeId, Path] = {}
+        self.granary = Granary(hostname)
+
+    def begin_test(self, datadirs: Dict[NodeId, Path]) -> None:
+        self.datadirs = datadirs
+        if _logging:
+            print("NodeKeyReuseGranary.begin_test() called.")
+        if _logging:
```
etam: redundant if

scripts/node_integration_tests/key_reuse.py
```diff
@@ -124,3 +141,96 @@ def _copy_keystore(datadir: Path, reuse_dir: Path) -> None:
...
+            print("Moving keys from granary to data-dirs")
+            self._recycle_keys()
+
+    def end_test(self) -> None:
+        print("NodeKeyReuseGranary.end_test() called.")
```
etam: `if _logging`?

scripts/node_integration_tests/key_reuse.py
```diff
@@ -124,3 +141,96 @@ def _copy_keystore(datadir: Path, reuse_dir: Path) -> None:
...
+        try:  # unlock the key
+            priv_key = decode_keyfile_json(keystore, PASSWORD.encode('utf-8'))
+        except ValueError:
+            raise WrongPassword
+
+        return Account(priv_key, ts)
+
+    @staticmethod
+    def _save_private_key(key, key_path: Path, password: str) -> None:
+        print("_save_private_key")
```
etam: `if _logging`?
also, those 2 prints can be 1

tests/factories/granary.py
```diff
@@ -0,0 +1,98 @@
...
+
+from golem_messages.cryptography import ECCx
+
+_logging = False
+
+
+class Account:
+    def __init__(self, raw_key, ts):
+        self.key = ECCx(raw_key)
+        random.seed()
```
etam: I don't think this is required.

tests/factories/granary.py
```diff
@@ -0,0 +1,98 @@
...
+
+        completed = subprocess.run(
+            cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            universal_newlines=True)
+        if _logging:
+            print('returncode:', completed.returncode)
+
+        if _logging:
```
etam: redundant if

tests/factories/granary.py
```diff
@@ -0,0 +1,98 @@
...
+            if completed.stderr:
+                print('stderr:', completed.stderr)
+            else:
+                print('stderr: EMPTY')
+
+        if completed.stdout:
```
etam: Suggested flow:
```
if something unusual happened:
    handle error

do the expected stuff
```

etam (changes requested):
Added some comments. Nothing critical though.

etam:
Is the Granary code available somewhere?

maaktweluit:
@etam thanks for the comments! will look into them :)

the granary is available here: https://github.com/golemfactory/golem-granary

scripts/node_integration_tests/key_reuse.py
maaktweluit: added

scripts/node_integration_tests/key_reuse.py
maaktweluit: added

scripts/node_integration_tests/key_reuse.py
maaktweluit: removed

scripts/node_integration_tests/key_reuse.py
maaktweluit: added

scripts/node_integration_tests/key_reuse.py
maaktweluit: added and merged

tests/factories/granary.py
maaktweluit: removed

tests/factories/granary.py
maaktweluit: good one! updated the flow

tests/factories/granary.py
maaktweluit: removed, think this was for an old "feature" to generate a key when none was available.

This is the responsibility of the `golemapp` now ( and maybe later `golem-granary` )

scripts/node_integration_tests/key_reuse.py
```diff
@@ -64,8 +64,15 @@ def enable(cls):
     def set_granary(cls, hostname):
         cls.granary_hostname = hostname
 
+class NodeKeyReuseBase():
+    def begin_test(self, datadirs: Dict[NodeId, Path]) -> None:
+        raise Exception("Not implemented")
```
etam: https://docs.python.org/3/library/exceptions.html#NotImplementedError

maaktweluit: fixed :)

scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,253 @@
...
+    def begin_test(self, datadirs: Dict[NodeId, Path]) -> None:
+        raise NotImplementedError()
+
+    def end_test(self) -> None:
+        raise NotImplementedError()
+
+
+SelfType = TypeVar('NodeKeyReuse')
+class NodeKeyReuse:
+    instance: Optional[SelfType] = None
```
etam: When referencing self type in type annotations, put it in quotes.
https://stackoverflow.com/a/33533514

maaktweluit: Nice! fixed

scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,251 @@
...
+
+from golem.core.keysauth import WrongPassword
+from tests.factories.granary import Granary, Account
+
+from scripts.node_integration_tests.playbooks.test_config_base import NodeId
+
+KEYSTORE_DIR = 'rinkeby/keys'
+KEYSTORE_FILE = 'keystore.json'
+TRANSACTION_FILE = 'tx.json'
+PASSWORD = 'dupa.8'
```
mfranciszkiewicz: Uhm, please change this

scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,251 @@
...
+                cls.provider = NodeKeyReuseGranary(cls.granary_hostname)
+            else:
+                print("key_reuse - local folder selected:", test_dir)
+                cls.provider = NodeKeyReuseLocalFolder(test_dir)
+        return cls.instance
+
+    @classmethod
+    def begin_test(cls, datadirs: Dict[NodeId, Path]):
+        if cls.enabled and cls.provider:
+            if _logging:
```
mfranciszkiewicz: Is it possible to move this `if` + `print` to a function (e.g. `log`) in a shared module (unless `_logging` has to be local).

scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,251 @@
...
+            self._copy_keystores()
+        except FileNotFoundError:
+            print('Copying keystores failed...')
+            return
+
+    def _recycle_keys(self) -> None:
+        if _logging:
+            print("Recycle keys")
+        # this is run before running second and later tests
+        for i, node_id in enumerate(self.datadirs):
```
mfranciszkiewicz: `enumerate` is not needed here

scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,251 @@
...
+        os.makedirs(str(dst_key_dir))
+        self._save_private_key(account.raw_key, dst_key_file, PASSWORD)
+        if account.transaction_store:
+            dst_trans_file.write_text(account.transaction_store)
+
+    def _copy_keystores(self):
+        # this is run after tests
+        # return key to granary as binary private key and transactions.json
+
+        for i, node_id in enumerate(self.datadirs):
```
mfranciszkiewicz: `enumerate` is not needed here

scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,251 @@
...
+            print('Copying keystores failed...')
+            return
+
+    def _recycle_keys(self) -> None:
+        if _logging:
+            print("Recycle keys")
+        # this is run before running second and later tests
+        for i, node_id in enumerate(self.datadirs):
+            account = self.granary.request_account()
+            if account is not None:
```
mfranciszkiewicz, line 202: If it is, a log message might be valuable

scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,251 @@
...
+
+        try:  # read tx.json
+            with open(src_ts_file, 'r') as f:
+                ts = f.read()
+        except FileNotFoundError:
+            if _logging:
+                print('No tx.json, continue')
+        try:  # read keystore.json
+            with open(src_key_file, 'r') as f:
+                keystore = f.read()
```
mfranciszkiewicz:
```suggestion
                keystore = json.load(f)
```

tests/factories/granary.py
```diff
@@ -0,0 +1,90 @@
...
+from eth_utils import encode_hex, decode_hex
+from ethereum.utils import sha3
+
+from golem_messages.cryptography import ECCx
+
+_logging = False
+
+
+class Account:
+    def __init__(self, raw_key, ts):
```
mfranciszkiewicz: Could you add type annotations, please? It's not obvious that `ts` is a string containing JSON.

scripts/node_integration_tests/key_reuse.py
```diff
@@ -0,0 +1,251 @@
...
+
+KEYSTORE_DIR = 'rinkeby/keys'
+KEYSTORE_FILE = 'keystore.json'
+TRANSACTION_FILE = 'tx.json'
+PASSWORD = 'dupa.8'
+
+_logging = False
+
+
+class NodeKeyReuseBase():
```
mfranciszkiewicz: Could you maybe provide some kind of distinction for this base class / child classes? Because `NodeKeyReuseBase` is not a parent of `NodeKeyReuse` but of `NodeKeyReuseLocalFolder` and `NodeKeyReuseGranary`

scripts/node_integration_tests/key_reuse.py
maaktweluit: This has been moved here from somewhere else in the code, but will change it :)

scripts/node_integration_tests/key_reuse.py
maaktweluit: Tried `logger` first but this does not work on `tearDown` in pytest :(

Will make it a private function for now

scripts/node_integration_tests/key_reuse.py
maaktweluit: why not? we need a counter,`i`, for this loop
otherwise it will give `node_id, data_dir` ( i think )

scripts/node_integration_tests/key_reuse.py
maaktweluit: good one! 
it will be just a warning but good to have a heads up the key is not from granary

scripts/node_integration_tests/key_reuse.py
maaktweluit: done

tests/factories/granary.py
maaktweluit: done

scripts/node_integration_tests/key_reuse.py
maaktweluit: - Added comments on all classes
- Renamed `NodeKeyReuse` -> `NodeKeyReuseConfig`

scripts/node_integration_tests/key_reuse.py
maaktweluit: This does not work, `json.load` returns a dict not a string

I think its overkill to use `json.dumps(json.load(f))` so reverted the change for now

maaktweluit: Wait, i see it broke because of the `json.loads()` later in the code, running another test on this

scripts/node_integration_tests/key_reuse.py
mfranciszkiewicz: `for node_id in self.datadirs` will iterate over keys

scripts/node_integration_tests/key_reuse.py
mfranciszkiewicz: Thank you! :)

scripts/node_integration_tests/key_reuse.py
maaktweluit: works! /shy

maaktweluit:
@mfranciszkiewicz thanks for the review, great comments!

I have addressed all of them and resolved the simple fixes, please resolve your other comments when happy with the fixes

scripts/node_integration_tests/key_reuse.py
maaktweluit: but we need the index(/counter), node_id and datadir in this loop to make the storage folders independent of the node_id's used for this test

mfranciszkiewicz: `i` is not used here at all. You're probably thinking about a different method.

scripts/node_integration_tests/key_reuse.py
maaktweluit: fixed!

Protected our future selves from this by splitting the functions and adding asserts

Please resolve if you like the fix ( only [this commit](https://github.com/golemfactory/golem/pull/4276/commits/1bfb41dbbcf7364261f331d9fb3fd4373e361f67) )

scripts/node_integration_tests/key_reuse.py
maaktweluit: haha /facepalm

you are right, fixed, testing now

scripts/node_integration_tests/key_reuse.py
maaktweluit: fixed

scripts/node_integration_tests/key_reuse.py
maaktweluit: works

etam (approved):

mfranciszkiewicz (approved):

mfranciszkiewicz (approved):

PR #4278, opened 2019-06-05
Implement subtasks restart cost estimation RPC

kmazurek:
- Added the RPC endpoint `comp.task.subtasks.estimated.cost`.
- Added `disable_concent` flag to the `comp.task.restart` endpoint.
- Fixed funds checking when performing a partial restart of a task (including previously failed subtasks).

### DIFF
--- a/golem/task/rpc.py
+++ b/golem/task/rpc.py
@@ -531,8 +531,12 @@ class ClientProvider:
 
     @rpc_utils.expose('comp.task.restart')
     @safe_run(_restart_task_error)
-    def restart_task(self, task_id: str, force: bool = False) \
-            -> typing.Tuple[typing.Optional[str], typing.Optional[str]]:
+    def restart_task(
+            self,
+            task_id: str,
+            force: bool = False,
+            disable_concent: bool = False
+    ) -> typing.Tuple[typing.Optional[str], typing.Optional[str]]:
         """
         :return: (new_task_id, None) on success; (None, error_message)
                  on failure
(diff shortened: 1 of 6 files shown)


golem/task/rpc.py
```diff
@@ -605,12 +606,20 @@ def restart_subtasks(
         :return: In case of any errors, returns the representation of the error
         (either a string or a dict). Otherwise, returns None.
         """
-        try:
-            task = self.task_manager.tasks[task_id]
-        except KeyError:
-            err_msg = f'Task not found: {task_id!r}'
-            logger.error(err_msg)
-            return err_msg
+        if not subtask_ids:
```
shadeofblue: @zakaprov actually, providing an empty array _should_ be legal here... in which case, we want to restart the task by repeating only the failed subtasks ... and only if we _do_ provide an array, we want to _also_ include the specified list of subtasks to restart...

shadeofblue:
@zakaprov -> generally, please run `node_integration_tests/tests/test_golem.py` on the branch to make sure

golem/task/rpc.py
kmazurek: You're right, let me revert this.

kmazurek:
@shadeofblue I ran the integration tests against this branch locally with all tests passing. Tried the same on Buildbot as well ([here](https://buildbot.golem.network/buildbot/#/builders/15/builds/717)), with three of the tests failing for reasons unrelated to my changes (geth nodes being out of sync).

shadeofblue (approved):
lgtm

PR #4297, opened 2019-06-06
Migrate payment to taskpayment

jiivan:
Migrate entries from `payment` table into `taskpayment`.

### DIFF
--- a/golem/database/schemas/030_migrate_payment_to_task_payment.py
+++ b/golem/database/schemas/030_migrate_payment_to_task_payment.py
@@ -1,0 +1,81 @@
+# pylint: disable=no-member,unused-argument
+import json
+import logging
+
+SCHEMA_VERSION = 29
+
+logger = logging.getLogger('golem.database')
+
+
+STATUS_MAPPING = {
+    1: 'awaiting',
+    2: 'sent',
+    3: 'confirmed',
+    4: 'overdue',
+}
+
+
+def migrate_payment(database, db_row):
+    details = json.loads(db_row['details'])
+    status = STATUS_MAPPING[db_row['status']]
+    cursor = database.execute_sql(
+        "INSERT INTO walletoperation"
+        " (tx_hash, direction, operation_type, status, sender_address,"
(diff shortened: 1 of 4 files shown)


golem/database/schemas/030_migrate_payment_to_task_payment.py
```diff
@@ -0,0 +1,81 @@
...
+        "  accepted_ts, settled_ts)"
+        " VALUES (?, ?, '', ?, ?, ?, datetime('now'),"
+        "         ?, ?)",
+        (
+            wallet_operation_id,
+            details['node_info']['key'],
+            db_row['subtask'],
+            db_row['value'],
+            db_row['created_date'],
+            # accepted_ts & settled_ts have mixed types & will be fixed
```
Krigpl: Wouldn't it make more sense to first fix the schema while the tables are empty and only after that fill it with proper values instead of some kind of fake ones?

Krigpl (commented):
This migration should be merged to develop at the same time as Golem starts using new tables instead of old ones, right? Otherwise there's a gap where old tables get filled and won't be migrated.

golem/database/schemas/030_migrate_payment_to_task_payment.py
jiivan: :+1: Will be fixed in #4308

jiivan:
@Krigpl I'm not sure if this is an issue in develop. Previous PR was big enough IMO. Maybe a little bit too big as some of my mistakes slipped through your scrupulous review :)

shadeofblue (approved):

Krigpl:
@jiivan I'd still recommend either delaying merging this PR until other changes are ready or (more puristic way) creating feature branch to which this and following PRs are going to be merged into, so keeping particular PRs small while still having all the changes merged at once.

While I agree that this isn't big of an issue in develop, it still may introduce confusion and unnecessary questions from developers themselves who are not aware of these changes, that's why my recommendation above.

jiivan:
In an offline discussion we clarified that this PR is ready to merge.

Issue #4301, opened 2019-06-07
glambda benchmark doesn't work


Krigpl:
```
INFO     [golem.task.benchmarkmanager        ] Running benchmark for glambda
WARNING  [golem.docker.task_thread           ] Task error - exit_code=1
stderr:
Failed to import the site module
Traceback (most recent call last):
  File "/usr/lib/python3.6/site.py", line 570, in <module>
    main()
  File "/usr/lib/python3.6/site.py", line 556, in main
    known_paths = addusersitepackages(known_paths)
  File "/usr/lib/python3.6/site.py", line 288, in addusersitepackages
    user_site = getusersitepackages()
  File "/usr/lib/python3.6/site.py", line 264, in getusersitepackages
    user_base = getuserbase() # this will also set USER_BASE
  File "/usr/lib/python3.6/site.py", line 254, in getuserbase
    USER_BASE = get_config_var('userbase')
  File "/usr/lib/python3.6/sysconfig.py", line 607, in get_config_var
    return get_config_vars().get(name)
  File "/usr/lib/python3.6/sysconfig.py", line 558, in get_config_vars
    _CONFIG_VARS['userbase'] = _getuserbase()
  File "/usr/lib/python3.6/sysconfig.py", line 205, in _getuserbase
    return joinuser("~", ".local")
  File "/usr/lib/python3.6/sysconfig.py", line 184, in joinuser
    return os.path.expanduser(os.path.join(*args))
  File "/usr/lib/python3.6/posixpath.py", line 249, in expanduser
    userhome = pwd.getpwuid(os.getuid()).pw_dir
KeyError: 'getpwuid(): uid not found: 1000'

tail of stdout:


WARNING  [golem.task.taskthread              ] Task computing error Subtask computation failed with exit code 1
WARNING  [golem.task                         ] Failed to compute benchmark Subtask computation failed with exit code 1
ERROR    [golem.task.benchmarkmanager        ] Unable to run glambda benchmark: Subtask computation failed with exit code 1
CRITICAL [golem.client                       ] Can't start network. Giving up.
Traceback (most recent call last):
  File "/home/admin_imapp/golem_repos/golem/golem/client.py", line 333, in start
    self.start_network()
  File "/home/admin_imapp/golem_repos/golem/golem/client.py", line 385, in start_network
    task_finished_cb=self._task_finished_cb,
  File "/home/admin_imapp/golem_repos/golem/golem/task/taskserver.py", line 131, in __init__
[...]

PR #4306, opened 2019-06-10
Fix GLambda benchmark

mplebanski:
Fixes https://github.com/golemfactory/golem/issues/4301

### DIFF
--- a/apps/glambda/glambdaenvironment.py
+++ b/apps/glambda/glambdaenvironment.py
@@ -26,5 +26,8 @@ class GLambdaTaskEnvironment(DockerEnvironment):
             volumes=[],
             binds={},
             devices=[],
-            environment={'HOME': '/home/user'}
+            # gVisor uses HOME variable before starting the image.
+            # Starting docker as a particular user (docker --user parameter)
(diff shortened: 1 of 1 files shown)


Krigpl:
You should at least link the issue, but ideally, any more descriptive summary should be included.

Krigpl (changes requested):
Putting `ENV HOME=/home/user` in the GLambda Dockerfile works for me, please check that.

mplebanski:
> Putting `ENV HOME=/home/user` in the GLambda Dockerfile works for me, please check that.

This is related to how we launch docker, why put it in Dockerfile?

Krigpl:
> This is related to how we launch docker, why put it in Dockerfile?

Why is this specific to GLambda?

mplebanski:
> > This is related to how we launch docker, why put it in Dockerfile?
> 
> Why is this specific to GLambda?

Because the runtime container is different.

Krigpl (approved):
This is also setting the `HOME` to a nonexistent directory. It'd be great to put any kind of comment in the code why does it is even needed there.

Krigpl (approved):
That looks good.

PR #4309, opened 2019-06-10
Transcoding video bundle

cameel:
This pull request adds a set of scripts for downloading and packaging videos for use in transcoding tests. It's completely independent of my other transcoding pull requests. Some of them may use files from this set but for now the assumption is that the files are provided by the developer from an external source.

See the README provided with the code for information on the usage and operation of the scripts.

The pull request includes a list of files for the bundle. The bundle is now at version 4. It differs from version 3 only slightly: it has an extra 4k file, one extra matroska file marked as good and cuts out shorter fragments of some videos to reduce size.

At this very moment version 4 of the prebuilt bundle has not been uploaded to the server yet so `download-bundle.sh` won't work unless you manually change the version to 3. I've had some minor technical problems - I'm going to do it tomorrow. Version 3 is still available and it's the version I used most extensively in my tests anyway.


cameel:
I see that tests are failing but I have no idea why. They fail only on Mac OS and on `assert ts.request_task()`. Which is weird because this pull request does not touch any Python code at all. I don't think the failure is in any way related to my changes.

cameel:
Bundle v4 is now up on the server.

nieznanysprawiciel:
I rebuilded tests. Probably they will pass after rebuild. It is known problem :(

cameel:
Thanks. Still failing though :(

nieznanysprawiciel (approved):

PR #4312, opened 2019-06-11
bring `node_name` back to `SubtaskState` ...

shadeofblue:


maaktweluit (approved):
LGTM!

Could you update the description to why this change is needed? Our future selves might need that later ;)

Krigpl:
fwiw I think retrieving the node's name from the node's id should be done as late as possible, i.e. just before it is really used. Core components do not benefit from passing the node's name all over the place.

shadeofblue:
@Krigpl well, as long as `SubtaskState` contains `node_name`, it would be good to have it filled correctly... apart from that, sure, it should indeed be refactored further to not need it at all

PR #4324, opened 2019-06-12
Task.initialize() executed in background

mfranciszkiewicz:
- task.rpc: run `Task.initialize` and `enqueue_new_task` in another thread
- task.taskstate: new `creating` and `errorCreating` `TaskStatus` vals
- initial TaskStatus is `creating`, the task is remembered immediately
- when task creation fails, `TaskStatus` is set to `errorCreating`


golem/core/deferred.py
```diff
@@ -1,12 +1,30 @@
 from queue import Queue, Empty
+from typing import Any
 
-from twisted.internet.defer import Deferred, TimeoutError
+from twisted.internet import defer
 from twisted.internet.task import deferLater
 from twisted.python.failure import Failure
 
 
+class DeferredSeq:
```
Wiezzel, line 10: Do I understand correctly that this class is supposed to compute `fn(...f3(f2(f1(None)))...)` and do it asynchronously?

mfranciszkiewicz: Yes, and by executing those functions in a separate thread. 

Since `deferred.callback` does not allow a `Deferred` object to be passed as an argument (in a callback chain), this class simply `yield`s on those before passing them further.

Krigpl:
Why do we need that right now? Is it still gonna be useful with the simple-task-api?

mfranciszkiewicz:
@Krigpl 

This PR solves some pending issues encountered by the CGI team. Until their app is refactored for the simple-task-api, this piece of code will prove useful to them because:

1) resource initialization may take a really long time, so we must not block the main thread
2) they need to be able to query the state that the task-in-creation is in, by returning the task id immediately and by introducing new `TaskStatus`es

Point 2 should still be relevant for simple-task-api.

Krigpl:
> Point 2 should still be relevant for simple-task-api.

simple-task-api is not affected by this issue, so that won't be relevant.

mfranciszkiewicz:
I agree only if we create tasks in a fire and forget fashion.

golem/core/deferred.py
```diff
@@ -1,12 +1,30 @@
...
+    def __init__(self, *fns) -> None:
+        self._seq = list(fns)
+
+    def execute(self) -> defer.Deferred:
+        from twisted.internet.threads import deferToThread
```
Wiezzel: Importing `deferToThread` does not import reactor (only *calling* it) so this import can be safely moved outside.

Wiezzel (approved):

golem/task/taskmanager.py
```diff
@@ -164,9 +176,17 @@ def create_task(self, dictionary, minimal=False):
                                                    minimal)
         definition.task_id = CoreTask.create_task_id(self.keys_auth.public_key)
         definition.concent_enabled = dictionary.get('concent_enabled', False)
-        builder = builder_type(self.node, definition, self.dir_manager)
 
-        return builder.build()
+        task = builder_type(self.node, definition, self.dir_manager).build()
+        task_id = task.header.task_id
+
+        logger.info("Creating task %r (%s)", type(task), task_id)
```
maaktweluit: could be formatted as kwargs:
```suggestion
        logger.info("Creating task. type=%r, id=%s", type(task), task_id)
```

golem/task/taskmanager.py
```diff
@@ -177,25 +197,26 @@ def get_task_definition_dict(self, task: Task):
...
+        task_state = self.tasks_states.get(task_id)
+
+        if not task_state:
+            task_state = TaskState(task)
+            self.tasks[task_id] = task
+            self.tasks_states[task_id] = task_state
+
+        if task_state.status is not TaskStatus.creating:
+            raise RuntimeError("Task {} has already been added"
                                .format(task.header.task_id))
```
maaktweluit: task_id is enough here, but you did not change this line..

golem/task/taskmanager.py
```diff
@@ -204,6 +225,15 @@ def add_new_task(self, task: Task, estimated_fee: int = 0) -> None:
                                  op=TaskOp.CREATED,
                                  persist=False)
 
+    def task_creation_failed(self, task_id: str, reason: str) -> None:
```
maaktweluit, line 228: should we log a warning / error here somewhere?

golem/task/taskmanager.py
```diff
@@ -347,7 +377,7 @@ def restore_tasks(self) -> None:
     def resources_send(self, task_id):
         self.tasks_states[task_id].status = TaskStatus.waiting
         self.notice_task_updated(task_id)
-        logger.info("Resources for task {} sent".format(task_id))
+        logger.info("Resources for task %s sent", task_id)
```
maaktweluit:
```suggestion
        logger.info("Resources for task send. id=%s", task_id)
```

golem/task/taskmanager.py
```diff
@@ -678,9 +713,9 @@ def get_trust_mod(self, subtask_id):
         if subtask_id in self.subtask2task_mapping:
             task_id = self.subtask2task_mapping[subtask_id]
             return self.tasks[task_id].get_trust_mod(subtask_id)
-        else:
-            logger.error("This is not my subtask {}".format(subtask_id))
-            return 0
+
+        logger.error("This is not my subtask %s", subtask_id)
```
maaktweluit:
```suggestion
        logger.error("This is not my subtask. id=%s", subtask_id)
```

maaktweluit (commented):
LGTM!

Some small nitty comments and one q

Will approve on reply / fix

golem/task/rpc.py
```diff
@@ -402,7 +403,7 @@ def enqueue_new_task(client, task, force=False) \
 
 def _create_task_error(e, _self, task_dict, **_kwargs) \
         -> typing.Tuple[None, typing.Union[str, typing.Dict]]:
-    logger.error("Cannot create task %r: %s", task_dict, e)
+    _self.client.task_manager.task_creation_failed(task_dict.get('id'), str(e))
```
etam, line 406: `task_dict` never contains id. See #4850

PR #4325, opened 2019-06-12
Remove key difficulty pow

Krigpl:
Doesn't provide any value at all right now, and can't be modified to do that.
R&D team will introduce a different approach against Sybil attacks.


golem/core/keysauth.py
```diff
@@ -53,9 +53,8 @@ class WrongPassword(Exception):
 class KeysAuth:
     """
     Elliptical curves cryptographic authorization manager. Generates
-    private and public keys based on ECC (curve secp256k1) with specified
-    difficulty. Private key is stored in file. When this file not exist, is
-    broken or contain key below requested difficulty new key is generated.
+    private and public keys based on ECC (curve secp256k1). Private key is
+    stored in file. When this file not exist or is broken.
```
Wiezzel: The last sentence of this comment doesn't have sense any more.

Wiezzel (approved):
:+1:

golem/core/keysauth.py
Krigpl: derp

PR #4327, opened 2019-06-12
Remove the 'duration' field from RPC Task dict representation

mfranciszkiewicz:
Due to performance issues that the frequent changes of `duration` are causing in the UI, this field is being removed.

`duration` is also redundant, since computation start time is already available and duration can be calculated manually.


mdtanrikulu (approved):
Approved!

PR #4332, opened 2019-06-13
Add paging to get-slow-argument.py

maaktweluit:
When pull requests got too big buildbot stopped building them.

Changed it so it reads the next pages of the review comments too.

### DIFF
--- a/scripts/get-slow-argument.py
+++ b/scripts/get-slow-argument.py
@@ -3,6 +3,7 @@
 # - input: pull_id from CI ( argv )
 # - output: argument to use for this test ( stdout )
 
+import os
 import sys
 import requests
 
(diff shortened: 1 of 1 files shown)


scripts/get-slow-argument.py
```diff
@@ -13,10 +13,30 @@
 # config vars
 required_approvals = 1
 
+_logging = False
```
Krigpl: When is this meant to be `True`?

maaktweluit: when it doesnt work and you need to see what is going on

Krigpl: maybe read the value from an env variable instead so you don't have to modify the code

Krigpl (approved):
vader_didnt_read

scripts/get-slow-argument.py
Krigpl: Or even better
- don't introduce the new `_log` function
- set `logger.propagate = False` when env variable is set

Krigpl: Oh, there's no logger here. Nevermind then (unless you wanna use logger).

maaktweluit: Good one! added :)

PR #4340, opened 2019-06-14
Requestor task models

Wiezzel:
These models are supposed to serve the purpose of persisting all necessary information about requested tasks. In the future, they will enable getting rid of pickles.

### DIFF
--- a/golem/model.py
+++ b/golem/model.py
@@ -11,7 +11,7 @@ from eth_utils import decode_hex, encode_hex
 import golem_messages
 from golem_messages import datastructures as msg_dt
 from golem_messages import message
-from golem_messages.datastructures import p2p as dt_p2p
+from golem_messages.datastructures import p2p as dt_p2p, masking
 from peewee import (
     BlobField,
     BooleanField,
@@ -33,6 +33,7 @@ from golem.core.simpleserializer import DictSerializable
 from golem.database import GolemSqliteDatabase
 from golem.ranking.helper.trust_const import NEUTRAL_TRUST
 from golem.ranking import ProviderEfficacy
+from golem.task import taskstate
 
 
 # TODO: migrate to golem.database. issue #2415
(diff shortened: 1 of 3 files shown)


Wiezzel:
I don't want to merge to develop because the code won't be used yet and probably while implementing the new TaskManager some changes would come up. Hence the epic banch as base.

golem/model.py
```diff
@@ -621,6 +622,74 @@ def __repr__(self):
...
         )
 
+####################
+# REQUESTOR MODELS #
+####################
+
+
+class RequestedTask(BaseModel):
+    task_id = CharField(primary_key=True)
+    app_id = CharField(null=False)
```
Krigpl, line 632: What values are gonna be used here?

golem/model.py
```diff
@@ -621,6 +622,74 @@ def __repr__(self):
...
+    status = StringEnumField(enum_type=taskstate.TaskStatus, null=False)
+
+    environment = CharField(null=False)
+    prerequisites = JsonField(null=False, default='{}')
+
+    task_timeout = IntegerField(null=False)  # milliseconds
+    subtask_timeout = IntegerField(null=False)  # milliseconds
+    start_time = UTCDateTimeField(null=True)
+
+    max_price = IntegerField(null=False)
```
Krigpl: Per hour? Per subtask? Per task?

golem/model.py
```diff
@@ -621,6 +622,74 @@ def __repr__(self):
...
+    estimated_cost = IntegerField(null=True)
```
Krigpl: How is that different than `max_price`?

golem/model.py
```diff
@@ -621,6 +622,74 @@ def __repr__(self):
...
+    estimated_fee = IntegerField(null=True)
+
+    max_subtasks = IntegerField(null=False)
+    concent_enabled = BooleanField(null=False, default=False)
+    mask = BlobField(null=False, default=masking.Mask().to_bytes())
```
Krigpl, line 648: Does `mask` need to be persistent? I imagine we could generate masks on the fly.

golem/model.py
```diff
@@ -621,6 +622,74 @@ def __repr__(self):
...
+    def deadline(self) -> Optional[datetime.datetime]:
+        if self.start_time is None:
+            return None
+        return self.start_time + \
+            datetime.timedelta(milliseconds=self.task_timeout)
+
+
+class ComputingNode(BaseModel):
+    node_id = CharField(primary_key=True)
+    name = CharField(null=False)
```
Krigpl, line 662: While I see the point of having a node name here, I'm somewhat reluctant to store it here. The problem with names is that they can change. And if that happens between computing two subtasks from the same subtask should the UI display the name (current) name or two different names?

Wiezzel:
I looked into `Task`, `TaskDefinition`, `TaskState`, and `SubtaskState` for candidates for model fields. Reasons why I omitted some of them:

`resources` – Can be established by listing files in the resource directory, no need to store in DB
`estimated_memory` – I don't think we will use that for new tasks (?)
`optimize_total` – I don't know what it does
`output_file` – Same as resources
`task_type` – Replaced with `app_id`
`run_verification` – Unsure if we need this
`options` – I don't know what it does
`docker_images` – Replaced by prerequisites
`compute_on` – Replaced by environment
`resource_size` – Can be obtained from filesystem, no need to store in DB
`progress` – We don't support this for subtasks, for tasks can be computed
`extra_data` – Replaced by prerequisites/payload
All the paths – Can be obtained using task_id/subtask_id and conventions

golem/model.py
```diff
@@ -621,6 +622,74 @@ def __repr__(self):
...
+        if self.start_time is None:
+            return None
+        return self.start_time + \
+               datetime.timedelta(milliseconds=self.task.subtask_timeout)
+
+    class Meta:
+        primary_key = CompositeKey('task', 'subtask_id')
+
+
+class RequestedSubtaskInput(BaseModel):
```
Krigpl: Why a separate model for it? Isn't it a simple list of paths?

golem/model.py
Wiezzel: Well, apps will have to have some identifiers, right? So these will be used here.

golem/model.py
Wiezzel: Per hour. I will update the field name to make it explicit.

golem/model.py
Wiezzel: Well, this is probably just `max_price` * `num_subtasks` * `subtask_timeout` so it might be removed, I suppose.

golem/model.py
Wiezzel: Mask is set a the creation of a task and periodically modified if there are too few offers. If we don't persist it then it will not work as intended.

golem/model.py
Wiezzel: The same one (current). So when a node starts computing a subtask for us we will just call `get_or_create()` and set the current name.

golem/model.py
Wiezzel: Initially I thought there might be more fields, but yeah, this can stored as a field of Subtask.

maaktweluit (approved):
LGTM!

Only a comment about some omitted fields:
- `estimated_memory` would be nice since blender tasks are prone to fail when there is not enough memory on the provider, not sure if this is implemented properly though..
- `progress` is also a nice feature to have per subtask but is not implemented at all now

Maybe we can make issues for these fields, or just leave them for now and add them when we need them

Wiezzel:
@maaktweluit `estimated_memory` seems to environment-specific or even app-specific. We need some more generic way to handle this.

Wiezzel:
@maaktweluit `progress` is nice but we can add it when it's actually supported.

Wiezzel:
@Krigpl I've just adjusted the code according to your comments.

mfranciszkiewicz:
`output_file` is an arbitrary output file location chosen by the user, we most probably still need that

Wiezzel:
@mfranciszkiewicz Added `output_file`.

Krigpl (approved):
lgtm, but I agree with not merging it to develop until it's been proved by real use case.

Issue #4341, opened 2019-06-14
Generating database migration fails


Wiezzel:
Since `Payments` model has been removed (#4240) but the corresponding table not, `peewee_migrate` is no longer able to generate migrations.
Running `./setup.py migration` results in the following error:
```
peewee.OperationalError: no such table: payment
```

PR #4343, opened 2019-06-14
Migrate deposit payments to TaskPayment

jiivan:


golem/model.py
```diff
@@ -359,24 +338,6 @@ def missing_amount(self):
         # pylint: disable=no-member
         return self.expected_amount - self.wallet_operation.amount
 
-
-class DepositPayment(BaseModel):
```
Krigpl, line 363: Either don't delete the model or delete the corresponding table in the migration as well.
This is the reason https://github.com/golemfactory/golem/issues/4341 happens.

Krigpl (commented):
I don't think it's safe to modify existing migrations.

jiivan:
> I don't think it's safe to modify existing migrations.

I also think so, but unfortunatelly they used custom fields.

Krigpl (approved):
waffle

PR #4345, opened 2019-06-14
Persistent topups

jiivan:


.gitignore
```diff
@@ -66,6 +66,7 @@ yacctab.py
 data/
 
 # Install and build
+apps/rendering/resources/taskcollector/Release/*
```
Krigpl: it's already there 2 lines below

golem/ethereum/transactionsystem.py
```diff
@@ -603,18 +667,29 @@ def withdraw(
                     available=self.get_available_gnt(),
                     currency=currency,
                 )
-            # TODO Create WalletOperation #4172
             tx_hash = self._sci.convert_gntb_to_gnt(
                 destination,
                 amount,
                 gas_price,
             )
+            if gas_price is None:
```
Krigpl: Please move this at the top of the function and remove lines 622-623 as well.
This will also prevent the case where gas price changes between sending the transaction and querying it here.

golem/ethereum/transactionsystem.py
```diff
@@ -603,18 +667,29 @@ def withdraw(
...
+            model.WalletOperation.create(
+                tx_hash=tx_hash,
+                direction=model.WalletOperation.DIRECTION.outgoing,
+                operation_type=model.WalletOperation.TYPE.transfer,
+                status=model.WalletOperation.STATUS.sent,
+                sender_address=self._sci.get_eth_address(),
+                recipient_address=destination,
+                amount=amount,
+                currency=model.WalletOperation.CURRENCY.GNT,
+                gas_cost=gas_price * self._sci.GAS_GNT_TRANSFER,
```
Krigpl, line 698: This is the upper bound, the final cost is known after it's mined.

golem/ethereum/paymentskeeper.py
```diff
@@ -64,12 +67,47 @@ def get_newest_payment(num: Optional[int] = None,
...
+                    model.WalletOperation.tx_hash == tx_hash,
+                    model.WalletOperation.operation_type  # noqa
+                    == model.WalletOperation.TYPE.transfer,
+                    model.WalletOperation.direction  # noqa
+                    == model.WalletOperation.DIRECTION.outgoing,
+                    model.WalletOperation.currency == currency
+                ).get()
+            operation.status = model.WalletOperation.STATUS.confirmed
+            operation.save()
+        except model.WalletOperation.DoesNotExist:
```
Krigpl, line 91: When do you expect this will happen?

golem/ethereum/paymentskeeper.py
```diff
@@ -64,12 +67,47 @@ def get_newest_payment(num: Optional[int] = None,
...
+    """Keeps information about payments for tasks
+       that should be processed and send or received.
+    """
 
     def __init__(self) -> None:
         """ Create new payments keeper instance"""
         self.db = PaymentsDatabase()
 
+    @staticmethod
+    def sent_transfer(
```
Krigpl: To be puristic `PaymentsKeeper`, as the description mentions, is about subtask payments, and this method is not.

golem/ethereum/transactionsystem.py
```diff
@@ -603,18 +667,29 @@ def withdraw(
...
                 amount,
                 gas_price,
             )
+            if gas_price is None:
+                gas_price = self.gas_price
+            model.WalletOperation.create(
+                tx_hash=tx_hash,
+                direction=model.WalletOperation.DIRECTION.outgoing,
+                operation_type=model.WalletOperation.TYPE.transfer,
+                status=model.WalletOperation.STATUS.sent,
```
Krigpl, line 693: I think you don't ever confirm this transaction.

Krigpl (changes requested):

golem/ethereum/transactionsystem.py
jiivan: It's confirmed in `PaymentsKeeper.sent_transfer()`

See: https://github.com/golemfactory/golem/blob/task_payment_topups/golem/ethereum/transactionsystem.py#L317-L328

golem/ethereum/transactionsystem.py
jiivan: Is it accessible in a `receipt` of a callback provided to `sci.subscribe_to_gnt_transfers()`?

golem/ethereum/paymentskeeper.py
jiivan: If transaction is triggered outside of golem (manually by user) or before this PR is merged.

golem/ethereum/paymentskeeper.py
jiivan: Should I move this to `TransactionSystem` or update the docstring?

Krigpl: Just update the docstring for now, since I ideally would aim into not having database operation directly in the TS

golem/ethereum/paymentskeeper.py
Krigpl: > transaction is triggered outside of golem (manually by user)

This invalidates a lot of assumptions in the TS and we assume this won't ever happen.

> before this PR is merged

Not good enough reason to keep it imo.

golem/ethereum/transactionsystem.py
Krigpl: Unfortunately, Ethereum doesn't provide gas cost in the receipt and we need to issue a request for it separately (there's SCI method for that).

golem/ethereum/transactionsystem.py
Krigpl: Right, I missed that.

golem/ethereum/paymentskeeper.py
```diff
@@ -64,12 +67,47 @@ def get_newest_payment(num: Optional[int] = None,
...
+            tx_hash: str,
+            sender_address: str,
+            recipient_address: str,
+            amount: int,
+            currency: model.WalletOperation.CURRENCY,
+    ):
+        try:
+            operation = model.WalletOperation.select() \
+                .where(
+                    model.WalletOperation.tx_hash == tx_hash,
```
Krigpl, line 86: `tx_hash` is enough to clearly identify the instance, so the rest of the `where` clause doesn't change anything.

golem/ethereum/transactionsystem.py
```diff
@@ -589,12 +627,40 @@ def withdraw(
...
+                gas_cost=gas_eth,
+            )
+
+            def on_eth_receipt(receipt):
+                if not receipt.status:
+                    log.error("Failed ETH withdrawal: %r", receipt)
+                    return
+                self._payments_keeper.sent_transfer(
+                    tx_hash=receipt.tx_hash,
+                    gas_amount=self.get_withdraw_gas_cost(
```
Krigpl: Use `receipt.gas_cost`

golem/ethereum/transactionsystem.py
```diff
@@ -589,12 +627,40 @@ def withdraw(
...
+                        amount,
+                        destination,
+                        currency,
+                    ),
+                    gas_price=self._sci.get_transaction_gas_price(
```
Krigpl: You know the gas price here, it's `gas_price` so you don't need to use that method (which issues an RPC request).

golem/ethereum/paymentskeeper.py
```diff
@@ -64,12 +67,31 @@ def get_newest_payment(num: Optional[int] = None,
...
+    """Keeps information about outgoing payments
+       that should be processed and send or received.
+    """
 
     def __init__(self) -> None:
         """ Create new payments keeper instance"""
         self.db = PaymentsDatabase()
 
+    @staticmethod
+    def sent_transfer(tx_hash: str, gas_amount: int, gas_price: Optional[int]):
```
Krigpl: maybe `confirm_transfer` would be more appropriate? I got confused about what this method does since there's also `sent` status.

golem/ethereum/transactionsystem.py
```diff
@@ -287,6 +287,42 @@ def _subscribe_to_events(self) -> None:
...
+            ),
+        )
+
+        self._sci.subscribe_to_gnt_transfers(
+            from_address=self._sci.get_eth_address(),
+            to_address=None,
+            from_block=from_block,
+            cb=lambda event: self._payments_keeper.sent_transfer(
+                tx_hash=event.tx_hash,
+                gas_amount=self._sci.GAS_GNT_TRANSFER,
```
Krigpl: That should get the transaction receipt and retrieve `gas_used` field from it. Unfortunately, web3 doesn't include this stuff in the event log object.

golem/ethereum/paymentskeeper.py
Krigpl: afaict you already always provide `gas_price` (as it should be), so maybe just provide a non-optional `gas_cost` and that's it.

golem/ethereum/incomeskeeper.py
```diff
@@ -15,6 +15,44 @@ class IncomesKeeper:
     """Keeps information about payments received from other nodes
     """
 
+    @staticmethod
+    def received_eth_transfer(
```
Krigpl: Those two methods are identical, so maybe DRY?

Krigpl (changes requested):

golem/ethereum/paymentskeeper.py
jiivan: `self._sci.get_transaction_gas_price()` can return `None`.

Krigpl: Not for mined transactions so you're safe here

tests/golem/ethereum/test_paymentskeeper.py
```diff
@@ -42,3 +44,35 @@ def test_subtasks_payments(self):
...
         payments = pd.get_subtasks_payments(['id1', 'id4', 'id2'])
         assert self._get_ids(payments) == ['id1', 'id2']
+
+
+class TestPaymentsKeeper(TestWithDatabase):
+    def setUp(self):
+        super().setUp()
+        self.payments_keeper = PaymentsKeeper()
+
+    def test_sent_transfer(self):
```
Krigpl: update method name

tests/golem/ethereum/test_incomeskeeper.py
```diff
@@ -289,3 +289,29 @@ def test_update_overdue_incomes_already_marked_as_overdue(self):
             income.wallet_operation.refresh().status,
             model.WalletOperation.STATUS.overdue,
         )
+
+    def test_received_eth_transfer(self):
```
Krigpl: Those two tests do not test different things the way they are implemented right now.
So either remove one of them or introduce some more assertions.

Krigpl (approved):
rickandmorty_yes

PR #4348, opened 2019-06-17
WASM Verification by Redundancy

mplebanski:
Implements https://github.com/golemfactory/golem/issues/4421


apps/glambda/glambdaenvironment.py
```diff
@@ -26,5 +26,8 @@ def get_container_config(self) -> Dict:
             volumes=[],
             binds={},
             devices=[],
-            environment={}
+            # gVisor uses HOME variable before starting the image.
+            # Starting docker as a particular user (docker --user parameter)
+            # does not set HOME and that's why we define it here.
+            environment={'HOME': '/home/user'} if is_linux() else {}
```
kubkon: How did `glambda` specifics creep in here? 😬

apps/wasm/task.py
```diff
@@ -1,20 +1,85 @@
...
+class VbrSubtask:
+    def __init__(self, id_gen, name, params, redundancy_factor):
+        self.id_gen = id_gen
+        self.name = name
+        self.params = params
+        self.result = None
+
+        self.subtasks = {}
+        self.verifier = BucketVerifier(
+            redundancy_factor, WasmTask._cmp_results, 0)
```
mbenke:
```suggestion
            redundancy_factor, WasmTask._cmp_results, referee_count=0)
```

apps/wasm/task.py
```diff
@@ -1,20 +1,85 @@
...
+        return s_id, deepcopy(self.params)
+
+    def get_instance(self, s_id) -> dict:
+        return self.subtasks[s_id]
+
+    def get_instances(self) -> List[str]:
+        return self.subtasks.keys()
+
+    def add_result(self, s_id, task_result):
+        self.verifier.add_result(self.subtasks[s_id]["actor"], task_result)
```
mbenke, line 85: This may throw an exception. Do we care?

apps/wasm/task.py
```diff
@@ -1,20 +1,85 @@
...
+from golem.task.taskbase import Task, AcceptClientVerdict
 from golem.task.taskstate import SubtaskStatus
+from golem.task.taskclient import TaskClient
+
+from .vbr import Actor, BucketVerifier, VerificationResult, NotAllowedError, MissingResultsError
+
+logger = logging.getLogger("apps.wasm")
+
+
+class VbrSubtask:
```
mbenke, line 42: Description needed

apps/wasm/task.py
```diff
@@ -1,20 +1,85 @@
...
+
+    def get_result(self):
+        return self.result
+
+    def is_finished(self) -> bool:
+        return self.verifier.get_verdicts() is not None
+
+    def get_verdicts(self):
+        verdicts = []
+        for actor, result, verdict in self.verifier.get_verdicts():
```
mbenke, line 96: what if `verifier.get_verdicts` returns `None`?

apps/wasm/task.py
```diff
@@ -85,51 +139,111 @@ def __init__(self, total_tasks: int, task_definition: WasmTaskDefinition,
...
-        output_dir_path = Path(self.options.output_dir, subtask_name)
+    def _find_vbrsubtask_by_id(self, subtask_id) -> Optional[VbrSubtask]:
+        for subtask in self.subtasks:
+            if subtask.contains(subtask_id):
+                return subtask
+
+        return None
+
+    def _cmp_results(result_list_a: List[Any], result_list_b: List[Any]) -> bool:
+        for r1, r2 in zip(result_list_a, result_list_b):
```
mbenke, line 199: Are we sure the files are in the same order on both lists?

apps/wasm/task.py
```diff
@@ -85,51 +139,111 @@ def __init__(self, total_tasks: int, task_definition: WasmTaskDefinition,
...
+            with open(r1, 'rb') as f1, open(r2, 'rb') as f2:
+                b1 = f1.read()
+                b2 = f2.read()
+                if b1 != b2:
+                    logger.info("Verification of task failed")
```
mbenke: This log message seems slightly misleading

apps/wasm/task.py
```diff
@@ -85,51 +139,111 @@ def __init__(self, total_tasks: int, task_definition: WasmTaskDefinition,
...
+                    return False
+
+        logger.info("Comparing: %s and %s", result_list_a, result_list_b)
+        logger.info("Verification of task was successful")
```
mbenke: Ditto

apps/wasm/task.py
```diff
@@ -85,51 +139,111 @@ def __init__(self, total_tasks: int, task_definition: WasmTaskDefinition,
...
+        return True
+
+    def computation_finished(self, subtask_id, task_result,
+                             verification_finished=None) -> None:
+        logger.info("Called in WasmTask")
```
mbenke: This should probably eventually go away (or at least be changed to a better message)

apps/wasm/task.py
```diff
@@ -85,51 +139,111 @@ def __init__(self, total_tasks: int, task_definition: WasmTaskDefinition,
...
+        s_ids = [s_id for s_id in subtask.get_instances()]
+        for s_id in s_ids:
+            instance = subtask.get_instance(s_id)
+            for actor, verdict in verdicts:
+                if actor is not instance['actor']:
+                    continue
+
+                if verdict == VerificationResult.SUCCESS:
+                    # pay up!
+                    logger.info("Accepting results for subtask %s", s_id)
```
mbenke, line 218: Maybe add provider id here

apps/wasm/task.py
```diff
@@ -85,51 +139,111 @@ def __init__(self, total_tasks: int, task_definition: WasmTaskDefinition,
...
+                    self.subtasks_given[s_id]['status'] = SubtaskStatus.finished
+                    TaskClient.assert_exists(actor.uuid, self.counting_nodes).accept()
+                else:
+                    logger.info("Rejecting results for subtask %s", s_id)
```
mbenke, line 224: Maybe add provider id here

apps/wasm/task.py
```diff
@@ -85,51 +139,111 @@ def __init__(self, total_tasks: int, task_definition: WasmTaskDefinition,
...
+                    self.subtasks_given[s_id]['status'] = SubtaskStatus.failure
```
mbenke, line 225: Is it the whole subtask, or its computation by one provider? If the former, this is dubious in the presence of referees.

apps/wasm/task.py
```diff
@@ -85,51 +139,111 @@ def __init__(self, total_tasks: int, task_definition: WasmTaskDefinition,
...
+
+        if not subtask.is_finished():
+            return
+
+        verdicts = subtask.get_verdicts()
+
+        s_ids = [s_id for s_id in subtask.get_instances()]
+        for s_id in s_ids:
+            instance = subtask.get_instance(s_id)
+            for actor, verdict in verdicts:
```
mbenke, line 212: In this loop, we change subtask status many times, potentially to different values; this smells.

apps/wasm/task.py
```diff
@@ -85,51 +139,111 @@ def __init__(self, total_tasks: int, task_definition: WasmTaskDefinition,
...
+        subtask.add_result(subtask_id, task_result)
+        self.subtasks_given[subtask_id] = {'status': SubtaskStatus.verifying}
+        WasmTask.CALLBACKS[subtask_id] = verification_finished
+
+        if not subtask.is_finished():
+            return
+
+        verdicts = subtask.get_verdicts()
+
+        s_ids = [s_id for s_id in subtask.get_instances()]
```
mbenke: Better name than `s_ids` needed, `instance_ids` maybe?

apps/wasm/task.py
```diff
@@ -85,51 +139,111 @@ def __init__(self, total_tasks: int, task_definition: WasmTaskDefinition,
...
+                    self.subtasks_given[s_id]['status'] = SubtaskStatus.failure
+                    TaskClient.assert_exists(actor.uuid, self.counting_nodes).reject()
+
+        # save the results but only if verification was successful
+        result = subtask.get_result()
+        if result is not None:
+            self.save_results(subtask.name, result)
+
+        for s_id in s_ids:
+            WasmTask.CALLBACKS.pop(s_id)()
```
mbenke: A comment on what is going on here might be helpful

apps/wasm/task.py
```diff
@@ -154,6 +268,151 @@ def filter_task_results(self, task_results, subtask_id, log_ext=".log",
...
+
+    def computation_failed(self, subtask_id: str, ban_node: bool = True):
+        subtask = self._find_vbrsubtask_by_id(subtask_id)
+        subtask.add_result(subtask_id, None)
+
+    def verify_task(self):
+        return self.finished_computation()
+
+    def get_total_tasks(self):
+        return ( WasmTask.REDUNDANCY_FACTOR + 1 ) * len(self.subtasks)
```
mbenke: We need to check what the intended semantics of this method is.

apps/wasm/task.py
```diff
@@ -154,6 +268,151 @@ def filter_task_results(self, task_results, subtask_id, log_ext=".log",
...
+
+    def get_active_tasks(self):
+        return 0
```
mbenke: why?

apps/wasm/task.py
```diff
@@ -154,6 +268,151 @@ def filter_task_results(self, task_results, subtask_id, log_ext=".log",
...
+
+    def get_tasks_left(self):
+        return 0
```
mbenke: why?

apps/wasm/task.py
```diff
@@ -154,6 +268,151 @@ def filter_task_results(self, task_results, subtask_id, log_ext=".log",
...
+
+    # pylint:disable=unused-argument
+    @classmethod
+    def get_subtasks(cls, part):
+        return dict()
```
mbenke: why?

apps/wasm/task.py
```diff
@@ -154,6 +268,151 @@ def filter_task_results(self, task_results, subtask_id, log_ext=".log",
...
+        Instead of tracking some aux variables, it polls VbrSubtasks
+        directly for their current state; i.e., whether they are finished,
+        or not.
+        """
+        num_total = self.get_total_tasks()
+        if num_total == 0:
+            return 0.0
+
+        num_finished = len(list(filter(lambda x: x.is_finished(), self.subtasks)))
+        logger.info("progress: %f", num_finished / num_total)
```
mbenke: this will probably log incorrect progress

apps/wasm/task.py
```diff
@@ -166,7 +425,7 @@ def build_full_definition(cls, task_type: 'CoreTaskTypeInfo',
         # Output is determined from 'output_dir' later on.
         dictionary['options']['output_path'] = ''
         # Subtasks count is determined by the amount of subtask info provided.
-        dictionary['subtasks_count'] = len(dictionary['options']['subtasks'])
+        dictionary['subtasks_count'] = (WasmTask.REDUNDANCY_FACTOR + 1) * len(dictionary['options']['subtasks'])
```
mbenke: This needs reconsideration

apps/wasm/vbr.py
```diff
@@ -0,0 +1,311 @@
...
+        result equal to None.
+
+        When user has added a result for each actor it reported by `add_actor` a side effect might be
+        the verdict being available or caller should continue adding actors and results.
+
+        Arguments:
+            actor {Actor} -- Actor who has computed the result
+            result {Any} --  Computation result
+        Raises:
+            UnknownActorError - raised when caller deliver an actor that was not previously reported by `add_actor` call.
```
mbenke: may also raise `ValueError`

apps/wasm/vbr.py
```diff
@@ -0,0 +1,311 @@
...
+        """
+
+    @abstractmethod
+    def get_verdicts(self) -> Optional[List[Tuple[Actor, Any, VerificationResult]]]:
+        """
+        Returns:
+            Optional[List[Any, Actor, VerificationResult]] -- If verification is resolved a list of 3-element
+            tuples (result reference, actor, verification_result) is returned. A None is returned
```
mbenke:
```suggestion
            tuples (actor, result reference, verification_result) is returned. A None is returned
```

apps/wasm/vbr.py
```diff
@@ -0,0 +1,311 @@
...
+            return True
+        return False
+
+    def __len__(self):
+        return len(self.values)
+
+
+class BucketVerifier(VerificationByRedundancy):
+    """Simple verification by redundancy: subtask is executed by 2 providers,
+with a possible third if no decision can be reached based on the first 2."""
```
mbenke: This comment is copied from another class and not really accurate

apps/wasm/vbr.py
```diff
@@ -0,0 +1,311 @@
...
+            found = False
+            for bucket in self.buckets:
+                if bucket.try_add(key=result, value=actor):
+                    found = True
+                    break
+
+            if not found:
+                self.buckets.append(Bucket(self.comparator, key=result, value=actor))
+
+        print(f'add_actor: {len(self.actors)} actors, {len(self.results)} results, {len(self.buckets)} buckets')
```
mbenke: This debug message should go away.

```suggestion
```

apps/wasm/vbr.py
```diff
@@ -0,0 +1,311 @@
...
+            self.verdicts = [(actor, self.results[actor], success if actor in winners else fail) for actor in self.actors]
+        elif self.majority - max_popularity <= self.referee_count:
+            self.verdicts = None
+            self.more_actors_needed = True
+        else:
+            self.verdicts = [(actor, self.results[actor], VerificationResult.UNDECIDED) for actor in self.actors]
+            self.more_actors_needed = False
+
+
+class SimpleSubtaskVerifier(VerificationByRedundancy):
```
mbenke: This class is not used (and implements a different API anyway - int comparator rather than bool comparator`).
I would remove it (or replace with correct version).

mbenke (changes requested):

apps/glambda/glambdaenvironment.py
mplebanski: Well I rebased to b0.20 and that was conflicted with our branch.

apps/wasm/task.py
mplebanski: I believe crashing hard is what we actually want in this case. Giving a wrong actor or a result for not reported actor is not what we allow in here.

apps/wasm/task.py
mplebanski: `VbrSubtask` has method `is_finished` to make sure it never happens.

apps/wasm/task.py
mplebanski: I don't know. @kubkon ?

apps/wasm/vbr.py
mplebanski: Yeah I see that API description is not aligned to implementation comments. I will remove impl comments and update API.

apps/wasm/vbr.py
mplebanski: Removed.

mplebanski:
@mbenke I modified `add_actor` but then reverted the changes. There was a problem that two instances of an Actor class with the same uuid were not equal (which is expected, but we forgot about it), thus duplicated actor was not detected. I added __eq__ and __hash__ operations to Actor class.

apps/wasm/vbr.py
```diff
@@ -0,0 +1,181 @@
...
+
+class Actor:
+    def __init__(self, uuid: str) -> None:
+        self.uuid = uuid
+
+    def __eq__(self, other):
+        return self.uuid == other.uuid
+
+    def __hash__(self):
+        return id(self)
```
mbenke: Why not 
```suggestion
        return hash(self.uuid)
```

apps/wasm/task.py
mbenke: ..and if `is_finished ` is not checked, we crash with a `TypeError`

apps/wasm/task.py
kubkon: AFAIK, yep, they ought to be unless something went horribly wrong in one Golem node; e.g., computation failed, etc.

apps/wasm/task.py
kubkon: IMHO any redundant checks in `get_verdicts` are not strictly required. All we need is to document the use-case in the docs, as in "call `is_finished` before calling `get_verdicts`".

apps/glambda/glambdaenvironment.py
kubkon: Ah, ok. We can fix that later then.

apps/wasm/task.py
kubkon: It's a remnant of the old code. But I agree, we should iron it out. Also, it probably should be `debug` log rather than `info`.

apps/wasm/task.py
kubkon: See above.

apps/wasm/task.py
kubkon: Again, see above. I've used it to trace calls through Golem. Not as easy as it looks :grin:

apps/wasm/task.py
kubkon: That's a good idea. We should probably also change from `info` to `debug`.

apps/wasm/task.py
kubkon: I'm not sure I understand your question. Could you elaborate?

apps/wasm/task.py
kubkon: Perhaps, although is that strictly necessary here? IHMO, it's debatable whether it improves the code hygiene here even.

apps/wasm/task.py
kubkon: We do? I'm not sure I can see that here. Could you explain?

apps/wasm/task.py
kubkon: Good catch!

apps/wasm/task.py
kubkon: Agreed! We need to agree whether we want this to be transparent wrt the redundancy; i.e., whether Golem should or shouldn't be aware of verification by redundancy at all.

apps/wasm/task.py
mplebanski: Resolved

apps/wasm/task.py
mplebanski: I removed it. There are other more informative logs left though.

kubkon (approved):

mplebanski:
@mbenke @kubkon Please approve for control build.

maaktweluit:
@mplebanski it has an approval so it should build, strange.. hit the rebuild ( on `hook_pr` ) to check https://buildbot.golem.network/buildbot/#/builders/12/builds/10804

kubkon:
Closing as it's been superseded by #4475.

PR #4349, opened 2019-06-17
fix the integration tests that rely on a different package (Blender s…

shadeofblue:
…cene) supplied to them

(those tests stopped using those scenes and reverted to the default `test_task_1` after the last refactoring

+ add a specific integration test to test a blender scene composed of multiple `.blend` files


scripts/node_integration_tests/helpers.py
```diff
@@ -127,6 +129,18 @@ def construct_test_task(task_package_name: str, task_settings: str) \
     return settings
 
 
+def scene_file_path(task_package_name: str, file_path: str):
```
etam:
```suggestion
def scene_file_path(task_package_name: str, file_path: str) -> str:
```

etam:
btw, this looks quite familiar: f9e2cb05812547c3921ead9c0214c01830bee370
I think I even like your approach more. You could revert the commit I mentioned, to make the code cleaner.

scripts/node_integration_tests/tests/test_golem.py
```diff
@@ -56,3 +56,9 @@ def test_large_result(self):
 
     def test_restart_failed_subtasks(self):
         self._run_test('golem.restart_failed_subtasks')
+
+    def test_main_scene_file(self):
+        self._run_test(
+            'golem.nested_column',
+            **{'task-package': 'column', 'task-settings': 'default'}
```
etam: Shouldn't those values be written in `TestConfig.__init__`?
Because if you run a test directly, using `run_test.py`, you're losing this config.

maaktweluit (approved):
LGTM!

etam (approved):
ok

PR #4351, opened 2019-06-17
Fixed negative task restart estimation cost

kmazurek:


Krigpl (changes requested):
Please add simple unit test for that

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -336,6 +336,19 @@ def test_backwards_compatibility_privkey(self):
         # Shouldn't throw
         self._make_ets(datadir=self.new_path / 'other', password=password)
 
+    def test_eth_for_batch_payment(self, *_):
```
Krigpl: `*_` doesn't seem to be needed

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -336,6 +336,19 @@ def test_backwards_compatibility_privkey(self):
...
+        payments_count = 2
+        eth_for_batch = self.ets.eth_for_batch_payment(payments_count)
+
+        self.assertGreater(eth_for_batch, 0)
+
+        with patch('golem.ethereum.transactionsystem.TransactionSystem.'
```
Krigpl: Don't patch methods of a component that you are testing. Patch external dependencies, in this case `sci` object.
Tests ideally should not rely on the internals (so the particular implementation of the method) and shouldn't fail when the corresponding method is refactored to use a different approach.

Krigpl (changes requested):

Krigpl (approved):
thanks
verynicegreatsuccess

PR #4353, opened 2019-06-17
Fix HyperV warning events

mfranciszkiewicz:
RPC serialization of the `events` enum is failing and the warning is not published.


shadeofblue (approved):
thank you

PR #4354, opened 2019-06-18
Disable Task.initialize execution timeout

mfranciszkiewicz:
Set timeout to `None` in `DeferredSeq.execute -> sync_wait` instead on relying on the default value (10 seconds)


maaktweluit (approved):
LGTM

PR #4367, opened 2019-06-19
node_integration_tests - Made 'cubes' the default scene

maaktweluit:
... this saves 200mb per test

The node_integration_tests are now generating a ~1GB zip file extracting to ~4GB

While investigating i found the default scene used is 43MB and its stored 3 to ~8 times per test

This PR changes:
- the default scene to `cubes` 
- the `test_jpg()` use the old default 43MB scene

results on `test_regular_task_run`
```
before:
223M	./node_integration_test

---

after:
12M	./node_integration_test
```

### DIFF
--- a/scripts/node_integration_tests/playbooks/golem/jpg/test_config.py
+++ b/scripts/node_integration_tests/playbooks/golem/jpg/test_config.py
@@ -1,6 +1,17 @@
+from scripts.node_integration_tests import helpers
+
+
 from ...test_config_base import TestConfigBase
 
 
 class TestConfig(TestConfigBase):
     def __init__(self):
         super().__init__(task_settings='jpg')
+
+    def update_task_dict(self):
+        self.task_package = 'test_task_1'
+        super().update_task_dict()
+        self.task_dict['main_scene_file'] = helpers.scene_file_path(
+            task_package_name='test_task_1',
+            file_path='wlochaty3.blend',
+        )
--- a/scripts/node_integration_tests/playbooks/test_config_base.py
+++ b/scripts/node_integration_tests/playbooks/test_config_base.py
@@ -97,7 +97,7 @@ class TestConfigBase:
             self.nodes[node_id] = make_node_config_from_env(node_id.value, i)
         self._nodes_index = 0
         self.nodes_root: 'Optional[Path]' = None
-        self.task_package = 'test_task_1'
+        self.task_package = 'cubes'
         self.task_settings = task_settings
         self.update_task_dict()
 


scripts/node_integration_tests/tests/test_golem.py
```diff
@@ -49,10 +49,7 @@ def test_task_output_directory(self):
         self._run_test('golem.task_output')
 
     def test_large_result(self):
-        self._run_test(
-            'golem.separate_hyperg',
-            **{'task-package': 'cubes', 'task-settings': '4k'},
-        )
+        self._run_test('golem.separate_hyperg', **{'task-settings': '4k'},)
```
shadeofblue: I'd still leave it set explicitly here - the default can be changed and we'll forget about it

scripts/node_integration_tests/tests/test_golem.py
```diff
@@ -34,7 +34,7 @@ def test_jpeg(self):
         self._run_test('golem.jpeg')
 
     def test_jpg(self):
-        self._run_test('golem.jpg')
+        self._run_test('golem.jpg', **{'task-package': 'test_task_1'},)
```
shadeofblue: I'm unsure if it's good if `test_jpg` and `test_jpeg` use different scenes...

shadeofblue:
@maaktweluit does this improve or worsen the execution times? 

(or is the difference insignificant?)

scripts/node_integration_tests/tests/test_golem.py
maaktweluit: ok, will put it back :)

scripts/node_integration_tests/tests/test_golem.py
maaktweluit: i do not think that is required, they are testing different things:
- `jpeg` is testing if blender can produce a `jpg` file
- `jpg` is testing if an invalid task type defaults back to PNG

Maybe we can rename `test_jpg` to `test_invalid_task_type` and use `abc` as extension to not cause any confusion

maaktweluit:
@shadeofblue the first tests shows that a single test ( test_regular_task_run ) is 1 minute faster, but we need to run it more to be sure

scripts/node_integration_tests/tests/test_golem.py
shadeofblue: ah, right... I wouldn't change the extension though

shadeofblue (approved):

scripts/node_integration_tests/tests/test_golem.py
etam: I think you should put such config changes in `TestConfig`. You'll miss this option, when running test manually with `run_test.py`.

maaktweluit: then maybe we should remove this option to pass config to `_run_test()`

will move it for now

PR #4370, opened 2019-06-19
node_integration_tests concent default disabled

### LINKED ISSUE #4174: Disable Concent by default in integration tests
Currently, Concent is enabled by default in all integration tests except for `no_concent`. Having the service on for all tests doesn't add much value since:
1. Testing Concent's availability can be done with a single test.
2. There are separate integration tests focused solely on the service's use cases.

Additionally, using Concent means the requestor has to create a deposit with each task created, increasing the running time of each test.

maaktweluit:
Resolves #4174

Seeing those warnings every day makes me want to fix this ;)


shadeofblue (approved):
sure

PR #4372, opened 2019-06-21
Add run_benchmark to environments

Krigpl:
And create `cpu_benchmark` Docker image which uses an existing minilight benchmark.


maaktweluit (approved):

golem/envs/docker/benchmark/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM golemfactory/base:1.5
+
+MAINTAINER Golem Tech <tech@golem.network>
+
+COPY minilight /golem/minilight
+COPY entrypoint.py /golem/
+
+ENTRYPOINT ["python3", "/golem/entrypoint.py"]
```
Wiezzel: Line feed missing.

golem/envs/docker/cpu.py
```diff
@@ -285,6 +288,12 @@ def _spawn_status_update_thread(_):
             f"Starting container '{self._container_id}' failed."))
         return deferred_start
 
+    def wait_until_stopped(self) -> Deferred:
```
Wiezzel, line 289: Maybe move this up in class hierarchy and make a part of `Runtime` interface?

Wiezzel (changes requested):
I think using stdout would be better than a JSON file in this case. Keep in mind that sharing directories is quite messy on Windows. If the shared directory is not a subdirectory of `CoputerRes` than a pop-up will appear asking for admin permissions (!). And here you're using a temporary directory.

golem/envs/docker/cpu.py
```diff
@@ -404,7 +411,7 @@ class DockerCPUEnvironment(Environment):
     MIN_MEMORY_MB: ClassVar[int] = 1024
     MIN_CPU_COUNT: ClassVar[int] = 1
 
-    SHARED_DIR_PATH: ClassVar[str] = '/golem'
+    SHARED_DIR_PATH: ClassVar[str] = '/golem/work'
```
Wiezzel, line 414: Is this consistent with Task API?

Krigpl: While it's not documented atm, this is what Blender assumes, so I guess this should be added as part of the Task API.

golem/envs/docker/cpu.py
```diff
@@ -505,6 +514,28 @@ def _clean_up():
...
+            tag=tag,
+            user=None if is_windows() else str(os.getuid()),
+            env={},
+        )
+        runtime = self.runtime(payload)
+        yield runtime.prepare()
+        yield runtime.start()
+        yield runtime.wait_until_stopped()
+        # Benchmark is supposed to output a single line containing a float value
+        score = float(list(runtime.stdout('utf-8'))[0])
```
Wiezzel:
```suggestion
        stdout = list(runtime.stdout('utf-8'))
```

golem/envs/docker/cpu.py
```diff
@@ -505,6 +514,28 @@ def _clean_up():
...
+        yield runtime.clean_up()
+        return score
```
Wiezzel:
```suggestion
        return float(stdout[0])
```

Wiezzel (commented):
A suggestion to move picking the first line and converting to float after `clean_up()` so that the container get cleaned up if something goes wrong with the output.

golem/envs/docker/cpu.py
Wiezzel: Maybe *the implementation* could be moved as well? It's not Docker-specific after all.

Krigpl: I dunno, I'm still considering using `docker.wait()` method since I'm not happy with this kind of active waiting and wouldn't want it to become the default implementation for all envs. Not insisting tho.

Wiezzel: OK, we don't have any other envs for now. We can reconsider this once we have some more.

Wiezzel (approved):

Issue #3960, opened 2019-03-05
error , variable`start_task` is None


ederenn:
####: Encountered on:
Mac OS, Golem 0.19.0 binaries, release version

Critical appears during computations of restarted task restarted by timed out subtasks.

```
2019-03-05 17:52:41 DEBUG    golem.task.taskmanager              check_next_subtask('10703c787ec9fafb158a2a26c4e8d2752bf82abcc11fb5cf917743af0cc3469c0d50afde64162575a7f6ed6b18f809a1d0438edcebe7fad5198574a6d060d3b4', 'Kelebron 3', '55df885e-3f5f-11e9-86d5-6015c9f98f86', 119947080605216656)
2019-03-05 17:52:41 DEBUG    golem.task.taskmanager              Should accept client verdict. verdict=AcceptClientVerdict.ACCEPTED, task=55df885e-3f5f-11e9-86d5-6015c9f98f86, node=10703c78..d060d3b4
2019-03-05 17:52:41 CRITICAL twisted                             Unhandled error in Deferred:
2019-03-05 17:52:41 CRITICAL twisted                             
Traceback (most recent call last):
  File "/Users/buildbot-worker/worker/buildpackage_macOS/build/.venv/lib/python3.6/site-packages/twisted/internet/defer.py", line 653, in _runCallbacks
  File "/Users/buildbot-worker/worker/buildpackage_macOS/build/golem/task/tasksession.py", line 575, in _offer_chosen
  File "/Users/buildbot-worker/worker/buildpackage_macOS/build/golem/task/taskmanager.py", line 400, in get_next_subtask
  File "/Users/buildbot-worker/worker/buildpackage_macOS/build/apps/blender/task/blenderrendertask.py", line 441, in query_extra_data
  File "/Users/buildbot-worker/worker/buildpackage_macOS/build/apps/rendering/task/framerenderingtask.py", line 379, in _choose_frames
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'
```

logs on debug
https://drive.google.com/open?id=1HlrOTpDGt24RVXl-ksmHihCiuqYl8_e7

shadeofblue:
@ederenn does this error have any influence on Golem's functioning?

ederenn:
tasks are finished but this error is repeated in logs in hundreds for one task

shadeofblue:
@ederenn is this still happening?

ederenn:
yes, it still does

[...]

PR #4381, opened 2019-06-24
Fixed failed subtask count when restarting a finished task

kmazurek:
Fixes: #3960

When copying the results of a restarted task, the number of failed subtasks
on the new task was being incorrectly set to the total subtasks count.
This resulted in the requestor attempting to assign more new subtasks
than necessary.

### DIFF
--- a/apps/rendering/task/renderingtask.py
+++ b/apps/rendering/task/renderingtask.py
@@ -190,6 +190,9 @@ class RenderingTask(CoreTask):
         exec_cmd(cmd)
 
     def _get_next_task(self):
+        logger.debug(f"_get_next_task. last_task={self.last_task}, "
+                     f"total_tasks={self.total_tasks}, "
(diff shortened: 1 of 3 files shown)


shadeofblue (approved):

apps/rendering/task/renderingtask.py
```diff
@@ -190,6 +190,9 @@ def _put_collected_files_together(self, output_file_name, files, arg):
         exec_cmd(cmd)
 
     def _get_next_task(self):
+        logger.debug(f"_get_next_task. last_task={self.last_task}, "
```
maaktweluit, line 193: please follow the [logging guidelines](https://github.com/golemfactory/golem-internal/wiki/Logging-guidelines), we do not want to format when the level is not enabled

kmazurek: Thanks for mentioning, fixed in #4431

PR #4387, opened 2019-06-25
send srr if fgtrf is in history

### LINKED ISSUE #4216: Requestor sends `ForceGetTaskResultFailed` as an attachment to `SubtaskResultsRejected`
## Rationale
Missing piece of UC2/UC3 interaction in Golem's Concent implementation

## Description

According to the Concent Service specs, a Requestor sends a `ForceGetTaskResultFailed` as an attachment to `SubtaskResultsRejected` when Concent UC3 (Forced Acceptance) is triggered after failure to download resources through the Concent in UC2.

### Technical description
* `on_force_subtask_results` should be updated so that: 
1) after it determines that it has no prior SRR or SRA
2) before trying to run  the whole download+verify flow, it first checks for prior `ForceGetTaskResultFailed` from Concent and sends a `SubtaskResultsRejected` attaching this message
3) and only after it doesn't have that message, proceed with the redo download+verify fallback (`_after_ack_report_computed_task`)

* ensure that such an `SRR` as generated in step 2 above doesn't cause the Provider to reach out to Concent with UC4 (as it will be immediately rejected) (only actual verification failure is honored) -> https://github.com/golemfactory/golem/issues/4217

### Additional tests
* add a unit test that verifies that an appropriate SRR with the attached FGTRF is sent

etam:
resolves #4216

### DIFF
--- a/golem/network/concent/received_handler.py
+++ b/golem/network/concent/received_handler.py
@@ -244,13 +244,28 @@ class TaskServerMessageHandler():
             task_id=msg.task_id
         )
         if not (sra or srr):
-            #  I can't remember verification results,
-            #  so try again and hope for the best
-            self._after_ack_report_computed_task(
-                report_computed_task=msg.ack_report_computed_task
-                .report_computed_task,
+            fgtrf = history.get(
+                message_class_name='ForceGetTaskResultFailed',
+                node_id=msg.provider_id,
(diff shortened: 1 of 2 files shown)


jiivan (approved):

shadeofblue (approved):
darth-vader-thumbs-up

PR #4392, opened 2019-06-27
+ Docker image integrity verifier

### LINKED ISSUE #4267: Pull Docker images by SHA digest
## Rationale
Right now, Docker images used by Golem are described using tags (e.g. `golemfactory/blender_verifier:1.3`). Relying on tags alone is error-prone and may lead to non-deterministic builds, since it's possible to push an entirely new Docker image under the same tag.

One example of such issue is #4143, where an incorrect Docker image under a specific tag got cached by Buildbot workers, causing tests to fail.

To avoid similar issues in the future, Golem could rely on not only tags but also the SHA digests of Docker images.

### Technical specification
For all docker environments (e.g. verifier, blender, wasm) a new field should be defined (e.g. `IMAGE_SHA_DIGEST`). This would contain the SHA of the expected Docker image to be used by Golem.
Initial image pulls should still be based on tags, making them more readable. The SHA sum check would be done as a second step, possibly failing or pulling the correct image if a checksum mismatch occurs.
#### Which application modules will be updated?
- `DockerEnvironment`
- `DockerManager`

shadeofblue:
more or less closes: https://github.com/golemfactory/golem/issues/4267


maaktweluit (approved):
LGTM!

Nice check :)

apps/image_integrity.ini
```diff
@@ -0,0 +1,27 @@
+# repository                    tag                 hash
```
Krigpl: This file may easily get desynchronized with the real state of the world (i.e. publishing a new image without adding a new tag to this file). Also these sections seem unnecessary and repetitive.

What I would suggest is to put the hashes into the already existing `images.ini` file.
- that solves the desynchronization issue since you'll be forced to update the hash at the time you update the tag
- no repetitive sections, since each branch will automatically have its own versions.

Krigpl (changes requested):

apps/image_integrity.ini
shadeofblue: @Krigpl `images.ini` defines only the images needed for the current branch, here we have a chance to verify several

I thought about it -> we could add a check later to verify that no references are present in `images.ini` that are not reflected in `image_integrity.ini` but that seems secondary to the most basic requirement

shadeofblue: @Krigpl addressed ...

maaktweluit (approved):

scripts/docker_integrity/verify.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def run_verification():
+
+    cnt_images, cnt_failures = verify_images()
+
+    if cnt_failures:
+        print(
+            f'{COLORS.RED}{cnt_failures} out of {cnt_images} images '
+            f'had modified hashes!{COLORS.RESET}'
```
Krigpl: `s/had/have/`?

scripts/docker_integrity/verify.py
```diff
@@ -0,0 +1,240 @@
...
+    if not images:
+        raise ConfigurationError(
+            "Could not parse Golem `images.ini`. Format has changed?"
+        )
+
+    return images
+
+
+def get_images() -> dict:
+    images: dict = {}
```
Krigpl, line 60: Is this a dict of `repo - > dict[tag -> hash]` ? Why not just a set of `[repo, tag, hash]`? And you can throw on duplicated entries and for two entries with the same image but different hashes, one will fail.

scripts/docker_integrity/verify.py
```diff
@@ -0,0 +1,240 @@
...
+    pass
+
+
+def get_golem_images() -> dict:
+    images: dict = {}
+
+    with open(GOLEM_IMAGES_FILE) as f:
+        for l in f:
+            m = re.match(
+                r"(?P<repo>[\w._/]+)\s+\S+\s+(?P<tag>[\w.]+)", l)
```
Krigpl, line 45: What's this extra `\S+\s+` is for? I can't parse it (and it's not in the other regex).
Also, I'd extract this regex, not to repeat it on the other place.

Krigpl (commented):
just nitpicking

scripts/docker_integrity/verify.py
shadeofblue: it's not the same regex since the files have differing syntaxes

shadeofblue: the `\S+` is to jump over the second column

requirements-test_to-freeze.txt
```diff
@@ -8,4 +8,5 @@ pylint==1.9.2
 pytest-cov
 pytest-timeout==1.2.1
 pytest==3.3.1
+requests==2.21.0
```
maaktweluit, line 11: this should also be added to `requirements-test.txt`, with or without version depending on how stable it is.

maaktweluit (changes requested):

scripts/docker_integrity/verify.py
shadeofblue: ugh... it's more complicated than that ;p

shadeofblue: @Krigpl anyway, updated

requirements-test_to-freeze.txt
shadeofblue: @maaktweluit it's _already_ there

maaktweluit (approved):

scripts/docker_integrity/verify.py
shadeofblue: meh...

Krigpl (approved):
mckayla

PR #4393, opened 2019-06-27
Fixed rpc tests after disabling concent

maaktweluit:
Caused by #4370 

- Mainnet was missing the balance check override
- Disabled key-reuse on mainnet test
- Testnet was still using the old concent override in playbook
- Moved old concent playbook to `test_rpc/concent`


scripts/node_integration_tests/playbooks/golem/rpc_test/concent/playbook.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Playbook(NodeTestPlaybook):
+    def step_verify_deposit_balance_call(self):
+        def on_success(result):
+            if result['value'] and result['status'] and result['timelock']:
+                print("Result correct %s" % result)
+                self.next()
+            else:
+                print("Unexpected result: %s" % result)
```
mfranciszkiewicz, line 12: Shouldn't we call `self.fail()` if the result is incorrect?

maaktweluit: i think it should retry, but concent is not my strongest side of golem.
In case of retry there is a sleep missing...

@shadeofblue what do you think? sleep or fail here?

shadeofblue: I think it should `fail` ... this is an artifact of the change that @etam made that changed the way playbooks are run (earlier, it would just run out of steps in this case)

shadeofblue (changes requested):
@maaktweluit it's strange that the _base_ playbook is the one that defines the empty step... I think it should be the other way round ... 

especially, that with your changes here, `concent` playbook is actually never used ...

scripts/node_integration_tests/playbooks/golem/rpc_test/concent/playbook.py
maaktweluit: fixed

scripts/node_integration_tests/tests/test_golem.py
```diff
@@ -16,9 +16,9 @@ def test_concent(self):
     def test_rpc(self):
         self._run_test('golem.rpc_test')
```
shadeofblue, line 17: with your changes in place, `golem.rpc_test.concent` must be used here

maaktweluit: added it as a separate test

maaktweluit:
@shadeofblue the "base" playbook is the default mode to run, `testnet + no concent`, i see this could also be the other way around but as discussed we leave it for now.

Resolved your other comment by adding the test

mfranciszkiewicz (approved):

PR #4396, opened 2019-06-27
node_integration_tests - Remove repeated part of datadirs

maaktweluit:
```
old: scripts.node_integration_tests.tests.test_golem.GolemNodeTest.test_frame_restart
new:                                      test_golem.GolemNodeTest.test_frame_restart
```

### DIFF
--- a/scripts/node_integration_tests/tests/base.py
+++ b/scripts/node_integration_tests/tests/base.py
@@ -49,7 +49,10 @@ class NodeTestBase(unittest.TestCase):
         self.test_dir = get_testdir() / self._relative_id()
 
     def _relative_id(self):
-        return self.id().replace(__name__ + '.', '')
+        id = self.id()
+        id = id.replace(__name__ + '.', '')
+        id = id.replace('scripts.node_integration_tests.tests.', '')
+        return id
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


Wiezzel (approved):
Looks ok. Do you still need to replace `__name__` though?

shadeofblue (approved):

maaktweluit:
Was looking at that `__name__` and it indeed seems not required.

@shadeofblue @etam do you know why it is there? can we safely remove it?

mfranciszkiewicz:
AFAICT replacing `__name__` was supposed to have a similar result, `GolemNodeTest.test_frame_restart` in this example.

mfranciszkiewicz (commented):
I think it would be nicer to dynamically resolve the required path or build the relative id from the scratch. Some helpful definitions from the `unittest` module:

```
def id(self):
    return "%s.%s" % (strclass(self.__class__), self._testMethodName)
```

```
def strclass(cls):
    return "%s.%s" % (cls.__module__, cls.__qualname__)
```

maaktweluit:
@Wiezzel @mfranciszkiewicz Based on the review comments i updated the PR to use `__name__` again for the replacement, so the test dir always starts with the file/module name.

mfranciszkiewicz (approved):

PR #4401, opened 2019-06-28
Fix Docker benchmark

### LINKED ISSUE #4395: DockerCPUEnvironment bechmark returns empty output
## Description

**Golem Version**: 6f09e1179 (develop)

**Golem-Messages version** (leave empty if unsure): 3.9.0

**Electron version** (if used): not aplicable

**OS** [e.g. Windows 10 Pro]: Linux

**Branch** (if launched from source): task_payment_rpc

**Mainnet/Testnet**: n/a

**Description of the issue**: Runtime is torn down on linux buildbot

_A clear and concise description of what went wrong, in which component, when and where._

**Actual result**:

`__init__.py                237 INFO     Runtime torn down.`

**Screenshots**:

```python
________________________ TestIntegration.test_benchmark ________________________
self = <tests.golem.envs.docker.cpu.test_integration.TestIntegration testMethod=test_benchmark>
    @inlineCallbacks
    def test_benchmark(self):
        config = DockerCPUConfig(work_dir=Path(tempfile.gettempdir()))
        env = DockerCPUEnvironment(config)
        yield env.prepare()
    
        Whitelist.add(env.BENCHMARK_IMAGE.split('/')[0])
>       score = yield env.run_benchmark()
tests/golem/envs/docker/cpu/test_integration.py:91: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.6/site-packages/twisted/internet/defer.py:1386: in _inlineCallbacks
    result = g.send(result)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <golem.envs.docker.cpu.DockerCPUEnvironment object at 0x7f3dd83ac3c8>
    @inlineCallbacks
    def run_benchmark(self) -> Deferred:
        image, tag = self.BENCHMARK_IMAGE.split(':')
        yield self.install_prerequisites(DockerPrerequisites(
            image=image,
            tag=tag,
        ))
        payload = DockerPayload(
            image=image,
            tag=tag,
            user=None if is_windows() else str(os.getuid()),
            env={},
        )
        runtime = self.runtime(payload)
        yield runtime.prepare()
        yield runtime.start()
        yield runtime.wait_until_stopped()
        # Benchmark is supposed to output a single line containing a float value
        stdout = list(runtime.stdout('utf-8'))
        yield runtime.clean_up()
>       return float(stdout[0])
E       IndexError: list index out of range
golem/envs/docker/cpu.py:537: IndexError
```

## Steps To Reproduce

Go to: https://buildbot.golem.network/buildbot/#/builders/9/builds/12938/steps/7/logs/stdio

## Expected behavior
No errors

Possibly related to #4185 or #4372

Krigpl:
Fixes #4395 

I dunno what's happening, because getting stdout of a stopped container should be deterministic but it's not. There's a possibility that the stdout will be empty for a while after stopping the container.

### DIFF
--- a/golem/envs/docker/cpu.py
+++ b/golem/envs/docker/cpu.py
@@ -531,10 +531,21 @@ class DockerCPUEnvironment(Environment):
         yield runtime.prepare()
         yield runtime.start()
         yield runtime.wait_until_stopped()
-        # Benchmark is supposed to output a single line containing a float value
-        stdout = list(runtime.stdout('utf-8'))
-        yield runtime.clean_up()
-        return float(stdout[0])
+        _, exit_code = runtime._inspect_container()
+        try:
+            if exit_code:
+                raise Exception(
+                    f'Benchmark run failed with exit code {exit_code}')
+            # Benchmark is supposed to output a single line containing
+            # a float value, but sometimes stdout is empty for a while after
+            # stopping the container
+            stdout = list(runtime.stdout('utf-8'))
+            while not stdout:
+                sleep(0.5)
+                stdout = list(runtime.stdout('utf-8'))
+            return float(stdout[0])
+        finally:
+            yield runtime.clean_up()
 
     @classmethod
     def metadata(cls) -> EnvMetadata:


golem/envs/docker/cpu.py
```diff
@@ -531,10 +531,21 @@ def run_benchmark(self) -> Deferred:
...
+                raise Exception(
+                    f'Benchmark run failed with exit code {exit_code}')
+            # Benchmark is supposed to output a single line containing
+            # a float value, but sometimes stdout is empty for a while after
+            # stopping the container
+            stdout = list(runtime.stdout('utf-8'))
+            while not stdout:
+                sleep(0.5)
+                stdout = list(runtime.stdout('utf-8'))
+            return float(stdout[0])
```
maaktweluit, line 546: Does this return break the `finally:`?
It feels neater to return outside of the `try:`

maaktweluit (commented):
LGTM!

One small q

golem/envs/docker/cpu.py
Krigpl: Nothing breaks `finally` and we still want to run `runtime.cleanup` when(/if) `float(stdout[0])` throws.

maaktweluit: ok, sounds good

maaktweluit (approved):

Wiezzel:
Whoa, that's some really ugly hack. It's fine for now if it works but I'd really love find some clean solution to this problem.

PR #4403, opened 2019-07-01
Rename DockerPayload

Krigpl:
Since it's a payload for the runtime let's call it RuntimePayload.
Also put all the relevant data into it, like directories bindings, etc.


golem/envs/__init__.py
```diff
@@ -56,23 +54,20 @@ class EnvEvent(NamedTuple):
 EnvEventListener = Callable[[EnvEvent], Any]
 
 
-class EnvConfig(DictSerializable, ABC):
```
Wiezzel: Why? We might need to serialize it.

golem/envs/__init__.py
```diff
@@ -56,23 +54,20 @@ class EnvEvent(NamedTuple):
...
+class EnvConfig(ABC):
     """ Environment-wide configuration. Specifies e.g. available resources. """
 
 
-class Prerequisites(DictSerializable, ABC):
```
Wiezzel: This one we definitely need. Prerequisites are supposed to be serialized and sent over the network in TaskHeader.

golem/envs/docker/__init__.py
```diff
@@ -1,41 +1,40 @@
...
 
 class DockerPrerequisites(DockerPrerequisitesData, Prerequisites):
-    """ This exists because NamedTuple must be single superclass """
-
-    def to_dict(self) -> Dict[str, Any]:
-        return self._asdict()
-
-    @staticmethod
-    def from_dict(data: Dict[str, Any]) -> 'DockerPrerequisites':
-        return DockerPrerequisites(**data)
```
Wiezzel: Please undo. Prerequisites are supposed to be serialized.

Wiezzel (changes requested):

Wiezzel (approved):
:+1:

Issue #4415, opened 2019-07-02
can't run binary testnet node; ERROR      Stopping because of 'Cannot start the client' error: No module named 'apps.glambda'


ederenn:
## Description

**Golem Version**:  GOLEM Version: 0.20.0+dev3.gc178317

**Golem-Messages version** (leave empty if unsure):   golem_messages Version: 3.7.0

**Electron version** (if used): 0.2.0 v9

**OS** [e.g. Windows 10 Pro]: all systems

**Branch** (if launched from source): -

**Mainnet/Testnet**: testnet 

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

It is not possible to start testnet Golem on any OS 
in `golem.log` 
```
2019-07-02 16:22:17 ERROR    golem.node                          Stopping because of 'Cannot start the client' error: No module named 'apps.glambda'
```

**Actual result**:

Golem is crashing

![Screenshot from 2019-07-02 16-32-04](https://user-images.githubusercontent.com/32061891/60521157-01051700-9ce7-11e9-826c-fd677471d927.png)

## Steps To Reproduce

1. Launch testnet node
2. Error modal displays in Gui
3. Check logs for error

## Logs and any additional context
```
2019-07-02 16:22:17 ERROR    golem.node                          Stopping because of 'Cannot start the client' error: No module named 'apps.glambda'
```

## Proposed Solution?
_(Optional: What could be a solution for that issue)_

mfranciszkiewicz:
Solution

Create empty `__init__.py` files at the following locations:
- `apps/glambda/__init__.py`
- `apps/glambda/benchmarks/__init__.py`
- `apps/glambda/task/__init__.py`

ederenn:
after fix another another thing is not working

```
2019-07-02 19:18:35 ERROR    golem.task.benchmarkmanager         Unable to run WASM benchmark: 
2019-07-02 19:18:35 CRITICAL golem.client                        Can't start network. Giving up.
Traceback (most recent call last):
  File "/home/buildbot-worker/worker/buildpackage_linux/build/golem/client.py", line 331, in start
  File "/home/buildbot-worker/worker/buildpackage_linux/build/golem/client.py", line 383, in start_network
[...]

PR #4416, opened 2019-07-02
Created __init__ files in glambda Python modules

kmazurek:
Fixes: #4415

This prevents the modules from being removed from Golem binaries.


kubkon (approved):
LGTM!

PR #4417, opened 2019-07-02
Asyncio reactor

mfranciszkiewicz:
Replaces the current event loop (Twisted) with the one based on `asyncio` (CPython core).

#### Introduced changes:
- `asyncioreactor` in place of a Twisted's platform-specific one
- upgraded `Twisted` and `Crossbar` to the newest versions
- upgraded `cryptography`, `cffi` and sub-dependencies of `pyinstaller`
- refactored the application start process in `golemapp.py` due to changes in Crossbar

#### Testing:
* **Full** Test suite

#### Possible regressions
* Relative slowdown on Windows (IOCP reactor dropped in favour of a select-based one)

#### Unlikely regressions
* RPC connection issues (GUI misbehaving)
* Network connection issues


golem/node.py
```diff
@@ -164,6 +164,7 @@ def __init__(self,  # noqa pylint: disable=too-many-arguments
         self._crossbar_serializer = crossbar_serializer
 
     def start(self) -> None:
+        from autobahn.wamp import ApplicationError
```
maaktweluit: maybe add a comment why this import is local only

golem/node.py
```diff
@@ -179,15 +180,21 @@ def on_rpc_ready() -> Deferred:
...
+                    sub_failure.trap(EnvironmentError)
+                    error_message = sub_failure.getErrorMessage()
 
                 logger.error(
                     """
                     There was a problem setting up the environment: {}
                     Golem will run with limited functionality to support
                     communication with local clients.
-                    """.format(sub_failure.getErrorMessage())
+                    """.format(error_message)
```
maaktweluit: nicer to use native logger formatting, but since its ERROR not so important

golem/rpc/__init__.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+
+def generate_rpc_certificate(data_dir: str):
+    from golem.rpc.cert import CertificateManager
+    from golem.rpc.common import CROSSBAR_DIR
+
+    cert_dir = os.path.join(data_dir, CROSSBAR_DIR)
```
maaktweluit, line 8: should we add `os.path` vs `pathlib` to the styleguide? would prefer to use only one

golem/rpc/router.py
```diff
@@ -55,35 +56,37 @@ def __init__(self,
...
 
-        self.node = Node(self.working_dir, reactor=reactor)
-        self.pubkey = self.node.maybe_generate_key(self.working_dir)
+        self.node = Node(personality=self.personality_cls,
+                         cbdir=self.working_dir,
+                         reactor=reactor)
 
-        workers = default_native_workers()
+        self.node.load_keys(self.working_dir)
+        self.node._config = self.config  # noqa # pylint: disable=protected-access
```
maaktweluit:
```suggestion
        self.node.load_config(None, self.config)
```

golemapp.py
```diff
@@ -111,35 +113,12 @@ def monkey_patched_getLogger(*args, **kwargs):
                   SerializerType.json.value,
               ]),
               help="Crossbar serializer (default: msgpack)")
-# Python flags, needed by crossbar (package only)
```
maaktweluit, line 114: yay, cleanup!

maaktweluit (changes requested):
LGTM!

Some small comments

golem/rpc/__init__.py
mfranciszkiewicz: This is function was moved from another file, from the time when we didn't commonly use `pathlib`

golem/node.py
mfranciszkiewicz: It doesn't need to be, it's just used in this one single place

golem/rpc/__init__.py
mfranciszkiewicz: It's a good idea to add this to the guidelines

golem/node.py
maaktweluit: ok, then no comment needed

golem/rpc/__init__.py
maaktweluit: ok, lets leave it for now and add it to the guidelines

maaktweluit (approved):

kmazurek:
I don't think I'm the right person to review these changes. I'd ask for a second opinion from someone more experienced with twisted and asyncio (@jiivan perhaps?).

jiivan (approved):
Looks good :+1: Have you tried running integration tests on this branch?

mfranciszkiewicz:
`scripts/node_integration_tests/tests/test_golem.py` pass on macOS and Linux

@maaktweluit reported a reoccurring issue from the older version of Twisted (reactor trying to register multiple writers for the same file descriptor) but I'm unable to reproduce it

mfranciszkiewicz:
I re-ran the suite multiple times on macOS and Linux and still cannot observe the reported issue. Tested with Python 3.6.4, 3.6.5 and 3.6.8.

The integration tests on buildbots behave similarly to the `develop` branch. Thus, the issue is most probably a local one.

Awaiting QA approval.

mfranciszkiewicz:
Golem seems to be stable, though the error still appears from time to time:
```
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/internet/asyncioreactor.py", line 136, in _readOrWrite
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/internet/abstract.py", line 271, in doWrite
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/internet/abstract.py", line 430, in stopWriting
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/internet/asyncioreactor.py", line 237, in removeWriter
  File "/usr/lib/python3.6/asyncio/selector_events.py", line 352, in remove_writer
    return self._remove_writer(fd)
  File "/usr/lib/python3.6/asyncio/selector_events.py", line 326, in _remove_writer
    self._selector.modify(fd, mask, (reader, None))
  File "/usr/lib/python3.6/selectors.py", line 261, in modify
    key = self.register(fileobj, events, data)
  File "/usr/lib/python3.6/selectors.py", line 405, in register
    key = super().register(fileobj, events, data)
  File "/usr/lib/python3.6/selectors.py", line 241, in register
    .format(fileobj, key.fd))
KeyError: '65 (FD 65) is already registered'
```

mfranciszkiewicz:
`removeWriter` seems to be a subject to a race condition. I've committed a patch suppressing the error.

golem/core/common.py
```diff
@@ -317,20 +317,33 @@ def config_logging(
...
 def install_reactor():
-
-    if is_windows():
-        from twisted.internet import iocpreactor
-        iocpreactor.install()
-    elif is_osx():
-        from twisted.internet import kqreactor
-        kqreactor.install()
+    from twisted.internet import asyncioreactor
+    asyncioreactor.install()
```
Wiezzel:
```suggestion
    asyncioreactor.install(asyncio.get_event_loop())
```

PR #4418, opened 2019-07-03
Expose ports from env runtimes

Krigpl:
So we can connect to the services running inside a Runtime.
Implemented for Linux and MacOS.

### DIFF
--- a/golem/envs/docker/cpu.py
+++ b/golem/envs/docker/cpu.py
@@ -703,7 +703,7 @@ class DockerCPUEnvironment(Environment):
         else:
             binds = None
 
-        if self._hypervisor.requires_ports_publishing():
+        if self._hypervisor.requires_ports_publishing() and payload.ports:
             port_bindings = {port: None for port in payload.ports}
         else:
             port_bindings = None
--- a/golem/envs/docker/cpu.py
+++ b/golem/envs/docker/cpu.py
@@ -131,6 +131,14 @@ class DockerInput(RuntimeInput):
         self._sock.close()
 
 
+class ContainerPortMapper:
+    def __init__(self, hypervisor: Hypervisor) -> None:
+        self._hypervisor = hypervisor
+
+    def get_port_mapping(self, container_id: str, port: int) -> Tuple[str, int]:
+        return self._hypervisor.get_port_mapping(container_id, port)
+
+
 class DockerCPURuntime(Runtime):
 
     CONTAINER_RUNNING: ClassVar[List[str]] = ["running"]
@@ -142,7 +150,8 @@ class DockerCPURuntime(Runtime):
             self,
             payload: DockerRuntimePayload,
             host_config: Dict[str, Any],
-            volumes: Optional[List[str]]
+            volumes: Optional[List[str]],
+            port_mapper: ContainerPortMapper,
     ) -> None:
         super().__init__(logger=logger)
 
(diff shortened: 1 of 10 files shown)


golem/envs/docker/cpu.py
```diff
@@ -683,11 +703,17 @@ def _create_host_config(
         else:
             binds = None
 
+        if self._hypervisor.requires_ports_publishing():
+            port_bindings = {port: None for port in payload.ports}
```
Wiezzel, line 708: `payload.ports` field is optional so this line would raise an error in case it is `None`. Need to add something like `(payload.ports or [])`. Or make `ports` non-optional and provide a default `[]` value.

Wiezzel (changes requested):
Therer's a "null pointer issue".
Also, please add some unit tests for `get_port_mapping` methods and expand the tests for `_create_host_config` so that they cover both cases when port publishing is required and when it's not.

Krigpl:
I'm against "unit" tests that test (and mock) internal methods like the `_create_host_config` family. This isn't unit test like.
But I'll conform with the existing style and add some.

Wiezzel:
@Krigpl I believe unit tests should test one method only. You can have "one public method" instead but this would make longer and harder (sometimes _a lot_) to write.

Krigpl:
Unit tests should tests public methods (and not necessarily just one method per test) only and not rely on how the code is structured internally.
Meaning that if you decide that you wanna inline private function like `_create_host_config` because suddenly it makes more sense to do that, thus not introducing any functional changes to the class, no unit test should break.

Tests that rely on internal code structure are really annoying when performing non-functional refactoring of the code.

Wiezzel:
I'd say it definitely should be one method per test. That's why they're called unit tests. They test single units of code. And units of code are methods/functions.

Wiezzel:
Also, you can never achieve full line or branch coverage without taking into account at least some implementation internals. That is to say, unit tests cannot be blackbox test.

Krigpl:
> And units of code are methods/functions.

Or classes. How would you unit test e.g. a hashmap implementation when limiting yourself to using one public method per test?

Krigpl:
> Also, you can never achieve full line or branch coverage without taking into account at least some implementation internals. That is to say, unit tests cannot be blackbox test.

Code coverage is a tool to help you write better unit tests. You can still use code coverage information to see what kind of scenario you need to test to achieve better coverage, you don't need to rely on the structure *in the test*.

Wiezzel:
>How would you unit test e.g. a hashmap implementation when limiting yourself to using one public method per test?

OK, that's a fine counterexample. But still in most cases it should be one method (and simple accessors don't count as methods).

Wiezzel:
Now, again regarding private method testing. It just usually makes test shorter and simpler.

First example: there's public method `a()` which calls private method `b()` which calls private method `c()`. Now, let's say there are 4 possible flows in each method (e.g. two if-s). If you test every method separately, you get full branch coverage with about 12 unit tests (4*3). On the other hand, if you treat all three as one code unit and test only via calling `a()` you should write 64 unit tests (4^3) to be able to say that you covered all possible scenarios.

Second example: how would you test the automatic status updates in `DockerCPURuntime` without creating tests for private methods? How much code would you need to cover all scenarios? I bet it would be a lot.

Wiezzel (approved):

Krigpl:
> Now, again regarding private method testing. It just usually makes test shorter and simpler.

I do agree with that. But as I mentioned it makes refactoring painful. My general approach to writing tests in Python is that I won't use approaches I wouldn't be able to use in more civilized languages.
It is also quite often the case that if writing tests is a painful process than the design of the module is likely just bad (not saying this is the case here, just my experience).

PR #4420, opened 2019-07-03
Added wasm and glambda to pyinstaller datas

kmazurek:
Resolves: #4415


kubkon (approved):
LGTM! Just a question though @zakaprov. This should be inclusive of any subdirs that we might find in any of the whitelisted dirs, yeah?

kmazurek:
@kubkon Yes, according to [this](https://pythonhosted.org/PyInstaller/spec-files.html#adding-data-files) part of the PyInstaller documentation this copies directories recursively. The WASM case seems to confirm this since `apps/wasm/test_data` contains only a subdirectory `input` with all the relevant data.

PR #4435, opened 2019-07-05
Implement Golem AppCallbacks

Krigpl:
Implement Task API's AppCallbacks interface using golem environments.

### DIFF
--- a/golem/task/appcallbacks/appcallbacks.py
+++ b/golem/task/appcallbacks/appcallbacks.py
@@ -1,0 +1,52 @@
+import abc
+from pathlib import Path
+from typing import Optional, Tuple, Type
+
+from golem_task_api import AppCallbacks
+
+from golem.core.deferred import sync_wait
+from golem.envs import Environment, Prerequisites, Runtime, RuntimePayload
+
+
+class TaskApiPayloadBuilder(abc.ABC):
+    @classmethod
+    @abc.abstractmethod
+    def create_payload(
+            cls,
+            prereq: Prerequisites,
+            shared_dir: Path,
+            command: str,
+            port: int,
+    ) -> RuntimePayload:
+        raise NotImplementedError()
+
+
+class EnvironmentCallbacks(AppCallbacks):
+    def __init__(
+            self,
+            env: Environment,
+            prereq: Prerequisites,
+            shared_dir: Path,
+            payload_maker: Type[TaskApiPayloadBuilder],
+    ) -> None:
+        self._shared_dir = shared_dir
+        self._prereq = prereq
+        self._env = env
+        self._payload_maker = payload_maker
+        self._runtime: Optional[Runtime] = None
+
+    def spawn_server(self, command: str, port: int) -> Tuple[str, int]:
+        runtime_payload = self._payload_maker.create_payload(
+            self._prereq,
+            self._shared_dir,
+            command,
+            port,
+        )
+        self._runtime = self._env.runtime(runtime_payload)
+        sync_wait(self._runtime.prepare())
+        sync_wait(self._runtime.start())
+        return self._runtime.get_port_mapping(port)
+
+    async def wait_after_shutdown(self) -> None:
+        assert self._runtime is not None
+        sync_wait(self._runtime.wait_until_stopped())
--- a/tests/golem/task/appcallbacks/test_docker.py
+++ b/tests/golem/task/appcallbacks/test_docker.py
@@ -7,7 +7,7 @@ from golem.envs.docker import DockerPrerequisites, DockerBind
 from golem.task.appcallbacks.docker import DockerTaskApiPayloadBuilder
 
 
-class TestDockerPayloadMaker(TestCase):
+class TestDockerPayloadBuilder(TestCase):
     def test_create_task_api_payload(self):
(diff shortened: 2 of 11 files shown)


golem/envs/__init__.py
```diff
@@ -509,6 +509,16 @@ def listen(self, event_type: EnvEventType, listener: EnvEventListener) \
         """ Register a listener for a given type of Environment events. """
         self._event_listeners.setdefault(event_type, set()).add(listener)
 
+    @abstractmethod
+    def create_task_api_payload(
```
Wiezzel: Having task-api-specific method in computational environment seems out of place.
I suggest changing the signature a little bit:
 * make `shared_dir` and `port` sequences instead of single values
 * make `command` optional

Then the method would be general enough to be called simply `create_payload` and don't bear any direct relationship to the task API.

golem/task/appcallbacks.py
```diff
@@ -0,0 +1,36 @@
...
+            env: Environment,
+            prereq: Prerequisites,
+            shared_dir: Path,
+    ) -> None:
+        self._shared_dir = shared_dir
+        self._prereq = prereq
+        self._env = env
+        self._runtime: Optional[Runtime] = None
+
+    def spawn_server(self, command: str, port: int) -> Tuple[str, int]:
```
Wiezzel: Actually, why isn't this method `async`? (I understand that for now it wouldn't make a difference, because `sync_wait` would be called anyway, but in the future it could be truly asynchronous).

tests/golem/envs/docker/cpu/test_env.py
```diff
@@ -640,3 +642,26 @@ def test_nonpublished_ports(self, local_client):
...
+
+
+class TestTaskApiPayload(TestDockerCPUEnv):
+    def test_create_task_api_payload(self):
+        prereq = DockerPrerequisites(image='image', tag='tag')
+        shared_dir = Path('shared_dir')
+        command = 'cmd'
+        port = 4444
+
+        runtime = self.env.create_task_api_payload(
```
Wiezzel: Why is called `runtime` when it's a payload?

tests/golem/envs/docker/cpu/test_env.py
```diff
@@ -640,3 +642,26 @@ def test_nonpublished_ports(self, local_client):
...
+            port,
+        )
+
+        self.assertEqual(prereq.image, runtime.image)
+        self.assertEqual(prereq.tag, runtime.tag)
+        self.assertEqual(command, runtime.command)
+        self.assertEqual([port], runtime.ports)
+        bind = \
+            DockerBind(source=shared_dir, target=f'/{api_constants.WORK_DIR}')
+        self.assertEqual([bind], runtime.binds)
```
Wiezzel: Payloads support equality comparison so all these could be done with one assertion. But I'm not sure if it's better.

golem/task/appcallbacks/appcallbacks.py
```diff
@@ -0,0 +1,52 @@
...
+        raise NotImplementedError()
+
+
+class EnvironmentCallbacks(AppCallbacks):
+    def __init__(
+            self,
+            env: Environment,
+            prereq: Prerequisites,
+            shared_dir: Path,
+            payload_maker: Type[TaskApiPayloadMaker],
```
Wiezzel: How will the code calling this constructor establish which `payload_maker` class should it provide?

Krigpl: I imagine that we'll have some mapping `"env name" -> env_instance`, so this could be simply extended to include this class as well.

Wiezzel: Well, we have `EnvironmentManager` but it doesn't know anything about payload makers. If you want it this way then you need to modify the manager. But I'm not sure if it's worth the trouble.

Krigpl: imo it is. Your approach cannot be applied - it's basically making the task api dependency implicit rather than explicit.
We need a method that takes prerequisites and outputs runtime payload for task api. I agree it shouldn't live in the `Environment` class and the proposed way is to basically have a full env description, similar to what we have right now in the `ini` files (but without the `ini` files). It would describe that the env class and particular payload maker are connected.

Wiezzel: OK, fair enough. So please adjust the `EnvironmentManager` to manage `TaskApiPayloadMaker`s as well.
And, if I may suggest a little name change, I think maybe `Builder` would be a better name than `Maker` because it's already an established term among coders.

tests/golem/task/appcallbacks/test_docker.py
```diff
@@ -0,0 +1,30 @@
+from pathlib import Path
+from unittest import TestCase
+
+from golem_task_api import constants as api_constants
+
+from golem.envs.docker import DockerPrerequisites, DockerBind
+from golem.task.appcallbacks.docker import DockerTaskApiPayloadBuilder
+
+
+class TestDockerPayloadMaker(TestCase):
```
Wiezzel:
```suggestion
class TestDockerPayloadBuilder(TestCase):
```

Wiezzel (approved):

PR #4440, opened 2019-07-08
Remove redundant provider client checks

tworec:
I found those redundant checks when I was testing #4177 with a single requestor and three providers.
It is a waste of resources to check things twice, but what's more important, we are also wasting developer time to read and extend such a code.

### DIFF
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -394,7 +394,7 @@ class TaskManager(TaskEventListener):
                          task_id: str,
                          estimated_performance: float,
                          price: int,
-                         offer_hash: bytes) \
+                         offer_hash: str) \
             -> Optional[ComputeTaskDef]:
         """ Assign next subtask from task <task_id> to node with given
         id <node_id>.
@@ -477,7 +477,7 @@ class TaskManager(TaskEventListener):
     def should_wait_for_node(self,
                              task_id: str,
                              node_id: str,
-                             offer_hash: Optional[bytes]) -> bool:
+                             offer_hash: str) -> bool:
         """ Check if the node has too many tasks assigned already """
         if not self.is_my_task(task_id):
             logger.debug(
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -394,7 +394,7 @@ class TaskManager(TaskEventListener):
                          task_id: str,
                          estimated_performance: float,
                          price: int,
-                         wtct_hash: Optional[bytes]) \
+                         wtct_hash: bytes) \
             -> Optional[ComputeTaskDef]:
         """ Assign next subtask from task <task_id> to node with given
         id <node_id>.
--- a/golem/task/taskmanager.py
+++ b/golem/task/taskmanager.py
@@ -394,7 +394,7 @@ class TaskManager(TaskEventListener):
                          task_id: str,
                          estimated_performance: float,
                          price: int,
-                         wtct_hash: bytes) \
+                         offer_hash: bytes) \
             -> Optional[ComputeTaskDef]:
         """ Assign next subtask from task <task_id> to node with given
         id <node_id>.
@@ -420,7 +420,7 @@ class TaskManager(TaskEventListener):
         if not self.task_needs_computation(task_id):
             return None
 
-        if self.should_wait_for_node(task_id, node_id, wtct_hash):
+        if self.should_wait_for_node(task_id, node_id, offer_hash):
             return None
 
         task = self.tasks[task_id]
@@ -477,7 +477,7 @@ class TaskManager(TaskEventListener):
     def should_wait_for_node(self,
                              task_id: str,
                              node_id: str,
-                             wtct_hash: Optional[bytes]) -> bool:
+                             offer_hash: Optional[bytes]) -> bool:
         """ Check if the node has too many tasks assigned already """
         if not self.is_my_task(task_id):
             logger.debug(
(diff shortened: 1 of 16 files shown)


golem/task/tasksession.py
```diff
@@ -277,96 +277,38 @@ def _react_to_want_to_compute_task(self, msg):
...
-            logger.debug(
-                "TaskFinished. task_id=%s, provider=%s",
-                msg.task_id,
-                node_name_id,
-            )
+            logger.debug("TaskFinished. %s", task_node_info)
             self._cannot_assign_task(msg.task_id, reasons.TaskFinished)
             return
 
-        if self.task_manager.should_wait_for_node(msg.task_id, self.key_id,
```
tworec, line 338: removing this because all its logic has already is covered upwards
- `is_my_task` directly in L270
- `should_accept_client` within `should_accept_provider`

golem/task/taskserver.py
```diff
@@ -704,14 +704,14 @@ class RejectedReason(Enum):
...
+    def should_accept_provider(  # pylint: disable=too-many-return-statements
             self,
-            node_id,
-            address,
-            task_id,
-            provider_perf,
-            max_resource_size,
-            max_memory_size):
+            node_id: str,
+            ip_addr: str,
```
mplebanski, line 710: Rest of the file uses `address` for a variable holding IP address. Please either revert it here or change it everywhere in the file.

golem/task/taskmanager.py
```diff
@@ -394,7 +394,7 @@ def get_next_subtask(self,
                          task_id: str,
                          estimated_performance: float,
                          price: int,
-                         wtct_hash: Optional[bytes]) \
```
mplebanski, line 397: Is this a logical equivalent of `offer_id`? If so maybe an `offer_id` would be a more readable name. I know this is out of scope of this PR, but we also recently introduced name wtct_hash at task level, which to me is a somewhat broken abstraction, because task now knows how one of the internal golem protocol messages is named and what is represents.

golem/task/tasksession.py
mplebanski: Since this is the only place where message WaitingForResults is sent it could be removed at all. Did you check how it affects provider side if this message is never received?

golem/task/taskserver.py
tworec: I've found just one IP address variable named `address` in this file -- argument of `set_last_message`

golem/task/taskmanager.py
tworec: fair and square. I'll rename it to `offer_hash`

golem/task/tasksession.py
tworec: yup. I have tested it with manual test on 1 Requestor and 3 Providers. And also running via local integration tests.

golem/task/taskserver.py
mplebanski: Methods: `set_last_message`, `new_connection`, `get_socket_addresses`

mplebanski:
@tworec OK please remove `_react_to_waiting_for_results` method, `WaitingForResultsTestCase` and `WaitingForResults` from `__set_msg_interpretations`, it's dead code now. @jiivan Shall it be removed from `golem-messages` also?

golem/task/tasksession.py
tworec: all local integration tests have passed

golem/task/taskserver.py
tworec: in `new_connection` the `address` is not a variable/argument but a `session` property, which I'd rather not rename, to not blow away this PR. But if you think it's worth doing, we can...

tworec: and `address` in `get_socket_addresses` can be a ip **or a hostname** afaict and check

mplebanski: I'd rather consider using `address` instead of `ip_addr` then. What do you think?

tworec: I think it is misleading. It was to me when I was reading this ...

mplebanski: OK. I agree this might be sometimes confused with eth address.

mplebanski (approved):

tworec:
thx

golem/task/tasksession.py
```diff
@@ -382,23 +325,18 @@ def _react_to_want_to_compute_task(self, msg):
...
+        task_node_info = "task_id=%r, node=%r" % (
+            msg.task_id, common.short_node_id(self.key_id))
         reasons = message.tasks.CannotAssignTask.REASON
         if not is_chosen:
-            logger.info(
-                "Provider not chosen by marketplace. task_id=%r, node=%r",
-                msg.task_id,
-                node_name_id,
-            )
+            logger.info("Provider not chosen by marketplace:%s", task_node_info)
```
jiivan, line 339: Missing whitespace

tworec: intentionally ;)

mplebanski:
Merging those changes to my dev branch creates a feeling that everything is progressing more slowly. A wild guess would be that providers are now not reacting to WaitingForResults with a offer resend. @tworec Could you check if you are experiencing the same in a manual setup? 
CC: @kubkon @mbenke

tworec:
@mplebanski Providers were not reacting to `WaitingForResults` even before this change, because this message was never sent. Requestor was sending `CannotAssignTask` with `NoMoreSubtasks` reason.

https://github.com/golemfactory/golem/pull/4440/files#diff-88ebb160824a1d131d3a3cd84874e290R290 

or in rare cases with `NotMyTask` reason
https://github.com/golemfactory/golem/pull/4440/files#diff-88ebb160824a1d131d3a3cd84874e290R270

Those were only two situations which could trigger sending `WaitingForResults`, but conditions for it was coded to be checked after them, so it could not be invoked

https://github.com/golemfactory/golem/pull/4440/files#diff-88ebb160824a1d131d3a3cd84874e290L338

PR #4443, opened 2019-07-09
Make env manager support payload builder

Krigpl:
Follow up to https://github.com/golemfactory/golem/pull/4435#discussion_r301437105

### DIFF
--- a/golem/task/envmanager.py
+++ b/golem/task/envmanager.py
@@ -1,20 +1,33 @@
-from typing import Dict, List
+from typing import Dict, List, NamedTuple, Type
 
 from golem.envs import EnvId, Environment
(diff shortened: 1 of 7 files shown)


Wiezzel (commented):
Seems alright, but we're gonna probably need some accessor method for payload builders.

Krigpl:
Probably, but I'm reluctant to introduce them without having a nonabstract use case (even now I'm not sure how/when certain public methods will be used, like `state` and `set_state`).

Wiezzel:
I'd say that something like this is quite a safe bet.
```python
def payload_builder(self, env_id: EnvId) -> Type[TaskApiPayloadBuilder]:
    return self._envs[env_id].payload_builder
```
But if you don't agree you can merge it as it is.

Wiezzel (approved):

PR #4449, opened 2019-07-10
Stop Golem from crashing on benchmark failure

mplebanski:
A rationale is that Golem should not crash on a failed benchmark. Ideally it should disable the app for this particular run. Since this is not straightforward to add I propose this PR which sets app's performance to 0. It will at least stabilize startup behavior. 

The context is GLambda's secure runtime for docker - gVisor, randomly crashing due to problems with permissions on `/sys/fs/cgroup/cpuset/docker/cpuset.cpu`. We are probably hitting issue described [here](https://bugzilla.redhat.com/show_bug.cgi?id=1609785#c1). From my trials one such cpuset mask refresh is e.g. running `docker run -it --cpuset-cpus=0 some_image`, although I'm not sure if it always works.

### DIFF
--- a/apps/glambda/glambdaenvironment.py
+++ b/apps/glambda/glambdaenvironment.py
@@ -1,9 +1,13 @@
+import filecmp
+import logging
 import shutil
 from typing import Dict
 
 from golem.core.common import is_linux
 from golem.docker.environment import DockerEnvironment
 from golem.environments.environment import SupportStatus, UnsupportReason
+
+logger = logging.getLogger(__name__)
 
 
 class GLambdaTaskEnvironment(DockerEnvironment):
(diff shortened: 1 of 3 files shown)


mplebanski:
A command line fix for crashing GLambda app:

`# cat /sys/fs/cgroup/cpuset/cpuset.cpus > /sys/fs/cgroup/cpuset/docker/cpuset.cpus`

mfranciszkiewicz:
@mplebanski ~Maybe it's a good idea to update the docker cpu set during installation?~

We should also disable the environment. That could be done with `pydispatcher` if it's difficult to reach the `EnvironmentsManager` instance.

shadeofblue (approved):
as far as I could tell, looks okay :)

mplebanski:
I moved the check to separate method and added FileNotFoundError exception handling.

mfranciszkiewicz (approved):

PR #4450, opened 2019-07-10
A different workaround for missing docker output issue

Wiezzel:
A slightly more elegant approach. Doesn't use sleep.

Here's the test for this PR: https://github.com/golemfactory/golem/commit/b08003653f395f68e09a43c5195b011a8f76f610#diff-60b01f4ff8ebc62e4b3a7185c33a38e0
And here's a triggered build for this test branch: https://buildbot.golem.network/buildbot/#/builders/9/builds/13330


Krigpl (approved):
thisisgine

maaktweluit (approved):
LGTM!

Ran the tests on the machine that broke 20+ times and it seems good

PR #4452, opened 2019-07-10
Docker environment integration tests cleanup

Wiezzel:
* Moved common common code to `setUp()`/`tearDown()`
* Extended timeout for test_io because DockerToolbox is slow

### DIFF
--- a/tests/golem/envs/docker/cpu/test_integration.py
+++ b/tests/golem/envs/docker/cpu/test_integration.py
@@ -1,12 +1,10 @@
 import socket
-import tempfile
-from pathlib import Path
 
 import pytest
 from twisted.internet.defer import inlineCallbacks
 from twisted.trial.unittest import TestCase
 
-from golem.envs import EnvStatus, RuntimeStatus
+from golem.envs import RuntimeStatus
 from golem.envs.docker import DockerPrerequisites, DockerRuntimePayload
 from golem.envs.docker.cpu import DockerCPUConfig, DockerCPUEnvironment
 from golem.envs.docker.whitelist import Whitelist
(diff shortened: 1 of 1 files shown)


tests/golem/envs/docker/cpu/test_integration.py
```diff
@@ -17,31 +17,31 @@
 @ci_skip
 class TestIntegration(TestCase, DatabaseFixture):
 
-    @pytest.mark.timeout(60)  # 60 sec should be well enough for this test
     @inlineCallbacks
-    def test_io(self):
-        # Set up environment
+    def setUp(self):
+        DatabaseFixture.setUp(self)
```
maaktweluit, line 20: Maybe its better to do a generic super setup?
```suggestion
        super().setUp()
```

tests/golem/envs/docker/cpu/test_integration.py
```diff
@@ -17,31 +17,31 @@
...
 
-        # Add environment cleanup to clean it if test goes wrong
-        def _clean_up_env():
-            if env.status() != EnvStatus.DISABLED:
-                env.clean_up()
-        self.addCleanup(_clean_up_env)
+    @inlineCallbacks
+    def tearDown(self):
+        yield self.env.clean_up()
+        DatabaseFixture.tearDown(self)
```
maaktweluit, line 32: same
```suggestion
        super().tearDown()
```

maaktweluit (approved):
LGTM!

2 small suggestions, not breaking so still approved

tests/golem/envs/docker/cpu/test_integration.py
Wiezzel: AFAIK calling `super().setUp()` would be equivalent to `TestCase.setUp()` here, not `DatabaseFixture.setUp()`. This is not what we want.

tests/golem/envs/docker/cpu/test_integration.py
Wiezzel: Ditto.

tests/golem/envs/docker/cpu/test_integration.py
```diff
@@ -17,31 +17,31 @@
 @ci_skip
 class TestIntegration(TestCase, DatabaseFixture):
 
-    @pytest.mark.timeout(60)  # 60 sec should be well enough for this test
     @inlineCallbacks
-    def test_io(self):
-        # Set up environment
+    def setUp(self):
+        DatabaseFixture.setUp(self)
         config = DockerCPUConfig(work_dirs=[Path(tempfile.gettempdir())])
```
Krigpl: Replace `tempfile.gettempdir` with `self.new_path`, otherwise it's not getting cleaned and there's no point in inheriting from TempDirFixture.

Krigpl (changes requested):

tests/golem/envs/docker/cpu/test_integration.py
Wiezzel: Well, I already wanted to do this but decided not to. Here's why:
1. We don't actually put anything in that directory for now so cleanup is not needed. `work_dirs` are just directories that Docker can use for sharing files but there's no test including file sharing.
2. There's point inheriting from `DatabaseFixture` because we want to use the database.
3. Using `self.new_path` would trigger a pop-up asking for admin permissions every single time a test is run. Using `tempdir` allows to do that just once.

Krigpl: 2. Imo if database is only used for the Whitelist it'd be better to just mock it. DatabaseFixture takes ~2-3s to set up.
3. lol ok. Please put that in the comment because this isn't obvious and someone might just change it in the future.

Wiezzel: Ad. 2 I didn't want to mock since it's supposed to be an integration test. I'd love to use an in-memory db but couldn't make it work on Windows.
Ad. 3 Actually, I've just come up with a better idea. We can use `self.new_path.parent.parent` which is `ComputerRes` directory that is shared anyway so no pop-ups will appear.

Wiezzel: @Krigpl like it?

Krigpl (approved):

tests/golem/envs/docker/cpu/test_integration.py
maaktweluit: i always thought it would do both of the parent classes, made a small test and it shows you are right

PR #4454, opened 2019-07-11
package hash/size cleanup

Krigpl:
ResourceServer::add_resources takes hash and size just in order to return them. This is unnecessary.

### DIFF
--- a/golem/resource/base/resourceserver.py
+++ b/golem/resource/base/resourceserver.py
@@ -46,13 +46,12 @@ class BaseResourceServer:
     def sync_network(self):
         self._download_resources()
 
-    def add_resources(self, pkg_path, pkg_sha1, res_id, pkg_size,  # noqa pylint:disable=too-many-arguments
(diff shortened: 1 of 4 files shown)


golem/resource/base/resourceserver.py
```diff
@@ -46,13 +46,12 @@ def get_distributed_resource_root(self):
     def sync_network(self):
         self._download_resources()
 
-    def add_resources(self, pkg_path, pkg_sha1, res_id, pkg_size,  # noqa pylint:disable=too-many-arguments
-                 client_options=None) -> Deferred:
+    def add_resources(self, pkg_path, res_id, client_options=None) -> Deferred:
```
Wiezzel, line 49: Maybe you could add type annotations for parameters?

Wiezzel (approved):

golem/resource/base/resourceserver.py
Krigpl: Meh, next time.

PR #4455, opened 2019-07-11
Added .coveragerc to exclude some lines

Wiezzel:
Also made all abstract methods rise NotImplementedError according to [the convention](https://docs.python.org/3/library/exceptions.html#NotImplementedError). As it turned out some of the supposedly abstract methods were actually being called some test adjustments were required.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,5 @@
+[report]
+exclude_lines =
+    pragma: no cover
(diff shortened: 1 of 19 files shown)


maaktweluit (approved):
LGTM!

mfranciszkiewicz (approved):

PR #4458, opened 2019-07-11
Confirm outgoing GNT

jiivan:
@igor discovered that GNT withdraw confirmation didn't work as expected.

### DIFF
--- a/golem/ethereum/paymentskeeper.py
+++ b/golem/ethereum/paymentskeeper.py
@@ -75,25 +75,6 @@ class PaymentsKeeper:
         """ Create new payments keeper instance"""
         self.db = PaymentsDatabase()
 
-    @staticmethod
-    def confirmed_transfer(
-            tx_hash: str,
-            gas_cost: int,
-    ):
-        try:
-            operation = model.WalletOperation.select() \
-                .where(
-                    model.WalletOperation.tx_hash == tx_hash,
-                ).get()
-            operation.status = model.WalletOperation.STATUS.confirmed
-            operation.gas_cost = gas_cost
-            operation.save()
-        except model.WalletOperation.DoesNotExist:
-            logger.warning(
-                "Got confirmation of unknown transfer. tx_hash=%s",
-                tx_hash,
-            )
-
     def get_list_of_all_payments(self, num: Optional[int] = None,
                                  interval: Optional[datetime.timedelta] = None):
         # This data is used by UI.
--- a/golem/ethereum/paymentskeeper.py
+++ b/golem/ethereum/paymentskeeper.py
@@ -105,3 +105,28 @@ class PaymentsKeeper:
             self,
             subtask_ids: Iterable[str]) -> List[model.TaskPayment]:
         return self.db.get_subtasks_payments(subtask_ids)
+
+    @staticmethod
+    def confirmed_transfer(
+            tx_hash: str,
+            successful: bool,
+            gas_cost: int,
+    ) -> None:
+        try:
+            operation = model.WalletOperation.select() \
+                .where(
+                    model.WalletOperation.tx_hash == tx_hash,
+                ).get()
+        except model.WalletOperation.DoesNotExist:
+            logger.warning(
+                "Got confirmation of unknown transfer. tx_hash=%s",
+                tx_hash,
+            )
+            return
+        if not successful:
+            logger.error("Failed transaction. tx_hash=%s", tx_hash)
+            operation.on_failed(gas_cost=gas_cost)
+            operation.save()
+            return
+        operation.on_confirmed(gas_cost=gas_cost)
+        operation.save()
--- a/golem/ethereum/transactionsystem.py
+++ b/golem/ethereum/transactionsystem.py
@@ -618,16 +618,15 @@ class TransactionSystem(LoopingCallService):
             receipt: 'sci_structs.TransactionReceipt',
             gas_price: Optional[int] = None,
     ):
-        tx_hash = f'0x{receipt.tx_hash}'
         try:
             operation = model.WalletOperation.select() \
                 .where(
-                    model.WalletOperation.tx_hash == tx_hash,
+                    model.WalletOperation.tx_hash == receipt.tx_hash,
                 ).get()
         except model.WalletOperation.DoesNotExist:
             log.warning(
                 "Got confirmation of unknown transfer. tx_hash=%s",
-                tx_hash,
+                receipt.tx_hash,
             )
             return
         if gas_price is None:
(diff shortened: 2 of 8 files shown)


golem/ethereum/transactionsystem.py
```diff
@@ -612,6 +612,36 @@ def get_withdraw_gas_cost(
             return self._sci.GAS_WITHDRAW
         raise ValueError('Unknown currency {}'.format(currency))
 
+    @sci_required()
+    def on_confirmed(
+            self,
+            receipt: 'sci_structs.TransactionReceipt',
+            gas_price: Optional[int] = None,
+    ):
+        tx_hash = f'0x{receipt.tx_hash}'
```
Krigpl: tx_hash from the receipt has `0x` already

golem/ethereum/transactionsystem.py
```diff
@@ -612,6 +612,36 @@ def get_withdraw_gas_cost(
             return self._sci.GAS_WITHDRAW
         raise ValueError('Unknown currency {}'.format(currency))
 
+    @sci_required()
+    def on_confirmed(
```
Krigpl: make it protected

golem/ethereum/transactionsystem.py
```diff
@@ -875,6 +901,7 @@ def concent_withdraw(self):
 
         def on_confirmed(_receipt) -> None:
```
Krigpl: drop `_` since now it's used

Krigpl (changes requested):

golem/ethereum/transactionsystem.py
jiivan: No it doesn't
```python
In [10]: raw_receipt
Out[10]: 
{'blockHash': b"\xcb\xcaI\xfb,u\xba/\xad\xa5ln\xa7\xdfYyDA'\xd2\x9bkN\x93\xa7y}\xc2.\x979\x9c",
 'blockNumber': 2940769,
 'gasUsed': 21000,
 'status': 1,
 'transactionHash': b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'}

In [11]: golem_sci.structs.TransactionReceipt(raw_receipt).tx_hash
Out[11]: '0000000000000000000000000000000000000000000000000000000000000000'
```

Krigpl: It does. `raw['transactionHash']` is not simply bytes, it's a special struct from web3 that when called `.hex()` on it, it generates a hex value with `0x` prefix.

jiivan: :-0

golem/ethereum/transactionsystem.py
```diff
@@ -319,30 +319,30 @@ def _subscribe_to_events(self) -> None:
             ),
         )
 
-        def _outgoing_gnt_transfer_confirmed(
-                event: 'sci_events.GntTransferEvent',
-        ):
-            assert self._sci is not None  # mypy :(
-            receipt: TransactionReceipt = self._sci.get_transaction_receipt(
-                event.tx_hash,
+        unconfirmed_query = model.WalletOperation.select() \
```
Krigpl: It'd be great to have this query somewhere else, like PaymentsKeeper

golem/ethereum/paymentskeeper.py
```diff
@@ -75,25 +75,6 @@ def __init__(self) -> None:
         """ Create new payments keeper instance"""
         self.db = PaymentsDatabase()
 
-    @staticmethod
-    def confirmed_transfer(
```
Krigpl, line 79: imo it's worth it to keep this method - thus keeping all the operations on models here and outside the transaction system

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -845,3 +847,50 @@ def test_nodeskeeper_record_not_present(self):
             ],
             self.ets.get_incomes_list(),
         )
+
+
+class TransactionConfirmationTest(TransactionSystemBase):
+    def setUp(self):
+        super().setUp()
+        self.ets = self._make_ets(provide_gntdeposit=True)
+        self.ets._gntb_balance = 100
```
Krigpl: Please use the `_refresh_balances` approach as in other tests.

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -845,3 +847,50 @@ def test_nodeskeeper_record_not_present(self):
...
+        self.sci.get_transaction_gas_price.return_value = 2
+
+    def test_transfer_eth(self):
+        self.sci.transfer_eth.return_value = self.tx_hash
+        self.ets.withdraw(
+            amount=1,
+            destination=random_eth_address(),
+            currency='ETH',
+        )
+        self.ets._on_confirmed(self.receipt)
```
Krigpl: Assert that `sci.on_transaction_confirmed` has been called and call the provided callbacks instead of invoking this function manually.

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -845,3 +847,50 @@ def test_nodeskeeper_record_not_present(self):
...
+
+    def test_deposit_transfer(self):
+        self.sci.deposit_payment.return_value = self.tx_hash
+        self.sci.get_deposit_value.return_value = 0
+        defer = self.ets.concent_deposit(
+            required=10,
+            expected=40,
+        )
+        deferred.sync_wait(defer)
+        self.ets._on_confirmed(self.receipt)
```
Krigpl: ditto

tests/golem/ethereum/test_transactionsystem.py
```diff
@@ -27,6 +32,21 @@
 PASSWORD = 'derp'
 
 
+def get_transaction_receipt(tx_hash, status=1):
+    return golem_sci.structs.TransactionReceipt(
+        raw_receipt={
```
Krigpl, line 37: Imo there's no reason for this function to be implemented this way, you could instead use `Mock(spec=TransactionReceipt)` and set fields you are interested in.
Especially that `blockHash` is still buggy just like `transactionHash` has been (raw bytes).

Krigpl (changes requested):

tests/golem/ethereum/test_transactionsystem.py
jiivan: It a simple struct. Why should I mock it?

Krigpl: To avoid the kind of bugs that you already introduced (incorrect types in the raw_receipt).

Krigpl (approved):

tests/golem/ethereum/test_transactionsystem.py
jiivan: I got your point. But reason I introduced that function was exactly that. I wasn't sure what was the correct type. Once this function is bug-free it will be a good factory for all tests here. Better than mock IMO.

PR #4462, opened 2019-07-12
Task API task creation RPC

Krigpl:


golem/task/taskserver.py
```diff
@@ -60,6 +60,7 @@
 from golem.task.taskconnectionshelper import TaskConnectionsHelper
 from golem.task.taskstate import TaskOp
 from golem.utils import decode_hex
+from .requestedtaskmanager import RequestedTaskManager
```
maaktweluit: The new style guide does not like relative imports: https://github.com/golemfactory/styleguide/blob/gh-pages/pyguide.md#22-imports

golem/task/requestedtaskmanager.py
```diff
@@ -0,0 +1,41 @@
+from pathlib import Path
+from typing import Dict, Any, List
+
+from dataclasses import dataclass
+
+TaskId = str
+
+
+@dataclass
+class CreateTaskParams:
```
Wiezzel, line 10: Could you change the param names to be consistent with `RequestedTask` model? https://github.com/golemfactory/golem/blob/5ff93ac96cffd5851d5a186d701ff6e385a0262f/golem/model.py#L637-L656
If you disagree on some naming choice you can of course propose a change in the model instead.

golem/task/requestedtaskmanager.py
```diff
@@ -0,0 +1,41 @@
...
+    timeout: int
+    subtask_timeout: int
+    concent_enabled: bool
+
+
+class RequestedTaskManager:
+    def create_task(
+            self,
+            params: CreateTaskParams,
+            task_params: Dict[str, Any],
```
Wiezzel: Maybe `golem_params` and `app_params`?

tests/golem/task/test_rpc_task_api.py
```diff
@@ -0,0 +1,132 @@
...
+
+        new_task_id = self.rpc.create_task_api_task(task_params, golem_params)
+        self.assertEqual(task_id, new_task_id)
+        self.requested_task_manager.create_task.assert_called_once_with(
+            mock.ANY,
+            task_params,
+        )
+        create_task_params = \
+            self.requested_task_manager.create_task.call_args[0][0]
+        self.assertEqual(
```
Wiezzel, line 74: Maybe instead of writing so many comparisons you could do something like this?
```python
self.assertEqual(
    create_task_params,
    requestedtaskmanager.CreateTaskParams(**golem_params))
```

Wiezzel (commented):
I think the module should be called `requested_task_manager` not `requestedtaskmanager`. At least that's what our [styleguide says](https://github.com/golemfactory/styleguide/blob/gh-pages/pyguide.md#s3.16-naming):
> `module_name`, `package_name` ....

Krigpl:
> I think the module should be called requested_task_manager not requestedtaskmanager. At least that's what our styleguide says:

Ugh, I think not a single package in Golem conforms to that style. So it's a matter whether we keep it consistent for now until we do a huge refactor or we make things messy. I went with consistency but can change it as well.

Wiezzel:
@Krigpl Well, ok. Let's keep it like that for now.

tests/golem/task/test_rpc_task_api.py
Krigpl: `golem_params` is a struct provided from the rpc call and I imagine bigints will be represented by strings as well as filepaths will be raw strings. This set of asserts make sure that these kinds of values are converted properly to the desired types.
Like here:
```
        self.assertEqual(
            [Path(r) for r in golem_params['resources']],
            create_task_params.resources,
        )
```

golem/task/requestedtaskmanager.py
Krigpl: Done, except for the `output_file` which should be `output_directory`.

tests/golem/task/test_rpc_task_api.py
Wiezzel: OK.

golem/task/requestedtaskmanager.py
```diff
@@ -0,0 +1,41 @@
...
+    def create_task(
+            self,
+            golem_params: CreateTaskParams,
+            app_params: Dict[str, Any],
+    ) -> TaskId:
+        """ Creates an entry in the storage about the new task and assigns
+        the task_id to it. The task then has to be initialized and started. """
+        raise NotImplementedError
+
+    def init_task(self, task_id: TaskId) -> None:
```
mfranciszkiewicz, line 33: This function should probably be asynchronous (`inlineCallbacks` / returns `Deferred`)

golem/task/rpc.py
```diff
@@ -499,6 +512,69 @@ def create_task(self, task_dict, force=False) \
...
+        @defer.inlineCallbacks
+        def init_task():
+            try:
+                self.requested_task_manager.init_task(task_id)
+
+                # ensure Concent deposit
+                min_amount, opt_amount = msg_helpers.requestor_deposit_amount(
+                    create_task_params.max_price_per_hour,
+                )
+                yield self.client.transaction_system.concent_deposit(
```
mfranciszkiewicz: Is it required for task types other than Blender?

golem/task/requestedtaskmanager.py
Krigpl: It most likely will be, just like some others in this class. For now it's just an interface, so I didn't bother to decide whether it's gonna be deferred or async.

golem/task/rpc.py
Krigpl: Technically no atm, but ideally yes.

mfranciszkiewicz: From what I know the existing code is going to be refactored to not deposit funds for tasks other than Blender. But we can leave this part as is for now and refactor it later.

Krigpl: I guess I can just get rid of that since Concent won't support task api tasks for a while anyway.

Wiezzel (approved):

mfranciszkiewicz (approved):

maaktweluit (approved):

PR #4469, opened 2019-07-15
Brass marketplace refactoring

mplebanski:


golem/task/tasksession.py
```diff
@@ -13,6 +12,8 @@
 from golem_messages import message
 from golem_messages import utils as msg_utils
 from pydispatch import dispatcher
+
+import twisted
```
maaktweluit: be carefull importing twisted on the top level of a file, maybe you can import `task` from `twisted.internet`? it looks like its the only thing used from there

mplebanski:
Work continued in #4512

PR #4483, opened 2019-07-16
Remove deadline param from funds locker

Krigpl:
Tasks are removed by `remove_task` after they time out so no need for doing that asynchronously as well.


maaktweluit (approved):
LGTM!

Nice cleanup :)

PR #4489, opened 2019-07-16
B0.20 b0.20.1

shadeofblue:


tests/golem/task/test_tasksession.py
```diff
@@ -1217,8 +1217,11 @@ def test_ctd_is_none(self, mock_cat, *_):
     def test_multi_wtct(self, *_):
         # given
         self.msg = msg_factories.tasks.WantToComputeTaskFactory(num_subtasks=3)
-        ctd = msg_factories.tasks.ComputeTaskDefFactory(resources=None)
-        self.ts.task_manager.get_next_subtask.return_value = ctd
+
+        def ctd(*args, **kwargs):
+            return msg_factories.tasks.ComputeTaskDefFactory(resources=None)
+
+        self.ts.task_manager.get_next_subtask.side_effect = ctd
```
kubkon, line 1224: Hmm, how come this was required here?

shadeofblue: @kubkon normally, CTD would be generated anew in each call to `get_next_subtask`... 

with the mock in place, the CTD is generated _once_ and then used as a return value for the mock and thus, iterations over subtasks operate on the same CTD and infuence each other... 

this change makes the CTD unique for each call again

kubkon: Hmm, I still can't see how that fixes it here. But that might be my lack of Python knowledge. So, wrapping

```python
msg_factories.tasks.ComputeTaskDefFactory(resources=None)
```
inside a separate scope (a function) causes a deep copy then?

jiivan (approved):
kawaczyherb

tests/golem/task/test_tasksession.py
shadeofblue: @kubkon nah, it's a difference between passing a function and passing a value ;) ...

kubkon: OK, my misunderstanding, looks good!

PR #4490, opened 2019-07-16
New ResourceManager class

mfranciszkiewicz:
[CLAY-384]

A `ResourceManager` class that is a simple proxy for the `simple-transfer` app.

- shares only a single file per call
- downloads only a single file per call
- uses an in-memory cache for shared files

### DIFF
--- a/golem/resource/client.py
+++ b/golem/resource/client.py
@@ -4,10 +4,10 @@ import os
 import socket
 import uuid
 from copy import deepcopy
-from twisted.internet.defer import Deferred
-from twisted.python.failure import Failure
 from types import MethodType
-from typing import Optional
+from typing import Optional, Dict, Iterable, Tuple
+from twisted.internet.defer import Deferred, succeed
+from twisted.python.failure import Failure
 
 import requests
 
@@ -33,7 +33,7 @@ class IClient(object):
     def get(self, content_hash, client_options=None, **kwargs):
         raise NotImplementedError
 
-    def id(self, client_options=None, *args, **kwargs):
+    def id(self, *args, client_options=None, **kwargs):
         raise NotImplementedError
 
 
@@ -220,19 +220,20 @@ class ClientHandler(metaclass=abc.ABCMeta):
 
 class DummyClient(IClient):
 
-    _resources = dict()
-    _paths = dict()
-    _id = "test"
+    _id = "dummy"
 
-    def add(self, files: dict, **_):
-        return self._add(files)
+    def __init__(self, *args, **kwargs):
+        del args, kwargs
+        self._resources = dict()
+        self._paths = dict()
 
-    def add_async(self, files: dict, **_):
-        deferred = Deferred()
-        deferred.callback(self._add(files))
-        return deferred
-
-    def _add(self, files: dict):
+    def add(
+            self,
+            files: Dict,
+            recursive=False,
+            **kwargs
+    ) -> str:
+        del recursive, kwargs
         from golem.core.fileshelper import common_dir
 
         resource_hash = str(uuid.uuid4())
@@ -240,16 +241,25 @@ class DummyClient(IClient):
         self._paths[resource_hash] = common_dir(files.keys())
         return resource_hash
 
-    def get(self,
+    def add_async(
+            self,
+            *args,
+            **kwargs
+    ) -> Deferred:
+        return succeed(self.add(*args, **kwargs))
+
+    def get(
+            self,
             content_hash: str,
             client_options: Optional[ClientOptions] = None,
-            filepath: Optional[str] = None,
-            **_) -> tuple:
-
+            **kwargs
+    ) -> Tuple[str, Iterable[str]]:
         from golem.core.fileshelper import copy_file_tree
 
         resource = self._resources[content_hash]
         path = self._paths[content_hash]
+        filepath = kwargs['filepath']
+
         files = [os.path.join(filepath, f) for f in resource.values()]
 
         if path != filepath:
@@ -257,21 +267,38 @@ class DummyClient(IClient):
 
         return content_hash, files
 
-    def id(self,
-           client_options: Optional[ClientOptions] = None,
-           *args, **kwargs) -> str:
+    def get_async(
+            self,
+            *args,
+            **kwargs
+    ) -> Deferred:
+        return succeed(self.get(*args, **kwargs))
 
-        return self._id
-
-    def cancel(self, content_hash: str):
+    def cancel(
+            self,
+            content_hash: str
+    ) -> str:
         return content_hash
 
-    @staticmethod
-    def cancel_async(content_hash: str):
-        deferred = Deferred()
-        deferred.callback(content_hash)
-        return deferred
+    def cancel_async(
+            self,
+            content_hash: str
+    ) -> Deferred:
+        return succeed(self.cancel(content_hash))
+
+    def id(
+            self,
+            *args,
+            client_options: Optional[ClientOptions] = None,
+            **kwargs
+    ) -> str:
+        del args, client_options, kwargs
+        return self._id
 
     @classmethod
-    def build_options(cls, **kwargs):
-        return ClientOptions(cls._id, 1)
+    def build_options(
+            cls,
+            **kwargs
+    ) -> ClientOptions:
+        del kwargs
+        return ClientOptions(cls._id, 1.)
--- a/golem/resource/client.py
+++ b/golem/resource/client.py
@@ -4,11 +4,11 @@ import os
 import socket
 import uuid
 from copy import deepcopy
-from types import MethodType
-from typing import Dict, Iterable, List, Optional, Tuple
-
-from twisted.internet.defer import Deferred, succeed
+from twisted.internet.defer import Deferred
 from twisted.python.failure import Failure
+from types import MethodType
+from typing import Optional
+
 import requests
 
 logger = logging.getLogger(__name__)
@@ -30,10 +30,10 @@ class IClient(object):
     def cancel(self, content_hash):
         raise NotImplementedError
 
-    def get(self, content_hash, filepath: str, client_options=None, **kwargs):
+    def get(self, content_hash, client_options=None, **kwargs):
         raise NotImplementedError
 
-    def id(self, *args, client_options=None, **kwargs):
+    def id(self, client_options=None, *args, **kwargs):
         raise NotImplementedError
 
 
@@ -220,24 +220,19 @@ class ClientHandler(metaclass=abc.ABCMeta):
 
 class DummyClient(IClient):
 
-    _id = "dummy"
+    _resources = dict()
+    _paths = dict()
+    _id = "test"
 
-    def __init__(
-            self,
-            *args,
-            **kwargs
-    ) -> None:
-        del args, kwargs
-        self._resources: Dict[str, List[str]] = dict()
-        self._paths: Dict[str, str] = dict()
+    def add(self, files: dict, **_):
+        return self._add(files)
 
-    def add(
-            self,
-            files: Dict[str, str],
-            recursive=False,
-            **kwargs
-    ) -> str:
-        del recursive, kwargs
+    def add_async(self, files: dict, **_):
+        deferred = Deferred()
+        deferred.callback(self._add(files))
+        return deferred
+
+    def _add(self, files: dict):
         from golem.core.fileshelper import common_dir
 
         resource_hash = str(uuid.uuid4())
@@ -245,21 +240,12 @@ class DummyClient(IClient):
         self._paths[resource_hash] = common_dir(files.keys())
         return resource_hash
 
-    def add_async(
-            self,
-            *args,
-            **kwargs
-    ) -> Deferred:
-        return succeed(self.add(*args, **kwargs))
-
-    def get(
-            self,
+    def get(self,
             content_hash: str,
-            filepath: str,
             client_options: Optional[ClientOptions] = None,
-            **kwargs
-    ) -> Tuple[str, Iterable[str]]:
-        del kwargs
+            filepath: Optional[str] = None,
+            **_) -> tuple:
+
         from golem.core.fileshelper import copy_file_tree
 
         resource = self._resources[content_hash]
@@ -271,38 +257,21 @@ class DummyClient(IClient):
 
         return content_hash, files
 
-    def get_async(
-            self,
-            *args,
-            **kwargs
-    ) -> Deferred:
-        return succeed(self.get(*args, **kwargs))
+    def id(self,
+           client_options: Optional[ClientOptions] = None,
+           *args, **kwargs) -> str:
 
-    def cancel(
-            self,
-            content_hash: str
-    ) -> str:
-        return content_hash
-
-    def cancel_async(
-            self,
-            content_hash: str
-    ) -> Deferred:
-        return succeed(self.cancel(content_hash))
-
-    def id(
-            self,
-            *args,
-            client_options: Optional[ClientOptions] = None,
-            **kwargs
-    ) -> str:
-        del args, client_options, kwargs
         return self._id
 
+    def cancel(self, content_hash: str):
+        return content_hash
+
+    @staticmethod
+    def cancel_async(content_hash: str):
+        deferred = Deferred()
+        deferred.callback(content_hash)
+        return deferred
+
     @classmethod
-    def build_options(
-            cls,
-            **kwargs
-    ) -> ClientOptions:
-        del kwargs
-        return ClientOptions(cls._id, 1.)
+    def build_options(cls, **kwargs):
+        return ClientOptions(cls._id, 1)
--- a/golem/resource/client.py
+++ b/golem/resource/client.py
@@ -5,10 +5,10 @@ import socket
 import uuid
 from copy import deepcopy
 from types import MethodType
-from typing import Optional, Dict, Iterable, Tuple
+from typing import Dict, Iterable, Optional, Tuple
+
 from twisted.internet.defer import Deferred, succeed
 from twisted.python.failure import Failure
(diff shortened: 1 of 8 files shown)


golem/resource/client.py
```diff
@@ -220,58 +220,85 @@ def _error(exc):
...
-        return deferred
-
-    def _add(self, files: dict):
+    def add(
+            self,
+            files: Dict,
+            recursive=False,
+            **kwargs
+    ) -> str:
+        del recursive, kwargs
```
Wiezzel: This `del` looks hacky. Why is it needed?

golem/resource/client.py
```diff
@@ -220,58 +220,85 @@ def _error(exc):
...
-            filepath: Optional[str] = None,
-            **_) -> tuple:
-
+            **kwargs
+    ) -> Tuple[str, Iterable[str]]:
         from golem.core.fileshelper import copy_file_tree
 
         resource = self._resources[content_hash]
         path = self._paths[content_hash]
+        filepath = kwargs['filepath']
```
Wiezzel: Maybe you could just change the signature in `IClient` to include `filepath`? Passing argument like this is obscure and kinda ugly.

golem/resource/resourcemanager.py
```diff
@@ -0,0 +1,120 @@
...
+        cached = self._cache.get(resolved_path)
+        if cached:
+            return succeed(cached)
+
+        deferred = self._client.add_async(
+            files={str(resolved_path): resolved_path.name},
+            client_options=client_options)
+        deferred.addCallback(partial(self._shared, resolved_path))
+
+        return deferred
```
Krigpl: Let's use inline callbacks for clarity

golem/resource/resourcemanager.py
```diff
@@ -0,0 +1,120 @@
...
+    def share(
+            self,
+            file_path: Path,
+            client_options: Optional[ClientOptions] = None,
+    ) -> Deferred:
+        """ Shares a single file; returns a future which resolves to
+            an assigned resource ID or a client-specific error """
+
+        resolved_path = file_path.resolve()
+        cached = self._cache.get(resolved_path)
```
Krigpl, line 43: afaik simple transfer deletes the files after some time out (3 days?)
meaning it can happen that the file will be deleted from simple transfer but you will return the invalid cached value.
maybe there should be an API call to simple transfer that asks whether given resource ID is still valid.

golem/resource/resourcemanager.py
```diff
@@ -0,0 +1,120 @@
...
+
+        resolved_path = directory.resolve(strict=False)
+
+        deferred = self._client.get_async(
+            content_hash=resource_id,
+            filepath=str(resolved_path),
+            client_options=client_options)
+        deferred.addCallback(self._downloaded)
+
+        return deferred
```
Krigpl: inline callbacks

golem/resource/resourcemanager.py
```diff
@@ -0,0 +1,120 @@
...
+            self,
+            resource_id: ResourceId,
+            directory: Path,
+            client_options: Optional[ClientOptions] = None,
+    ) -> Deferred:
+        """ Downloads a single resource to a given directory;
+            returns a future which resolves to Tuple[ResourceId, Path]
+            or a client-specific error """
+
+        resolved_path = directory.resolve(strict=False)
```
Krigpl, line 66: maybe also throw if path is a file and not a directory

Krigpl (changes requested):
Not sure about the timeout thing.

golem/resource/resourcemanager.py
```diff
@@ -0,0 +1,120 @@
...
+            host: str,
+    ) -> None:
+        self._client = self.Client(port, host)
+        self._cache: Dict[Path, ResourceId] = dict()
+        self._cache_rev: Dict[ResourceId, Path] = dict()
+
+    @classmethod
+    def build_client_options(
+            cls,
+            peers=None,
```
Wiezzel: Type of`peers`?

golem/resource/resourcemanager.py
mfranciszkiewicz: IMO this timeout should be controlled by Golem, not simple-transfer. It forgets the files after restarting anyway.

golem/resource/resourcemanager.py
mfranciszkiewicz: Or simply forward the call and fail with simple-transfer

mfranciszkiewicz: Or it may just be handled by simple-transfer, less extra management here

mfranciszkiewicz:
I'd like to hear more voices on replacing the deferred's with inlineCallbacks

Krigpl:
> I'd like to hear more voices on replacing the deferred's with inlineCallbacks

What are the cons of doing that?

mfranciszkiewicz:
@Krigpl There’s nothing wrong with implementing inlineCallbacks. It’s a matter of spending time refactoring vs doing something else.

golem/resource/resourcemanager.py
```diff
@@ -0,0 +1,120 @@
...
+        return cls.Client.build_options(
+            peers=peers, **kwargs)
+
+    def share(
+            self,
+            file_path: Path,
+            client_options: Optional[ClientOptions] = None,
+    ) -> Deferred:
+        """ Shares a single file; returns a future which resolves to
+            an assigned resource ID or a client-specific error """
```
Wiezzel: From [our styleguide](https://github.com/golemfactory/styleguide/blob/gh-pages/pyguide.md#382-functions-and-methods):

> The docstring should be imperative ("""Fetch rows from a Bigtable.""") rather than descriptive ("""Fetches rows from a Bigtable.""").

golem/resource/client.py
```diff
@@ -220,58 +220,85 @@ def _error(exc):
...
 
-    def add_async(self, files: dict, **_):
-        deferred = Deferred()
-        deferred.callback(self._add(files))
-        return deferred
-
-    def _add(self, files: dict):
+    def add(
+            self,
+            files: Dict,
```
Wiezzel: `Dict[str,str]` ?

golem/resource/client.py
mfranciszkiewicz: Oh, it’s a way of handling unused arguments from our new guidelines.

Wiezzel:
I'm also in favor of using `inlineCallbacks`. It makes code easier to understand because it mimics typical synchronous code. And you wouldn't need things like returning arguments back from `_shared()` or `_downloaded()`.

golem/resource/client.py
Wiezzel: Do think it's reasonable? Maybe we should adjust the guidelines? If not, just please put some comment (`#  unused arguments`).

Wiezzel: Actually, I thought that `pylint` won't complain at all about unused arguments if they are added for compatibility with superclass interface.

mfranciszkiewicz: In general: `del` makes sure that arguments marked as unused in a different way cannot be used at all (e.g. `_recrusive`). It's simply unusual for our codebase.

> Actually, I thought that pylint won't complain at all about unused arguments if they are added for compatibility with superclass interface.

And here unused arguments are listed explicitly

Wiezzel: Well, ok.

mfranciszkiewicz: I actually wanted to create a PR regarding unused arguments but in the end I think it does make sense. It's not about keeping `pylint` happy but enforcing the promise of not using the arguments at all.

golem/resource/resourcemanager.py
Krigpl: It's still an issue if Golem runs for more than 3 days at the moment.

golem/resource/resourcemanager.py
```diff
@@ -0,0 +1,118 @@
...
+
+        return cls.Client.build_options(
+            peers=peers, **kwargs)
+
+    @inlineCallbacks
+    def share(
+            self,
+            file_path: Path,
+            client_options: Optional[ClientOptions] = None,
+    ):
```
Krigpl, line 38: return type

golem/resource/resourcemanager.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @inlineCallbacks
+    def download(
+            self,
+            resource_id: ResourceId,
+            directory: Path,
+            client_options: Optional[ClientOptions] = None,
+    ):
+        """ Download a single resource to a given directory;
+            return a future which resolves to Tuple[ResourceId, Path]
```
Krigpl: resource_id is unchanged from the argument so no need to return it, just the filepath is fine.

golem/resource/resourcemanager.py
```diff
@@ -0,0 +1,118 @@
...
+
+        return self._parse_download_response(response)
+
+    @inlineCallbacks
+    def drop(
+            self,
+            resource_id: ResourceId,
+    ):
+        """ Stop a single resource ID from being shared;
+            return a future which resolves to a ResourceId or
```
Krigpl: not valid anymore

golem/resource/resourcemanager.py
```diff
@@ -0,0 +1,118 @@
...
+from golem.network.hyperdrive.client import HyperdriveAsyncClient
+from golem.resource.client import ClientOptions
+
+ResourceId = NewType('ResourceId', str)
+Peers = Iterable[Dict[str, Tuple[str, int]]]
+
+
+class ResourceManager:
+
+    Client = HyperdriveAsyncClient
```
Krigpl: Just make a `Client` instance a dependency in the constructor. There's no reason for this approach, and it will also make testing much easier.

Krigpl (changes requested):

golem/resource/resourcemanager.py
mfranciszkiewicz: It's a `GeneratorType` type with no subtype which I consciously omitted here. If I put the expected return type, the linter complains.

golem/resource/resourcemanager.py
```diff
@@ -26,95 +26,93 @@ def __init__(
...
+        yield self._client.cancel_async(resource_id)
+        self._dropped(resource_id)
 
     def _shared(
             self,
             path: Path,
             resource_id: ResourceId,
-    ) -> ResourceId:
-        """ Caches a resource ID and passes it down the callback chain """
+    ) -> None:
```
Wiezzel: I'd suggest renaming to `add_to_cache` or (even better) inlining this methods as it's only used once if see correctly.

golem/resource/resourcemanager.py
```diff
@@ -26,95 +26,93 @@ def __init__(
...
+        """ Cache the resource ID """
 
         self._cache[path] = resource_id
         self._cache_rev[resource_id] = path
-        return resource_id
+
+    def _dropped(
+            self,
+            resource_id: ResourceId,
+    ) -> None:
```
Wiezzel: I'd suggest renaming to `remove_from_cache` or (even better) inlining this methods as it's only used once if see correctly.

golem/resource/resourcemanager.py
mfranciszkiewicz: > There's no reason for this approach

There's also no reason for providing the client instance in the constructor since there is actually only one `Client` type used in Golem.

> it will also make testing much easier

One line was required to change the tested `Client` class.

golem/resource/resourcemanager.py
mfranciszkiewicz: And will patching `simple-transfer` solve the discussed issue?

golem/resource/client.py
```diff
@@ -4,10 +4,10 @@
 import socket
 import uuid
 from copy import deepcopy
-from twisted.internet.defer import Deferred
-from twisted.python.failure import Failure
 from types import MethodType
-from typing import Optional
+from typing import Optional, Dict, Iterable, Tuple
```
maaktweluit: it used to be alphabetically sorted, but makes more sense to do module imports after the system ones

maybe move the whitespace from line 11 up as well to group `twisted` and `requests`

golem/resource/client.py
```diff
@@ -220,32 +220,42 @@ def _error(exc):
...
-        deferred.callback(self._add(files))
-        return deferred
-
-    def _add(self, files: dict):
+    def add(
+            self,
+            files: Dict[str, str],
+            recursive=False,
+            **kwargs
+    ) -> str:
```
maaktweluit: this seems to fit on one line

maaktweluit (commented):
LGTM!

Some nitty comments, will approve on reply

I see a few more functions where the signature fits on one line, can mark them all if you want :)

golem/resource/client.py
mfranciszkiewicz: It's for the sake of consistency. I actually should have updated `__init__` as well.

mfranciszkiewicz: But I can make them fit in one line if you will.

maaktweluit: haha, was looking for consistency and then saw the `__init__`..

I do not mind, both are OK with me :)

Wiezzel: Actually, our guidelines encourage one parameter per line.

golem/resource/resourcemanager.py
Krigpl: Please familiarize yourself with the benefits of dependency injection pattern.
I wouldn't consider mocking private methods in tests as a good approach when a simpler and cleaner alternative exists.
By injecting the Client you'll be able to mock the instance appropriately and avoid code like this: 
``` 
    with mock.patch.object(
        self.resource_manager._client,
        'add_async',
        wraps=self.resource_manager._client.add_async
    ) as add_async:
```

> There's also no reason for providing the client instance in the constructor since there is actually only one Client type used in Golem.

This is the case now, there's no reason to believe it'll always be the case and then it'll have to be refactored.

mfranciszkiewicz: > Please familiarize yourself with the benefits of dependency injection pattern.

This assumption here was truly not needed.

> I wouldn't consider mocking private methods in tests as a good approach when a simpler and cleaner alternative exists.

It wouldn't be `DummyClient` that I would replace the `Client` with, just some other implementation. This code has nothing to do with passing a client instance as an argument. I simply didn't create a new one.

> This is the case now, there's no reason to believe it'll always be the case and then it'll have to be refactored.

And yet somehow it is always discouraged in the reviews.

Krigpl: In the `TestResourceManager` we wanna test the functionality of the `ResourceManager` class that it makes expected calls to the some `Client` instance and we're not interested how the `Client ` instance is in fact implemented.

Thus if we simply change the constructor to
```
    def __init__(
            self,
            client: HyperdriveAsyncClient,
    ) -> None:
        self._client = client
        self._cache_rev: Dict[ResourceId, Path] = dict()
```
we sacrifice absolutely nothing but we gain:
- ability to write clean unit tests. We can just pass a Mock object here and assert in the tests that expected methods have been called in different scenarios - exactly what we want to test.
- flexibility to have two ResourceManager instances with different clients in ever needed in the future for free
- no need for unnecessary DummyClient and DummyResourceManager

golem/resource/resourcemanager.py
Krigpl: Yeah, making the `timeout` a param of the `upload` command would be the best solution imo.

tests/golem/resource/test_resourcemanager.py
```diff
@@ -0,0 +1,62 @@
...
+        sample_dir.mkdir(parents=True)
+
+        self.resource_manager.download(resource_id, sample_dir)
+        assert self.client.get_async.called
+
+    def test_drop(self):
+        sample_path = self.new_path / "sample.txt"
+
+        resource_id = self.resource_manager.share(sample_path).result
+        assert sample_path in self.resource_manager._cache
```
Krigpl, line 59: The way to test `drop` without accessing protected members would be to call `share`, then `drop`, then `share` again and assert that the second `share` calls a method on the client.

tests/golem/resource/test_resourcemanager.py
```diff
@@ -0,0 +1,62 @@
...
+        assert isinstance(
+            self.resource_manager.build_client_options(),
+            ClientOptions)
+
+    def test_share(self):
+        sample_path = self.new_path / "sample.txt"
+
+        result = self.resource_manager.share(sample_path).result
+        assert isinstance(result, str)
+        assert self.resource_manager._cache[sample_path] == result
```
Krigpl, line 35: `test_share_cached` asserts exactly that but without accessing the protected member so I'd say this test is redundant.

Krigpl (commented):
Thanks, this looks super clear and easy now imo.

golem/resource/resourcemanager.py
mfranciszkiewicz: `timeout` is a part of `ClientOptions`, the problematic part is the enforced sweep, which also breaks the current implementation.

tests/golem/resource/test_resourcemanager.py
mfranciszkiewicz: That's true if we really do not want to access protected members in unit tests. But it hides the intent behind the test and this is more explicit.

golem/resource/resourcemanager.py
Krigpl: Oh okay, I didn't realize there are two kinds of timeouts here.

Krigpl (approved):

Wiezzel (approved):
👍

Krigpl:
fwiw I'm fine with merging it as is and patching simple-transfer later if this is what's blocking this PR.

mfranciszkiewicz:
`simple-transfer` has been updated but not released. I'll introduce the required changes in a separate PR.

PR #4496, opened 2019-07-17
Build output path from input parameters

nieznanysprawiciel:
Golem built output path from task name, which limited filename to 26 signs and was problematic for using in demo.


magdasta (commented):
Looks good, have you tested it?

bidzyyys (approved):

nieznanysprawiciel:
I tested this only with WebService backend.

magdasta (approved):

PR #4501, opened 2019-07-18
Implement new env performance before sending WTCT

maaktweluit:
Compute the new app benchmark if needed. Send the new app benchmark score in WTCT so that it can be used in the new subtask creation flow.

Steps taken:
- Added new env as return value to `TaskServer.get_environment_by_id()`
- Added (new) `EnvironmentManager.get_performance()`
  - Stores performance scores in the same table as old EnvironmentManager
  - returns Deferred -> float if benchmark is cached or started, None when benchmark is running
- Added performance to `CompTaskInfo` so it does not have to be calculated after the task is completed 
- Made a task on the clay board for the minimum env performance

Steps left:
- [x] TODO: Decide on `deferToThread` or `sync_wait` for running the benchmark
- [x] Run unit and integration tests for old code
- [x] Add unit tests for new code

### DIFF
--- a/golem/task/envmanager.py
+++ b/golem/task/envmanager.py
@@ -76,9 +76,12 @@ class EnvironmentManager:
         """ Gets the performance for the given environment
             Checks the database first, if not found it starts a benchmark
             Return value Deferred resulting in a float
-            or None when the benchmark is already running"""
+            or None when the benchmark is already running. """
         if self._running_benchmark:
             return None
+
+        if not self.enabled(env_id):
+            raise Exception("Requested performance for disabled environment")
 
         perf = None
         try:
@@ -87,13 +90,17 @@ class EnvironmentManager:
         except Performance.DoesNotExist:
             pass
 
-        env = self._envs[env_id]
+        env = self._envs[env_id].instance
         self._running_benchmark = True
 
         try:
             result = yield env.run_benchmark()
         except Exception:
-            logger.error('failed to run benchmark. env=%r', env_id)
+            logger.error(
+                'failed to run benchmark. env=%r',
+                env_id,
+                exc_info=True
+            )
             raise
         finally:
             self._running_benchmark = False
--- a/golem/task/envmanager.py
+++ b/golem/task/envmanager.py
@@ -83,7 +83,6 @@ class EnvironmentManager:
         if not self.enabled(env_id):
             raise Exception("Requested performance for disabled environment")
 
-        perf = None
         try:
             perf = Performance.get(Performance.environment_id == env_id)
             return perf.value
--- a/golem/task/envmanager.py
+++ b/golem/task/envmanager.py
@@ -1,7 +1,12 @@
-from typing import Dict, List, NamedTuple, Type
+import logging
+from typing import Dict, List, NamedTuple, Type, Union
 
 from golem.envs import EnvId, Environment
+from golem.model import Performance
 from golem.task.task_api import TaskApiPayloadBuilder
+from twisted.internet.defer import Deferred
+
+logger = logging.getLogger(__name__)
 
 
 class EnvEntry(NamedTuple):
@@ -41,6 +46,8 @@ class EnvironmentManager:
 
     def enabled(self, env_id: EnvId) -> bool:
         """ Get the state (enabled or not) for an Environment. """
+        if env_id not in self._state:
+            return False
         return self._state[env_id]
 
     def set_enabled(self, env_id: EnvId, enabled: bool) -> None:
@@ -60,3 +67,25 @@ class EnvironmentManager:
 
     def payload_builder(self, env_id: EnvId) -> Type[TaskApiPayloadBuilder]:
         return self._envs[env_id].payload_builder
+
+    def get_performance(self, env_id) -> Union[Deferred, float]:
+        perf = None
+        try:
+            perf = Performance.get(Performance.environment_id == env_id)
+        except Performance.DoesNotExist:
+            pass
+
+        if perf is None or perf.value is None:
+            def _save_performance(raw_perf):
+                Performance.update_or_create(env_id, raw_perf)
+
+            def _benchmark_error(_e):
+                logger.error('failed to run benchmark. env=%r', env_id)
+
+            env = self._envs[env_id]
+            deferred = env.run_benchmark()
+            deferred.addCallback(_save_performance)
+            deferred.addErrback(_benchmark_error)
+            return deferred
+
+        return perf.value
--- a/golem/task/envmanager.py
+++ b/golem/task/envmanager.py
@@ -1,10 +1,11 @@
 import logging
-from typing import Dict, List, NamedTuple, Type, Union
+from typing import Dict, List, NamedTuple, Type
+
+from twisted.internet.defer import Deferred, inlineCallbacks
 
 from golem.envs import EnvId, Environment
 from golem.model import Performance
 from golem.task.task_api import TaskApiPayloadBuilder
-from twisted.internet.defer import Deferred
 
 logger = logging.getLogger(__name__)
 
@@ -20,6 +21,7 @@ class EnvironmentManager:
     def __init__(self):
         self._envs: Dict[EnvId, EnvEntry] = {}
         self._state: Dict[EnvId, bool] = {}
+        self._running_benchmark: bool = False
 
     def register_env(
             self,
@@ -45,7 +47,8 @@ class EnvironmentManager:
             self.set_enabled(env_id, enabled)
 
     def enabled(self, env_id: EnvId) -> bool:
-        """ Get the state (enabled or not) for an Environment. """
+        """ Get the state (enabled or not) for an Environment.
+            Also returns false when the environment is not registered"""
         if env_id not in self._state:
             return False
         return self._state[env_id]
@@ -68,7 +71,12 @@ class EnvironmentManager:
     def payload_builder(self, env_id: EnvId) -> Type[TaskApiPayloadBuilder]:
         return self._envs[env_id].payload_builder
 
-    def get_performance(self, env_id) -> Union[Deferred, float]:
+    @inlineCallbacks
+    def get_performance(self, env_id) -> Deferred:
+        """ Gets the performance for the given environment
+            Checks the database first, if not found it starts a benchmark
+            Return value Deferred resulting in a float
+            or None when the benchmark is already running"""
         perf = None
         try:
             perf = Performance.get(Performance.environment_id == env_id)
@@ -76,16 +84,26 @@ class EnvironmentManager:
             pass
 
         if perf is None or perf.value is None:
-            def _save_performance(raw_perf):
-                Performance.update_or_create(env_id, raw_perf)
-
-            def _benchmark_error(_e):
-                logger.error('failed to run benchmark. env=%r', env_id)
+            if self._running_benchmark:
+                return None
 
             env = self._envs[env_id]
-            deferred = env.run_benchmark()
-            deferred.addCallback(_save_performance)
-            deferred.addErrback(_benchmark_error)
-            return deferred
+            self._running_benchmark = True
+
+            try:
+                result = yield env.run_benchmark()
+            except Exception:
+                logger.error('failed to run benchmark. env=%r', env_id)
+                raise
+            finally:
+                self._running_benchmark = False
+
+            Performance.update_or_create(env_id, result)
+            logger.info(
+                'finshed running benchmark. env=%r, score=%r',
+                env_id,
+                result
+            )
+            return result
 
         return perf.value
--- a/golem/task/envmanager.py
+++ b/golem/task/envmanager.py
@@ -77,33 +77,31 @@ class EnvironmentManager:
             Checks the database first, if not found it starts a benchmark
             Return value Deferred resulting in a float
             or None when the benchmark is already running"""
+        if self._running_benchmark:
+            return None
+
         perf = None
         try:
             perf = Performance.get(Performance.environment_id == env_id)
+            return perf.value
         except Performance.DoesNotExist:
             pass
 
-        if perf is None or perf.value is None:
-            if self._running_benchmark:
-                return None
+        env = self._envs[env_id]
+        self._running_benchmark = True
 
-            env = self._envs[env_id]
-            self._running_benchmark = True
+        try:
+            result = yield env.run_benchmark()
+        except Exception:
+            logger.error('failed to run benchmark. env=%r', env_id)
+            raise
+        finally:
+            self._running_benchmark = False
 
-            try:
-                result = yield env.run_benchmark()
-            except Exception:
-                logger.error('failed to run benchmark. env=%r', env_id)
-                raise
-            finally:
-                self._running_benchmark = False
-
-            Performance.update_or_create(env_id, result)
-            logger.info(
-                'finshed running benchmark. env=%r, score=%r',
-                env_id,
-                result
-            )
-            return result
-
-        return perf.value
+        Performance.update_or_create(env_id, result)
+        logger.info(
+            'Finshed running benchmark. env=%r, score=%r',
+            env_id,
+            result
+        )
+        return result
--- a/golem/task/taskserver.py
+++ b/golem/task/taskserver.py
@@ -309,7 +309,16 @@ class TaskServer(
 
         self._last_task_request_time = time.time()
         self.task_computer.stats.increase_stat('tasks_requested')
-        self._request_task(task_header)
+
+        def _request_task_error(e):
+            logger.error(
+                "Failed to request task: task_id=%r, exception=%r",
+                task_header.task_id,
+                e
+            )
+        # Unyielded deferred, fire and forget requesting a new task
+        deferred = self._request_task(task_header)
+        deferred.addErrback(_request_task_error)  # pylint: disable=no-member
 
     @inlineCallbacks
     def _request_task(self, theader: dt_tasks.TaskHeader) -> Deferred:
--- a/golem/task/taskserver.py
+++ b/golem/task/taskserver.py
@@ -39,7 +39,7 @@ from golem.envs import Environment as NewEnv
 from golem.envs.docker.cpu import DockerCPUConfig
 from golem.envs.docker.non_hypervised import NonHypervisedDockerCPUEnvironment
 from golem.marketplace import OfferPool
-from golem.model import Performance, TaskPayment
+from golem.model import TaskPayment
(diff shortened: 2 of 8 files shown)


golem/task/envmanager.py
```diff
@@ -41,6 +46,8 @@ def set_state(self, state: Dict[EnvId, bool]) -> None:
 
     def enabled(self, env_id: EnvId) -> bool:
         """ Get the state (enabled or not) for an Environment. """
+        if env_id not in self._state:
```
maaktweluit, line 52: @Wiezzel this was to re-use `.enabled()` function as `has_environment()` an alternative is to make it a separate function, what do you think?

golem/task/envmanager.py
```diff
@@ -60,3 +67,25 @@ def environment(self, env_id: EnvId) -> Environment:
...
+
+        if perf is None or perf.value is None:
+            def _save_performance(raw_perf):
+                Performance.update_or_create(env_id, raw_perf)
+
+            def _benchmark_error(_e):
+                logger.error('failed to run benchmark. env=%r', env_id)
+
+            env = self._envs[env_id]
+            deferred = env.run_benchmark()
```
maaktweluit: TODO: Test this deferred chain

golem/task/envmanager.py
```diff
@@ -60,3 +67,25 @@ def environment(self, env_id: EnvId) -> Environment:
 
     def payload_builder(self, env_id: EnvId) -> Type[TaskApiPayloadBuilder]:
         return self._envs[env_id].payload_builder
+
+    def get_performance(self, env_id) -> Union[Deferred, float]:
```
Wiezzel: Really don't like the `Union` thing here. I think it should always return a `Deferred`. Also , using `@inlineCallbacks` would make the code nicer.
I think @Krigpl would agree.

golem/task/envmanager.py
Wiezzel: Looks fine.

golem/task/envmanager.py
maaktweluit: For the return value: 
I agree if we go for the `deferToThread` route, with a `sync_wait` we have to know if the value is already there or needs 1 minute to complete.

For the `@inlineCallbacks`:
You can not return a `Deferred` using inline callbacks, what part should be using inlineCallbacks as you propose?

Wiezzel: I meant something like this:

```python
    @inlineCallbacks
    def get_performance(self, env_id) -> Deferred:
        try:
            perf = Performance.get(Performance.environment_id == env_id)
            if perf.value is not None:
                return perf.value
        except Performance.DoesNotExist:
            pass

        env = self._envs[env_id]
        try:
            raw_perf = yield env.run_benchmark()
        except:
            logger.exception('failed to run benchmark. env=%r', env_id)
            return None  # Or re-raise maybe?
            
        Performance.update_or_create(env_id, raw_perf)
        return raw_perf
```

maaktweluit: Ok, thanks!

will update it like this after we decide on `sync_wait` vs `deferToThread`, becuase currently the `yield` is out of this function for the `deferToThread` approach

maaktweluit: Updated

golem/task/envmanager.py
```diff
@@ -60,3 +70,40 @@ def environment(self, env_id: EnvId) -> Environment:
...
+            Checks the database first, if not found it starts a benchmark
+            Return value Deferred resulting in a float
+            or None when the benchmark is already running"""
+        perf = None
+        try:
+            perf = Performance.get(Performance.environment_id == env_id)
+        except Performance.DoesNotExist:
+            pass
+
+        if perf is None or perf.value is None:
```
Krigpl: Why would `perf.value` be `None`? Let's assume an invariant that if an entry exists in the DB it is not `None`. Then you can just return it above.

golem/task/envmanager.py
```diff
@@ -60,3 +70,40 @@ def environment(self, env_id: EnvId) -> Environment:
...
+            if self._running_benchmark:
+                return None
```
Krigpl: Why would that return `None` in that case? Just make it return the same deferred instance as the previous call thus making the caller agnostic to whether a benchmark is already running or not.

iow make `self._running_benchmark` hold a deferred instance and just yield it here. Assuming it's for the same env type ofc.

golem/task/envmanager.py
maaktweluit: In the case where the value was cleared. This depends on if the row will be deleted or set to `None`.

When i think about it deleting would be cleaner here, will update it.

golem/task/envmanager.py
maaktweluit: This is to not send multiple WTCT messages

The caller is fire and forget, so it needs to return None when already running and wait for benchmark + send WTCT only for the first time triggered

Krigpl: In lines 345-348 you wait for the benchmark result for the old env but not wait for the new one, why is that?

maaktweluit: that is a booboo on my side, fixed it for my next commit, thanks!

golem/task/taskserver.py
```diff
@@ -324,6 +339,16 @@ def _request_task(self, theader: dt_tasks.TaskHeader) -> Optional[str]:
...
 
+            # Check performance
+            performance = None
+            if isinstance(env, OldEnv):
+                performance = env.get_performance()
+            else:  # NewEnv
+                env_mgr = self.task_keeper.new_env_manager
+                performance = yield env_mgr.get_performance(env_id)
+            if performance is None:
+                return None
```
mfranciszkiewicz, line 374: Could you add a log explaining that we're not requesting a task because the benchmark is running?

golem/task/taskserver.py
```diff
@@ -373,7 +398,7 @@ def _request_task(self, theader: dt_tasks.TaskHeader) -> Optional[str]:
             return theader.task_id
         except Exception as err:  # pylint: disable=broad-except
             logger.warning("Cannot send request for task: %s", err)
-            logger.debug("Detailed traceback", exc_info=True)
+            logger.info("Detailed traceback", exc_info=True)
```
mfranciszkiewicz: Despite `info`, tracebacks are some scary beasts that look really serious

golem/task/taskserver.py
```diff
@@ -771,7 +796,11 @@ def add_forwarded_session_request(self, key_id, conn_id):
 
     def get_min_performance_for_task(self, task: Task) -> float:
         env = self.get_environment_by_id(task.header.environment)
-        return env.get_min_accepted_performance()
+        if isinstance(env, OldEnv):
+            return env.get_min_accepted_performance()
+        # NewEnv
+        # TODO: Implement minimum performance in new env
```
mfranciszkiewicz, line 831: Maybe `config_desc` instead? I feel like min performance is a setting that does not belong to a computation environment.

golem/task/envmanager.py
```diff
@@ -60,3 +70,42 @@ def environment(self, env_id: EnvId) -> Environment:
 
     def payload_builder(self, env_id: EnvId) -> Type[TaskApiPayloadBuilder]:
         return self._envs[env_id].payload_builder
+
+    @inlineCallbacks
+    def get_performance(self, env_id) -> Deferred:
```
mfranciszkiewicz, line 75: IMO this function should prevent running benchmarks if the environment is disabled (as this class manages the state of environments)

golem/task/envmanager.py
```diff
@@ -15,6 +21,7 @@ class EnvironmentManager:
     def __init__(self):
         self._envs: Dict[EnvId, EnvEntry] = {}
         self._state: Dict[EnvId, bool] = {}
+        self._running_benchmark: bool = False
```
mfranciszkiewicz, line 24: One thing to consider for the future: we may want to be able to execute computations in multiple environments simultaneously (e.g. GPU and gWASM). This flag will prevent requesting tasks for both if any benchmark is running. We may want to discuss a different approach to this issue.

golem/task/taskserver.py
maaktweluit: good one, fixed

golem/task/taskserver.py
maaktweluit: whoops, this was to debug a test, for some reason the `debug` level did not show..

reverted!

golem/task/taskserver.py
maaktweluit: I was thinking this setting belongs to the app, not the env. Already made an issue on the board to fix this, do you think it is better to use `config_desc` here instead of `0.0`?

golem/task/envmanager.py
maaktweluit: good one, added

golem/task/envmanager.py
maaktweluit: I was thinking this could become part of the new taskcomputer. That should be the component responsible for knowing what env can run what computation at what time

Will make an issue on the board

golem/task/envmanager.py
```diff
@@ -60,3 +70,45 @@ def environment(self, env_id: EnvId) -> Environment:
...
+            Checks the database first, if not found it starts a benchmark
+            Return value Deferred resulting in a float
+            or None when the benchmark is already running. """
+        if self._running_benchmark:
+            return None
+
+        if not self.enabled(env_id):
+            raise Exception("Requested performance for disabled environment")
+
+        perf = None
```
Krigpl: This variable is unnecessary.

tests/golem/task/test_envmanager.py
```diff
@@ -75,8 +83,77 @@ def test_set_state(self):
...
+        EnvManagerBaseTest,
+        DatabaseFixture,
+        TwistedTestCase
+):
+    @inlineCallbacks
+    def test_get_performance_running(self):
+        # Given
+        env_id = "env1"
+        self.register_env(env_id)
+        env = self.manager._envs[env_id]
```
Krigpl: `register_env` returns the env instance so you don't have to do that

tests/golem/task/test_envmanager.py
```diff
@@ -75,8 +83,77 @@ def test_set_state(self):
...
+        self.manager._running_benchmark = True
+
+        # When
+        result = yield self.manager.get_performance(env_id)
+
+        # Then
+        env.instance.run_benchmark.assert_not_called()
+        self.assertTrue(result is None)
```
Krigpl: There's `assertIsNone`

tests/golem/task/test_envmanager.py
```diff
@@ -75,8 +83,77 @@ def test_set_state(self):
...
+    def test_get_performance_disabled_env(self):
+        # Given
+        env_id = "env1"
+        self.register_env(env_id)
+        env = self.manager._envs[env_id]
+        self.manager.set_enabled(env_id, False)
+
+        # When
+
+        with self.assertRaises(Exception):
```
Krigpl: Always prefer `assertRaisesRegex` and provide at least partial exception message, otherwise a simple syntax error will also make that check pass.

tests/golem/task/test_envmanager.py
```diff
@@ -75,8 +83,77 @@ def test_set_state(self):
...
+        # Given
+        env_id = "env1"
+        self.register_env(env_id)
+        env = self.manager._envs[env_id]
+        self.manager.set_enabled(env_id, True)
+
+        Performance.update_or_create(env_id, 300.0)
+
+        # When
+        yield self.manager.get_performance(env_id)
```
Krigpl: Imo you should assert that the returned value is the value from the DB.

golem/task/taskserver.py
```diff
@@ -276,36 +289,31 @@ def request_task_by_id(self, task_id: str) -> None:
...
 
         task_header = self.task_keeper.get_task(self.requested_tasks)
         if task_header is None:
-            return None
+            return
 
         self._last_task_request_time = time.time()
         self.task_computer.stats.increase_stat('tasks_requested')
-        return self._request_task(task_header)
+        self._request_task(task_header)
```
Krigpl: Always put a comment when unyielded deferred is the expected behavior. Also provide an error handler.

golem/task/envmanager.py
maaktweluit: good one, removed :)

tests/golem/task/test_envmanager.py
maaktweluit: i tried that at first and ran into errors, will test it again and leave a comment if there is a reason for it

tests/golem/task/test_envmanager.py
maaktweluit: Tried `assertNone` and it didnt work.. :(
TIL, thanks!

tests/golem/task/test_envmanager.py
maaktweluit: No errors this time, was probably related to something else..

fixed, thanks!

tests/golem/task/test_envmanager.py
maaktweluit: good one, updated!

tests/golem/task/test_envmanager.py
maaktweluit: added

golem/task/taskserver.py
maaktweluit: done

maaktweluit:
@Krigpl thanks for the review!

fixed all the comments and rebased to the latest develop, please check again

golem/task/taskserver.py
```diff
@@ -276,36 +289,40 @@ def request_task_by_id(self, task_id: str) -> None:
             return
         self._request_task(task_header)
 
-    def _request_random_task(self) -> Optional[str]:
+    def _request_random_task(self):
```
Wiezzel:
```suggestion
    def _request_random_task(self) -> None:
```

tests/golem/task/test_envmanager.py
```diff
@@ -75,8 +83,78 @@ def test_set_state(self):
...
+        env, _ = self.register_env(env_id)
+        env.get_benchmark = MagicMock(side_effect=Exception)
+
+        self.manager.set_enabled(env_id, True)
+
+        # When
+        yield self.manager.get_performance(env_id)
+
+        # Then
+        env.run_benchmark.assert_called_once()
```
Krigpl: That's not what you wanna assert in this test - you wanna assert that the exception is propagated (returned deferred holds an exception).

tests/golem/task/test_envmanager.py
```diff
@@ -75,8 +83,78 @@ def test_set_state(self):
...
+        # Then
+        env.run_benchmark.assert_not_called()
+        self.assertEqual(result, perf)
+
+    @inlineCallbacks
+    def test_get_performance_benchmark_error(self):
+        # Given
+        env_id = "env1"
+        env, _ = self.register_env(env_id)
+        env.get_benchmark = MagicMock(side_effect=Exception)
```
Krigpl: An alternative way is `env.get_benchmark.return_value = Exception`

tests/golem/task/test_envmanager.py
```diff
@@ -75,8 +83,78 @@ def test_set_state(self):
...
+class TestEnvironmentManagerDB(  # pylint: disable=too-many-ancestors
+        EnvManagerBaseTest,
+        DatabaseFixture,
+        TwistedTestCase
+):
+    @inlineCallbacks
+    def test_get_performance_running(self):
+        # Given
+        env_id = "env1"
+        env, _ = self.register_env(env_id)
```
Krigpl: These two lines are common for all tests, you can move them to `setUp`

golem/task/taskserver.py
```diff
@@ -335,6 +362,17 @@ def _request_task(self, theader: dt_tasks.TaskHeader) -> Optional[str]:
...
                 return None
 
+            # Check performance
+            performance = None
+            if isinstance(env, OldEnv):
+                performance = env.get_performance()
+            else:  # NewEnv
+                env_mgr = self.task_keeper.new_env_manager
+                performance = yield env_mgr.get_performance(env_id)
+            if performance is None:
```
Wiezzel, line 372: Would this branch ever fire in a real case scenario? As far as I can see `EnvManager.get_performance` is only called above and we're waiting for the result. So to get a `None` this method (`_request_task`) must have been called simultaneously from two threads.

golem/task/taskserver.py
```diff
@@ -335,6 +362,17 @@ def _request_task(self, theader: dt_tasks.TaskHeader) -> Optional[str]:
                     )
                 return None
 
+            # Check performance
+            performance = None
+            if isinstance(env, OldEnv):
```
Wiezzel, line 367: To have a method returning `Union` and then do a `isinstance` check looks like an anti-pattern to me. Maybe better add some simple `get_performance()` method?

tests/golem/task/test_envmanager.py
maaktweluit: Good catch, updated it so it checks if the raised exception is passed

tests/golem/task/test_envmanager.py
maaktweluit: is there any added value to using return_value over side_effect?
i like side_effect for exceptions since its not really returning but raising

tests/golem/task/test_envmanager.py
maaktweluit: done :)

golem/task/taskserver.py
maaktweluit: Yes, that is the real scenario, to avoid duplicate benchmarks and callbacks:

- `_request_task()` is a un-yielded deferred, fire and forget
- `sync_network` is called every second
- `task_request_interval` is 5 seconds
- benchmark timeout is 60 seconds

Based on these numbers i think its likely that the `_request_task()` is called multiple times on slow computers ( benchmark takes more then 5 seconds )

golem/task/taskserver.py
maaktweluit: It makes it very easy to remove the old stuff, just remove the union and the OldEnv checks.

Was thinking of adding a separate `_get_performance()` method but it is only used in one place.

I agree with the comment but still think this is the better of the evils, open to suggestions :)

golem/task/taskserver.py
maaktweluit: Nice one, fixed :)

golem/task/taskserver.py
Wiezzel: OK, thanks for clarifications.

Wiezzel (approved):

Krigpl (approved):

mfranciszkiewicz (approved):

PR #4510, opened 2019-07-18
New TaskComputer

Wiezzel:
Created NewTaskComputer class for computing tasks created with the new Task API. Support for old-style tasks is kept as well. TaskComputerAdapter was introduced to dispatch tasks between new and old
task computer.


Wiezzel:
@Krigpl @mfranciszkiewicz Unit tests for the NewTaskComputer are yet to come but you can already start the review.

golem/task/taskserver.py
```diff
@@ -419,7 +419,7 @@ def resource_failure(self, task_id: str, reason: str) -> None:
 
         self.task_computer.task_interrupted()
         self.send_task_failed(
-            self.task_computer.assigned_subtask['subtask_id'],
+            self.task_computer.assigned_subtask_id,
```
Krigpl: This is suspicious. It either means that `assigned_subtask_id` doesn't get cleared after `task_interrupted` or it is cleared and it's empty here.

golem/task/taskcomputer.py
```diff
@@ -42,28 +48,404 @@ def __init__(self):
...
+        return self._new_computer.assigned_subtask_id \
+            or self._old_computer.assigned_subtask_id
+
+    def start_computation(self) -> None:
+        if self._new_computer.has_assigned_task():
+            # Fire and forget because compute resolves when computation ends
+            task_id = self.assigned_task_id
+            subtask_id = self.assigned_subtask_id
+            computation = self._new_computer.compute()
+            self._handle_computation_results(task_id, subtask_id, computation)
```
Krigpl, line 127: That returns deferred but it's not yielded anywhere - expected? If so, please put a comment.

golem/task/taskcomputer.py
```diff
@@ -42,28 +48,404 @@ def __init__(self):
...
+    def get_progress() -> Optional[ComputingSubtaskStateSnapshot]:
+        return None  # Not supported because it's app-specific
+
+    def get_current_computing_env(self) -> Optional[EnvId]:
+        if self._assigned_task is None:
+            return None
+        return self._assigned_task.env_id
+
+    @defer.inlineCallbacks
+    def change_config(
```
Krigpl, line 404: So what's the decision with `change_config`? I believe it doesn't necessarily belong here but it's too awkward to remove just now?

golem/task/taskcomputer.py
Wiezzel: That's on purpose. The `computation` deferred resolves once the computation is *complete* not when it's started. And this method is supposed to just start the computation. I remember adding a comment about it but must've deleted it by accident.

golem/task/taskcomputer.py
Wiezzel: There's no final decision yet. @mfranciszkiewicz is researching this topic. This is most probably a temporary solution. And it's compatible with existing code.

golem/task/taskserver.py
Wiezzel: Hmm... This is actually a race condition. In most cases `assigned_subtask_id` won't be cleaned *yet*. 😛 
Good catch!

golem/task/taskcomputer.py
```diff
@@ -42,28 +48,404 @@ def __init__(self):
...
+        assigned_task = self._assigned_task
+        assert assigned_task is not None
+
+        task_api_service = self._get_task_api_service(
+            task_id=assigned_task.task_id,
+            env_id=assigned_task.env_id,
+            prereq_dict=assigned_task.prereq_dict
+        )
+        # FIXME: Remove when ProviderAppClient implements shutdown
+        # pylint: disable=protected-access
```
maaktweluit: when a disable is on its own line you need to re-enable it.

You can also add it to the end of the line with `  # noqa pylint: disable=protected-access`

requirements_to-freeze.txt
```diff
@@ -60,4 +60,4 @@ ujson==1.35
 urllib3==1.24.3
 web3==4.2.1
 zope.interface==4.4.2
-zxcvbn-python
\ No newline at end of file
+zxcvbn-python
```
maaktweluit, line 63: Also add the version you tested with to the `requirements.txt`

tests/golem/task/test_taskcomputeradapter.py
```diff
@@ -0,0 +1,371 @@
...
+        self.env_manager = mock.Mock(spec_set=EnvironmentManager)
+        self.adapter = TaskComputerAdapter(
+            task_server=self.task_server,
+            env_manager=self.env_manager
+        )
+
+
+class TestInit(TaskComputerAdapterTestBase):
+
+    def test(self):
```
maaktweluit: Not sure if this is required by pytest but i think its neat to have a name for the test
```suggestion
    def test_init(self):
```

maaktweluit (approved):
LGTM!

Some small nitty comments, still approving

golem/task/taskserver.py
Krigpl: I still think it's better to get this fixed - like by grabbing the subtask_id before calling `task_interrupted` and using that value here.

golem/task/taskcomputer.py
Krigpl: It still needs an error handler, right? twisted complains when there are unyielded deferreds with unhandled errors.

golem/task/taskserver.py
Wiezzel: Yea, for sure.

requirements_to-freeze.txt
Wiezzel: I don't know what happened here because I didn't touch it. Probably some merge glitch.

golem/task/taskcomputer.py
Wiezzel: `_handle_computation_results()` has a catch-all block so it shouldn't raise any error. I can put the `_send_results()` inside this block as well.

Wiezzel: Wait, there *is* a comment but somehow it got three lines up.

requirements_to-freeze.txt
maaktweluit: ok, strange.. then please revert this line :)

Wiezzel: I've just added a missing newline here.  I think there's no need to revert.

golem/task/taskcomputer.py
mfranciszkiewicz: Let's keep the method in this PR

requirements_to-freeze.txt
maaktweluit: ahh, i see.. 🤦‍♂

tests/golem/task/test_newtaskcomputer.py
```diff
@@ -0,0 +1,331 @@
...
+
+    @property
+    def task_deadline(self):
+        return int(time.time()) + 3600
+
+    @property
+    def subtask_deadline(self):
+        return int(time.time()) + 3600
+
+    def get_task_header(self, **kwargs):
```
Krigpl: I guess you could just use factories from GM, here and for ctd, and remove these helpers.

tests/golem/task/test_newtaskcomputer.py
```diff
@@ -0,0 +1,331 @@
...
+
+    def get_compute_task_def(self, **kwargs):
+        return ComputeTaskDef(
+            subtask_id=kwargs.get('subtask_id') or self.subtask_id,
+            extra_data=kwargs.get('subtask_params') or self.subtask_params,
+            performance=kwargs.get('performance') or self.performance,
+            deadline=kwargs.get('subtask_deadline') or self.subtask_deadline
+        )
+
+    def patch_async(self, name, *args, **kwargs):
```
Krigpl: What is `async` supposed to mean here?

Wiezzel: It supposed to mean that these `patch` method is suitable for asynchronous functions unlike the ordinary `@patch` decorator.

tests/golem/task/test_newtaskcomputer.py
Wiezzel: Will take a look at these factories.

tests/golem/task/test_newtaskcomputer.py
Krigpl: afaict it's only used with non-async functions here so I got confused, that's why I asked.

tests/golem/task/test_newtaskcomputer.py
Wiezzel: Well, after a look I don't see much value in using these factories because they generate some random ID's and I want to use particular ones so that I can easily make asserts without passing them around.

Krigpl: You can set the values of the fields you're interested in. Like `TaskHeaderFactory(task_id=<your id>)`, so basically very similar logic to this one here.
Not insisting but I feel it's somewhat redundant to have these helpers.

Krigpl (approved):

golem/task/taskcomputer.py
```diff
@@ -42,28 +48,395 @@ def __init__(self):
...
+            self._assigned_task = None
+            self._runtime = None
+            self._task_finished_callback()
+
+    def _get_task_dir(self, task_id: str, env_id: EnvId) -> Path:
+        return self._work_dir / env_id / task_id
+
+    def _get_task_api_service(
+            self,
+            task_id: str,
```
Krigpl: These params are not needed, you can just grab it from the currently assigned task. Same for `get_task_dir`.
This is only meant to be used for the current task anyway.

Wiezzel: Done.

golem/task/taskcomputer.py
```diff
@@ -42,28 +48,399 @@ def __init__(self):
...
+        )
+        ProviderTimer.start()
+
+    def compute(self) -> defer.Deferred:
+        assigned_task = self._assigned_task
+        assert assigned_task is not None
+
+        task_api_service = self._get_task_api_service()
+        # FIXME: Remove when ProviderAppClient implements shutdown
+        self._runtime = task_api_service._runtime  # noqa pylint: disable=protected-access
```
Krigpl: You'll set `_runtime` to `None` here. You should move this line to after calling `compute` below.

golem/task/taskcomputer.py
```diff
@@ -42,28 +48,399 @@ def __init__(self):
...
+        self._new_computer = NewTaskComputer(
+            env_manager=env_manager,
+            work_dir=task_server.get_task_computer_root(),
+            task_finished_callback=finished_cb,
+            stats_keeper=self.stats
+        )
+        sync_wait(self._new_computer.prepare())
+
+        # Should this node behave as provider and compute tasks?
+        self.compute_tasks = task_server.config_desc.accept_tasks \
```
mfranciszkiewicz, line 78: `compute_tasks` should become a `@property` because afair `task_server.config_desc.in_shutdown` is changed dynamically during runtime

mfranciszkiewicz (approved):
One small thing

golem/task/taskcomputer.py
Wiezzel: This is a part of the config and config needs to be explicitly changed be calling `change_config`. That has been the case with the old `TaskComputer`.

If some code modifies the `in_shutdown` setting without calling `update_config` then probably it should be fixed but that's not a part of this PR.

maaktweluit: `change_config` is called on the client after updating this setting, so should be good :)

https://github.com/golemfactory/golem/blob/develop/golem/node.py#L377

tests/golem/task/test_taskcomputeradapter.py
```diff
@@ -0,0 +1,361 @@
...
+from golem.task.envmanager import EnvironmentManager
+from golem.task.taskcomputer import TaskComputerAdapter
+from golem.task.taskserver import TaskServer
+from tests.factories.taskserver import ClientConfigDescriptor
+
+
+class TaskComputerAdapterTestBase(TwistedTestCase):
+
+    @mock.patch('golem.task.taskcomputer.IntStatsKeeper')
+    @mock.patch('golem.task.taskcomputer.TaskComputer')
```
Krigpl: How about spec_set?
```suggestion
    @mock.patch('golem.task.taskcomputer.TaskComputer', mock.Mock(spec_set=TaskComputer))
```

same below

PR #4512, opened 2019-07-19
Introduce RequestorBrassMarketStrategy

### LINKED ISSUE #4521: Refactor Brass marketplace and remove Twisted dependencies

mplebanski:
Closes #4521


golem/task/tasksession.py
```diff
@@ -311,16 +313,39 @@ def _react_to_want_to_compute_task(self, msg):
...
+            ).addErrback(_on_error)
+            logger.info(
+                "Will select providers for task %s in %.1f seconds",
+                msg.task_id,
+                self.task_server.config_desc.offer_pooling_interval
+            )
+
+        market_strategy.add(
+            Offer(msg, msg.task_id, self.key_id, ProviderStats(0),
+                  current_task.header.max_price, msg.price, self)
```
Krigpl, line 347: Most of the times when you feel like passing `self` to a constructor of a different object there's huge probability there's an issue with the design. This is the case here, the way `context` is used is completely unnecessary. Just do `self._offer_chosen` on line 320.

golem/task/tasksession.py
```diff
@@ -311,16 +313,39 @@ def _react_to_want_to_compute_task(self, msg):
...
-        logger.debug("Offer accepted & added to pool. offer=%s", offer)
-        d.addCallback(functools.partial(self._offer_chosen, msg=msg))
-        d.addErrback(golem_async.default_errback)
+        market_strategy = get_task_market_strategy(self.task_manager,
+                                                   current_task)
+
+        if market_strategy\
+                .get_task_offer_count(msg.task_id) == 0:
+            # This is a first offer for given task_id, schedule resolution.
+            from twisted.internet import reactor
```
Krigpl, line 331: There's `golem.deferred.call_later` for that.

Krigpl (changes requested):
fwiw, removing twisted from marketplace and adjusting tasksession should be one PR and the rest of the refactoring another one. Keep PRs modular.

golem/task/tasksession.py
mplebanski: @Krigpl Unfortunately If I call `self._offer_chosen` on line 320 it will be all done in a context of a single provider. Glad you looked into it, we are looking for a solution that would not require us to pass `self` to another object.

Krigpl: Quite. Then one way would be to simply replace context with the generic callback function
(which is effectively same as the previous approach with deferreds).

mfranciszkiewicz: @mplebanski The task session object should not be required at all - we have a task session message queue in place. Besides, it would be a better idea to store `weakref`s to sessions if truly needed.

mplebanski: @mfranciszkiewicz Could you elaborate on `task session object should not be required at all`?

mplebanski:
> fwiw, removing twisted from marketplace and adjusting tasksession should be one PR and the rest of the refactoring another one. Keep PRs modular.

I am closing this in order to open a PR for refactor only. https://github.com/golemfactory/golem/pull/4533

PR #4515, opened 2019-07-19
Transcoding: all files, codecs, framerates, resolutions

cameel:
This pull request adds all the remaining changes done as a part of the previous batch of tasks (tasks A-E) and tasks I2 and O2 on Golem side. It **does not** contain the implementation of the other most recent tasks. Those will be separate pull requests. The changes include:
1. Test improvements
    - Tests now do not change the transcoding parameters they do not test. For example the resolution change tests use the codec and container from the source file instead of always transcoding to `h264/mp4`.
        - If preserving the parameters is not possible (e.g. the codec can be decoded but not encoded by ffmpeg) the test is skipped.
    - Tests check if the target parameters are supported and if not - the test is skipped. So changing the supported codecs, resolutions, codec conversions, etc. in `ffmpeg-tools` may change the set of conversion tests that are actually performed here.
        - In case of the conversion that is the main subject of the test (e.g. codec in case of codec change tests or resolution in case of resolution change tests), the test first makes sure that the right exception (e.g. `UnsupportedVideoCodecConversion`, `InvalidResolution`) is raised before the test is skipped. If the code does not detect the combination as unsupported, the test fails.
    - Added custom names for each case in parameterized tests (`testcase_func_name`). Now the name contains the source and target parameters (e.g. codec/container, resolution, frame rate) which makes it easier to tell which particular case failed just by the name of the test.
2. More tested combinations:
    - Codec change tested for every codec for which transcoding is possible. Each codec tested with one container.
    - Full list of files from `transcoding-video-bundle` v4 added to the transcoding integration tests.
    - Testing with more resolutions.
    - Testing with more frame rates.
3. Disabled or relaxed checks in tests:
    - Ignoring `frame_count`, `codec_name`, `sample_rate`, `sample_format` and `bitrate` for audio streams.
    - Ignoring `pixel_format` for video streams. Also `frame_count` in codec change and frame rate change tests
    - Ignoring `codec_name` for subtitle streams.
    - If an attribute of the input or output file has no value in metadata returned by ffprobe it's ignored in comparisons even if it does have a value for the other file. This was causing a lot of false-negatives.
    - Frame rate comparisons now have a tolerance of +/- 0.5 Hz.
    - A workaround to make tests treat `h265` and `hevc` as the same codec.
    - `relative_start_time` instead of `start_time` in FfprobeReport.
    - For `program_count` zero is now considered the same as a missing value. This way the attribute is actually compared only if the input or output file has at least 2 programs. Otherwise conversions from a format that supports programs to the one that does not and vice versa produce differences in tests.
    - Duration tolerance increased to +/- 1 second.
4. Frame rate validations.
5. More specific exception types raised by transcoding steps. This lets us know at a glance at which stage a particular test failed if it ended with an error returned from the Docker container.

### Things of note
1. Running tests of the files from `transcoding-video-bundle` is disabled for now (they are skipped). If you want to run the full tests for now you have to manually remove the `skip()` commands from the beginning of each test. They will be enabled once subsequent pull requests make them pass on the full set.
2. The newly added resolutions will have to be be added to `ffmpeg-tools` first (https://github.com/golemfactory/ffmpeg-tools/pull/9). Note that resolution change test won't fail without them but they'll be skipped which makes it useless.
3. The choice of containers is pretty arbitrary. I made sure that each supported codec is tested but the containers were assigned only based on what I thought looked like a common container for that codec. I also tried to choose a different container for each codec where possible.

### Dependencies
This pull request requires changes from https://github.com/golemfactory/ffmpeg-tools/pull/9 which has not been merged yet.


golem/testutils.py
```diff
@@ -424,7 +417,7 @@ def _run_test_job(self, task, root_dir, params):
         self._log_docker_logs(dtt)
 
         if dtt.error:
-            raise DockerTestJobFailure(dtt.error_msg)
+            raise ffmpegTranscodingError(dtt.error_msg)
```
nieznanysprawiciel: TestTaskIntegration should be general enough to be used with other use cases as well. So it should throw more generic exception then ffmpegTranscodingError.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
```diff
@@ -103,46 +225,116 @@ def test_split_and_merge_with_codec_change(self,
...
+        (
+            (video, resolution)
+            for video in VIDEO_FILES  # pylint: disable=undefined-variable
+            for resolution in (
+                [320, 240],
+                [640, 260],
+                [640, 480],
+            )
+        ),
+        testcase_func_name=lambda testcase_func, param_num, param: (
```
nieznanysprawiciel, line 267: Can you hide name generation, by, for example, creating function that will do this? This code isn't necessary, to understand test, but takes space and makes code less readable.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
```diff
@@ -103,46 +225,116 @@ def test_split_and_merge_with_codec_change(self,
...
-        operation.exclude_from_diff({'video': {'bitrate', 'frame_count'}})
-        (_input_report, _output_report, diff) = operation.run(video_file)
-        self.assertEqual(diff, [])
+        operation.request_resolution_change(video["resolution"])
+        operation.exclude_from_diff(
+            TestFfmpegIntegration.ATTRIBUTES_NOT_PRESERVED_IN_CONVERSIONS)
+        operation.exclude_from_diff({'video': {'frame_count'}})
+        operation.enable_treating_missing_attributes_as_unchanged()
+
+        if not Container.is_supported(video['container'].value):
```
nieznanysprawiciel, line 238: Could you hide in function skips logic? These skips makes test logic harder to read, but the aren't main purpose of these tests.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
```diff
@@ -103,46 +225,116 @@ def test_split_and_merge_with_codec_change(self,
...
+
+        supported_conversions = source_codec.get_supported_conversions()
+        if video_codec.value in supported_conversions:
+            (_input_report, _output_report, diff) = operation.run(
+                video["path"])
+            self.assertEqual(diff, [])
+        else:
+            with self.assertRaises(UnsupportedVideoCodecConversion):
+                operation.run(video["path"])
+            pytest.skip("Video codec conversion not supported")
```
nieznanysprawiciel, line 255: Checking if transcoding raises exception is for sure good idea. But what is purpose of skipping test after we checked, that exception is raised?

tests/apps/ffmpeg/task/test_ffmpegintegration.py
```diff
@@ -103,46 +225,116 @@ def test_split_and_merge_with_codec_change(self,
...
+            pytest.skip("Source container not supported")
+        if not Container.is_supported(container.value):
+            pytest.skip("Target container not supported")
+        if not video['container'].is_supported_video_codec(source_codec.value):
+            pytest.skip("Source video codec not supported by the container")
+        if not container.is_supported_video_codec(video_codec.value):
+            pytest.skip("Target video codec not supported by the container")
+
+        supported_conversions = source_codec.get_supported_conversions()
+        if video_codec.value in supported_conversions:
```
nieznanysprawiciel, line 248: Maybe VideoCodec should have can_convert function. It takes to much code to check so simple thing.

tests/apps/ffmpeg/task/utils/ffprobe_report.py
```diff
@@ -513,7 +576,15 @@ def _diff_attributes(cls,
                 getattr(expected_stream_report, attribute),
             )
 
+            skip_missing_value = (
```
nieznanysprawiciel: These new added lines are the same as code in other reports implementation. The same is with _diff_attributes methods, they look very similar between different classes. Is it possible, to refactor it, so there is no code duplication and these classes would share common implementation?

nieznanysprawiciel (commented):
I would be useful to have some tests checking if ffmpeg integration raises correct exceptions. It should be one test per case. Since we won't run tests on full dataset, the need to be separated from tests on video bundle.

golem/testutils.py
cameel: Good point. I added this class because I wanted it to be perfectly clear that the step that failed was the transcoding performed by the provider. `DockerTestJobFailure` seems a bit too generic for that. I see that in practice only `_run_test_job()` raises it so it's not ambiguous but when I saw it in logs I never felt that certainty. It always registered in my mind as "some docker job failed somewhere". Maybe it's the matter of getting used to the fact that it refers **the** test job (the one that's part of the task) and not **a** test job that could be executed anywhere in the base class.

We'll change it back.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
cameel: The reason I left it here was that these are hard to make reusable - `params[0][x]` values must match the positional function parameters. And it's just a single string that would be a single line if not for the 80 character limit. I felt that moving this to a place where it's not next to the function signature was making it even harder to understand than it is now.

But yeah, I'm not really satisfied with how it turned out. We have actually already changed that - just not in this pull request. @pawelkisielewicz, while moving the tests for the whole set of videos to a separate module, had to move it to a function to avoid duplication. So it'll come with that pull request.

I can take it out from there and copy it here if you want but I'd recommend to just wait for that upcoming pull request (looks like it's going to pass my internal review soon).

tests/apps/ffmpeg/task/test_ffmpegintegration.py
cameel: Sure. You're right, this obscures the more important test logic. It will result in 4 extra functions tightly coupled to the corresponding tests though. That's because the conditions are not identical in each test - I didn't want overcomplicate this and replicate the whole validation logic here. But maybe that's still better.

I was actually thinking about moving more stuff out of these tests but I didn't want to bury it under too many layers. We already have the three utility classes created just to make the tests leaner... I think that simplificaction might be more effective after we're done with all the upcoming changes. We could then take a step back, look at it and decide how to better organize it.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
cameel: It's a bit of a hack. Ideally, this would be two separate tests:
- One would check if the conversion works when the target is supported. And would skip the test if it's not supported.
- The other would check that for unsupported targets we get an exception.

The problem is that it would add lots of duplication. These tests are very long and the two would be almost identical - only the part in the `if` would be different.

I could also remove the `skip()` and assume that the check for the exception is just a part of the test. But having it marked as skipped gives you an indication that the conversion was not really tested - which was useful to me when working on it.

Another option would be to make the test that expects the exception simpler - use less combinations, hard-code some parameters, etc. I felt that it would still leave too much duplicated code.

So I just didn't see a perfect solution here and I chose the one that seemed the least bad to me. I can change it if you prefer one of the alternatives.

tests/apps/ffmpeg/task/utils/ffprobe_report.py
cameel: Sure. If you think it's worth it, we can do this.

I left it like this because originally it was just a couple of assignments and a simple condition that was awkwardly broken into multiple lines due to line length limit. And it's hard to give it a good and precise name that wouldn't be longer than the condition itself. The function call including passing parameters to it and assigning results might not be that much shorter than the code itself.

But it grew a bit over time as new parameters were being added so maybe it makes more sense now. I'll see what we can do.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
cameel: Maybe. It could be written as: `video_codec.value in source_codec.get_supported_conversions()` which is a bit longer than `source_codec.can_convert(video_codec.value)` but not dramatically so. The reason it's not on one line is that when I originally wrote this condition, it was longer - it had an `and` part. Then I realized that I did not need the other part and removed it. I must have not noticed back then that it made it short enough to fit on one line or I would change it.

Should we define it? If such a function was available, I'd use it but I'd suggest just making it a single line here.

cameel:
> I would be useful to have some tests checking if ffmpeg integration raises correct exceptions. It should be one test per case. Since we won't run tests on full dataset, the need to be separated from tests on video bundle.

Sure. Do you mean exceptions like `ffmpegExtractSplitError` and others newly added in this pull request? Or the exceptions that validations from `ffmpeg-tools` throw?

tests/apps/ffmpeg/task/test_ffmpegintegration.py
nieznanysprawiciel: We can wait for next PRs

tests/apps/ffmpeg/task/test_ffmpegintegration.py
nieznanysprawiciel: > But having it marked as skipped gives you an indication that the conversion was not really tested - which was useful to me when working on it
Ok. This convinces me We can leave it as is.
Additional tests checking only exceptions will solve problem.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
nieznanysprawiciel: It's not big deal here.
But Codec class could have such function.
It requires changes in ffmpeg-tools, so better let's do this later.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
nieznanysprawiciel: Ok this can wait for nexts PRs if you prefer this way.

nieznanysprawiciel:
I meant rather validation exceptions. If there are some specific conditions that should throw for example ffmpegExtractSplitError, than we should test it too.
These tests should use video that is already in repository.
And I don't want to test all combination, but only each type of error once.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
cameel: OK. In that case I'll add an extra task to refactor these tests a little after we're done with the others.

tests/apps/ffmpeg/task/test_ffmpegintegration.py
cameel: OK. Great.

cameel:
OK. So we'll add tests for the following validations:
- Unsupported target video codec
- Unsupported target audio codec
- Unsupported target container
- Target container does not support the target video codec
- Target container does not support the target audio codec
- Target container is an exclusive demuxer
- Unsupported target frame rate
- Unsupported target resolution

I looked closer at `ffmpegExtractSplitError` and similar exceptions but you're right, I can't see any specific cases we could test. They're just wrappers for exceptions or error codes from docker jobs and represent mostly unpredictable failures or problems with the source file.

pawelkisielewicz:
There is one problem with validation: Unsupported target audio codec. In this case `AudioCodecNotSupportedByContainer` should be raised, but it is checked only if there is at least one audio stream in file. If there is no audio stream, it is not checked and validation automatically passes. 
The problem is that both files from git repository does not have any audio stream, so with this file it is not possible to test this Exception. What should be done in this case?

nieznanysprawiciel:
Do we have any very small file that could be added to repo?

cameel:
Yes. There are several options. Which one do you prefer?

1. There are many random sites with test videos. Easy to find small files but license may be a problem:
    - `small.mp4` from [standaloneinstaller.com > Big list of sample videos for testers](https://standaloneinstaller.com/blog/big-list-of-sample-videos-for-testers-124.html)
        - license: not stated, it only says "sample videos for 3gp, avi, flv, m4v, mkv, mov,mp4, mpg, mpeg, mts, webm, wmv, vob gathered from various sources from internet".
        - direct link: [`small.mp4`](http://mirrors.standaloneinstaller.com/video-sample/small.mp4) 0.17 MB (many other formats available in sizes between 0.17 and 0.44 MB encoded with various video codecs).
    - https://sample-videos.com/ has multiple 1 MB samples from Big Buck Bunny with audio and encoded with various video codecs.
2. Clips from archive.org. There are lots of videos, often with Creative Commons licenses. The downside is that I could not find a way to filter by file size or duration and most files seem to be big (25 MB and over). If we spend some time on it I think we could find something very small but the smallest ones I found after some browsing were:
    - [archive.org > test file mp4](https://archive.org/details/Pbtestfilemp4videotestmp4)
        - license: [CC BY-NC-SA 2.5](https://creativecommons.org/licenses/by-nc-sa/2.5/)
        - direct link: [`video_test.ogv`](https://archive.org/download/Pbtestfilemp4videotestmp4/video_test.ogv) 909.2 kB (slightly bigger .mp4 version also available).
    - [archive.org > Sample Video 1280x 720 2mb](https://archive.org/details/SampleVideo1280x7202mb_201905)
        - license: not specified but the clips are from [Big Buck Bunny](https://en.wikipedia.org/wiki/Big_Buck_Bunny) which uses [CC BY 3.0](https://creativecommons.org/licenses/by/3.0/)
        - direct link: [`SampleVideo_1280x720_2mb.ogv`](https://ia903001.us.archive.org/16/items/SampleVideo1280x7202mb_201905/SampleVideo_1280x720_2mb.ogv) 1 MB (a 2 MB .mp4 version is also available)
3. The whole Big Buck Bunny is available at https://peach.blender.org/download/. We could use ffmpeg or any other software to extract a very small fragment ourselves.
4. We could just record a very short video with a phone or a webcam.

I think that option 3 would be the best choice.

nieznanysprawiciel:
We need video that is as small as possible and is able to be in public repo.
Choose whatever you want.

golem/testutils.py
pawelkisielewicz: Ok, it is changed back to previous version

tests/apps/ffmpeg/task/utils/ffprobe_report.py
pawelkisielewicz: Done, the whole duplication is removed, logic is now in one function called in both places

magdasta:
Can you point out specific lines where these test are implemented (test method headers)? I couldn't find some of them

> * Unsupported target video codec
> * Unsupported target audio codec
> * Unsupported target container
> * Target container does not support the target video codec
> * Target container does not support the target audio codec
> * Target container is an exclusive demuxer
> * Unsupported target frame rate
> * Unsupported target resolution

cameel:
Yeah, sorry, it looks like we only have one of the two tests that were supposed to validate the audio codec and even that one validates the source audio codec rather than the target one. We'll fix that.

So:
- Unsupported target video codec: already covered by an existing test `test_task_invalid_params()` (https://github.com/golemfactory/golem/pull/4515/files#diff-89cec46e17c0203f8be527fa5a95b5a4L264-L279)
- Unsupported target audio codec: **missing**
- Unsupported target container: `test_invalid_container_should_raise_proper_exception()` (https://github.com/golemfactory/golem/pull/4515/files#diff-89cec46e17c0203f8be527fa5a95b5a4R596-R610)
- Target container does not support the target video codec:  `test_unsupported_target_video_codec()` (https://github.com/golemfactory/golem/pull/4515/files#diff-89cec46e17c0203f8be527fa5a95b5a4R526-R540)
- Target container does not support the target audio codec: `test_unsupported_audio_codec_should_raise_proper_exception()` (https://github.com/golemfactory/golem/pull/4515/files#diff-89cec46e17c0203f8be527fa5a95b5a4R612-R624) (**checks the wrong thing**)
- Target container is an exclusive demuxer: `test_unsupported_target_container_if_exclusive_demuxer()` (https://github.com/golemfactory/golem/pull/4515/files#diff-89cec46e17c0203f8be527fa5a95b5a4R542-R557)
- Unsupported target frame rate: `test_invalid_frame_rate_should_raise_proper_exception()` (https://github.com/golemfactory/golem/pull/4515/files#diff-89cec46e17c0203f8be527fa5a95b5a4R578-R594)
- Unsupported target resolution: `test_invalid_resolution_should_raise_proper_exception()` (https://github.com/golemfactory/golem/pull/4515/files#diff-89cec46e17c0203f8be527fa5a95b5a4R559-R576)

cameel:
As for the video: we extracted a clip from the Big Buck Bunny movie (about 170 kB). Its Creative Commons license requires attribution so we have included a short README that says where the file originally came from.

magdasta:
Also, there's a problem with linter failing for tests

cameel:
Pylint warning fixed. One line was not properly wrapped.

magdasta:
Linter is failing again

cameel:
Linter problems fixed. Sorry for that.

The missing test has been added and now the audio tests check target audio parameters instead of validating the source file.

By the way - it's impossible to get `UnsupportedAudioCodec` outside of tests in Golem because it won't let ffmpeg-tools raise it. `_assert_codec_container_support()` in `TranscodingTask` does a part of the same validation and raises `AudioCodecNotSupportedByContainer` instead (used to be `TranscodingTaskBuilderException` before this pull request). Now that ffmpeg-does validation, `_assert_codec_container_support()` seems superfluous and probably should be removed.

One more important thing. @pawelkisielewicz found a small bug [in the corresponding pull request in ffmpeg-tools](https://github.com/golemfactory/ffmpeg-tools/pull/9) which has already been merged. Here's a fix: https://github.com/golemfactory/ffmpeg-tools/pull/13. Without this fix the test for "target container is an exclusive demuxer" will not pass when `--runslow` is used.

nieznanysprawiciel:
@cameel  _assert_codec_container_support() is not so superflouos. It gives us chance to reject task before it's created.
I don't understand why test_task_invalid_audio_params works. It should raise exception while creating task_definition...

cameel:
> _assert_codec_container_support() is not so superflouos. It gives us chance to reject task before it's created.

It's superflous in the sense that in its current form it does not do anything above what ffmpeg-tools already does. It just checks `is_supported_audio_codec()` and `is_supported_video_codec()`. But if you mean that you want to keep it that way because you want to have a place to add some further restrictions that are very specific to golem-app business logic and would not make sense in a generic library like ffmpeg-tools then fine, that makes sense. But in that case I'd suggest renaming it to make it clear that it was meant to work that way. It wasn't obvious to me.

> I don't understand why test_task_invalid_audio_params works. It should raise exception while creating task_definition...

The structure of this test is basically the same as in `test_task_invalid_params()` - one of several integration tests that were already present when we started working with that code - which is located just above the block of a few new tests added in this pull request.

We used this structure instead of relying on `SimulatedTranscodingOperation` (used in the other tests that check validations) because STO's methods expect codec enums rather than strings. Using enums is in my opinion better in most cases - but here we specifically needed to use a spurious codec that would be rejected as invalid. We could have extended STO to handle codecs as strings but I did not think it was worth it just to test this one special case.

As to how it works - yes, the exception should be raised when a task definition is being created by the requestor and not when the task is already being executed by the provider - but that's exactly what happens in the test. The thing is that the task def object is being created inside `self.execute_task(task_def)`. `self._create_task_def_for_transcoding()` just creates a dict that is later used by `execute_task()` to create the actual task def object.

The confusing thing is probably that we refer to the dict created by `_create_task_def_for_transcoding()` as "task def". Maybe we should use a more specific name? `task_def_dict`, `task_def_input`, `proto_task_def` or something like that? `task_def_def` would be great in my opinion but probably even more confusing for a different reason :)

nieznanysprawiciel:
But creating task definition should raise AudioCodecNotSupportedByContainer and test checks if it raises UnsupportedAudioCodec

cameel:
Creating task definition could raise either of them, depending on which validation runs first:
- `build_full_definition()` calls `_assert_codec_container_support()` which would raise `AudioCodecNotSupportedByContainer`.
- `initialize()` calls `validate_transcoding_params()` which would raise `UnsupportedAudioCodec`.

Both of them run on requestor's machine and I consider them to be a part of the code that creates task definition (one produces task def, the other splits the video and updates the def)

So it's only a matter of which runs first. In this situation I would indeed expect `AudioCodecNotSupportedByContainer` as you pointed out because `build_full_definition()` is supposed to run first.

@pawelkisielewicz I remember seeing it in my review but I don't remember what the reason was off the top of my head and I am unable to run the code right now to check. Could you chime in and explain <s>why it's not raising `UnsupportedAudioCodec`</s>? EDIT: I meant "why it's raising `UnsupportedAudioCodec`?" of course.

pawelkisielewicz:
There are 2 validations, raising:
`AudioCodecNotSupportedByContainer` - in golem code, normally this runs first
`UnsupportedAudioCodec` - in ffmpeg code, this runs after. 

I agree that `AudioCodecNotSupportedByContainer` should be raised, it is checked in `test_unsupported_audio_codec_should_raise_proper_exception`. Here task definition is build by golem code. 
But ffmpeg-tools is separate repository, has its own validations and only one way to get `UnsupportedAudioCodec` is to do it like in `test_task_invalid_audio_params`. It has very similar structure to `test_task_invalid_params`. And the task was to test all validations.

nieznanysprawiciel:
Both tests test_task_invalid_audio_params and test_task_invalid_params use execute_task, which calls building task definition first and then initializes task.
I can't see any difference between these tests and others, that would cause, that AudioCodecNotSupportedByContainer is not raised.

cameel:
@nieznanysprawiciel I'm back at my computer now so finally I could run the code, look at the stack trace of the exception and give you a full explanation.

In short: `build_full_definition()` raises `UnsupportedAudioCodec` rather than `AudioCodecNotSupportedByContainer` when a codec is not listed in `AudioCodec`/`VideoCodec`. That's because it tries to convert a string to an enum value using `from_name()` and it fails. `AudioCodec._missing_()` gets called and it simply raises `UnsupportedAudioCodec`.

That's how this always worked but it if you want us to change it so that `AudioCodecNotSupportedByContainer` is raised instead, it looks pretty straightforward. Here's the relevant fragment from `build_full_definition()`:

``` python
            ac = audio_options.get('codec')
            vc = video_options.get('codec')

            audio_params = TranscodingTaskOptions.AudioParams(
                AudioCodec.from_name(ac) if ac else None,
                audio_options.get('bit_rate'))

            video_params = TranscodingTaskOptions.VideoParams(
                VideoCodec.from_name(vc) if vc else None,
                video_options.get('bit_rate', presets.get('video_bit_rate')),
                video_options.get('frame_rate'),
                video_options.get('resolution'))

            output_container = Container.from_name(options.get(
                'container', presets.get('container')))

            cls._assert_codec_container_support(audio_params.codec,
                                                video_params.codec,
                                                output_container)
```

We could simply move `_assert_codec_container_support()` above the `audio_params` assignment and give it the raw codec name before we even try to get an enum value. This way we would never get `UnsupportedAudioCodec` here (unless someone adds a value not present in `AudioCodec`/`VideoCodec` to `_CONTAINER_SUPPORTED_CODECS` in ffmpeg-tools by mistake but we could make such a mistake easy to spot in development by adding an assert there).

As @pawelkisielewicz said above, this is actually the only way he managed to trigger `UnsupportedAudioCodec` and `UnsupportedVideoCodec` in Golem's code so after this change it should no longer be possible to stumble into it.

Let me know if you want this change.

nieznanysprawiciel:
No, there are no changes needed. I only wanted to know why it works.

cameel:
OK. Great.

PR #4520, opened 2019-07-22
Mark test_fails as skip since it is expected to fail

### LINKED ISSUE #4516: `script/node_integration_test/tests/test_fails.py` fails [duh]
## Description

**Golem Version**: 0.20.1+
**Branch** (if launched from source): b0.20.1

**Description of the issue**:

the test fails - which might be its purpose ... in which case, it should be skipped during normal execution of tests and only executed when some parameter is passed or whatnot... 

## Steps To Reproduce
1. `pytest -sv scripts/node_integration_tests/`
2. `scripts/node_integration_tests/tests/test_fails.py::TestFails::test_fails FAILED`

## Expected behavior
The test should not fail - or if failing is its purpose - should be skipped during normal execution of the test suite

## Proposed Solution?
add some required parameter to have this test included in the test run?

maaktweluit:
Resolves #4516

### DIFF
--- a/scripts/node_integration_tests/tests/test_fails.py
+++ b/scripts/node_integration_tests/tests/test_fails.py
@@ -1,3 +1,9 @@
-class TestFails:
+from unittest import TestCase
+
+import pytest
+
+
+class TestFails(TestCase):
+    @pytest.mark.skip("A test to check teardowns and is skipped by default")
     def test_fails(self):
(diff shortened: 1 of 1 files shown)


shadeofblue (approved):
lgtm -> though, it would be good to cherry-pick that into `b0.20`, given that the issue is also present there

PR #4527, opened 2019-07-23
Task API subtask creation flow

Krigpl:
Adjust react to want to compute task to Task API flow. Yet another two if conditions.

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -21,6 +21,14 @@ class CreateTaskParams:
     concent_enabled: bool
 
 
+@dataclass
+class SubtaskDefinition:
+    subtask_id: SubtaskId
+    resources: List[str]
+    params: Dict[str, Any]
+    deadline: int
+
+
 class RequestedTaskManager:
     def create_task(
             self,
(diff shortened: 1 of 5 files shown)


golem/task/tasksession.py
```diff
@@ -259,14 +264,15 @@ def _cannot_assign_task(self, task_id, reason):
...
-
+        task_id = msg.task_id
         reasons = message.tasks.CannotAssignTask.REASON
 
         if msg.concent_enabled and not self.concent_service.enabled:
             self._cannot_assign_task(msg.task_id, reasons.ConcentDisabled)
             return
 
-        if not self.task_manager.is_my_task(msg.task_id):
+        if not self.task_manager.is_my_task(msg.task_id) and \
```
maaktweluit: `msg.` is no longer needed here ( there are more places but these lines are changed )

maaktweluit (approved):
LGTM!

One nitty comment, still approving

golem/task/requestedtaskmanager.py
```diff
@@ -45,11 +53,32 @@ def task_exists(self, _task_id: TaskId) -> bool:  # noqa pylint: disable=no-self
...
+        raise NotImplementedError
+
     def get_subtasks_outputs_dir(self, task_id: TaskId) -> Path:
         """ Return a path to the directory where subtasks outputs should be
         placed. """
         raise NotImplementedError
 
+    def has_pending_subtasks(self, task_id: TaskId) -> Path:
+        """ Return True is there are pending subtasks waiting for
+        computation. """
```
Wiezzel: Why the `Path` return type? And how does this differ from `is_task_finished`?

golem/task/tasksession.py
```diff
@@ -345,35 +362,63 @@ def _offer_chosen(  # pylint: disable=too-many-locals
         )
         offer_hash = binascii.hexlify(msg.get_short_hash()).decode('utf8')
         for _i in range(msg.num_subtasks):
-            ctd = self.task_manager.get_next_subtask(
-                self.key_id, msg.task_id, msg.perf_index, msg.price, offer_hash)
+            if self.requested_task_manager.task_exists(task_id):
```
Wiezzel: This method is a little bit too long. Maybe it could be somehow split?

Wiezzel (approved):
Looks fine.

golem/task/requestedtaskmanager.py
Krigpl: `Path` is to be fixed, good catch.
`han_pending_subtasks` tells you whether next call to `get_subtask` will give you a proper subtask definition, while task_finished is when a task is finalized, i.e. all subtasks computed and verified, or the task timed out, or was aborted, etc.
It may happen that `has_pending_subtasks` will return False, because all subtasks all currently being computed or verified, but some will fail verification and next call to `has_pending_subtasks` will return True again. All that while task_finished is False.

PR #4528, opened 2019-07-23
[O5] Modify decorator removing test files to remove them by default

cameel:
This is a simple change that flips the meaning of the `@remove_temporary_dirtree_if_test_passed` decorator (now called `@keep_temporary_dirtree_if_test_failed`). Now temporary files are automatically deleted unless you use the decorator. And even then we only keep them if the test fails.

### Dependencies
This pull request is based on #4515 and should be merged after it.

**NOTE**: I have set the base branch of this pull request to `transcoding-all-files-codecs-framerates-resolutions` (#4515) rather than `CGI/transcoding/master`. Please remember to change the base branch back to `CGI/transcoding/master` before you merge.

### DIFF
--- a/golem/testutils.py
+++ b/golem/testutils.py
@@ -196,10 +196,11 @@ class PEP8MixIn(object):
 def keep_temporary_dirtree_if_test_failed(fun):
     @wraps(fun)
     def wrapper(self, *args, **kwargs):
-        fun(self, *args, **kwargs)
-        # If test fails, we won't reach this point, but tearDown
-        # will be called and directories won't be removed.
-        self.REMOVE_TMP_DIRS = True
+        try:
+            fun(self, *args, **kwargs)
+        except BaseException:
+            self.REMOVE_TMP_DIRS = False
+            raise
     return wrapper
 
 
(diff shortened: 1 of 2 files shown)


golem/testutils.py
```diff
@@ -218,7 +219,7 @@ def setUp(self):
 
         # Assume that test failed. @dont_remove_dirs_on_failed_test decorator
         # will set this variable to True on the end of test.
```
nieznanysprawiciel: Comment not updated

nieznanysprawiciel (changes requested):
remove_temporary_dirtree_if_test_passed is still present in some tests in test_ffmpegintegration.py

nieznanysprawiciel:
> remove_temporary_dirtree_if_test_passed is still present in some tests in test_ffmpegintegration.py

Ok I changed it already

golem/testutils.py
pawelkisielewicz: Sorry for this, already updated

nieznanysprawiciel (approved):

cameel:
OK. Great. This pull request sits right on top of `CGI/transcoding/master` so it can be merged right way.

I have changed the base of this pull request to `CGI/transcoding/master` since the `transcoding-all-files-codecs-framerates-resolutions` has already been merged.

Sorry for the old decorator name. Looks like we did not update this pull request after we added new tests in a pull request below it.

nieznanysprawiciel:

PR #4529, opened 2019-07-23
+ integration test against an uneven resolution division (400/3)

shadeofblue:
+ support for multi-frame tasks in the output verification step


kmazurek (approved):
ok

PR #4545, opened 2019-07-25
Blender inconsistent subtask sizes

magdasta:
Fix for problem with blender tasks being negatively verified for uneven divisions (400x400 pixels, 3 subtasks) and multiple frames


kmazurek (approved):
LGTM, one minor suggestion would be to add type hints to the methods being added (`get_parts_in_frame`, `get_subtask_y_border`) and to `get_min_max_y`.

PR #4549, opened 2019-07-26
Avoid storing resources_options in TaskCompKeeper

Krigpl:
It's completely unnecessary since we can just pass TTC and grab it from there.


maaktweluit (approved):
Nice cleanup!

PR #4551, opened 2019-07-29
Remove unused fields from WaitingTaskResult

Krigpl:
Simply unnecessary.


Wiezzel (approved):
Do we still need to return these from `task_result_manager.create` then?

Krigpl:
ffs, this is somehow used by concent.

Issue #4526, opened 2019-07-23
Provider usage reporting for gWASM subtasks (GR-183)


shadeofblue:
## Description

US. As a Golem requestor I need to get usage  report for every subtask completed in order to calculate payment and evaluate provider according to my market strategy.

It includes change in RCT message. It can be extra data. (possibly it should be a separate field?)

As soon as a subtask verification is finished, usages of providers who passed verification should be reported to the market strategy module.

AC. PR sent to review.

etam:
What if provider is lying about resources usage?

Wiezzel:
@etam AFAIR it is covered in the Pay-as-you-Golem document.

etam:
@Wiezzel where is this document?

Wiezzel:
@etam Can't find the link now. Ask the R&D team.

PR #4559, opened 2019-07-30
CPU usage reporting in Docker jobs

kmazurek:
Part of: #4526

This modifies the WASM task entrypoint to use GNU time for obtaining
user and system CPU time usage measurements.
Usage reports are saved to a file in a directory mounted within the job Docker container (`/golem/stats/resource_usage.json`).
A new version of base and WASM Docker images will need to be released in order for this to work (GNU `time` is not included by default in the `ubuntu` image).


kmazurek:
Just wanted to share an idea for an alternative approach: instead of making the stats gathering on a per-app basis, we could have a common way of measuring the usage. In that scenario, the stats gathering program invocation would be added to a task's entrypoint within Golem core (`DockerJob` or `DockerTaskThread` would be the likely candidates).
For this purpose we could still use GNU `time` (although it's a bit messy when it comes to formatting, see my changes) or include a different tool in the base Docker image  (e.g. [docker-stats-on-exit-shim](https://github.com/delcypher/docker-stats-on-exit-shim), based on cgroups) and use it.
The main benefit here is having a more predictable format for the usage report file (which will need to be parsed once a task is computed so that the stats can be included in `ReportComputedTask`).

mplebanski:
I like the second approach better because one can think of stats as something generic for every computation and that would also simplify handling the case where stats do not exist because the app is not providing them (are gWasm specific like it is now). 

Data format for statistics could be standardized by a dataclass definition. One could e.g. pick the format returned by `docker-stats-at-exit-shim` or make up his own. 

What do @Wiezzel & @mfranciszkiewicz think about modifying `entrypoint` in `DockerJob`? 

Nevertheless, what you have done so far aligns with our actual requirements for gWasm usage marketplace.

kmazurek:
One extra point for a common way of collecting stats: potential app developers (including gWASM apps) won't have to implement it in their project unless they need some custom stats (which we won't support atm anyway).
I've created a tool ([repo here](https://github.com/golemfactory/docker-cgroups-stats)) for collecting Cgroups stats. It's based on `docker-stats-at-exit-shim` with a bunch of modifications (e.g. only including specific Cgroup subsystems, using Go modules for building).
This tool can be included in the base Docker image by adding a stage to build it and copy the binary to the final image.

kmazurek:
I've updated the proposed implementation to use the new approach. @mfranciszkiewicz and/or @Wiezzel please take a second look.

Wiezzel (approved):
Looks fine.
However, this has to be only a temporary solution because it's incompatible with the Task API. We cannot assume that app developers would use our base image. Any within-container stats collection is unreliable when the image is provided by an untrusted third party. As the final solution I would suggest periodically querying the Docker API for container resource usage and aggregating on Golem side.

kmazurek:
I see, as discussed - collecting usage stats externally should be part of the changes required for Task API. Before merging this I'm going to update all Docker images to use the new base image.

PR #4561, opened 2019-07-31
Task API computation flow

Krigpl:
Yet another two if conditions.

### DIFF
--- a/golem/task/taskcomputer.py
+++ b/golem/task/taskcomputer.py
@@ -10,7 +10,7 @@ from threading import Lock
 
 from dataclasses import dataclass
 from golem_messages.message.tasks import ComputeTaskDef, TaskHeader
-from golem_task_api import ProviderAppClient
+from golem_task_api import ProviderAppClient, constants as task_api_constants
 from pydispatch import dispatcher
 from twisted.internet import defer
 
@@ -118,6 +118,13 @@ class TaskComputerAdapter:
     def support_direct_computation(self, value: bool) -> None:
         self._old_computer.support_direct_computation = value
 
+    def get_task_resources_dir(self) -> Path:
+        if not self._new_computer.has_assigned_task():
+            raise ValueError(
+                'Task resources directory only available when a task-api task '
+                'is assigned')
+        return self._new_computer.get_task_resources_dir()
+
     def start_computation(self) -> None:
         if self._new_computer.has_assigned_task():
             task_id = self.assigned_task_id
@@ -143,7 +150,7 @@ class TaskComputerAdapter:
             self._task_server.send_results(
                 subtask_id=subtask_id,
                 task_id=task_id,
-                result=[output_file],
+                task_api_result=output_file,
             )
         except Exception as e:  # pylint: disable=broad-except
             self._task_server.send_task_failed(
@@ -264,6 +271,9 @@ class NewTaskComputer:
         if self._assigned_task is None:
             return None
         return self._assigned_task.subtask_id
+
+    def get_task_resources_dir(self) -> Path:
+        return self._get_task_dir() / task_api_constants.NETWORK_RESOURCES_DIR
 
     def _is_computing(self) -> bool:
         return self._computation is not None
--- a/golem/task/taskserver.py
+++ b/golem/task/taskserver.py
@@ -21,6 +21,7 @@ from golem_messages import exceptions as msg_exceptions
 from golem_messages import message
 from golem_messages.datastructures import tasks as dt_tasks
 from pydispatch import dispatcher
+from twisted.internet import defer
 from twisted.internet.defer import inlineCallbacks, Deferred, \
     TimeoutError as DeferredTimeoutError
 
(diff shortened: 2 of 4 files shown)


golem/task/taskserver.py
```diff
@@ -478,7 +492,7 @@ def resource_failure(self, task_id: str, reason: str) -> None:
 
     def send_results(self, subtask_id, task_id, result):
```
Wiezzel: I think this method is kind of obscure. I'd suggest the following changes to make the interface clearer:
* Get rid of the `result` dict. Having a dict that always holds just one key is pointless. And also dicts are evil.
* Introduce two optional parameters instead: `result` and `task_api_result` that would hold data which is now in `result['data']` and `result['task_api_data']` respectively.
* Provider type annotation for all parameters.

golem/task/taskcomputer.py
```diff
@@ -118,6 +118,13 @@ def support_direct_computation(self) -> bool:
     def support_direct_computation(self, value: bool) -> None:
         self._old_computer.support_direct_computation = value
 
+    def get_task_resources_dir(self) -> Path:
+        if not self._new_computer.has_assigned_task():
+            raise Exception(
```
Wiezzel: I think `ValueError` would be more appropriate.

tests/golem/task/test_taskcomputeradapter.py
```diff
@@ -360,6 +360,25 @@ def test_both_computers_reconfigured(self):
         )
 
 
+class TestTaskResourcesDir(TaskComputerAdapterTestBase):
+    def test_old_assigned(self):
+        self.new_computer.has_assigned_task.return_value = False
+        self.old_computer.has_assigned_task.return_value = True
+        with self.assertRaisesRegex(
+            Exception,
```
Wiezzel: Don't forget to update if you change the exception type in the method.

golem/task/taskcomputer.py
Krigpl: I dunno, `ValueError` is about incorrect arguments, but there are no arguments here.
> Raised when an operation or function receives an argument that has the right type but an inappropriate value

golem/task/taskserver.py
Krigpl: Didn't feel like doing that but okay https://github.com/golemfactory/golem/pull/4563

golem/task/taskcomputer.py
Wiezzel: I think `ValueError` is used in Python to indicate invalid state as well. See [this answer](https://stackoverflow.com/a/1701327).

Krigpl: Ok.

golem/task/taskserver.py
```diff
@@ -495,12 +515,18 @@ def send_results(self, subtask_id: str, task_id: str, result: List[Path]):
...
             last_sending_trial=last_sending_trial,
             delay_time=delay_time,
             owner=header.task_owner)
 
-        self._create_and_set_result_package(wtr)
+        if result:
+            self._create_and_set_result_package(wtr)
+        else:
+            resource_id = \
+                sync_wait(self.new_resource_manager.share(task_api_result))
```
mfranciszkiewicz, line 527: I'd avoid using `sync_wait` in this case. E.g. sharing a ~295 MB file takes 1.4s on my MBP. It doesn't seem like much but the thread will be blocked for that time.

Krigpl: Afaict this is consistent with the current approach where we also upload the results synchronously.

mfranciszkiewicz: I don't believe we do. If the function blocks, it's probably executed in another thread.

Krigpl: `_create_and_set_result_package` -> `task_result_manager.create` -> `EncryptedResultPackageManager::create` -> `resource_manager.add_file` -> `HyperdriveResourceManager::add_file` with `async=False` -> `HyperdriveResourceManager::_add_files_sync` -> `HyperdriveClient::add` which issues the post request synchronously.

mfranciszkiewicz: Well, it looks like someone broke the implementation. Package creation itself used to (and should) be done in a separate thread.

Krigpl: According to the blame it was synchronous for at least past two years.

Krigpl: Not saying it shouldn't be asynchronous, because it should, it's just I went with the consistent behavior instead of introducing a new one. This can be improved in a separate PR.

mfranciszkiewicz: I don't have the information on who broke it or when. Digging through the history eats up too much time.

https://github.com/golemfactory/golem/blob/90ef893f4e6b2877881ba1f7f350be6c4aa43c50/golem/task/tasksession.py#L907-L912

mfranciszkiewicz: Actually the package is still created in a separate thread.

`TaskComputer.task_computed` -> `TaskServer.send_result` is called explicitly in `TaskThread.run` (or `DockerTaskThread.run` -> `DockerTaskThread._task_computed`)

`TaskThread` inherits from `threading.Thread`

tests/golem/task/test_taskserver.py
```diff
@@ -1163,6 +1182,30 @@ def test_already_assigned(
         dispatcher_mock.send.assert_not_called()
         logger_mock.error.assert_called()
 
+    def test_task_api(
+            self, _logger_mock, _dispatcher_mock,
```
maaktweluit, line 1186: one argument per line from the style guide

maaktweluit (commented):
LGTM!

One nitty comment and will approve when the threading question is solved

Krigpl:
@maaktweluit I consider the threading issue resolved.

maaktweluit (approved):
ok, thanks!

Wiezzel (approved):

mfranciszkiewicz:
@Krigpl Please create a follow-up PR solving the `sync_wait` issue.

Krigpl:
@mfranciszkiewicz Didn't you find that this runs in a separate thread anyway so it doesn't effectively matter much anyway at the moment?

mfranciszkiewicz:
@Krigpl ATM it runs in a separate thread through `DockerTaskThread`. `NewTaskComputer` doesn't make use of the `DockerTaskThread` class.

Issue #4456, opened 2019-07-11
Golem doesn't notify about new version of an app


ederenn:
## Description

**Golem Version**: 0.20.0

**Golem-Messages version** 3.7.0

**Electron version** (if used): 0.2.0

**OS** [e.g. Windows 10 Pro]: Any os

**Branch** (if launched from source): -

**Mainnet/Testnet**:

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

When the new version on Golem is released there modal about new app vesion is not displayed

## Expected behavior

Golem should inform user that new app version is available in a modal similar to this one

<img width="514" alt="Screen Shot 2019-07-11 at 14 56 27" src="https://user-images.githubusercontent.com/32061891/61053944-cf7b0280-a3ee-11e9-879a-b2c5e8979ec9.png">

ederenn:
notification worked for me on upgrade from 0.20.0 to 0.20.1

PR #4562, opened 2019-07-31
Allow bootstrap nodes to send partial version tags

jiivan:
fixes: #4456

### DIFF
--- a/golem/network/p2p/peersession.py
+++ b/golem/network/p2p/peersession.py
@@ -21,10 +21,16 @@ logger = logging.getLogger(__name__)
 
 def compare_version(client_ver):
     try:
-        v_client = semantic_version.Version(client_ver)
+        v_client = semantic_version.Version(client_ver, partial=True)
     except ValueError:
         logger.debug('Received invalid version tag: %r', client_ver)
         return
+    # Convert partial version to complete one
+    v_client = semantic_version.Version(
+        f"{v_client.major}"
+        f".{v_client.minor if v_client.minor else 0}"
+        f".{v_client.patch if v_client.patch else 0}",
+    )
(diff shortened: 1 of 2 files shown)


golem/network/p2p/peersession.py
```diff
@@ -21,10 +21,16 @@
 
 def compare_version(client_ver):
     try:
-        v_client = semantic_version.Version(client_ver)
+        v_client = semantic_version.Version(client_ver, partial=True)
     except ValueError:
         logger.debug('Received invalid version tag: %r', client_ver)
         return
+    # Convert partial version to complete one
+    v_client = semantic_version.Version(
```
kmazurek, line 29: I think this conversion is not necessary. According to [how versions are compared](https://python-semanticversion.readthedocs.io/en/latest/reference.html#semantic_version.Version.__cmp__) in `semantic_version`, when at least one of the compared versions is partial and it's actually missing some components (e.g. patch or minor), the missing parts will be considered equal to their equivalents on the other side of the comparison. This matches our use case, since we only trigger a signal when the partial version is higher than `gconst.GOLEM_VERSION` (and that is still covered).

jiivan: I'm doing this because this value is propagated via an rpc event and it was always full version. I don't wanna risk electron problems and also wanna make life easier for frontend/apps devs.

https://github.com/golemfactory/golem/blob/b0.20-version/golem/client.py#L278-L280

kmazurek: Ok, got it - perhaps you could mention that in the comment here, just to make sure it's not considered redundant by someone else.

kmazurek (approved):
lgtm

PR #4563, opened 2019-07-31
Clean up TaskServer::send_results

Krigpl:
Don't use single key dict.


Wiezzel (approved):
Thank you @Krigpl, very cool!

PR #4564, opened 2019-07-31
Implement ProviderAppClient.shutdown in NewTaskComputer

maaktweluit:
Depends on https://github.com/golemfactory/task-api/pull/20

Replace the `NewTaskComputer._runtime` hack with a proper implementation of shutdown.

- Only store the running `ProviderAppClient` in `self._app_client`
- Call `ProviderAppClient.shutdown()` on any form of error, to ensure it is shut(ting) down
- ~use `self.has_assigned_task()` where possible over `is None` check~

### DIFF
--- a/golem/task/task_api/__init__.py
+++ b/golem/task/task_api/__init__.py
@@ -54,6 +54,9 @@ class EnvironmentTaskApiService(TaskApiService):
         return self._runtime.get_port_mapping(port)
 
     def running(self) -> bool:
+        """
+        Checks if the service is 'closable' thus needs to be shutdown on errors
+        """
         return self._runtime is not None and self._runtime.status() in [
             RuntimeStatus.CREATED,
             RuntimeStatus.PREPARING,
--- a/golem/task/task_api/__init__.py
+++ b/golem/task/task_api/__init__.py
@@ -5,7 +5,13 @@ from typing import Optional, Tuple, Type
 from golem_task_api import TaskApiService
 
 from golem.core.deferred import sync_wait
-from golem.envs import Environment, Prerequisites, Runtime, RuntimePayload
+from golem.envs import (
+    Environment,
+    Prerequisites,
+    Runtime,
+    RuntimePayload,
+    RuntimeStatus,
+)
 
 
 class TaskApiPayloadBuilder(abc.ABC):
@@ -46,6 +52,15 @@ class EnvironmentTaskApiService(TaskApiService):
         sync_wait(self._runtime.prepare())
         sync_wait(self._runtime.start())
         return self._runtime.get_port_mapping(port)
+
+    def running(self) -> bool:
+        return self._runtime is not None and self._runtime.status() in [
+            RuntimeStatus.CREATED,
+            RuntimeStatus.PREPARING,
+            RuntimeStatus.PREPARED,
+            RuntimeStatus.STARTING,
+            RuntimeStatus.RUNNING,
+        ]
 
     async def wait_until_shutdown_complete(self) -> None:
         assert self._runtime is not None
--- a/golem/task/taskcomputer.py
+++ b/golem/task/taskcomputer.py
@@ -10,10 +10,7 @@ from threading import Lock
 
 from dataclasses import dataclass
 from golem_messages.message.tasks import ComputeTaskDef, TaskHeader
-from golem_task_api import (
-    ProviderAppClient,
-	constants as task_api_constants
-)
+from golem_task_api import ProviderAppClient, constants as task_api_constants
 from pydispatch import dispatcher
 from twisted.internet import defer
 
--- a/golem/task/taskcomputer.py
+++ b/golem/task/taskcomputer.py
@@ -371,10 +371,12 @@ class NewTaskComputer:
             ProviderTimer.finish()
             self._computation = None
             self._assigned_task = None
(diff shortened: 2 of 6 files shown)


golem/task/taskcomputer.py
```diff
@@ -288,19 +286,17 @@ def task_given(
 
     def compute(self) -> defer.Deferred:
         assigned_task = self._assigned_task
-        assert assigned_task is not None
+        assert self.has_assigned_task()
```
Wiezzel: `has_assigned_task()` has not been used on purpose here. I believe typechecker doesn't understand this and will complain when accessing `assigned_task`'s fields.

golem/task/taskcomputer.py
```diff
@@ -231,11 +231,9 @@ def __init__(
         self._work_dir = work_dir
         self._task_finished_callback = task_finished_callback
         self._stats_keeper = stats_keeper or IntStatsKeeper(CompStats)
-        self._app_client = ProviderAppClient()
+        self._app_client: Optional[TaskApiService] = None
```
Wiezzel: Is it client or service? Name says one, type says the other.

golem/task/taskcomputer.py
```diff
@@ -310,7 +306,7 @@ def compute(self) -> defer.Deferred:
     @defer.inlineCallbacks
     def _wait_until_computation_ends(self) -> defer.Deferred:
         assigned_task = self._assigned_task
-        assert assigned_task is not None
+        assert self.has_assigned_task()
```
Wiezzel: Ditto.

golem/task/taskcomputer.py
```diff
@@ -361,7 +357,12 @@ def _wait_until_computation_ends(self) -> defer.Deferred:
             ProviderTimer.finish()
             self._computation = None
             self._assigned_task = None
-            self._runtime = None
+            # This check should not be needed but is there to please the linter
+            if self._app_client is not None:
```
Wiezzel: An assert would be better.

golem/task/taskcomputer.py
maaktweluit: that is a bummer, ok, will change it back

golem/task/taskcomputer.py
maaktweluit: client :) fixed

golem/task/taskcomputer.py
maaktweluit: an assert could raise an exception and leave an invalid state, the if check ensures the next lines are also executed

Updated it for now, but not sure what is best

golem/task/taskcomputer.py
Wiezzel: Well, optional attributes are like this. If only we had Rust-style state machines...

golem/task/taskcomputer.py
Wiezzel: If it's for the typechecker then it should *never* occur that `_app_client` is `None`. Also, why do we need it to be a field instead of just using a local variable?

Wiezzel: And, actually, this is being called in a wrong place, I believe. We don't need shutdown if the computation ended successfully, right?

maaktweluit: It used to be called from 3 places, but after some refactoring iti s only 2, `compute` and `wait_for_compute`.

Since one calls the other we can make it a local variable now :)

golem/task/task_api/__init__.py
```diff
@@ -47,6 +53,15 @@ def start(self, command: str, port: int) -> Tuple[str, int]:
         sync_wait(self._runtime.start())
         return self._runtime.get_port_mapping(port)
 
+    def running(self) -> bool:
```
Wiezzel, line 56: I assumes this means something like 'not stopped yet' since `CREATED` Runtime could hardly be considered as 'running'. Maybe a small comment could be placed here to make it explicit.

golem/task/taskcomputer.py
```diff
@@ -361,7 +359,8 @@ def _wait_until_computation_ends(self) -> defer.Deferred:
             ProviderTimer.finish()
             self._computation = None
             self._assigned_task = None
-            self._runtime = None
+            if not success:
+                yield app_client.shutdown()
```
Wiezzel, line 375: This may raise errors. I'm thinking maybe it should be called after `task_finished_callback` so we don't miss the callback in case something goes wrong?

golem/task/task_api/__init__.py
maaktweluit: yes, more like "not stopped yet" it is used to check if shutdown needs to be called or not.
During created it has some things to clean up so shutdown is required

With grpclib i found they name it `closable` maybe that is a better name here too but that would be for next task-api release

golem/task/taskcomputer.py
maaktweluit: was thinking about try/catching it again but i like your proposal more :D

golem/task/task_api/__init__.py
maaktweluit: @Wiezzel please resolve this comment if you approve of the docs i have added

Wiezzel (approved):

mfranciszkiewicz (approved):

PR #4565, opened 2019-07-31
Remove unused ACLs

mplebanski:


mplebanski:
IIRC `disallow_node_timeout_seconds` and `disallow_ip_timeout_seconds` are never set to something else then `None`, thus this is dead code. 

Also `should_accept_client` is keeping track of who's rejected and a record of that is updated in `_mark_subtask_failed` in CoreTask. I opt with apps having control over who is rejected and for what.

Krigpl:
This is something you can set from the `app_config` so I doubt this is a deadcode.
Leaving that to @jiivan

PR #4567, opened 2019-07-31
Skipping disable_concent flag for subtask restarts on active tasks

kmazurek:
Currently it's possible to skip checking for Concent deposit when restarting a range of subtasks from an active task. This has been changed so that the deposit check always depends on the original Concent settings in the task's definition.


etam (approved):
ok

jiivan (commented):
LGTM, two remarks

1. Maybe it's too late since this rpc is already used by electron, but I generally prefer positive flags. I.e. `concent_enabled` instead of `disable_concent`.
1. I would love to see a test checking this controversial condition `task_state.status.is_active() == True; disable_concent=True; task.header.concent_enabled=True`.

PR #4569, opened 2019-08-01
Fix VbR should_accept_client

### LINKED ISSUE #4568: `should_accept_client` prematurely committing providers in gWasm app
Since `should_accept_client` method of `taskbase.py:Task` is called both before and after offer sorting the gWasm `should_accept_client` implementation should not commit providers to subtasks just yet, the entire commitment logic should be moved somewhere else.

mplebanski:
Closes #4568


mbenke (approved):
Resolves the problem of some offers being prematurely dropped

PR #4574, opened 2019-08-01
[P2] Allow transcoding to any resolution with the same aspect ratio

cameel:
This pull request contains minor adjustments to transcoding integration tests required to make them work after https://github.com/golemfactory/ffmpeg-tools/pull/11. In particular the removed `list_matching_resolutions()` function is not longer used.

### Dependencies
This pull request is based on #4573 and should be merged after it. This is just for ease or development - It does not really depend on that base pull request and could be merged without it if rebased.

For the code to work you also need a `golemfactory/ffmpeg-experimental` image built with https://github.com/golemfactory/ffmpeg-tools/pull/11. Tests in the CI won't pass until it's released on DockerHub.

**NOTE**: I have set the base branch of this pull request to `transcoding-ignore-some-parameters-for-problematic-files` (#4573) rather than `CGI/transcoding/master`. Please remember to change the base branch back to `CGI/transcoding/master` before you merge.

PR #4575, opened 2019-08-01
[P6a] Audio channel layout validations

cameel:
A simple change required to make transcoding validations work after https://github.com/golemfactory/ffmpeg-tools/pull/12. `validate_transcoding_params()` now requires access to video metadata.

### Dependencies
This pull request is based on #4574 and should be merged after it. This is just for ease or development - It does not really depend on that base pull request and could be merged without it if rebased.

For the code to work you also need a `golemfactory/ffmpeg-experimental` image built with https://github.com/golemfactory/ffmpeg-tools/pull/12. Tests in the CI won't pass until it's released on DockerHub.

**NOTE**: I have set the base branch of this pull request to `transcoding-allow-transcoding-to-any-resolution-with-same-aspect-ratio` (#4574) rather than `CGI/transcoding/master`. Please remember to change the base branch back to `CGI/transcoding/master` before you merge.

### DIFF
--- a/apps/transcoding/task.py
+++ b/apps/transcoding/task.py
@@ -133,7 +133,11 @@ class TranscodingTask(CoreTask):  # pylint: disable=too-many-instance-attributes
             # matter which we choose.
             dst_params = self._get_extra_data(0)["targs"]
 
-            validation.validate_transcoding_params(src_params, dst_params)
+            validation.validate_transcoding_params(
+                src_params,
+                dst_params,
+                video_metadata
+            )
 
         except validation.InvalidVideo as e:
             logger.error(e.response_message)
--- a/apps/transcoding/task.py
+++ b/apps/transcoding/task.py
@@ -121,20 +121,12 @@ class TranscodingTask(CoreTask):  # pylint: disable=too-many-instance-attributes
         try:
             validation.validate_video(video_metadata)
 
-            src_params = meta.create_params(
-                meta.get_format(video_metadata),
-                meta.get_resolution(video_metadata),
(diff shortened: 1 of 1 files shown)


apps/transcoding/task.py
```diff
@@ -133,7 +133,11 @@ def initialize(self, dir_manager: DirManager):
             # matter which we choose.
             dst_params = self._get_extra_data(0)["targs"]
 
-            validation.validate_transcoding_params(src_params, dst_params)
+            validation.validate_transcoding_params(
+                src_params,
+                dst_params,
+                video_metadata
+            )
```
nieznanysprawiciel: If we require video_metadata to validate transcoding, we should remove src_params parameter from this function. All these parameters are created from metadata anyway.

nieznanysprawiciel (changes requested):

apps/transcoding/task.py
cameel: OK. Good point. Would it be fine to just move the `meta.create_params()` call into `validate_transcoding_params()` or would you prefer to go even further and not create this dict at all and make `validate_transcoding_params()` always get values directly from `video_metadata`?

nieznanysprawiciel: Moving meta.create_params seems to be the easiest way

pawelkisielewicz: Fixed. Please see also changes in ffmpeg-tools, it was required because of this change:
https://github.com/golemfactory/ffmpeg-tools/pull/12

nieznanysprawiciel:

PR #4576, opened 2019-08-02
Correct linter name in summary

jiivan:
In linter summary, flake8 is wrongly labelled as pycodestyle.


kmazurek (approved):
Thanks for fixing the test!

Issue #4481, opened 2019-07-16
Adapt HyperdriveClient to simple-transfer error reporting


kmazurek:
## Rationale
With the release of Golem 0.20, [hyperg](https://github.com/golemfactory/golem-hyperdrive) was replaced with [simple-transfer](https://github.com/golemfactory/simple-transfer). While the API for both is identical, `simple-transfer` has changed the way errors are reported. Golem's `HyperdriveClient` needs to be adapted to handle these errors correctly.

### Technical specification
Generally, `simple-transfer` reports errors back to the client  using the appropriate HTTP status codes. This means that for general failures a `500 Internal Server Error` will be returned. Similarly, for invalid requests the server will respond with a `400 Bad Request` status.

In particular, when getting a non-existent resource (via `/api/download` command), `simple-transfer` will return the HTTP status code 500 in a fail-fast manner. This is in contrast with the previous implementation which would hang indefinitely (or until timeout is reached) in such a situation.

#### Which application modules will be updated?
`HyperdriveClient`, `HyperdriveAsyncClient` (`golem/network/hyperdrive/client.py`)

#### Additional tests
Hyperdrive client tests (`tests/golem/network/hyperdrive/test_hyperdrive_client.py`) will need to be updated to include more specific error cases.

shadeofblue:
irrelevant

PR #4577, opened 2019-08-02
[CLAY-399] hyperg: unify shared resource lifetimes

mfranciszkiewicz:
~**Do not merge**~
**Requires `simple-transfer` v0.3.5+ ~(not released)~**

Adds support for the 2 new API calls:
- GET `/resources`
    Response:
    ```
    {
        "hash": hash,
        "files": n_files,
        "totalSize": size,
        "validTo": valid_to
    }
    ```
- GET `/resources/{resource_id}`
    Response:
    ```
    {
        "hash": hash,
        "files": files,
        "totalSize": size,
        "validTo": valid_to
    }
    ```

The latter will be used for validating the lifetimes of cache entries in the new `ResourceManager`.

### DIFF
--- a/golem/network/hyperdrive/client.py
+++ b/golem/network/hyperdrive/client.py
@@ -2,17 +2,19 @@ import json
 import logging
 import math
 from ipaddress import AddressValueError, ip_address
-from typing import Optional, Dict, Tuple, List, Iterable, Callable, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 import collections
 
 import requests
 from requests import HTTPError
-from twisted.internet.defer import Deferred
+from twisted.internet.defer import inlineCallbacks
 
 from golem_messages import helpers as msg_helpers
+from twisted.web.client import readBody
+from twisted.web.http_headers import Headers
 
-from golem.core import golem_async
+from golem.core.golem_async import AsyncHTTPRequest
 from golem.resource.client import IClient, ClientOptions
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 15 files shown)


golem/network/hyperdrive/client.py
```diff
@@ -126,9 +124,16 @@ def cancel(self, content_hash):
...
-    def _request(self, **data):
-        response = requests.post(url=self._url,
-                                 headers=self._headers,
+    def _request(
+            self,
+            endpoint: str = DEFAULT_ENDPOINT,
+            **data
+    ) -> Dict:
+        if endpoint and endpoint[0] == '/':
+            endpoint = endpoint[1:]
```
Wiezzel, line 131: AFAIK multiple slashes are allowed in URLs and shouldn't cause problems.

golem/network/hyperdrive/client.py
```diff
@@ -142,110 +147,128 @@ def _request(self, **data):
...
+            self,
+            endpoint: str = HyperdriveClient.DEFAULT_ENDPOINT,
+            method: bytes = b'POST',
+            params: Optional[Dict] = None,
+            parser: Optional[Callable[[Dict], Any]] = None,
+    ):
+        body = None
+
+        if endpoint and endpoint[0] == '/':
+            endpoint = endpoint[1:]
```
Wiezzel, line 245: AFAIK multiple slashes are allowed in URLs and shouldn't cause problems.

Wiezzel (approved):

golem/network/hyperdrive/client.py
```diff
@@ -37,27 +39,23 @@ class HyperdriveClient(IClient):
...
-        super(HyperdriveClient, self).__init__()
-
-        # API destination address
-        self.host = host
-        self.port = port
+        super().__init__()
         # connection / read timeout
         self.timeout = timeout
-
         # default POST request headers
```
maaktweluit: This comment is outdated now, think it can be removed
```suggestion
        # API destination address
```

golem/network/hyperdrive/client.py
```diff
@@ -142,110 +147,128 @@ def _request(self, **data):
...
+        if endpoint and endpoint[0] == '/':
+            endpoint = endpoint[1:]
+        if params:
+            body = json.dumps(params).encode(self.ENCODING)
+
+        uri = f'{self._url}/{endpoint}'.encode(self.ENCODING)
+        response = yield AsyncHTTPRequest.run(
+            method,
+            uri=uri,
+            headers=self.RAW_HEADERS,
```
maaktweluit, line 253: What is the difference between `RAW_HEADERS` and `HEADERS`?
```suggestion
            headers=self.HEADERS,
```

maaktweluit (commented):
LGTM!

Some small comments, will approve on reply

golem/network/hyperdrive/client.py
mfranciszkiewicz: `RAW_HEADERS` are an instance of `Headers` class, which serializes the given dict to proper binary data. `Headers` instance is an argument to Twisted's HTTP request function.

golem/network/hyperdrive/client.py
```diff
@@ -142,110 +147,128 @@ def _request(self, **data):
...
+        )
 
-        def on_error(body):
-            try:
-                decoded = body.decode('utf-8')
-            except Exception as exc:  # pylint: disable=broad-except
-                _result.errback(exc)
-            else:
-                _result.errback(HTTPError(decoded))
+        try:
```
kmazurek, line 257: In relation to #4481: I think it would make sense to check to status code of the response at this point, raising an error if necessary. The error could be a custom class including the status code so that it gets propagated to the caller. This would allow for error disambiguation once `simple-transfer` returns more specific status codes.

mfranciszkiewicz: I believe this should be covered in a follow up PR.

PR #4578, opened 2019-08-02
[CLAY-399] hyperg: obligatory share timeouts [2/3]

mfranciszkiewicz:
Enforces timeouts on each shared resource. Golem is in control over resource lifetimes and no longer relies on defaults set in `simple-transfer` (3 days).

### DIFF
--- a/golem/network/hyperdrive/client.py
+++ b/golem/network/hyperdrive/client.py
@@ -24,12 +24,10 @@ def to_hyperg_peer(host: str, port: int) -> Dict[str, Tuple[str, int]]:
     return {'TCP': (host, port)}
 
 
-def round_timeout(value: Optional[Union[int, float]]) -> Optional[int]:
-    if not isinstance(value, (int, float)):
-        return None
-
-    value_int = int(math.ceil(value))
-    return value_int if value_int > 0 else None
+def round_timeout(value: Union[int, float]) -> int:
+    if not isinstance(value, (int, float)) or value <= 0:
+        raise ValueError(f"Invalid timeout: {value}")
+    return int(math.ceil(value))
 
 
 class HyperdriveClient(IClient):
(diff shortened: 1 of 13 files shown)


golem/network/hyperdrive/client.py
```diff
@@ -24,12 +24,10 @@ def to_hyperg_peer(host: str, port: int) -> Dict[str, Tuple[str, int]]:
...
 
-def round_timeout(value: Optional[Union[int, float]]) -> Optional[int]:
-    if not isinstance(value, (int, float)):
-        return None
-
-    value_int = int(math.ceil(value))
-    return value_int if value_int > 0 else None
+def round_timeout(value: Optional[Union[int, float]]) -> int:
+    if not isinstance(value, (int, float)) or value <= 0:
+        raise ValueError(f"Invalid timeout: {value}")
```
Wiezzel, line 31: If it's not allowed to be `None` then it shouldn't be `Optional`. Also, type-checking with `if`s shouldn't be a thing. That's what mypy is for.

golem/network/hyperdrive/client.py
```diff
@@ -70,22 +68,20 @@ def addresses(self):
         return addresses
 
     def add(self, files, client_options=None, **kwargs):
-        timeout = client_options.timeout if client_options else None
         response = self._request(
             command='upload',
             id=kwargs.get('id'),
             files=files,
-            timeout=round_timeout(timeout)
+            timeout=round_timeout(client_options.timeout)
```
Wiezzel, line 77: If `client_options` are expected to be non-`None` please update the function signature.

golem/network/hyperdrive/client.py
```diff
@@ -70,22 +68,20 @@ def addresses(self):
...
         return response['hash']
 
     def restore(self, content_hash, client_options=None, **kwargs):
-        timeout = client_options.timeout if client_options else None
         response = self._request(
             command='upload',
             id=kwargs.get('id'),
             hash=content_hash,
-            timeout=round_timeout(timeout)
+            timeout=round_timeout(client_options.timeout)
```
Wiezzel, line 86: If `client_options` are expected to be non-`None` please update the function signature.

golem/network/hyperdrive/client.py
mfranciszkiewicz: This function is shared with HyperdriveClient and the old HyperdriveResourceManager. The signature will be updated when those entities are removed.

golem/network/hyperdrive/client.py
mfranciszkiewicz: Same as above

golem/network/hyperdrive/client.py
mfranciszkiewicz: Same as above

golem/network/hyperdrive/client.py
maaktweluit: maybe we can add a TODO to remember out future selves

maaktweluit (approved):
LGTM!

PR #4592, opened 2019-08-08
Auto-setup wrapper for environments

Wiezzel:
A decorator that automagically prepares and cleans up environment when it's necessary. That means in particular:
  * Running benchmarks
  * Installing prerequisites
  * Using runtimes (runtime usage starts with calling `prepare()` and ends with calling `clean_up()`, creation is not included)

Some comments:
  * `DockerCPUEnvironment` is not ready to use this decorator because runtime creation assumes the environment is enabled. This however could be quite easily refactored.
  * The wrapper currently assumes 'infinite parallel factor' which means it can handle running benchmark, installing prerequisites, and running arbitrary number of runtimes, all at once. However, with small adjustment we could make the wrapper enforce exclusion between these (if needed).

### DIFF
--- a/golem/envs/auto_setup.py
+++ b/golem/envs/auto_setup.py
@@ -1,0 +1,195 @@
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Optional,
+    Tuple,
+    Type
+)
+
+from twisted.internet.defer import (
+    Deferred,
+    DeferredLock,
+    inlineCallbacks
+)
+
+from golem.envs import (
+    CounterId,
+    CounterUsage,
+    EnvConfig,
+    EnvEventListener,
+    EnvEventType,
+    Environment,
+    EnvMetadata,
+    EnvStatus,
+    EnvSupportStatus,
+    Prerequisites,
+    Runtime,
+    RuntimeEventListener,
+    RuntimeEventType,
+    RuntimeInput,
+    RuntimeOutput,
+    RuntimePayload,
+    RuntimeStatus
+)
+
+
+class RuntimeSetupWrapper(Runtime):
+
+    def __init__(
+            self,
+            runtime: Runtime,
+            start_usage: Callable[[], Deferred],
+            end_usage: Callable[[], Deferred]
+    ) -> None:
+        self._runtime = runtime
+        self._start_usage = start_usage
+        self._end_usage = end_usage
+
+    @inlineCallbacks
+    def prepare(self) -> Deferred:
+        yield self._start_usage()
+        yield self._runtime.prepare()
+
+    @inlineCallbacks
+    def clean_up(self) -> Deferred:
+        yield self._runtime.clean_up()
+        yield self._end_usage()
+
+    def start(self) -> Deferred:
+        return self._runtime.start()
+
+    def wait_until_stopped(self) -> Deferred:
+        return self._runtime.wait_until_stopped()
+
+    def stop(self) -> Deferred:
+        return self._runtime.stop()
+
+    def status(self) -> RuntimeStatus:
+        return self._runtime.status()
+
+    def stdin(self, encoding: Optional[str] = None) -> RuntimeInput:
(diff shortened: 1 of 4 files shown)


golem/envs/auto_setup.py
```diff
@@ -0,0 +1,195 @@
...
+        def __init__(self, *args, **kwargs) -> None:
+            self._env = wrapped(*args, **kwargs)  # type: ignore
+            self._num_users = 0
+            self._lock = DeferredLock()
+
+        @inlineCallbacks
+        def _start_usage(self) -> Deferred:
+            yield self._lock.acquire()
+            try:
+                if self._num_users < 1:
```
maaktweluit: i think its more clear to check equal to 0, any reason to use < 1?
```suggestion
                if self._num_users == 0:
```

Wiezzel: No particular reason. Can change.

tests/golem/envs/test_auto_setup.py
```diff
@@ -0,0 +1,214 @@
...
+    def test_supported(self):
+        self.assertEqual(self.wrapped_cls.supported(), self.env_cls.supported())
+
+    def test_metadata(self):
+        self.assertEqual(self.wrapped_cls.metadata(), self.env_cls.metadata())
+
+    def test_parse_prerequisites(self):
+        prereq_dict = {'key': 'value'}
+        self.assertEqual(
+            self.wrapped_cls.parse_prerequisites(prereq_dict),
```
maaktweluit: one line one thought, would call:
```
result = self.wrapped_cls.parse_prerequisites(prereq_dict)
assert ...
```

This will also separate `Given`, `When` and `Then`

Wiezzel: OK

tests/golem/envs/test_auto_setup.py
```diff
@@ -0,0 +1,214 @@
...
+        payload = Mock(spec_set=RuntimePayload)
+        config = Mock(spec_set=EnvConfig)
+        runtime = self.wrapped_env.runtime(payload, config)
+        self.assertEqual(runtime.status(), self.runtime.status())
+
+    def test_runtime_stdin(self):
+        payload = Mock(spec_set=RuntimePayload)
+        config = Mock(spec_set=EnvConfig)
+        runtime = self.wrapped_env.runtime(payload, config)
+        stdin = runtime.stdin('utf-7')
```
maaktweluit, line 168: lol, mocked encoding? ^^

golem/envs/auto_setup.py
maaktweluit: i dont mind, both works, was just a thought to make it more readable, only change if you agree

tests/golem/envs/test_auto_setup.py
Wiezzel: Well, 'utf-8' might become a default value some time, so using something weird like 'utf-7' makes a better test.

maaktweluit: :) i like it

golem/envs/auto_setup.py
```diff
@@ -0,0 +1,195 @@
...
+
+        @classmethod
+        def supported(cls) -> EnvSupportStatus:
+            return wrapped.supported()
+
+        def status(self) -> EnvStatus:
+            return self._env.status()
+
+        def prepare(self) -> Deferred:
+            raise AttributeError('prepare and clean_up not supported')
```
mfranciszkiewicz, line 131: AFAICT these implementations of `prepare` and `clean_up` make the wrapper incompatible with `EnvironmentTaskApiService`; they'll cause the `start` method to fail. Please provide more insight on this.

Wiezzel: `EnvironmentTaskApiService` doesn't call any of these methods. It calls `Runtime.start()` but that's a different story.

mfranciszkiewicz (approved):

PR #4596, opened 2019-08-08
Localhost envrionment

Wiezzel:
A simple test environment capable of running GRPC service on localhost. Both requestor and provider services are supported. RPC methods return predefined values specified in prerequisites.

### DIFF
--- a/tests/golem/envs/localhost.py
+++ b/tests/golem/envs/localhost.py
@@ -1,0 +1,276 @@
+import asyncio
+import logging
+from pathlib import Path
+from threading import Thread
+from typing import Optional, Dict, Any, Tuple, List
+
+from dataclasses import dataclass, asdict, field
+from golem_task_api import RequestorAppHandler, ProviderAppHandler, entrypoint
+from golem_task_api.server import AppServer
+from golem_task_api.structs import Subtask
+from twisted.internet import defer, threads
+
+from golem.envs import (
+    CounterId,
+    CounterUsage,
+    EnvConfig,
+    EnvId,
+    EnvironmentBase,
+    EnvMetadata,
+    EnvSupportStatus,
+    Prerequisites,
+    Runtime,
+    RuntimeBase,
+    RuntimeInput,
+    RuntimeOutput,
+    RuntimePayload
+)
+from golem.task.task_api import TaskApiPayloadBuilder
+
+logger = logging.getLogger(__name__)
+
+
+class LocalhostConfig(EnvConfig):
+
+    def to_dict(self) -> dict:
+        return {}
+
+    @staticmethod
+    def from_dict(data: dict) -> 'LocalhostConfig':
+        return LocalhostConfig()
+
+
+@dataclass
+class LocalhostPrerequisites(Prerequisites):
+    compute_results: Dict[str, str] = field(default_factory=dict)
+    benchmark_result: float = 0.0
+    subtasks: List[Subtask] = field(default_factory=list)
+    verify_results: Dict[str, Tuple[bool, Optional[str]]] \
+        = field(default_factory=dict)
+
+    def to_dict(self) -> dict:
+        return asdict(self)
+
+    @staticmethod
+    def from_dict(data: dict) -> 'LocalhostPrerequisites':
+        return LocalhostPrerequisites(**data)
+
+
+@dataclass
+class LocalhostPayload(RuntimePayload):
+    command: str
+    shared_dir: Path
+    prerequisites: LocalhostPrerequisites
+
+
+class LocalhostPayloadBuilder(TaskApiPayloadBuilder):
+
+    @classmethod
+    def create_payload(
+            cls,
+            prereq: Prerequisites,
+            shared_dir: Path,
+            command: str,
+            port: int
+    ) -> RuntimePayload:
+        assert isinstance(prereq, LocalhostPrerequisites)
+        return LocalhostPayload(
+            command=command,
+            shared_dir=shared_dir,
+            prerequisites=prereq
+        )
+
+
+class LocalhostAppHandler(RequestorAppHandler, ProviderAppHandler):
+
+    def __init__(self, prereq: LocalhostPrerequisites) -> None:
+        self._compute_results = prereq.compute_results
+        self._benchmark_result = prereq.benchmark_result
+        self._subtasks = prereq.subtasks
+        self._verify_results = prereq.verify_results
+
+    async def create_task(
+            self,
+            task_work_dir: Path,
+            max_subtasks_count: int,
+            task_params: dict
+    ) -> None:
+        pass
+
+    async def next_subtask(self, task_work_dir: Path) -> Subtask:
+        return self._subtasks.pop(0)
+
+    async def verify(
+            self,
+            task_work_dir: Path,
+            subtask_id: str
+    ) -> Tuple[bool, Optional[str]]:
+        return self._verify_results[subtask_id]
+
+    async def discard_subtasks(
+            self,
+            task_work_dir: Path,
+            subtask_ids: List[str]
+    ) -> List[str]:
+        return []
+
+    async def run_benchmark(self, work_dir: Path) -> float:
+        return self._benchmark_result
+
+    async def has_pending_subtasks(self, task_work_dir: Path) -> bool:
+        return bool(self._subtasks)
+
+    async def compute(
+            self,
+            task_work_dir: Path,
+            subtask_id: str,
+            subtask_params: dict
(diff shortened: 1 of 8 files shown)


golem/task/task_api/local_grpc.py
```diff
@@ -0,0 +1,302 @@
...
+    def stderr(self, encoding: Optional[str] = None) -> RuntimeOutput:
+        raise NotImplementedError
+
+    def get_port_mapping(self, port: int) -> Tuple[str, int]:
+        return '127.0.0.1', port
+
+    def usage_counters(self) -> Dict[CounterId, CounterUsage]:
+        return {}
+
+    def call(self, alias: str, *args, **kwargs) -> defer.Deferred:
```
Krigpl: rm

golem/task/task_api/local_grpc.py
```diff
@@ -0,0 +1,302 @@
...
+        raise NotImplementedError
+
+
+class LocalGrpcEnvironment(Environment):
+
+    """ This environment is capable of spawning Task API services on localhost.
```
Krigpl: I'm not so sure about this approach. Environments and DockerEnvironment in particular are built in a Task API agnostic way - while this environment is not. Also `grpc` doesn't come up anywhere in the code as the dependency (since it's hidden in the task api) so I'd drop `grpc` in the name and use `HostEnvironment` or `LocalhostEnvironment` instead.

An alternative approach to implement this env would be to make prerequisites contain the async function to run in a separate asyncio loop (effectively similar to the `command` param in the DockerEnv). That'd make the environment task api agnostic and could be even placed under `golem/envs` as a general-purpose environment.

Wiezzel: 1. 👍 for renaming to `LocalhostEnvironment`.
2. How would make the prerequisites `DictSerializable` if they contained a function?

Krigpl: 2. Right, forgot about that. I guess it could be similar to docker's way -  we effectively identify the binary/function there as a docker image which we expect is available on the host and the command to run. We could have a predefined set of functions executable on the host and prerequisites would just contain the identifier. I dunno, might not be worth the trouble atm, but the current approach feels weird to me.

Wiezzel: Well, it might seem weird but keep in mind this code is just for testing. Predefined set of functions doesn't look like a good alternative to me. What kind of functions would that be to keep it non-Task API specific?
Another solution which I considered would be to put a fully qualified name of a function in the prerequisites and dynamically import it. But that's kinda hacky.

Krigpl: If it's just for testing and I'd be fine if we put that class under `tests/`.

Wiezzel: Maybe `golem/tools`? That's where we keep things like `TestWithReactor` or `TestWithDatabase`.

Krigpl: tbh I have no idea why the classes you mentioned aren't under `tests/` (imo they should).

Wiezzel: Maybe that's because they are not test code *per se* but something you use to write test code? Nvm, I can put my code in `tests/`.

golem/task/task_api/local_grpc.py
Wiezzel: That's dependent on #4592

tests/golem/envs/localhost.py
```diff
@@ -0,0 +1,303 @@
...
+    def create_payload(
+            cls,
+            prereq: Prerequisites,
+            shared_dir: Path,
+            command: str,
+            port: int
+    ) -> RuntimePayload:
+        assert isinstance(prereq, LocalhostPrerequisites)
+        # Parse the command e.g. 'provider 1234' -> 'PROVIDER'
+        service_type, _ = command.upper().split()
```
mfranciszkiewicz: You can ignore a variable count of the arguments with
```suggestion
        service_type, *_ = command.upper().split()
```

tests/golem/envs/localhost.py
```diff
@@ -0,0 +1,303 @@
...
+        return defer.succeed(None)
+
+    def stop(self) -> defer.Deferred:
+        assert self._server is not None
+        assert self._server_loop is not None
+        assert self._server_thread is not None
+        try:
+            self._server_loop.run_until_complete(self._server.stop())
+        except Exception:  # pylint: disable=broad-except
+            return defer.fail()
```
mfranciszkiewicz, line 183: It might be a good idea to fail with the exception raised

tests/golem/envs/localhost.py
```diff
@@ -0,0 +1,303 @@
...
+        return '127.0.0.1', port
+
+    def usage_counters(self) -> Dict[CounterId, CounterUsage]:
+        return {}
+
+    def call(self, alias: str, *args, **kwargs) -> defer.Deferred:
+        raise NotImplementedError
+
+
+class LocalhostEnvironment(EnvironmentBase):
```
mfranciszkiewicz, line 206: IMO test environments should live in a separate package / namespace for the clarity of its purpose. E.g. if / when we introduce a `HostDirectEnvironment`, the two living next to each other may cause unnecessary confusion.

tests/golem/envs/localhost.py
Wiezzel: I believe that's exactly what `fail()` does here. It should automatically capture current error.

tests/golem/envs/localhost.py
Wiezzel: Well, it's in `tests` directory, according to @Krigpl's suggestion. Doesn't it make clear that it's a test-only environment? If not, where do you suggest it should be put?

tests/golem/envs/localhost.py
mfranciszkiewicz: Unfortunately, no. It's a shortcut for
```
def fail(result=None):
    ...
    d = Deferred()
    d.errback(result)
    return d
```

Wiezzel: Well, it is. But `d.errback(None)` does it. 😉

Wiezzel: You can try it:
```python
class TestFailure(TwistedTestCase):

    @inlineCallbacks
    def test_failure(self):
        def _foo():
            try:
                raise OSError()
            except:
                return defer.fail()

        with self.assertRaises(OSError):
            yield _foo()
```

tests/golem/envs/localhost.py
mfranciszkiewicz: @Wiezzel Right - somehow I missed the `tests` prefix. Sorry for the confusion.

mfranciszkiewicz (approved):

tests/golem/envs/localhost.py
```diff
@@ -0,0 +1,303 @@
...
+                port=self._port,
+                handler=self._app_handler
+            )
+        else:
+            self._server = RequestorAppServer(
+                work_dir=self._work_dir,
+                port=self._port,
+                handler=self._app_handler
+            )
+        await self._server.start()
```
Krigpl: What's wrong with reusing the `entrypoint` logic from the Task API ?

Wiezzel: Nothing's wrong, I think. Might do that.

tests/golem/envs/localhost.py
```diff
@@ -0,0 +1,287 @@
...
+    def create_payload(
+            cls,
+            prereq: Prerequisites,
+            shared_dir: Path,
+            command: str,
+            port: int
+    ) -> RuntimePayload:
+        assert isinstance(prereq, LocalhostPrerequisites)
+        # Parse the command e.g. 'provider 1234' -> 'PROVIDER'
+        service_type, _ = command.upper().split()
```
Krigpl: I think you don't need this whole service type logic anymore, just store the command and pass it to the entrypoint. Which is a good thing anyway since there's no need to inspect command's internals.

Wiezzel: Quite indeed.

tests/golem/envs/localhost.py
```diff
@@ -0,0 +1,279 @@
...
+
+class LocalhostRuntime(RuntimeBase):
+
+    def __init__(
+            self,
+            payload: LocalhostPayload,
+    ) -> None:
+        super().__init__(logger)
+        self._command = payload.command
+        self._port = payload.port
```
Krigpl: Not used anymore

Krigpl (approved):

PR #4599, opened 2019-08-09
Introduce ProviderMarketStrategy

mplebanski:
Introduce ProviderMarketStrategy (and its subclasses) to allow provider to have specific market strategy for every task type.

This PR is part of the usage market epic.


mbenke (approved):
LGTM

apps/wasm/task.py
```diff
@@ -28,7 +29,8 @@
 from apps.core.task.coretaskstate import Options, TaskDefinition
 from apps.wasm.environment import WasmTaskEnvironment
 from golem.marketplace.wasm_marketplace import (
-    RequestorWasmMarketStrategy, UsageReport
+    RequestorWasmMarketStrategy, UsageReport,
```
maaktweluit: one import per line according to the style guide ( when they do not fit on a single one )

golem/marketplace/__init__.py
```diff
@@ -1,7 +1,18 @@
 # pylint: disable=unused-import
 from .marketplace import Offer, ProviderPerformance  # noqa
-from .marketplace import RequestorMarketStrategy  # noqa
-from .brass_marketplace import RequestorBrassMarketStrategy  # noqa
-from .wasm_marketplace import RequestorWasmMarketStrategy  # noqa
+from .marketplace import (  # noqa
```
maaktweluit, line 2: you can add `Offer` and `ProviderPerformance` to this block

golem/marketplace/brass_marketplace.py
```diff
@@ -52,8 +53,34 @@ def resolve_task_offers(cls, task_id: str) -> Optional[List[Offer]]:
...
         def payment_computer(price: int):
             return price * task.header.subtask_timeout
+
         return payment_computer
+
+
+class ProviderBrassMarketStrategy(ProviderMarketStrategy):
+
+    @classmethod
+    def calculate_price(cls, pricing: ProviderPricing, max_price: int,
```
maaktweluit: linebreaks

golem/marketplace/wasm_marketplace.py
```diff
@@ -1,13 +1,17 @@
 import logging
-from typing import (
-    Callable, List, Dict, ClassVar, Tuple, Optional,
-    Iterable, TYPE_CHECKING
-)
+from typing import (Callable, List, Dict, ClassVar, Tuple, Optional, Iterable,
```
maaktweluit, line 2: this is OK, typing is exempt from the one var per line rule ( just FYI )

tests/golem/marketplace/test_marketplace.py
```diff
@@ -104,12 +103,20 @@ class TestRequestorBrassMarketStrategy(TestCase):
 
     @staticmethod
     def _mock_offer():
-        return Offer(
-            provider_id='provider_1',
-            provider_performance=ProviderPerformance(100),
-            max_price=5000,
-            price=5000
-        )
+        mock_offer = MagicMock(spec_set=[
```
maaktweluit, line 106: can we use `spec=TYPE` here?

maaktweluit (approved):
LGTM!

Few style comments ( that we should start catching with the linter ).
Logic seems good, so approving :)

apps/wasm/task.py
mbenke: done

golem/marketplace/__init__.py
mbenke: done

golem/marketplace/brass_marketplace.py
mbenke: done

tests/golem/marketplace/test_marketplace.py
mbenke: It doesn't work  for me, but I am no pytest expert. Leaving it as is, if you find a way to make it work, be my guest

PR #4584, opened 2019-08-05
Cgi/verifier/scripts for addional tests

nieznanysprawiciel:
### This PR introduces following changes:
1. Scripts for extended tests of blender verifier
2. Semi integrations tests utilities
3. Tests reproducing some known bugs in blender verifier

### Detailed description:
1. Scripts are checking different subtasks splits on predefined checker scene
    * Tests check if verifier computes crops coordinates correctly and if crops are correctly matched with verified image region.
   * Tests don't use verifier verdict, but instead compare produced crops pixel by pixel with reference crops. This bypasses problem with not ideal verification quality.
   * Scripts execution lasts very long, so they are not included in standard test set. They will be ran once and all discovered bugs will be reproduced using semi-integrations tests. This means that scripts don't affect Golem in any way.

2. Semi integrations tests utilities:
    * Test utilities help write tests running full application flow (I mean golem integrations from app directory).
    * Utils use TaskManager class to emulate golem behavior as close as possible.
    * Tests don't need network and external nodes. Providers are mocked on local computer.
    * keep_testdir_on_fail decorator was added to TempDirFixture, to easier check content of test temporary directories, when debugging.
    * These utils are used by transcoding branches as well, so merging them will help us to easier merge to develop in the future.

3. Tests reproducing some known bugs in blender verifier
    * Test reproducing bugs found so far by verifier scripts.
    * Reproduced bugs are already fixed on develop, this PR adds only reproduction that uses semi-integration tests utils.

PR #4601, opened 2019-08-09
Reproduction of bug in reputation system - database LocalRank creation

nieznanysprawiciel:
This PR adds test that reproduces bug related to LocalRank database entry creation.

# Description:
Creation of LocalRank database object should fill it with default value for provider_efficacy. But this default value changes if someone calls update_provider_efficacy function in database_manager.py. All new database entries are created with updated value.

It seems that all new objects are referencing the same provider_efficacy object. Probably this is caused by database that makes only shallow copy of default value (peewee.py line 4764).

# Workaround:
This bug was originally discovered in https://github.com/golemfactory/golem/pull/4584. There were dependencies between tests that caused tests/golem/test_model.py:TestLocalRank:test_default_fields test fails. TestBlenderIntegration tested whole workflow of blender task and called provider reputation updates functions. Because these test cases were executed before test_model.py module, database default value was already changed.

To avoid this issue, I changed update_provider_efficacy function in database_manager.py (line 233), to make deepcopy of provider_efficacy field.
But real solution to this problem probably requires changes in database itself.

### DIFF
--- a/golem/model.py
+++ b/golem/model.py
@@ -397,6 +397,14 @@ class TaskPayment(BaseModel):
 ##################
 
 
+def provider_efficacy_producer():
+    def producer():
+        return ProviderEfficacy(0., 0., 0., 0.)
+    # peewee-migrate expects a '__self__' attribute
+    producer.__self__ = ProviderEfficacy
+    return producer
+
+
 class LocalRank(BaseModel):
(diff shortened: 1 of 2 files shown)


mfranciszkiewicz:
PR solving the issue: https://github.com/golemfactory/golem/pull/4606

magdasta (approved):

PR #4605, opened 2019-08-09
Add requestor stats RPC endpoint

### LINKED ISSUE #4582: Add an RPC endpoint with requestor stats
## Rationale
As part of the settings screen revamp in `golem-electron` a new section containing requestor statistics will be added. Currently, requestor stats are collected and sent to the monitor service, but are not available through an RPC endpoint. Therefore, in order for client apps to have access to them, a new endpoint needs to be added.

## Description
At the moment, the endpoint is expected to return the following statistics about tasks:
- number of tasks requested by the node currently in the network,
- total number of requested tasks which were successfully computed,
- total number of requested tasks which failed.

Also, the result of the RPC call should include a breakdown of failure reasons for requested subtasks. There are three subtask statuses which are of interest and their counts should be returned:
- timeout,
- verification failure,
- other error.

Currently, an RPC endpoint with provider statistics is exposed under `comp.tasks.stats`. The requestor stats endpoint should be placed in the same namespace, for example: `comp.tasks.requestor_stats`.

kmazurek:
Resolves: #4582

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -902,7 +902,32 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
                                                               single=single)
 
     @rpc_utils.expose('comp.tasks.stats')
-    def get_task_stats(self) -> Dict[str, Any]:
+    def get_task_stats(
+            self,
+            stats_type: Optional[str] = 'provider'
+    ) -> Dict[str, Any]:
+        class StatsType(msg_datastructures.StringEnum):
+            provider = enum.auto()
+            requestor = enum.auto()
+
+        if stats_type:
+            try:
+                stats_type = StatsType(stats_type)
+            except ValueError:
+                logger.error('Invalid stats type: %r', stats_type)
+                return {}
+
+        if stats_type == StatsType.provider:
+            return self.get_provider_stats()
+        elif stats_type == StatsType.requestor:
+            return self.get_requestor_stats()
+
+        return {
+            'provider': self.get_provider_stats(),
+            'requestor': self.get_requestor_stats(),
+        }
+
+    def get_provider_stats(self) -> Dict[str, Any]:
         return {
             'provider_state': self.get_provider_status(),
             'in_network': self.get_task_count(),
@@ -914,7 +939,6 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             'subtasks_with_timeout': self.get_comp_stat('tasks_with_timeout'),
         }
 
-    @rpc_utils.expose('comp.tasks.requestor_stats')
     def get_requestor_stats(self) -> Dict[str, Any]:
         tm: TaskManager = self.task_manager
         tasks_in_network = len(tm.get_tasks_headers())
--- a/golem/client.py
+++ b/golem/client.py
@@ -1,6 +1,7 @@
 # pylint: disable=too-many-lines
 
 import collections
+import dataclasses
 import enum
 import logging
 import sys
@@ -916,11 +917,13 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
     @rpc_utils.expose('comp.tasks.requestor_stats')
     def get_requestor_stats(self) -> Dict[str, Any]:
         tm: TaskManager = self.task_manager
+        tasks_in_network = len(tm.get_tasks_headers())
         current_stats = tm.requestor_stats_manager.get_current_stats()
         finished_stats = tm.requestor_stats_manager.get_finished_stats()
         return {
-            'current': current_stats.asdict(),
-            'finished': finished_stats.asdict(),
+            'tasks_in_network_cnt': tasks_in_network,
+            'current_tasks_stats': dataclasses.asdict(current_stats),
+            'finished_tasks_stats': dataclasses.asdict(finished_stats),
         }
 
     def get_supported_task_count(self) -> int:
--- a/golem/client.py
+++ b/golem/client.py
@@ -919,8 +919,8 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
         current_stats = tm.requestor_stats_manager.get_current_stats()
         finished_stats = tm.requestor_stats_manager.get_finished_stats()
         return {
-            'current': current_stats._asdict(),
-            'finished': finished_stats._asdict(),
+            'current': current_stats.asdict(),
+            'finished': finished_stats.asdict(),
         }
 
     def get_supported_task_count(self) -> int:
--- a/golem/client.py
+++ b/golem/client.py
@@ -913,6 +913,16 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             'subtasks_with_timeout': self.get_comp_stat('tasks_with_timeout'),
         }
 
+    @rpc_utils.expose('comp.tasks.requestor_stats')
+    def get_requestor_stats(self) -> Dict[str, Any]:
+        tm: TaskManager = self.task_manager
+        current_stats = tm.requestor_stats_manager.get_current_stats()
+        finished_stats = tm.requestor_stats_manager.get_finished_stats()
+        return {
+            'current': current_stats._asdict(),
+            'finished': finished_stats._asdict(),
+        }
+
     def get_supported_task_count(self) -> int:
         if self.task_server:
             return len(self.task_server.task_keeper.supported_tasks)
--- a/golem/client.py
+++ b/golem/client.py
@@ -904,7 +904,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
     @rpc_utils.expose('comp.tasks.stats')
     def get_task_stats(
             self,
-            stats_type: Optional[str] = 'provider'
+            stats_type: Optional[str] = None
     ) -> Dict[str, Any]:
         class StatsType(msg_datastructures.StringEnum):
             provider = enum.auto()
--- a/golem/task/taskrequestorstats.py
+++ b/golem/task/taskrequestorstats.py
@@ -1,9 +1,9 @@
 import logging
 import time
+import typing
 from collections import defaultdict
 from dataclasses import dataclass, replace
 from threading import Lock
-from typing import NamedTuple, Optional
 
 from pydispatch import dispatcher
 
@@ -15,7 +15,7 @@ __all__ = ['RequestorTaskStatsManager']
 
 logger = logging.getLogger(__name__)
 
-TaskMsg = NamedTuple("TaskMsg", [("ts", float), ("op", Operation)])
+TaskMsg = typing.NamedTuple("TaskMsg", [("ts", float), ("op", Operation)])
 
 
 class SubtaskInfo:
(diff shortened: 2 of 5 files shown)


kmazurek:
In the current (draft) version the endpoint's return value has the following structure:
```
{
    "current": {
        "collected_results_cnt": 0,
        "failed_subtasks_cnt": 0,
        "finished_task_cnt": 0,
        "not_downloadable_subtasks_cnt": 0,
        "requested_subtasks_cnt": 0,
        "tasks_cnt": 0,
        "timed_out_subtasks_cnt": 0,
        "verified_results_cnt": 0,
        "work_offers_cnt": 0
    },
    "finished": {
        "failed": [
            0,
            0.0
        ],
        "finished_ok": [
            0,
            0.0
        ],
        "finished_with_failures": [
            0,
            0.0
        ]
    }
}
```

- the values in `finished` object refer to finished tasks, the tuples contain the count of tasks (int) and the total time spent (float) for each key,
- the tuples mentioned above will be replaced with objects with keys to indicate what the values are (this requires some refactoring of the stats models),
- to get the number of subtasks which failed at verification (from `current` object): `collected_results_cnt - verified_results_cnt`

kmazurek:
Here's the updated response structure (for requestor stats):
```
{
  "current_tasks_stats": {
    "collected_results_cnt": 0,
    "failed_subtasks_cnt": 0,
    "finished_task_cnt": 0,
    "not_downloadable_subtasks_cnt": 0,
    "requested_subtasks_cnt": 0,
    "tasks_cnt": 0,
    "timed_out_subtasks_cnt": 0,
    "verified_results_cnt": 0,
    "work_offers_cnt": 0
  },
  "finished_tasks_stats": {
    "failed": {
      "tasks_cnt": 0,
      "total_time": 0
    },
    "finished_ok": {
      "tasks_cnt": 0,
      "total_time": 0
    },
    "finished_with_failures": {
      "tasks_cnt": 0,
      "total_time": 0
    }
  },
  "tasks_in_network_cnt": 0
}
```

etam (approved):
ok

golem/task/taskrequestorstats.py
```diff
@@ -222,43 +223,39 @@ def in_progress_subtasks_count(self) -> int:
...
-    ("failed_subtasks_cnt", int),
-    ("work_offers_cnt", int)])
-CurrentStats.__doc__ = """Statistics about a set of tasks
-
-Intended to be used as a summary of information from a set of
-`TaskStats`, this is periodically sent to the monitor.
-"""
-
-EMPTY_CURRENT_STATS = CurrentStats(0, 0, 0, 0, 0, 0, 0, 0, 0)
+@dataclass
```
tworec, line 226: cool, #TIL such a python feature exists ;)

golem/client.py
```diff
@@ -901,7 +902,32 @@ def get_task_preview(self, task_id, single=False):
                                                               single=single)
 
     @rpc_utils.expose('comp.tasks.stats')
-    def get_task_stats(self) -> Dict[str, Any]:
+    def get_task_stats(
```
tworec, line 905: You are preserving API here to be back-compatible, but the result is not.

If you are breaking compatibility, then it's better to break it on the API level, to show API users that things have changed and they need to adjust client code.

btw. for now we are assuming that the only client for our RPC is our own electron, but there is golemcli and some of our active users were asking about RPC documentation, so in the future, we should be cautious about changes to RPC API

golem/task/taskrequestorstats.py
```diff
@@ -277,7 +274,7 @@ def update_current_stats_with_task(
     """
```
tworec: see 3 lines above
```
``current`` is a tuple
```
it is not any more

golem/task/taskrequestorstats.py
```diff
@@ -390,8 +386,8 @@ class RequestorTaskStats:
     def __init__(self):
         self.tasks = defaultdict(
             TaskInfo)  # type: DefaultDict[str, TaskInfo]
```
tworec: can you pls change this old-style type hint into a proper one?

tworec (changes requested):
First, pls notice that RPC changes affect not only Electron, but (a new) `golemcli` too. Are you volunteer to integrate this into it?

Second, do you plan it for `b0.20.1` as the tag in #4582 suggests?

Third, those names 
```
    "collected_results_cnt": 0,
    "failed_subtasks_cnt": 0,
    "finished_task_cnt": 0,
    "not_downloadable_subtasks_cnt": 0,
    "requested_subtasks_cnt": 0,
    "tasks_cnt": 0,
    "timed_out_subtasks_cnt": 0,
    "verified_results_cnt": 0,
    "work_offers_cnt": 0
```
aren't self-explanatory to me, even though I've spent much time reading it within the last 1,5 years. 
You can notice there are four types of objects: tasks, subtasks, results and work offers. What are their definitions? What is the result for: subtask or task? (I know the answer, but it is not implicit from field name)

Fourth, provider stats are in different form/shape/structure than requestor stats. This is highly misleading/surprising.

Fifth, requestor stats are not persisted on disk, and provider stats are stored within golem.db
Shouldn't we made both persisted to not present randomly changing numbers on the electron?

Sixth, see my note about API back compatibility.

Sorry for this overwhelming review comment, but if we want to do it, lets do it well, or do not do it at all.

golem/client.py
kmazurek: I agree we should have our RPC endpoints documented and reviewed, I'm planning to push for this after the 0.21 release. So far, RPC API changes are (or in some cases, should be) mentioned in release notes, which will be the case for this change.
Regarding this endpoint - what is your suggestion here? I think the parameter `stats_type` should remain optional, returning all stats by default.

kmazurek:
@tworec
1. I'm aware that this introduces breaking changes to any RPC consumers, including `golem-client`. I'm willing to update it to match the new API.
2. I believe that tag is no longer valid, it's already been updated.
3. Which of these field names do you consider unclear? What would be your recommendations for them? I maintained the existing field names since, as far as I can tell, they are being used by monitor and I didn't want to break compatibility on that front.
4. This task is not related to provider stats, therefore I left them as they were. Besides, the only possible change I can see here would be renaming the stats field to follow some common convention.
5. Good point, I think it's worth the time investment at this point to make requestor stats persistent as well.
6. See my answer in comments.

golem/task/taskrequestorstats.py
kmazurek: It's a new feature in Python 3.7, we're using a 3.6 backport library for it.

golem/client.py
tworec: I'd prefer to not change `comp.tasks.stats` endopoint. Let it return what is used to.

But if you want to upgrade naming and structure of the returned json (which is good direction) that is better to change the signature, so the clients are not able to call it in a way they used to... its better to break compatibility as early/fast as possible

tworec (commented):
> @tworec
> 
> 1. I'm aware that this introduces breaking changes to any RPC consumers, including `golem-client`. I'm willing to update it to match the new API.

Great

> 2. I believe that tag is no longer valid, it's already been updated.

This is most important thing. I was afraid that this is going to be included in next coming small release

> 3. Which of these field names do you consider unclear? What would be your recommendations for them? I maintained the existing field names since, as far as I can tell, they are being used by monitor and I didn't want to break compatibility on that front.
```
    "collected_results_cnt": 0,
```
is this related to results of tasks or subtasks?
```
    "not_downloadable_subtasks_cnt": 0,
```
this is rather about not downloadable **results** of subtasks, so the name is kinda misleading, because it does not indicate that subtask was successful, but result is not downloadable 
```
    "requested_subtasks_cnt": 0,
```
I do not remember what does this mean, but name suggests that this is all subtasks requestor has send in reponse to WTCTs, also counting failures, timeouts and retries, right?
```
    "tasks_cnt": 0,
```
requested or supported tasks?


> 4. This task is not related to provider stats, therefore I left them as they were.

The point is, you did not. They have changed too. And `golemcli` is not working with this branch.
```
ERROR [golem_rpc_api::rpc] on comp.tasks.stats unable to parse: [Object({"provider": Object({"provider_state": Object({"status": String("Not accepting tasks")}), "in_network": Number(0), "supported": Number(0), "subtasks_computed": Array([Number(0), Number(222)]), "subtasks_accepted": Array([Number(0), Number(0)]), "subtasks_rejected": Array([Number(0), Number(0)]), "subtasks_with_errors": Array([Number(0), Number(21)]), "subtasks_with_timeout": Array([Number(0), Number(7)])}), "requestor": Object({"tasks_in_network_cnt": Number(0), "current_tasks_stats": Object({"tasks_cnt": Number(0), "finished_task_cnt": Number(0), "requested_subtasks_cnt": Number(0), "collected_results_cnt": Number(0), "verified_results_cnt": Number(0), "timed_out_subtasks_cnt": Number(0), "not_downloadable_subtasks_cnt": Number(0), "failed_subtasks_cnt": Number(0), "work_offers_cnt": Number(0)}), "finished_tasks_stats": Object({"finished_ok": Object({"tasks_cnt": Number(0), "total_time": Number(0.0)}), "finished_with_failures": Object({"tasks_cnt": Number(0), "total_time": Number(0.0)}), "failed": Object({"tasks_cnt": Number(0), "total_time": Number(0.0)})})})})]: missing field `provider_state`

```
using golemcle release ver 0.1.9  
```
RUST_LOG=debug golemcli status
```


> Besides, the only possible change I can see here would be renaming the stats field to follow some common convention.

yup.

> 5. Good point, I think it's worth the time investment at this point to make requestor stats persistent as well.

yup.

> 6. See my answer in comments.

thx

kmazurek:
Okay, so for the next steps there are three main areas here.
#### 1. How should the API change?
As discussed offline with @tworec, I think all stats should be accessible via a single RPC endpoint (`comp.tasks.stats`). However, the stats type to be retrieved should be specified **explicitly** by the caller, and the RPC will only return one type of stats at a time. There will be no default value returned.
#### 2. How should the stats be structured?
As pointed out, some of the names for the stats fields are ambiguous and may lead to confusion. Given this, I'm going to consider renaming these fields and/or changing the structure of the responses.
#### 3. How should the stats be stored?
Similarly to point 2., this requires some more research. More specifically, I'm going to compare the provider and requestor stats and see whether the approach used for serialising provider stats can also be applied here.

tworec:
> Okay, so for the next steps there are three main areas here.
> 
> #### 1. How should the API change?
> As discussed offline with @tworec, I think all stats should be accessible via a single RPC endpoint (`comp.tasks.stats`). However, the stats type to be retrieved should be specified **explicitly** by the caller, 

so far, so good

> and the RPC will only return one type of stats at a time.

It would be optimal in sense of API calls to give the ability to list desired stats to fetch ie. `['requesor', 'provider', 'concent']` (concent is just an example)

> There will be no default value returned.


> #### 2. How should the stats be structured?
> As pointed out, some of the names for the stats fields are ambiguous and may lead to confusion. Given this, I'm going to consider renaming these fields and/or changing the structure of the responses.

If stats were hierarchical it will increase readability 

> 
> #### 3. How should the stats be stored?
> Similarly to point 2., this requires some more research. More specifically, I'm going to compare the provider and requestor stats and see whether the approach used for serialising provider stats can also be applied here.

👍 

> Here's the updated response structure (for requestor stats):
> 
> ```
> {
>   "current_tasks_stats": {
>  --- 
>  cut for brevity
>  ---
>   "tasks_in_network_cnt": 0
> }
> ```

pls. paste full response in comments, not only `requestor`

Issue #3980, opened 2019-03-11
BLENDER_NVGPU doesn't work on develop > 0.19


kbeker:
```bash
INFO     [golem.task.benchmarkmanager        ] Running benchmark for BLENDER_NVGPU
WARNING  [golem.task.taskthread              ] Task computing error 400 Client Error: Bad Request ("Unknown runtime specified nvidia")
WARNING  [golem.task                         ] Failed to compute benchmark 400 Client Error: Bad Request ("Unknown runtime specified nvidia")
ERROR    [golem.task.benchmarkmanager        ] Unable to run BLENDER_NVGPU benchmark: 400 Client Error: Bad Request ("Unknown runtime specified nvidia")
CRITICAL [golem.client                       ] Can't start network. Giving up.
Traceback (most recent call last):
  File "/home/kbeker/GITHUB/codepoets/golem/golem/client.py", line 311, in start
    self.start_network()
  File "/home/kbeker/GITHUB/codepoets/golem/golem/client.py", line 363, in start_network
    task_finished_cb=self._task_finished_cb,
  File "/home/kbeker/GITHUB/codepoets/golem/golem/task/taskserver.py", line 117, in __init__
    finished_cb=task_finished_cb)
  File "/home/kbeker/GITHUB/codepoets/golem/golem/task/taskcomputer.py", line 78, in __init__
    sync_wait(deferred, BENCHMARK_TIMEOUT)
  File "/home/kbeker/GITHUB/codepoets/golem/golem/core/deferred.py", line 33, in sync_wait
    result.raiseException()
  File "/home/kbeker/.virtualenvs/golem/lib/python3.6/site-packages/twisted/python/failure.py", line 385, in raiseException
    raise self.value.with_traceback(self.tb)
Exception: 400 Client Error: Bad Request ("Unknown runtime specified nvidia")
```

Every time, when I want to run `golemapp` benchmark raises with above exception.
System:

```bash
Operating System: Manjaro Linux
Kernel: Linux 4.19.20-1-MANJARO
```

VGA controller:
```bash
01:00.0 VGA compatible controller: NVIDIA Corporation GP106BM [GeForce GTX 1060 Mobile 6GB] (rev a1) (prog-if 00 [VGA controller])
        Subsystem: Micro-Star International Co., Ltd. [MSI] GP106M [GeForce GTX 1060 Mobile 6GB]
        Flags: bus master, fast devsel, latency 0, IRQ 146
[...]

PR #4608, opened 2019-08-12
nvidia-docker2 deprecation fix script

mfranciszkiewicz:
The script helps migrating from the deprecated `nvidia-docker2` to `nvidia-container-toolkit`. The deprecation issue was (indirectly) reported in https://github.com/golemfactory/golem/issues/3980#issuecomment-519118949.

Actions performed by the script:
- remove `nvidia-docker2`
- remove `docker-ce` from being on hold
- upgrade `docker-ce`
- install `nvidia-container-toolkit`
- write `nvidia` runtime definition to `/etc/docker/daemon.json`
- restart the daemon

### DIFF
--- a/Installer/Installer_Linux/nvgpu_fix.sh
+++ b/Installer/Installer_Linux/nvgpu_fix.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+set -e
+
+declare -r PYTHON=python3
+
+function info_msg()
+{
+    echo -e "\e[92m$@\e[39m"
+}
+
+function install_runtime()
+{
+    code=$(cat <<EOC
+import json
+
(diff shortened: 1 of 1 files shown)


maaktweluit (approved):
LGTM!

Maybe that ugly inline python script can be replaced with `jq` https://stedolan.github.io/jq/ but its not installed by default :(

mfranciszkiewicz:
@maaktweluit I agree, inlining Python code is not the most elegant solution. I was considering `jq` but installation / conditional removal when not already installed doesn't seem pretty either.

mfranciszkiewicz:
This script doesn't need to be merged, only linked to when needed by the user. Thank you for the review.

PR #4609, opened 2019-08-13
Fix saved ports in the KnownHosts table

Krigpl:
Noticed that while inspecting my local `KnownHosts` table - all of the entries have had invalid ports.
`self.port` is the port used by the other side to connect to us, `self.listen_port` is the one that we wanna save, because that's where the other side listens.


maaktweluit (approved):
LGTM!

PR #4611, opened 2019-08-13
Provide new envs performance in node's metadata

Krigpl:
And that'll automatically make the netmasking use proper values for new envs.

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -371,8 +371,6 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             self.keys_auth,
             connect_to_known_hosts=self.connect_to_known_hosts
         )
-        self.p2pservice.add_metadata_provider(
-            'performance', self.environments_manager.get_performance_values)
 
         self.task_server = TaskServer(
             self.node,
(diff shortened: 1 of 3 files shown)


maaktweluit (approved):
LGTM!

Should we also update this `p2pservice.metadata` after the benchmarks are ran?

Krigpl:
@maaktweluit These are metadata providers meaning that they are invoked everytime metadata is needed. So after rerunning the benchmarks they will return most recent values too.

maaktweluit:
@Krigpl Thanks for the clarification, it indeed gets the latest data this way

golem/client.py
```diff
@@ -385,6 +383,18 @@ def start_network(self):
             task_finished_cb=self._task_finished_cb,
         )
 
+        def get_performance_values():
+            values = self.environments_manager.get_performance_values()
+            new_env_manager = self.task_server.task_keeper.new_env_manager
+            for new_env in new_env_manager.environments():
+                env_id = new_env.metadata().id
+                value = new_env_manager.get_cached_performance(env_id)
+                if value:
```
Wiezzel:
```suggestion
                if value is not None:
```
Performance might be `0.0` and we want to include it then.

Wiezzel (commented):
Looks OK, but what about this part?
>Redesign the way the benchmark scores are stored for the known peers. Create a separate column for benchmark scores to make it more explicit.

Krigpl:
@Wiezzel I did not understand the goal of this part. Do we wanna a single column which still contains all the benchmarks of new evns? So still a json value? That's what we have now so there's no benefit to that.

Wiezzel (approved):

PR #4612, opened 2019-08-13
Add TaskResult class

mplebanski:
In the usage market, computation outcome is analysed from two viewpoints: correctness  (actual result) and resource usage. This PR introduces the TaskResult dataclass encapsulating these data elements.

### DIFF
--- a/golem/task/taskbase.py
+++ b/golem/task/taskbase.py
@@ -1,11 +1,9 @@
 import abc
 import logging
 from enum import Enum
-from typing import (
-    List,
-    Optional,
-    Type)
+from typing import Callable, Dict, List, Optional, Type
 
+from dataclasses import dataclass
 import golem_messages
 from golem_messages.datastructures import tasks as dt_tasks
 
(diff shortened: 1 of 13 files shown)


golem/task/taskbase.py
```diff
@@ -87,6 +85,12 @@ def notify_update_task(self, task_id):
         pass
 
 
+@dataclass
+class TaskResult:
+    files: List[str] = []
```
Krigpl: You can't do default values like that with lists and dicts - they'll always be the same object. Use `dataclass.field` with factory instead.

Krigpl (changes requested):
Please, *do* provide a description that summarizes the motivation and reasoning behind the changes in the PR.

golem/task/taskbase.py
mbenke: Addressed in 96c2800 and 898c8f0

Krigpl (approved):

mbenke (approved):

PR #4615, opened 2019-08-14
[D5/Z3] Transcoding: test with the number of subtasks equal to the number of key frames

cameel:
This pull request adds a test that splits each input video into the maximum possible number of segments.

Note that key frame numbers are hard-coded as a part of the input data set to make the test faster and less complex. The test video set is expected to change very rarely so this should not be a problem.

### Dependencies
This pull request is based on #4575 and should be merged after it. This is just for ease or development - It does not really depend on that base pull request and could be merged without it if rebased.

**NOTE**: I have set the base branch of this pull request to `transcoding-audio-channel-layout-validations` (#4575) rather than `CGI/transcoding/master`. Please remember to change the base branch back to `CGI/transcoding/master` before you merge.


nieznanysprawiciel:

Issue #439, opened 2016-06-30
JS files for Ethereum are missing in the package


chfast:
These 2 files needs to be packaged:
- golem/ethereum/genesis_golem.json
- golem/ethereum/mine_pending_transactions.js

PR #4616, opened 2019-08-14
manifest cleanup

tworec:
I've noticed `MANIFEST.in` file, which has two suspicious lines
```
include golem/ethereum/genesis_golem.json
include golem/ethereum/mine_pending_transactions.js
```

while we do not have those files in repo (there is even Issue #439 by @chfast on that).
furthermore it's last line is also strange
```
recursive-include apps *.blend *.jpg *.lxs *.ply *.lxo *.lxm *.lxv *.txt
```

because there is no files with jpg,lxs,ply,lxo,lxm nor lxv extension within `apps` folder
can be checked by
```
find -E apps -regex ".*\.(jpg|lxs|ply|lxo|lxm|lxv)$" 
```

and only textfiles are from `wasm` usecase which I'm kinda sure are not intended to be included...
so this line should be
```
recursive-include apps *.blend
```

maybe it is a good time to clean it up...

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,4 +3,4 @@ include requirements.txt
 include README.md
 include LICENSE.txt
 include loggingconfig.py
-recursive-include apps *.blend
+recursive-include apps *.blend *.txt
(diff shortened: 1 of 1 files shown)


maaktweluit (approved):
looks legacy to me

mfranciszkiewicz (approved):

chfast (approved):

Issue #4595, opened 2019-08-08
Refactor TaskBuilder from class to function


etam:
`TaskBuilder` classes are used in `TaskManager.create_task` in this way: 
```python
builder_type(...args...).build()
```
This suggests, that it shouldn't be a class, but a function.

etam:
Also it should be more clear, where task definition can be altered. Right now creating task works like this:
1. a dictionary comes from rpc
2. it's converted into `TaskDefinition` using `TaskBuilder.build_definition`, which is a `classmethod`.
3. `TaskDefinition` object is passed to `TaskBuilder` instance.
4. `TaskDefinition.get_task_kwargs` is used by `TaskBuilder` to produce kwargs, that are passed to `Task` constructor.

The original definition is altered in steps 2 and 4. This whole process could be greatly simplified.

(`TaskDefinition`, `TaskBuilder` and `Task` may also refer to types derived from them)

etam:
![me_trying_to_understand_how_tasks_are_created](https://user-images.githubusercontent.com/220752/62781200-bb8ef300-bab7-11e9-9212-32dfcaa65a07.jpg)

Issue #4003, opened 2019-03-14
Cannot create tasks with subtasks counts > 50


shadeofblue:
identified on 0.19.0, still present on 0.20 RC

* [x] verify the exact conditions governing the current (blender ?) subtask count limits
    * there are two pieces of this functionality that would need an update, both of which call `apps.rendering.task.framerenderingtask.calculate_subtasks_count`:
        * `_validate_task_dict` in `golem.task.rpc` called from `prepare_and_validate_task_dict` which in turn is called by the `comp.task.create`, `comp.task.restart` and indirectly by `comp.task.subtasks.restart` -> this one is pretty much broken since the `task_dict`'s `options` does not have the `frame_count` property (which is filled only when the appropriate task builder is run later in the process of task creation) - the result is that the logic that should be used for animations (rounding towards highest value, lower or equal to the requested number of subtasks which divides by the number of frames without remainder - in case subtask count > frames count - and a different calculation for subtasks < frames) is never used and both single frames and animations use the single-frame logic for task dictionary validation (`1 < number of subtasks < 50`) -> most weirdly (for both single frames and animations) when the subtask count is outside the limits, a default value of `20` is outputted as "optimal" (imvho, a maximum should be used when the number was above 50...)
        * `FrameRenderingTaskBuilder._calculate_total` ... which most likely uses the logic "correctly" and, again correctly (though silently) applies this logic to the enqueued task -> effectily enforcing the aforementioned animation logic to an already-"validated" task dict ... -> which causes weird artifacts -> e.g. when you define a task with 5 frames and 4 subtasks, 3 fragments are created, which means 3 subtasks are run... 
            * more interestingly, `RenderingTaskBuilder._calculate_total` (overridden in `FrameRenderingTaskBuilder`) contains some of the same code that's featured in `framerenderingtask.calculate_subtasks_count` ... is that even used?

* [ ] establish what the subtask count limits function should be - maybe ask: @badb @mfranciszkiewicz  
* [ ] fix/implement reasonable subtask count limits

shadeofblue:
any pull requests related to this issue should go into the `b0.20.1` branch for now

etam:
There is also an open question: if provided subtasks_count is illegal, should creating task be aborted, or `subtasks_count` corrected?
Because right now it looks like different authors of different parts had different answers to this question.

etam:
Related: https://github.com/golemfactory/golem-electron/issues/253

etam:
[...]

PR #4617, opened 2019-08-16
Cleanup subtasks count

etam:
This PR is required for #4003. It cleans up how `subtasks_count` is validated and corrected during task creation.

Most of the noise comes from removing `CoreTask.total_tasks` field. It's redundant with `CoreTask.task_definition.subtasks_count`. The main reason for removing it was to remove calculation from `RenderingTaskBuilder.get_task_kwargs` (see also #4595). There is a method `CoreTask.get_total_tasks()`, which should be used and could be overridden by deriving classes (like `WasmTask` does).

This also affects golemfactory/golem-electron#253. ~~The existing `comp.task.subtasks.count` RPC method is removed.~~ (EDIT: it's restored for now, but should be removed soon) Instead a new one is created `comp.task.create.dry_run`, so that client might check not only `subtasks_count` but all other parameters as well.

### DIFF
--- a/apps/appsmanager.py
+++ b/apps/appsmanager.py
@@ -2,32 +2,45 @@ import os
 from configparser import ConfigParser
 from collections import OrderedDict
 from importlib import import_module
+from typing import (
+    Dict,
+    List,
+    TYPE_CHECKING,
+    Tuple,
+    Type,
+)
 
 from golem.config.active import APP_MANAGER_CONFIG_FILES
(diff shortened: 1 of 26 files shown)


etam:
`create_dry_run` for rust cli: https://github.com/golemfactory/golem-client/pull/83

shadeofblue (approved):
thank_you_roadsign

PR #4618, opened 2019-08-16
Docker GPU environment

mfranciszkiewicz:
New version of `DockerGPUEnvironment`, focusing on `nvidia-docker-runtime` support


golem/task/taskcomputer.py
```diff
@@ -416,19 +430,29 @@ def change_config(
...
+        docker_cpu.update_config(DockerCPUConfig(**config_dict))
+        yield docker_cpu.prepare()
+
+        if not self._env_manager.enabled(DockerGPUEnvironment.ENV_ID):
+            return
+
+        docker_gpu = self._env_manager.environment(DockerGPUEnvironment.ENV_ID)
+        yield docker_gpu.clean_up()
+        docker_gpu.update_config(DockerGPUConfig(**config_dict))
+        yield docker_gpu.prepare()
```
Wiezzel: To avoid restarting the VM twice it should be done in this order:
```python
yield docker_cpu.clean_up()
yield docker_gpu.clean_up()
docker_cpu.update_config()
docker_gpu.update_config()
yield docker_cpu.prepare()
yield docker_gpu.prepare()
```

Wiezzel (commented):
At first glance it looks fine, but I will have to get a second or even third look since this is a lot of code.

One potential issue is a hidden dependency between CPU and GPU environments. If you prepare both and the clean up just one of them, the other would be unusable (because of the VM being disabled) while still having 'enabled' status. However, this might possibly be resolved using the auto-setup wrapper (#4592) by sharing a single usage counter between two environments.

golem/task/taskcomputer.py
mfranciszkiewicz: In general, that's true. But in practice there are serious difficulties with supporting GPUs on desktop hypervisors, so we're only allowing GPUs in VM-less environments (Linux). This part needs to be refactored either way.

mfranciszkiewicz:
~There is one more general issue to address with the new environments - a "partial" support mode. I.e. requestors may not need the specialized hardware to request tasks and verify subtask results. It's not only for GPUs, TEEs are affected as well.~

EDIT:
"Partial" support shouldn't be resolved on the Environment level but e.g. by the application specifying separate environments for requesting and computing in the manifest file.

golem/envs/docker/gpu.py
```diff
@@ -0,0 +1,147 @@
...
+@dataclass
+class DockerNvidiaGPUConfig(DockerGPUConfig):
+
+    gpu_vendor: str = field(
+        default=nvidia.VENDOR)
+    gpu_devices: List[str] = field(
+        default_factory=lambda: copy(nvidia.DEFAULT_DEVICES))
+    gpu_caps: List[str] = field(
+        default_factory=lambda: copy(nvidia.DEFAULT_CAPABILITIES))
+    gpu_requirements: List[Tuple[str, str]] = field(
```
maaktweluit: any reason to use a `List[Tuple]` over a `Dict[str, str]`?

golem/envs/docker/vendor/nvidia.py
```diff
@@ -0,0 +1,94 @@
...
+
+    raise ValueError(f"Invalid {VENDOR} GPU device names: {devices}")
+
+
+def validate_capabilities(caps: List[str]) -> None:
+    if not caps:
+        raise ValueError(f"Missing {VENDOR} GPU caps: {caps}")
+
+    special_count = sum([c in SPECIAL_CAPABILITIES for c in caps])
+    special_plus_others = special_count > 0 and len(caps) > 1
```
maaktweluit: Not super clear this variable is a bool, at first i thought it should be the sum of all caps
```suggestion
    has_mixed_caps = special_count > 0 and len(caps) > 1
```

tests/golem/envs/docker/gpu/test_config.py
```diff
@@ -0,0 +1,132 @@
...
+            config.gpu_requirements)
+
+    def test_container_config_dict(self):
+        config = DockerNvidiaGPUConfig(
+            gpu_vendor=nvidia.VENDOR,
+            gpu_devices=['0', '1'],
+            gpu_caps=['display', 'video'],
+            gpu_requirements=[
+                ('driver', '>=400.0'),
+                ('brand', 'Tesla'),
```
maaktweluit: haha

tests/golem/envs/docker/vendor/test_nvidia.py
```diff
@@ -0,0 +1,158 @@
+from unittest import mock, TestCase
+
+from golem.envs.docker.vendor import nvidia
+
+
+# FIXME: update the test after nvgpu logic is moved out of the apps folder
```
maaktweluit, line 6: is there an issue for this fixme?

maaktweluit (commented):
LGTM!

Feels a bit strange for the `CPU_config` to be superclass of `GPU_config` was thinking they had a shared base.

Will approve on reply, no changes required

tests/golem/envs/docker/vendor/test_nvidia.py
mfranciszkiewicz: There is none - I can create one. This should be resolved when apps are migrated to the Task API.

mfranciszkiewicz:
@maaktweluit `DockerGPUConfig` really is a superset of `DockerCPUConfig` - the GPU environment also needs CPU, RAM and working dirs to be configured. I see the CPU environment as a base for other environments. Both configs may share a common base, but in case of the CPU config that would only be a name change.

golem/envs/docker/gpu.py
mfranciszkiewicz: No particular reason, `List[Tuple]` seemed to be simpler. But I'll refactor it to `Dict` for convenience.

maaktweluit:
Ok, clear :D thanks for the fixes, approving!

maaktweluit (approved):

golem/envs/docker/gpu.py
```diff
@@ -0,0 +1,147 @@
...
+            environment=environment,
+        )
+
+
+class DockerGPURuntime(DockerCPURuntime):
+
+    def usage_counters(self) -> Dict[CounterId, CounterUsage]:
+        raise NotImplementedError
+
+    def call(self, alias: str, *args, **kwargs) -> Deferred:
```
Wiezzel: `call` has been removed from `Runtime` interface.

golem/task/taskcomputer.py
```diff
@@ -244,15 +245,28 @@ def __init__(
...
-        docker_env = self._env_manager.environment(DockerCPUEnvironment.ENV_ID)
-        yield docker_env.prepare()
+        docker_cpu = self._env_manager.environment(DockerCPUEnvironment.ENV_ID)
+        yield docker_cpu.prepare()
+
+        if not self._env_manager.enabled(DockerGPUEnvironment.ENV_ID):
+            return
+
+        docker_gpu = self._env_manager.environment(DockerGPUEnvironment.ENV_ID)
+        yield docker_gpu.prepare()
```
Wiezzel: This method was removed in #4651

golem/task/taskcomputer.py
```diff
@@ -244,15 +245,28 @@ def __init__(
...
+        docker_cpu = self._env_manager.environment(DockerCPUEnvironment.ENV_ID)
+        if docker_cpu.status() is not EnvStatus.DISABLED:
+            yield docker_cpu.clean_up()
+
+        if not self._env_manager.enabled(DockerGPUEnvironment.ENV_ID):
+            return
+
+        docker_gpu = self._env_manager.environment(DockerGPUEnvironment.ENV_ID)
+        if docker_gpu.status() is not EnvStatus.DISABLED:
+            yield docker_gpu.clean_up()
```
Wiezzel: This method was removed in #4651

golem/task/taskcomputer.py
Wiezzel: OK, once #4651 is merged this is obsolete.

golem/envs/docker/gpu.py
mfranciszkiewicz: There is no support for it in Task API and it needs to be implemented.

Wiezzel: I'm not sure if it's needed any longer. All RPC calls are being made by designated methods from `ProviderAppClient` and `RequestorAppClient`.

mfranciszkiewicz: It's a custom app call, made by the requestor on provider's app instance.

mfranciszkiewicz: Extending `ProviderAppClient` is what I had in mind. We'd also need for Golem to listen for requestor app requests and send them to the provider.

Wiezzel: Oh, OK, sure then.

Wiezzel (approved):

PR #4621, opened 2019-08-19
node_integration_tests: lowered the comp.tasks timeout from 60 to 2 minutes

maaktweluit:
Creating a task should never take this long

### DIFF
--- a/scripts/node_integration_tests/playbooks/base.py
+++ b/scripts/node_integration_tests/playbooks/base.py
@@ -307,7 +307,7 @@ class NodeTestPlaybook:
             new_tasks = tasks - self.known_tasks
             if len(new_tasks) != 1:
                 print("Cannot find the new task ({})".format(new_tasks))
-                time.sleep(1)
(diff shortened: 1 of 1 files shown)


Wiezzel (approved):
Sure, why not.

shadeofblue (approved):
may slightly increase output clutter but agree with the premise ... 

might I suggest an intermediate value of e.g. 3 seconds sleep?

mfranciszkiewicz (approved):

maaktweluit:
@shadeofblue no extra clutter for me, it already gets the task the first time, so normally the timer is not even used.

Why 3 seconds over 1? happy to change it for a good cause :)

shadeofblue:
@maaktweluit I was just concerned that if - in some cases - the creation takes e.g. half a minute, you'd get 30 messages ... which seems undesirable ..

maaktweluit:
@shadeofblue thanks, updated :)

PR #4625, opened 2019-08-20
bump messages to 3.11 (always require contract address in the sig has…

shadeofblue:
…h data)

### DIFF
--- a/tests/golem/task/test_concent_logic.py
+++ b/tests/golem/task/test_concent_logic.py
@@ -62,8 +62,8 @@ class TaskToComputeConcentTestCase(testutils.TempDirFixture):
         self.msg.generate_ethsig(self.requestor_keys.raw_privkey)
         self.ethereum_config = EthereumConfig()
         self.msg.sign_all_promissory_notes(
-            deposit_contract_address=getattr(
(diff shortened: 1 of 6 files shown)


tests/golem/task/test_concent_logic.py
```diff
@@ -225,7 +224,11 @@ def test_no_concent_promissory_note_sig(self, send_mock, *_):
         )
 
     def test_bad_promissory_note_sig(self, send_mock, *_):
-        self.msg.sign_promissory_note(self.different_keys.raw_privkey)
+        self.msg.sign_promissory_note(
+            deposit_contract_address=getattr(
+                self.ethereum_config, 'deposit_contract_address'),
```
Krigpl: What's the benefit here of using `getattr`? Attribute's name is not dynamic and both will fail if it doesn't exist. And this doesn't benefit from static analysis.
Same in other places.

Krigpl (approved):

tests/golem/task/test_concent_logic.py
shadeofblue: there must have been a reason ... I think ...

PR #4629, opened 2019-08-22
Extract event loop from twisted.reactor

jiivan:
AsyncIO event loop is created in `common.install_reactor()`
and can be extracted from `twisted.internet.reactor` when
needed (in threads).


shadeofblue (approved):
strange apple

PR #4632, opened 2019-08-22
Requested Task Manager

maaktweluit:
Implementation of the new `RequestedTaskManager`

- [x] Does not use pickles, but models from #4340 
- [x] Uses CoreTask only statically when needed
- [x] Implements all app_client calls

Open ends from FIXME's:
- `# FIXME: Blender creates preview files here`
- golem_params.max_subtasks: `# FIXME: This is a separate argument now, delete here?`

Fixed ends:
- `# FIXME: Where to put resources in database`? Not in database, just copy the file = DONE
- `# FIXME: Where to put app_params in database`? new field in database = DONE


golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +50,254 @@ class SubtaskDefinition:
 
 
 class RequestedTaskManager:
+    def __init__(self, env_manager: EnvironmentManager, public_key, root_path):
+        logger.debug('RequestedTaskManager(public_key=%r, root_path=%r)',
+                     public_key, root_path)
+        self._dir_manager = DirManager(root_path)
```
Krigpl, line 103: I don't think we should use `DirManager` here - the directory structure is defined by Task API and this class doesn't know anything about that.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +50,254 @@ class SubtaskDefinition:
...
             app_params: Dict[str, Any],
     ) -> TaskId:
         """ Creates an entry in the storage about the new task and assigns
         the task_id to it. The task then has to be initialized and started. """
-        raise NotImplementedError
+        logger.debug('create_task(golem_params=%r, app_params=%r)',
+                     golem_params, app_params)
+
+        task = RequestedTask.create(
+            task_id=CoreTask.create_task_id(self._public_key),
```
Krigpl: Just use `idgenerator.generate_id(public_key)` directly, it's awkward that this class depends on `CoreTask`.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +50,254 @@ class SubtaskDefinition:
...
+            max_subtasks=task.max_subtasks,
+            max_price_per_hour=task.max_price_per_hour,
+            concent_enabled=task.concent_enabled,
+        )
+        app_client = await self._get_app_client(task.environment)
+        logger.debug('init_task(task_id=%r) before creating task', task_id)
+        await app_client.create_task(
+            task.task_id,
+            task.max_subtasks,
+            asdict(task_params),
```
Krigpl: `app_params` should be here, the whole `task_params` variable is invalid here

golem/task/requestedtaskmanager.py
maaktweluit: Good one, fully agree! will replace it with a TaskApiDirManager or smth

golem/task/requestedtaskmanager.py
maaktweluit: nice, updated

golem/task/requestedtaskmanager.py
maaktweluit: thanks, that makes the function a lot smaller, updated

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +50,241 @@ class SubtaskDefinition:
...
+            task=task,
+            subtask_id=result.subtask_id,
+            status=SubtaskStatus.starting,
+            # payload='{}',
+            # inputs='[]',
+            start_time=default_now(),
+            price=task.max_price_per_hour,
+            computing_node=computing_node,
+        )
+        return subtask
```
Krigpl: This isn't a `SubtaskDefinition`

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +50,241 @@ class SubtaskDefinition:
...
+            asyncio.get_event_loop()
+        except Exception:
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
+        future = asyncio.ensure_future(asyncio.wait(shutdown_futures))
+        deferred = deferred_from_future(future)
+        logger.debug('quit() deferred=%r', deferred)
+        return deferred
+
+    async def _get_app_client(self, env_id: EnvId) -> RequestorAppClient:
```
Krigpl: This should cache app clients per app id, not env id.

golem/task/requestedtaskmanager.py
maaktweluit: i was thinking of replacing the `SubtaskDefenition` with `model.RequestedSubtask` since they have a some fields in common.

Do you think using `SubtaskDefenition` has added value here?

golem/task/requestedtaskmanager.py
maaktweluit: At first i had it as `app_id`, not sure why i changed it..

Will put it back, thanks!

golem/task/requestedtaskmanager.py
Krigpl: Better encapsulation. `SubtaskDefinition` is supposed to contain just the data needed for the correct computation while `model.RequestedSubtask` is `RequestedTaskManager`s implementation detail, can contain whatever data it's needed for any reasons and thus shouldn't be a part of the public interface.

maaktweluit: thanks, updated :)

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +51,262 @@ class SubtaskDefinition:
...
+        result = await app_client.verify(task.task_id, subtask_id)
+        if result:
+            subtask.status = SubtaskStatus.finished
+            finished_subtasks = RequestedSubtask.select(
+                fn.Count(RequestedSubtask.subtask_id)
+            ).where(
+                RequestedSubtask.task == task
+                and RequestedSubtask.status == SubtaskStatus.finished
+            )
+            if finished_subtasks >= task.max_subtasks:
```
Krigpl: This is incorrect. `max_subtasks` is the upper bound for the number of subtasks and the application is free to create fewer subtasks than that.
Currently, the right way to check whether a task is finished is to check whether all subtasks are finished + there no pending subtasks.

maaktweluit: boohoo, it looked so nice and simple :P

Updated the check :)

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
 
 
 class RequestedTaskManager:
+    def __init__(self, env_manager: EnvironmentManager, public_key, root_path):
```
Wiezzel: Plz add type annotations for all parameters and reformat to one param per line.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+            name=golem_params.name,
+            status=TaskStatus.creating,
+            environment=golem_params.environment,
+            # prerequisites='{}',
+            task_timeout=golem_params.task_timeout,
+            subtask_timeout=golem_params.subtask_timeout,
+            start_time=default_now(),
+            max_price_per_hour=golem_params.max_price_per_hour,
+            max_subtasks=golem_params.max_subtasks,
+            # concent_enabled = BooleanField(null=False, default=False),
```
Wiezzel: Let's be explicit that concent is not supported for Task API.
```suggestion
            concent_enabled=False,
```

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+
+        # FIXME: Is RTM responsible for managing test tasks?
+
+        app_client = await self._get_app_client(task.app_id, task.environment)
+        logger.debug('init_task(task_id=%r) before creating task', task_id)
+        await app_client.create_task(
+            task.task_id,
+            task.max_subtasks,
+            task.app_params,
+        )
```
Wiezzel, line 176: `RequestorAppClient.create()` should return env_id and prerequisites dict. We probably need a task for this.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
-        raise NotImplementedError
+        logger.debug('create_task(golem_params=%r, app_params=%r)',
+                     golem_params, app_params)
+
+        task = RequestedTask.create(
+            task_id=idgenerator.generate_id(self._public_key),
+            app_id=golem_params.app_id,
+            name=golem_params.name,
+            status=TaskStatus.creating,
+            environment=golem_params.environment,
```
Wiezzel: Why is this in `golem_params`? This should be returned from `RequestorAppClient.create()`.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+            raise RuntimeError(
+                f"Task not active, can not verify. task_id={task_id}")
+        subtask = RequestedSubtask.get(
+            RequestedSubtask.subtask_id == subtask_id)
+        # FIXME, check if subtask_id belongs to task
+        assert subtask.task == task
+        app_client = await self._get_app_client(task.app_id, task.environment)
+        subtask.status = SubtaskStatus.verifying
+        subtask.save()
+        result = await app_client.verify(task.task_id, subtask_id)
```
Wiezzel: A try/catch would be nice here (setting failure status).

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+                f"Task not active, can not abort. task_id={task_id}")
+        task.status = TaskStatus.aborted
+        task.save()
+        subtasks = self._get_pending_subtasks(task_id)
+        for subtask in subtasks:
+            ProviderComputeTimers.finish(subtask.subtask_id)
+            subtask.status = SubtaskStatus.cancelled
+            subtask.save()
+
+        # self.notice_task_updated(task_id, op=TaskOp.ABORTED)
```
Wiezzel, line 349: Shouldn't the app be notified about task getting aborted? If so, we probably need a task for this as well.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+        logger.debug('quit() clients=%r', self._app_clients)
+        if not self._app_clients:
+            logger.debug('No clients to clean up')
+            return succeed(None)
+        shutdown_futures = [
+            app.shutdown() for app in self._app_clients.values()
+        ]
+        logger.debug('quit() futures=%r', shutdown_futures)
+        # FIXME: error when running in another thread.
+        # this fixes it, but is it the right way?
```
Wiezzel: This is kinda :nauseated_face: 
Maybe you can use `get_event_loop` from `golem.core.golem_async`? (see #4629)

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+            ProviderComputeTimers.finish(subtask.subtask_id)
+            subtask.status = SubtaskStatus.cancelled
+            subtask.save()
+
+        # self.notice_task_updated(task_id, op=TaskOp.ABORTED)
+
+        await self._shutdown_app_client(task.app_id)
+
+    def quit(self) -> Deferred:
+        # FIXME: make async not Deferred?
```
Wiezzel: Yep, totally. All other methods are async, so this should be as well. We should avoid having mixed asyncio/twisted interfaces.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+            env_id,
+            app_id
+        )
+        if not self._env_manager.enabled(env_id):
+            raise RuntimeError(
+                f"Error connecting to app: {env_id}. environment not enabled")
+        env = self._env_manager.environment(env_id)
+        payload_builder = self._env_manager.payload_builder(env_id)
+        prereq = env.parse_prerequisites(
+            {"image": "blenderapp", "tag": "latest"}  # FIXME: hardcoded :(
```
Wiezzel: No worry, that would be implemented in `AppManager` soon.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+        )
+        shared_dir = self._dir_manager.root_path
```
Wiezzel: Most probably this not the path you want to share. Each app should have its subdirectory.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+            payload_builder=payload_builder,
+            prereq=prereq,
+            shared_dir=shared_dir
+        )
+
+    @staticmethod
+    def _get_unfinished_subtasks_for_node(
+            task_id: TaskId,
+            computing_node: ComputingNode
+    ) -> None:
```
Wiezzel:
```suggestion
    ) -> int:
```

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+        ).where(
+            RequestedSubtask.computing_node == computing_node,
+            RequestedSubtask.task_id == task_id,
+            RequestedSubtask.status != SubtaskStatus.finished,
+        ).scalar()
+        logger.debug('unfinished subtasks: %r', unfinished_subtask_count)
+        return unfinished_subtask_count
+
+    @staticmethod
+    def _get_pending_subtasks(task_id: TaskId) -> None:
```
Wiezzel: Fix return type annotation.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,345 @@ class SubtaskDefinition:
...
+        )
+
+    async def _shutdown_app_client(self, app_id) -> None:
+        # Check if app completed all tasks
+        unfinished_tasks = RequestedTask.select(
+            fn.Count(RequestedTask.task_id)
+        ).where(
+            RequestedTask.app_id == app_id,
+            # FIXME: duplicate list with TaskStatus.is_active()
+            RequestedTask.status.in_([
```
Wiezzel, line 432: Maybe `not_in_` would be shorter?

golem/task/taskcomputer.py
```diff
@@ -298,7 +298,7 @@ def compute(self) -> defer.Deferred:
         assert assigned_task is not None
 
         task_api_service = self._get_task_api_service()
-        app_client = ProviderAppClient(service=task_api_service)
+        app_client = ProviderAppClient.create(service=task_api_service)
```
Wiezzel: That's been fixed in #4634

golem/task/requestedtaskmanager.py
maaktweluit: done

golem/task/requestedtaskmanager.py
maaktweluit: does this belong to the task `start/stop client`? then i'll pick this up in the next PR ( using this one only for the app_client calls)

golem/task/requestedtaskmanager.py
maaktweluit: i added the missing fields as input param's for now, will get it from the create() in the next PR

golem/task/requestedtaskmanager.py
maaktweluit: Good one, added

golem/task/requestedtaskmanager.py
maaktweluit: Good thought!

Was thinking the shutdown would suffice for now..

Will make a task about adding abort notification to the app

golem/task/requestedtaskmanager.py
maaktweluit: nice, will leave it for now then :)

golem/task/requestedtaskmanager.py
maaktweluit: :) will add this in the start / stop PR too

golem/task/requestedtaskmanager.py
maaktweluit: good one, will add this to the new DirManager that was proposed earlier in this PR

golem/task/requestedtaskmanager.py
maaktweluit: added

golem/task/requestedtaskmanager.py
maaktweluit: what do you mean?
there are 8 total TaskStatus'es so listing 3 seems shorter

golem/task/taskcomputer.py
maaktweluit: will rebase over it :)

maaktweluit (commented):
@Wiezzel thanks for your comments, added / fixed all but:
- Dir manager related ( fixing in this PR )
- Start / stop related ( will move this code to a new PR )

golem/task/requestedtaskmanager.py
maaktweluit: fixed

golem/task/taskcomputer.py
```diff
@@ -214,10 +213,14 @@ def change_config(
             config_desc=config_desc,
             in_background=in_background))
 
-    def quit(self) -> None:
-        sync_wait(self._new_computer.clean_up())
+    @defer.inlineCallbacks
```
Krigpl: Why do you touch this file in this PR?

golem/task/requestedtaskmanager.py
Krigpl: Considering this method is introduced here but nowhere called I'd say either implement it properly or leave it completely out of this PR. There's no value in having a partially implemented non used method.

golem/task/requestedtaskmanager.py
Krigpl: Since this class is not designed to be thread safe, the caller of this method, if calling from a different thread should take care of that issue, not the class.

golem/task/requestedtaskmanager.py
Wiezzel: I don't think so. Start/stop client cares about *requestor* env and prerequisites and I mean the *provider* part here.

golem/task/requestedtaskmanager.py
```diff
@@ -30,60 +53,360 @@ class SubtaskDefinition:
...
+        logger.debug('has_pending_subtasks(task_id=%r)', task_id)
+        task = RequestedTask.get(RequestedTask.task_id == task_id)
+        app_client = await self._get_app_client(task.app_id, task.environment)
+        return await app_client.has_pending_subtasks(task.task_id)
 
-    def get_next_subtask(
+    async def get_next_subtask(
             self,
             task_id: TaskId,
+            computing_node: ComputingNode
```
Krigpl: I'd be inclined not to use `molde.ComputingNode` here, since again, I consider models to be implementation detail and not part of the interface. i.e. changing underlying database shouldn't affect the interface imo.

golem/task/requestedtaskmanager.py
Wiezzel: OK, nvm.

Krigpl (commented):
This is looking good - my integration script is mostly happy already expect for the directory management.

golem/task/taskcomputer.py
maaktweluit: whoops, this was supposed to be reverted, will clean it up

maaktweluit:
@Krigpl thanks for the comments, it does not give a text box to reply to them separately so will do it here:
- `quit()` will remove from this PR and split out to the next one where it should be used
- `threads` agree, will also be moved and ( hopefully ) fixed in the next PR
- `model.*` as argument: good comment, will fix this in this PR

Will still add a new DirManager before un-drafting this PR

golem/task/requestedtaskmanager.py
maaktweluit: Ok, could you make a task for it? think you can describe it better then me

golem/task/requestedtaskmanager.py
maaktweluit: Reverted it for this PR, will bring it back and fix it next one

golem/task/requestedtaskmanager.py
maaktweluit: updated to use a local dataclass instead

golem/task/requestedtaskmanager.py
```diff
@@ -21,6 +41,12 @@ class CreateTaskParams:
     concent_enabled: bool
 
 
+@dataclass
+class ComputingNodeDefenition:
```
Krigpl: typo

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +55,383 @@ class SubtaskDefinition:
...
+            task=task,
+            subtask_id=result.subtask_id,
+            status=SubtaskStatus.starting,
+            payload=result.params,
+            inputs=result.resources,
+            start_time=default_now(),
+            price=task.max_price_per_hour,
+            computing_node=node,
+        )
+        deadline = subtask.start_time \
```
Krigpl: That should be `min`'d with task's deadline, I think.

tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -0,0 +1,327 @@
...
+    ComputingNode,
+    CreateTaskParams,
+    RequestedTaskManager,
+    ComputingNodeDefenition,
+)
+from golem.task.taskstate import TaskStatus, SubtaskStatus
+from golem.testutils import DatabaseFixture
+
+
+class TestRequestedTaskManager(DatabaseFixture, TwistedTestCase):
```
Krigpl, line 32: `RequestedTaskManager` doesn't use Twisted at all, why do you want to use it for tests? Just use `pytest`'s `asyncio` plugin.

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +55,383 @@ class SubtaskDefinition:
...
+            raise RuntimeError(f"Task {task_id} has already been initialized")
+
+        # FIXME: Blender creates preview files here
+
+        logger.debug(
+            'init_task(task_id=%r) - preparing directories. app_id=%s',
+            task_id,
+            task.app_id,
+        )
+        self._dir_manager.prepare_task_dir(task.app_id, task_id)
```
Krigpl: That needs to go into `create_task` since we need resources dir to put the resources into before calling `create` on task api.

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: This is for the using the `DatabaseFixture`, since its based off `unittest.TestCase` it does not work with `@pytest.mark.asyncio`

Tried to fix it quickly but it opened a whole new can of worms, so moving these fixes to a separate PR.

golem/task/requestedtaskmanager.py
maaktweluit: fixed

golem/task/requestedtaskmanager.py
maaktweluit: i agree! added it

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +55,383 @@ class SubtaskDefinition:
...
 
     def get_subtasks_outputs_dir(self, task_id: TaskId) -> Path:
         """ Return a path to the directory where subtasks outputs should be
         placed. """
-        raise NotImplementedError
+        task = RequestedTask.get(RequestedTask.task_id == task_id)
+        return self._dir_manager.get_network_results_dir(task.app_id, task_id)
 
-    def has_pending_subtasks(self, task_id: TaskId) -> bool:
+    async def has_pending_subtasks(self, task_id: TaskId) -> bool:
```
Krigpl, line 222: I think this should first check whether the task is still active before issuing the rpc call. Otherwise this may spawn the runtime that won't be shutdown anywhere later.

golem/task/requestedtaskmanager.py
maaktweluit: good one, moved!

golem/task/requestedtaskmanager.py
maaktweluit: you are right, the check was already in `get_next_subtask(` so i moved it here.

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: https://github.com/golemfactory/golem/pull/4648

golem/task/requestedtaskmanager.py
maaktweluit: new DirManager is added, please check again

golem/task/requestedtaskmanager.py
maaktweluit: New DirManager is in place

tests/golem/task/test_requestedtaskmanager.py
Krigpl: Alternatively, just use `pytest`'s `tmpdir` fixture for the temporary directory.

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +55,384 @@ class SubtaskDefinition:
...
+        computation at the given moment. If there are the next call to
         get_next_subtask will return properly defined subtask. It may happen
         that after not having any pending subtasks some will become available
         again, e.g. in case of failed verification a subtask may be marked
         as pending again. """
-        raise NotImplementedError
+        logger.debug('has_pending_subtasks(task_id=%r)', task_id)
+        task = RequestedTask.get(RequestedTask.task_id == task_id)
+        if not task.status.is_active():
+            raise RuntimeError(
```
Krigpl: I'm not sure that should throw - if the task is not active it doesn't have any pending subtasks, so simple `False` is fine imo.

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: good one, will update the `testutils` to use this too in the linked PR.

Would like to keep this PR as is for now so we can merge it

golem/task/requestedtaskmanager.py
maaktweluit: haha, raise them all!

good point, the next_subtask can raise but here False is good enough

tests/golem/task/test_requestedtaskmanager.py
Krigpl: sure, sounds good

tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -0,0 +1,330 @@
...
+        self.env_manager = Mock(spec=EnvironmentManager)
+        self.public_key = str.encode('0xdeadbeef')
+        self.rtm = RequestedTaskManager(
+            env_manager=self.env_manager,
+            public_key=self.public_key,
+            root_path=self.new_path
+        )
+
+    def tearDown(self):
+        RequestedSubtask.delete().execute()
```
Krigpl: I don't think you need that. Each test will work on its own fresh DB.

tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -0,0 +1,330 @@
...
+    @staticmethod
+    def _add_has_pending_subtasks_to_client_mock(mock_client, result=True):
+        f = asyncio.Future()
+        f.set_result(result)
+        mock_client.has_pending_subtasks = Mock(return_value=f)
+
+    def _add_next_subtask_to_client_mock(self, mock_client):
+        result = Mock(spec=Subtask)
+        result.params = '{}'
+        result.resources = '[]'
```
Krigpl: These shouldn't be strings.

tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -0,0 +1,330 @@
...
+        )
+
+    def _create_task(self):
+        golem_params = self._build_golem_params()
+        app_params = {}
+        task_id = self.rtm.create_task(golem_params, app_params)
+        return task_id
+
+    def _mock_client_create(self):
+        mock_client = Mock(spec=RequestorAppClient)
```
Krigpl: Try to use this approach for dealing with mocking async methods https://github.com/golemfactory/task-api/blob/master/python/tests/test_integration.py#L14

I think it'll be cleaner than those verbose mock futures and `_patch_async`s.

Krigpl (commented):
Looks good - couple of comments regarding cleaner tests tho.

golem/task/requestedtaskmanager.py
Wiezzel: Done.

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +55,383 @@ class SubtaskDefinition:
     deadline: int
 
 
+class DirManager:
```
Wiezzel, line 59: Maybe `RequestorDirManager` to be more specific? We already have one `DirManager` and maybe could use another one for management of provider directories.

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +55,383 @@ class SubtaskDefinition:
...
+        task_resources_dir = task_dir / constants.RESOURCES_DIR
+        task_resources_dir.mkdir()
+        task_net_resources_dir = self.get_network_resources_dir(app_id, task_id)
+        task_net_resources_dir.mkdir()
+        task_results_dir = task_dir / constants.RESULTS_DIR
+        task_results_dir.mkdir()
+        task_net_results_dir = self.get_network_results_dir(app_id, task_id)
+        task_net_results_dir.mkdir()
+
+    def get_network_resources_dir(self, app_id, task_id):
```
Wiezzel: Let's go with the subtask inputs/subtask outputs naming since as I think we agreed upon it.

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +55,383 @@ class SubtaskDefinition:
...
+            status=SubtaskStatus.starting,
+            payload=result.params,
+            inputs=result.resources,
+            start_time=default_now(),
+            price=task.max_price_per_hour,
+            computing_node=node,
+        )
+        task_deadline = task.deadline
+        if task_deadline is None:
+            raise RuntimeError(f"Task not started. task_id={task_id}")
```
Wiezzel: Why is this check done after creating the subtask? Wouldn't it leave application in an invalid state?

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +55,383 @@ class SubtaskDefinition:
...
+        if result:
+            subtask.status = SubtaskStatus.finished
+        else:
+            subtask.status = SubtaskStatus.failure
+        subtask.save()
+
+        if result:
+            # Check if task completed
+            if not await self.has_pending_subtasks(task_id):
+                if not self._get_pending_subtasks(task_id):
```
Wiezzel, line 329: Why are both checks needed?

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +55,383 @@ class SubtaskDefinition:
...
+                'app_client created for app_id=%r, clients=%r',
+                app_id, self._app_clients[app_id])
+        return self._app_clients[app_id]
+
+    def _get_task_api_service(
+            self,
+            app_id: str,
+            env_id: EnvId
+    ) -> EnvironmentTaskApiService:
+        # FIXME: Stolen from golem/task/taskcomputer.py:_get_task_api_service()
```
Wiezzel: There's no more `_get_task_api_service()`  in `taskcomputer`

golem/task/taskserver.py
```diff
@@ -155,7 +155,11 @@ def __init__(self,
             apps_manager=apps_manager,
             finished_cb=task_finished_cb,
         )
-        self.requested_task_manager = RequestedTaskManager()
+        self.requested_task_manager = RequestedTaskManager(
+            env_manager=new_env_manager,
+            root_path=TaskServer.__get_task_manager_root(client.datadir),
```
Wiezzel, line 178: Fun fact: this returns the same path as `self.get_task_computer_root()`

tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -0,0 +1,330 @@
...
+    def test_create_task(self):
+        # given
+        golem_params = self._build_golem_params()
+        app_params = {}
+        # when
+        task_id = self.rtm.create_task(golem_params, app_params)
+        # then
+        row = RequestedTask.get(RequestedTask.task_id == task_id)
+        assert row.status == TaskStatus.creating
+        assert row.start_time < default_now()
```
Wiezzel: I know it's boring but I'd put some more assertions on row attributes here. Maybe check if the task directory was initialized as well?

golem/task/requestedtaskmanager.py
Krigpl: fwiw I'm not sure if it's even worth making it a separate class, just moving the functions to the RTM would be fine imo. This class won't be used anywhere else and getters like `get resources dir/network_resources dir/network results dir` are simple redirections.
Not insisting tho.

Krigpl: (since effectively the only functionality here is the `prepare_task_dir` method)

Wiezzel: Actually, I think this class belongs to the Task API repository. And know what? That logic is already there:
https://github.com/golemfactory/task-api/blob/2dccb9f4cb237e2f1676caecaa83ced2ea9da9a6/python/golem_task_api/testutils/tasklifecycleutil.py#L100-L114

golem/task/requestedtaskmanager.py
Krigpl: AppManager is now in develop so let's do it properly.

tests/golem/task/test_requestedtaskmanager.py
Krigpl:
```
__________________ TestRequestedTaskManager.test_create_task __________________
self = <tests.golem.task.test_requestedtaskmanager.TestRequestedTaskManager testMethod=test_create_task>
    def test_create_task(self):
        # given
        golem_params = self._build_golem_params()
        app_params = {}
        # when
        task_id = self.rtm.create_task(golem_params, app_params)
        # then
        row = RequestedTask.get(RequestedTask.task_id == task_id)
        assert row.status == TaskStatus.creating
>       assert row.start_time < default_now()
E       AssertionError: assert datetime.datetime(2019, 9, 2, 9, 45, 14, 217336, tzinfo=datetime.timezone.utc) < datetime.datetime(2019, 9, 2, 9, 45, 14, 217336, tzinfo=datetime.timezone.utc)
E        +  where datetime.datetime(2019, 9, 2, 9, 45, 14, 217336, tzinfo=datetime.timezone.utc) = <golem.model.RequestedTask object at 0x0000022568380208>.start_time
E        +  and   datetime.datetime(2019, 9, 2, 9, 45, 14, 217336, tzinfo=datetime.timezone.utc) = default_now()
tests\golem\task\test_requestedtaskmanager.py:65: AssertionError
```
Please use freeze_gun when testing time sensitive stuff.

Krigpl (changes requested):

tests/golem/task/test_requestedtaskmanager.py
Wiezzel: Good remark. However, I'm not sure if `freeze_gun` works correctly with `TwistedTestCase`.

golem/task/requestedtaskmanager.py
maaktweluit: haha, yeah, i copied it from there ;)

Sounds like its best to move the dirmanager to task-api ( since we should not use `testutils`) in production code

will have a try at this

golem/task/requestedtaskmanager.py
maaktweluit: good one, did not think of changing the given interface but will do it :D

golem/task/requestedtaskmanager.py
maaktweluit: this can only happen when `start_time` is `None` and should never happen, its a safety check to remove optional, maybe we can just assert it or remove the check all together?

golem/task/taskserver.py
maaktweluit: yep, i think i had even more options to pick where to get the same dir :P iirc i picked this one since it was most purpose specifically named ( task manager )

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: good one! the dir-manager was extracted later so did not add test asserts for it yet

maaktweluit: will try it, if it does not work i'll make the assert `<=` ;)

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: thanks, was looking for a solve like this :D

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: iirc they are also strings as default values in the models, what should they be? dict and list? will update them :)

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: i think i made a mistake in a test and needed this, you are probably correct, will try to remove ot

maaktweluit: you are right, removed

golem/task/taskserver.py
Krigpl: imo RTM shouldn't share the directory with the old tm

tests/golem/task/test_requestedtaskmanager.py
Krigpl: I dunno about the model and this isn't a struct from the model - this is task api struct and params is a dict and resources is a list.

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: will postpone this fix to the other PR with the `tempdir` fixes, it was the last missing part of getting that to work and do not want to open that can of worms in this PR again

golem/task/requestedtaskmanager.py
maaktweluit: one asks the app, the other asks the local database, they should always be in sync but you never know, maybe it can have some better error handling when they are out of sync

golem/task/requestedtaskmanager.py
maaktweluit: :( but still stolen it from there :P will check if there is a new place to link to, otherwise i'll link to the SHA it was stolen from ;)

maaktweluit: updated fn name :)

golem/task/requestedtaskmanager.py
Wiezzel: If it should never occur I'd go with simple assert.

golem/task/requestedtaskmanager.py
Wiezzel: OK, let's leave it like this.

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: added a test

maaktweluit: `freezegun` installed

golem/task/requestedtaskmanager.py
maaktweluit: ok, fixed

golem/task/taskserver.py
maaktweluit: i think it should for now, since the docker shares are all configured to only be available in this directory

Due to the structure the RTM will always use other sub-directories so it is still somehow separated

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: thanks! i used the database models first for these objects, after refactoring i did not change the values

Updated now

golem/task/requestedtaskmanager.py
maaktweluit: i started looking into this and the change became too big, since it also requires re-release and testing of the task-api and blenderapp packages.

Made a new issue for this improvement, linking to this comment.
We can discuss it in refinement tomorrow and pick it up separately

golem/task/requestedtaskmanager.py
maaktweluit: done

maaktweluit:
@Krigpl @Wiezzel Thanks for the reviews! solved most comments, moved 2 bigger things to the next issues and linked to the comments for our future selves :)

Please check again

tests/golem/task/test_requestedtaskmanager.py
Krigpl: I insist. This is how much you gain, it's much simpler: https://github.com/golemfactory/golem/compare/mwu/requested-taskmanager...rtm_tests_async?expand=1
Feel free to just merge it to this branch.

golem/task/requestedtaskmanager.py
Krigpl: @maaktweluit  This ain't no done :)

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: haha, ok if you insist...

This does not remove the `_patch_async`, but indeed looks a lot cleaner

Your branch does give a warning now: any idea where that is from? 

``` 
tests/golem/task/test_requestedtaskmanager.py::TestRequestedTaskManager::test_init_task
  /Users/mwu-gol/src/golem-project/golem/tests/golem/task/test_requestedtaskmanager.py:85: RuntimeWarning: coroutine 'AsyncMock.__call__' was never awaited
    assert mock_client.create_task.called_once_with(
```

golem/task/requestedtaskmanager.py
maaktweluit: ahh, didnt see it was on develop

maaktweluit: fixed! so many comments missed this one :)

tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -0,0 +1,335 @@
...
+        mock_client = self._mock_client_create()
+        self._add_create_task_to_client_mock(mock_client)
+
+        task_id = self._create_task()
+        # when
+        yield self._coro_to_def(self.rtm.init_task(task_id))
+        row = RequestedTask.get(RequestedTask.task_id == task_id)
+        # then
+        assert row.status == TaskStatus.creating
+        assert mock_client.create_task.called_once_with(
```
Krigpl:
```suggestion
        mock_client.create_task.assert_called_once_with(
```
And the warning disappears (I dunno why, tho).

maaktweluit: thanks, that solved it :) merged your branch into this one

golem/task/taskserver.py
Krigpl: That is not true actually - only the directory from the `get_task_computer_root` is available in docker. So we need to change the docker config to include RTM's directory too. And then it could be a separate one.

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +56,389 @@ class SubtaskDefinition:
...
+            subtask.save()
+
+        # self.notice_task_updated(task_id, op=TaskOp.ABORTED)
+
+        await self._shutdown_app_client(task.app_id)
+
+    async def _get_app_client(
+            self,
+            app_id: str,
+            env_id: EnvId
```
Krigpl: You don't need the env param anymore since it's determined by the app.

golem/task/requestedtaskmanager.py
Krigpl: But @Wiezzel is still right that `environment` doesn't belong in golem_params, so please remove it from there.

golem/task/requestedtaskmanager.py
maaktweluit: good one, fixed!

should have added a fixme, remember thinking about this before

golem/task/taskserver.py
maaktweluit: What is not true?

Both `__get_task_manager_root` and `get_task_computer_root` return the same folder ( 2x hardcoded :( ) and this is the one shared by docker

Sharing on windows is quite a pain, so changing the docker config would be a separate issue. I'm not sure if its worth it to make multiple shares since the new tasks will replace the old, leaving only one share in the end.

Krigpl: Ok, I didn't realize that those two methods return the same directory...

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +56,386 @@ class SubtaskDefinition:
...
+            task_timeout=golem_params.task_timeout,
+            subtask_timeout=golem_params.subtask_timeout,
+            start_time=default_now(),
+            max_price_per_hour=golem_params.max_price_per_hour,
+            max_subtasks=golem_params.max_subtasks,
+            # Concent is explicitly enabled for task_api for now...
+            concent_enabled=False,
+            # mask = BlobField(null=False, default=masking.Mask().to_bytes()),
+            output_directory=golem_params.output_directory,
+            # FIXME: Where to move resources?
```
Krigpl: Create the directories earlier and copy the resources.

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +56,386 @@ class SubtaskDefinition:
...
+            resources=golem_params.resources,
+            # FIXME: add app_params?
```
Krigpl: What is this FIXME about?

maaktweluit: was a reminder to double check if it was OK to add the app_params in the database like this, think its OK, shall i remove the fixme?

golem/task/requestedtaskmanager.py
maaktweluit: this was also about the field in the database, not sure if we need this anymore

Wiezzel (approved):

golem/task/requestedtaskmanager.py
Krigpl: We still need to copy the resources, but I doubt we need the inputs in the DB.

golem/task/requestedtaskmanager.py
Krigpl: Yeah

tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -0,0 +1,306 @@
...
+        self.public_key = str.encode('0xdeadbeef')
+        self.rtm = RequestedTaskManager(
+            env_manager=self.env_manager,
+            app_manager=self.app_manager,
+            public_key=self.public_key,
+            root_path=self.new_path
+        )
+
+    def tearDown(self):
+        super().tearDown()
```
Krigpl: Not really needed

golem/task/requestedtaskmanager.py
maaktweluit: removed inputs from DB, added copy

golem/task/requestedtaskmanager.py
maaktweluit: done

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: pylint kept telling me this too, i kept thinking it might be useful some day..

removed!

golem/task/requestedtaskmanager.py
```diff
@@ -29,61 +57,389 @@ class SubtaskDefinition:
...
+            name=golem_params.name,
+            status=TaskStatus.creating,
+            environment=golem_params.environment,
+            # prerequisites='{}',
+            task_timeout=golem_params.task_timeout,
+            subtask_timeout=golem_params.subtask_timeout,
+            start_time=default_now(),
+            max_price_per_hour=golem_params.max_price_per_hour,
+            max_subtasks=golem_params.max_subtasks,
+            # Concent is explicitly enabled for task_api for now...
```
Krigpl: Disabled : )

maaktweluit: fixed

golem/database/schemas/035_requestor_models.py
```diff
@@ -0,0 +1,79 @@
...
+        name = pw.CharField(max_length=255, null=True)
+        status = pw.StringEnumField()
+        prerequisites = pw.JsonField(default='{}')
+        task_timeout = pw.IntegerField()
+        subtask_timeout = pw.IntegerField()
+        start_time = pw.UTCDateTimeField(null=True)
+        max_price_per_hour = pw.IntegerField()
+        max_subtasks = pw.IntegerField()
+        concent_enabled = pw.BooleanField(default=False)
+        prerequisites = pw.JsonField(default='[]')
```
Krigpl: This is duplicated with line 31. Also what is this field for?

tests/golem/task/test_requestedtaskmanager.py
Krigpl: Not necessarily in this PR (although please do not forget), but try to just use `async_test` decorator like https://github.com/golemfactory/golem/blob/develop/tests/golem/task/task_api/test_task_api_service.py#L39
I created it specifically to integrate async tests with `unittest.TestCase` and works just fine.

golem/database/schemas/035_requestor_models.py
Wiezzel: Line 31 is correct. It should be a dict. These are serialized provider prerequisites.

maaktweluit: removed line 38 :)

Krigpl (approved):

tests/golem/task/test_requestedtaskmanager.py
maaktweluit: added it to the list of the asyncio conversion PR :) 
https://github.com/golemfactory/golem/pull/4648

PR #4633, opened 2019-08-22
Improvements in Localhost environment

Wiezzel:
* `LocalhostPrerequisites` contain functions instead of return values because those were not flexible enough for more complicated test scenarios (e.g. timeouts, errors).
* Running GRPC server in a subprocess instead of a thread provides a great speed up and allows to implement proper `stop()`.

### DIFF
--- a/tests/golem/envs/localhost.py
+++ b/tests/golem/envs/localhost.py
@@ -1,7 +1,8 @@
 import asyncio
 import logging
+import signal
+from multiprocessing import Process
 from pathlib import Path
-from threading import Thread
 from typing import Optional, Dict, Any, Tuple, List, Awaitable, Callable
 
 from dataclasses import dataclass, asdict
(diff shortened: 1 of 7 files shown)


Wiezzel:
Actually, `LocalhostRuntime.stop` still isn't working but let's ignore this for now.

tests/golem/envs/localhost.py
```diff
@@ -168,17 +168,17 @@ def _spawn_server(self) -> None:
...
         self._started()
         return defer.succeed(None)
 
     def stop(self) -> defer.Deferred:
-        assert self._server is not None
         assert self._server_loop is not None
         assert self._server_thread is not None
         try:
-            self._server_loop.run_until_complete(self._server.stop())
+            # FIXME: This raises 'Cannot close a running event loop'
```
maaktweluit: did you try to await stop first?

Krigpl: Alternatively, you can try not calling `close` altogether since afaiu it's not necessary and `stop` should suffice.

Wiezzel: @maaktweluit How can I await stop? `call_soon_threadsafe` doesn't return an awaitable object.

maaktweluit: ahh, that is a bummer.. did not check the docs, was just a random thought

all examples from a quick search run close in the finally of the try block. Maybe we can move stop and close to the same function and use `.call_soon_threadsafe(` only once?

Wiezzel: >Maybe we can move stop and close to the same function and use .call_soon_threadsafe( only once?

maaktweluit: hahahaha

Ok, i'm out of ideas for now..

Wiezzel: Went with multiprocessing in the end so this is deprecated.

Krigpl (approved):
lgtm, please update the summary to reflect what's really changed

PR #4634, opened 2019-08-22
Various fixes for NewTaskComputer

Wiezzel:
* Do not attempt to send computation results if task was failed or cancelled (because it will obviously fail).
* Call `task_keeper.task_started` and `task_keeper.task_ended` when computing tasks with the `NewTaskComputer`.
* Call task finished callback at the very end of handling computation results.
* Create `ProviderAppClient` the proper (i.e. asynchronous way).
* Convert `app_client.shutdown` future to deferred.
* Use minimum timeout of 1 second for computation, even if deadline has already passed. For unknown reasons deadlines of 0 or less don't work as expected (i.e. don't cancel the future immediately) for GRPC calls.
* Use `addCallbacks` instead of `addBoth` for gathering resources in `TaskServer` (`addBoth` is meant to add a **single callback** that handles both correct results and failures). Also included the `consumeErrors` option for `gatherResults` to avoid having 'Unhandled error in deferred' logs.

### DIFF
--- a/golem/task/taskcomputer.py
+++ b/golem/task/taskcomputer.py
@@ -60,6 +60,7 @@ class TaskComputerAdapter:
     ) -> None:
         self.stats = IntStatsKeeper(CompStats)
         self._task_server = task_server
+        self._finished_cb = finished_cb
         self._old_computer = TaskComputer(
             task_server=task_server,
             stats_keeper=self.stats,
@@ -69,7 +70,6 @@ class TaskComputerAdapter:
         self._new_computer = NewTaskComputer(
             env_manager=env_manager,
             work_dir=task_server.get_task_computer_root(),
-            task_finished_callback=finished_cb,
             stats_keeper=self.stats
         )
         sync_wait(self._new_computer.prepare())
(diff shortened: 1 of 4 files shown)


Wiezzel:
I rebased the branch to reap the fruit of fixes without waiting for #4633 to be fixed.

Krigpl (approved):
obama_approves

golem/task/taskcomputer.py
```diff
@@ -60,6 +60,7 @@ def __init__(
     ) -> None:
         self.stats = IntStatsKeeper(CompStats)
         self._task_server = task_server
+        self._finished_cb = finished_cb
```
maaktweluit, line 63: why is the Adapter now responsible for calling the `_finished_cb`?

I think the new TaskComputer should keep this so its still there when the Adapter is removed

maaktweluit (commented):
LGTM!

One small q, will approve on reply or fix

golem/task/taskcomputer.py
Wiezzel: Nope, this should be moved to `TaskServer` once the adapter is removed. And that's because the callback ought to be called after `send_results` or `send_task_failed`. What this callback actually does is shutting down Golem so its place is at the very end of the whole pipeline.

maaktweluit: Ok, sounds great! Thanks for the clear reply

maaktweluit (approved):

PR #4635, opened 2019-08-22
Tests for integration between TaskServer and NewTaskComputer

Wiezzel:


tests/golem/task/test_taskserver.py
```diff
@@ -1687,3 +1696,222 @@ def test_get_min_performance_for_task(self):
...
+        except AttributeError:
+            pass
+        asyncio.set_event_loop(asyncio.new_event_loop())
+        install_reactor()
+
+    @classmethod
+    def cleanUpClass(cls):
+        uninstall_reactor()
+
+    @patch('golem.task.taskserver.TaskHeaderKeeper', spec=TaskHeaderKeeper)
```
Krigpl, line 1720: Curious - how do you decide when to use spec vs spec_set ?

tests/golem/task/test_taskserver.py
```diff
@@ -1687,3 +1696,222 @@ def test_get_min_performance_for_task(self):
         # Then
         self.ts.get_environment_by_id.assert_called_once()
         self.assertEqual(result, 0.0)
+
+
+class TestNewTaskComputerIntegration(
```
Krigpl, line 1701: Considering this is an integration test and doesn't share anything with the rest of the tests here I'd say to put this class in its own file.

tests/golem/task/test_taskserver.py
```diff
@@ -1687,3 +1696,222 @@ def test_get_min_performance_for_task(self):
...
+        testutils.TestWithClient,
```
Krigpl, line 1702: Depends how much you wanna cover by this test, but I'd say client is not needed for testing the integration between these two components.

tests/golem/task/test_taskserver.py
Wiezzel: Quite simple: if using `spec_set` raises an error then I go with `spec`. :stuck_out_tongue: It usually boils down to assigning some fields of the mocked object since `spec_set` won't allow it.

tests/golem/task/test_taskserver.py
Wiezzel: Oh yes, this one is tricky. So as it may surprise you, `TestWithClient` doesn't instantiate the `Client` class. It just creates a mock with client's spec. And I need one for creating a `TaskServer`. So why not use it?

tests/golem/task/test_taskserver.py
Wiezzel: OK, fair enough.

tests/golem/task/test_taskserver.py
Krigpl: Right, but that's the point of spec_set, not to allow setting attributes that do not exist. Where do we do that?

Wiezzel: For `TaskHeaderKeeper` we're using its `task_headers` field. That won't work with `spec_set`.

For `TaskManager` we're assigning `apps_manager` and `comp_task_keeper` fields. That won't work with `spec_set` either.

Both could be probably resolved by editing the appropriate classes (making fields private and providing accessor methods) but I don't think it's a part of this PR.

tests/golem/task/test_taskserver.py
mfranciszkiewicz: Actually, it would be great to separate integration tests from the rest of the unit tests. Not necessarily in this PR though.

mfranciszkiewicz (approved):
Those pesky unit tests and IMO we're done here

Krigpl (approved):

PR #4638, opened 2019-08-22
[P1] Move transcoding integration tests to a separate file

cameel:
This pull request extracts the codec change / resolution change / frame rate change / segment count tests from `test_ffmpegintegration` into a separate base class and then adds a separate module that runs them on the files from `transcoding-video-bundle`. Now `test_ffmpegintegration` still runs these tests but only on the videos stored in the repository and should be fast enough to be used as a part of CI.

### Notes
1) This pull request does not remove `@ci_skip` from `test_ffmpegintegration`. I left it in so that you can decide for yourself whether you really want to enable these tests in CI.

### Dependencies
This pull request is based on #4615 and should be merged after it. This is just for ease or development - It does not really have to depend on anything between it and `CGI/transcoding/master` but putting it directly on master and then having you merged it with those other branches would be a pain and give you tons of conflicts.

The branches below depend on https://github.com/golemfactory/ffmpeg-tools/pull/12 so tests won't pass before that gets merged.

**NOTE**: I have set the base branch of this pull request to `transcoding-task-d5-z3-test-with-the-number-of-subtasks-equal-to-the-number-of-key-frames` (#4615) rather than `CGI/transcoding/master`. Please remember to change the base branch back to `CGI/transcoding/master` before you merge.


tests/apps/ffmpeg/task/ffmpeg_integration_base.py
```diff
@@ -0,0 +1,358 @@
...
+        # about it and it’s likely to be a bug in ffmpeg. So bad output is
+        # unfortunately the expected result here.
+        'standalone-tra3106[mjpeg,720x496,17s,v1a0s0d0,i1525p1016b1016,29.97fps][segment1of17].avi': {'video': {'frame_rate'}},
+        # This wmv3/wmv file has incorrect FPS in metadata (or at least ffprobe
+        # returns an incorrect value). This makes ffmpeg (incorrectly) double
+        # the number of frames while keeping the FPS in output file metadata the
+        # same as in the original. We can’t help it. We’re going to assume that
+        # this is the expected result, i.e. garbage in, garbage out.”
+        'standalone-catherine[wmv3+wmav2,180x140,42s,v1a1s0d0,i637p1257b631,_][segment1of6].wmv': {'video': {'frame_rate'}},
+    }
```
nieznanysprawiciel: This looks like ignores specific for video bundle - shouldn't be in base class.

tests/apps/ffmpeg/task/ffmpeg_integration_base.py
```diff
@@ -0,0 +1,358 @@
...
+        operation.request_resolution_change(video["resolution"])
+        operation.exclude_from_diff(
+            FfmpegIntegrationBase.ATTRIBUTES_NOT_PRESERVED_IN_CONVERSIONS)
+        operation.exclude_from_diff({'video': {'frame_count'}})
+        operation.enable_treating_missing_attributes_as_unchanged()
+
+        filename = os.path.basename(video['path'])
+        if filename in self._IGNORED_ATTRIBUTES_OF_BROKEN_FILE:
+            operation.exclude_from_diff(
+                self._IGNORED_ATTRIBUTES_OF_BROKEN_FILE[filename])
```
nieznanysprawiciel: Probably we don't need to ignore anything in basic test set??

tests/apps/ffmpeg/task/ffmpeg_integration_base.py
```diff
@@ -0,0 +1,358 @@
...
+
+        if not Container.is_supported(video['container'].value):
+            pytest.skip("Source container not supported")
+        if not Container.is_supported(container.value):
+            pytest.skip("Target container not supported")
+        if not video['container'].is_supported_video_codec(source_codec.value):
+            pytest.skip("Source video codec not supported by the container")
+        if not container.is_supported_video_codec(video_codec.value):
+            pytest.skip("Target video codec not supported by the container")
```
nieznanysprawiciel: We don't won't to skip tests in basic test set. Skips are usefull only in extended test set.

tests/apps/ffmpeg/task/ffmpeg_integration_base.py
```diff
@@ -0,0 +1,358 @@
...
+
+        supported_conversions = source_codec.get_supported_conversions()
+        if video_codec.value in supported_conversions:
+            (_input_report, _output_report, diff) = operation.run(
+                video["path"])
+            self.assertEqual(diff, [])
+        else:
+            with self.assertRaises(UnsupportedVideoCodecConversion):
+                operation.run(video["path"])
+            pytest.skip("Video codec conversion not supported")
```
nieznanysprawiciel: Are there any conversions that are expected to fail in basic test set ?

tests/apps/ffmpeg/task/ffmpeg_integration_base.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def split_and_merge_with_codec_change(
+            self,
+            video,
+            video_codec,
+            container):
+        # FIXME: These tests should be re-enabled once all the fixes needed
+        # to make them pass are done and merged.
+        if video["path"].startswith("videos/"):
+            pytest.skip("Files from transcoding-video-bundle disabled for now")
```
nieznanysprawiciel: Code for video bundle only. By the way it would be better to have single switch somwhere to enable video bundle test. The best option is flag added to pytest call, but if it requires to much efford, switch could be somwhere in TestFfmpegIntegrationFullBundleSet class.

nieznanysprawiciel (changes requested):
### Please read this comment, before you dive into details!!

This PR is not what I expected, maybe I didn't describe enough details.

We need 2 groups of tests: first are integration tests that will help us avoid regression, second group are tests using video bundle, that will be probably used for further experiments or will be run ocasionally on image updates.

Since there're different goals, we need different implementation and in this PR, testing functions were combined in base class.
What I expect from ffmpegintegration test:
- Someone who reads test body, knows what is happening. Now he must check base class implementation. Implementation should not be hidden
- In regression tests we know what are expected results and we don't need to deduce them. Here are many switches that can skip test in some cases (for example whenn ffmpeg-tools support changes). That means that if something changes, test silently fails. Developer should know that behavior changed and fix test by himself.
- Implementation shouldn't contain any things that are necessary in video bundle tests, but are not necessary in regression tests. There're some things that I listed review, but please check other parts of code as well.

This means that tests code will be partially duplicated between regression tests and bundle tests - and that's ok for me. Of course common functionalities can be moved to base class as long as it gives us better readability. For example moving test name generation really worked and now code looks better. But in general I follow rule, that test should explicite show important things, that it does.

And just a side note: ci_skip is a little bit confusing and it skips test only on mac and windows, where we can't run docker. On linux ci it still runs.

cameel:
Right. You make very good points. I focused too much on reducing duplication but there's some value in those test sets being more explicit and independent of each other too. I have just pushed some changes that hopefully alleviate that.

I moved the test bodies out of the base class and cut down the regression tests so that they contain no code that is not relevant. I have also fixed minor mistakes that I noticed along the way (bugs introduced to functions building test names when they were converted from lambdas, test videos from the repo being on both lists, etc.). You can see my commit descriptions for a detailed list of changes.

> And just a side note: ci_skip is a little bit confusing and it skips test only on mac and windows, where we can't run docker. On linux ci it still runs.

Oh, so this means that the tests on full video bundle will be running in CI once we enable the bundle (and we were going to enable it once all the tests pass). Do we want to do something about it?

tests/apps/ffmpeg/task/ffmpeg_integration_base.py
cameel: Done. Now bundle tests and regression have separate implementations and I removed this bit from the former.

tests/apps/ffmpeg/task/ffmpeg_integration_base.py
cameel: Yes. Every test has at least one unsupported value on its parameter list. For this particular test it's `(VideoCodec.THEORA, Container.c_OGG)`. Conversion to theora is not on the list of available conversions of any codec because it almost always resulted in a bad file in my tests.

I solved this by duplicating the codec/container lists and removing unsupported values from parameter lists of tests that are in `TestFfmpegIntegration`.

tests/apps/ffmpeg/task/ffmpeg_integration_base.py
cameel: Now this is no longer checked in regression tests.

As for the bundle tests: I removed the condition since it's no longer necessary and put the decorator over the whole class. Is that OK for now? It does make it possible to enable/disable those tests in a single place so it's no worse than having a hard-coded switch in class body.

A command-line parameter obviously would be better and adding it does not seem complicated (should work the same way as `@runslow`/`--runslow`) but I'd prefer to do it last so that @pawelkisielewicz can focus on finishing the current tasks.

tests/apps/ffmpeg/task/ffmpeg_integration_base.py
cameel: Done. Now implementations are separate and regression tests no longer have these conditions.

tests/apps/ffmpeg/task/ffmpeg_integration_base.py
cameel: Done. Now bundle tests and regression have separate implementations and I removed this bit from the former.

cameel:
I have just noticed that this branch fails linter checks after my changes even though it passed `./lint.sh` on my machine. One thing was wrong import order (though it should have been detected already earlier; lintdiff is broken as usual). The other was missing `parameterized` in `requirements-lint.txt`. The latter is weird because it did not happen on branches this one is based on and they use that library too. I guess something must have changed in the code that buildbot runs. All I can see now is that it installs stuff from `requirements-lint.txt` which did not contain `parameterized` until now. I suspect that it used to install either `requirements-test.txt` or `requirements-lint_to-freeze.txt` before, either of which would install `parameterized`.

So I'm adding the commit that fixes it here but it's possible that it may have to be cherry-picked to branches below if new commits with review fixes get pushed there and linter starts failing.

cameel:
I have just noticed that this branch fails linter checks after my changes even though it passed `./lint.sh` on my machine. One thing was wrong import order (though it should have been detected already earlier; lintdiff is broken as usual). The other was missing `parameterized` in `requirements-lint.txt`. The latter is weird because it did not happen on branches this one is based on and they use that library too. I guess something must have changed in the code that buildbot runs. All I can see now is that it installs stuff from `requirements-lint.txt` which did not contain `parameterized` until now. I suspect that it used to install either `requirements-test.txt` or `requirements-lint_to-freeze.txt` before, either of which would install `parameterized`.

So I'm adding the commit that fixes it here but it's possible that it may have to be cherry-picked to branches below if new commits with review fixes get pushed there and linter starts failing.

nieznanysprawiciel:
>Oh, so this means that the tests on full video bundle will be running in CI once we enable the bundle (and we were going to enable it once all the tests pass). Do we want to do something about it?

I see you added @pytest.mark.skip. This is probably good enough solution to disable these tests.

nieznanysprawiciel (approved):

nieznanysprawiciel:

PR #4639, opened 2019-08-22
[P4] Transcoding: add option for stripping potentially problematic subtitle and data streams to transcoding task

cameel:
This pull request adds two new prameters to the transcoding task. The parameters allow stripping `data`/`subtitle` streams not present on our whitelist from the video (https://github.com/golemfactory/ffmpeg-tools/pull/15).

The feature is disabled by default in the task because we're essentially stripping information that can be important to the user. In many cases, however, this information is not really critical and user most likely prefers it to be removed than his transcoding to fail so we're now giving him that option. 

### Dependencies
This pull request is based on #4638 and should be merged after it. This is just for ease or development - It does not really depend on that base pull request and could be merged without it if rebased.

The feature depends on changes added in https://github.com/golemfactory/ffmpeg-tools/pull/15. Tests won't pass until that pull request is merged.

**NOTE**: I have set the base branch of this pull request to `transcoding-task-p1-move-integration-tests-to-a-separate-file` (#4638) rather than `CGI/transcoding/master`. Please remember to change the base branch back to `CGI/transcoding/master` before you merge.

### DIFF
--- a/apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
+++ b/apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
@@ -174,20 +174,27 @@ def do_replace(input_file,
                replacement_source,
                output_file,
                stream_type,
-               container=None):
+               container=None,
+               strip_unsupported_data_streams=False,
+               strip_unsupported_subtitle_streams=False):
 
     commands.replace_streams(
         input_file,
         replacement_source,
         output_file,
         stream_type,
-        container)
+        container,
+        strip_unsupported_data_streams,
+        strip_unsupported_subtitle_streams)
 
 
 def do_merge_and_replace(input_file,
                          chunks,
                          output_file,
-                         container=None):
+                         container=None,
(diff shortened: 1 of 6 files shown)


nieznanysprawiciel (changes requested):
Stripping streams happens in merge step. It is far to late from user perspective., because everything is computed and than whole transcoding fails. We need additional validation before task is sent to providers. Probably a good place would be validation after split, when file metadata is already availble.

cameel:
Yeah, I know. We're going to add a validation that rejects a video if it contains subtitle and data stream types not on the whitelist and user does not explicitly set a parameter that tells Golem to strip them. But that's P3. I originally wanted P3 to be done before or together with P4 but P4 went through my review faster and I did not see a reason to hold one back because of the other.

If you prefer to have both in one go, I can just push commits from P3 here once they are ready instead of creating a separate pull request. I was hoping that I'd have those commits by today but unfortunately I still found bits that need to be changed so it went back to @pawelkisielewicz for another round.

nieznanysprawiciel:

PR #4642, opened 2019-08-26
Skip test_localhost_env (until fixed)

Wiezzel:


Krigpl (approved):
youdidntseeanythingpinguin

tests/golem/envs/test_localhost_env.py
```diff
@@ -22,6 +23,7 @@
 )
 
 
+@unittest.skip('To be fixed')
```
Krigpl: If it's gonna take a while you can create a github issue and put the # here

Wiezzel: I won't forget about it since it's blocking #4633 #4634 and #4635

etam: I insist.
It will be useful to have such issue reported on github, to have a reference if something similar happens in the future.

I guess it's related to twisted and reactors, so it's not a unique case.

Wiezzel: There you go.
#4643

etam (approved):
ok

PR #4644, opened 2019-08-26
Create Task API directories before computing the task

Krigpl:
They should exist before we start computing the task.


Wiezzel (approved):
Seems OK, but maybe add an assertion for that in `TestTaskGiven`?

Krigpl:
@Wiezzel 
yes

Issue #2424, opened 2018-03-20
Remove build_dictionary method from CoreTask


badb:
It's there for backward compatibility only. The rendering tasks should move to overriding their own TaskDefinitions instead of overriding `build_dictionary`.

PR #4647, opened 2019-08-27
Enable concent only for blender 2.79

### LINKED ISSUE #4546: Concent Service should be disallowed (and disabled by default) for non-Blender tasks
## Rationale
As long as the Concent Service supports only Blender 2.79 tasks, we need to ensure that any other tasks can only be added with `concent_enabled=False` and when `concent_enabled` is not provided it's set to `True` iff concent service is enabled on Requestor's end _and_ if a task at hand is a Blender 2.79 task.

#### Setting `concent_enabled` by default
When adding tasks when `concent_enabled` is not specified in the task dict, set the `concent_enabled` flag on the task to `True` only if:
1. Concent Service is enabled on this Requestor node
2. The task is a Blender (2.79) task

otherwise, set `concent_enabled` to `False`

#### `concent_enabled` set in the task dict
If the `concent_enabled` flag is set to `True` in the task dict passed to `comp.task.create`, the `concent_enabled` should only be allowed if:
1. Concent Service is enabled on this Requestor node
2. The task is a Blender (2.79) task

otherwise, such task should be disallowed - similarly to how it's currently disallowed to create a Concent-enabled task when the requestor has Concent disabled

https://github.com/golemfactory/golem/blob/e593affaae07ec32f3622f8a1ba53382aa9c0b7e/golem/task/rpc.py#L96-L106

### User Interface

(Most likely, the Electron front-end should also be aware that it can only set `concent_enabled` to `True` for Blender 2.79 tasks)

### Additional tests
* test to verify the correct setting of the `concent_enabled` flag on the `TaskHeader` by default (when not provided in the task dict) when Concent Service is enabled on Requestor's node for: Blender (`True`), gWASM (`False`) and gLambda (`False`)
* test to verify that setting `concent_enabled` to True in the task dict causes the task to be rejected by the `comp.task.create` RPC for gWASM and gLambda and (if missing) that it succeeds for Blender 2.79

#### Base test process
** QA limited to the single feature (in case of very trivial changes)

#### Prerequisites
* Ensure you have a Requestor that's Concent-enabled
* Ensure you have a Provider that's Concent-enabled and for which `golemcli debug rpc golem.concent.required_as_provider` returns `True`

#### Concent enabled for Blender
* Add a Blender task
* The task should be correctly added and computed
* Verify (somehow -> debug logs?) that the task indeed created as `concent-enabled: True`)

#### Concent disabled for gWASM
* Add a gWASM task
* The task should be correctly added and computed
* Verify (somehow -> debug logs?) that the task indeed created as `concent-enabled: False`)

#### Concent disabled for gLambda
* Add a gLambda task
* The task should be correctly added and computed
* Verify (somehow -> debug logs?) that the task indeed created as `concent-enabled: False`)

#### `concent-enabled` allowed for Blender
* Add a Blender task for which the `concent_enabled: True` is specified in the task definition (json)
* The task should be correctly added and computed

#### `concent-enabled` disallowed for gWASM
* Add a gWASM task for which the `concent_enabled: True` is specified in the task definition (json)
* Task creation should fail

#### `concent-enabled` disallowed for gLambda
* Add a gLambda task for which the `concent_enabled: True` is specified in the task definition (json)
* Task creation should fail

### Possible regressions
* any functionality related to Concent enabling and disabling should be checked to still be working correctly and that tasks are correctly created as concent-enabled and concent-disabled as expected

### Development
* [x] Feature implemented
* [x] Feature unit/integration tests implemented

### Dev QA
* [x] Basic tests by the developer pass
* [x] Unit tests pass
* [ ] Golem integration tests pass
* [ ] Concent integration tests pass
* [ ] Concent acceptance tests pass

### QA team
* [ ] Concent enabled for Blender
* [ ] Concent disabled for gWASM
* [ ] Concent disabled for gLambda
* [ ] `concent-enabled` allowed for Blender
* [ ] `concent-enabled` disallowed for gWASM
* [ ] `concent-enabled` disallowed for gLambda

## Issues encountered during QA
_When adding an issue here, please update testing scenarios and QA progress above_
* [ ] link to issue...

shadeofblue:
closes: #4546

### DIFF
--- a/apps/appsmanager.py
+++ b/apps/appsmanager.py
@@ -4,7 +4,7 @@ from collections import OrderedDict
 from importlib import import_module
 
 from golem.config.active import (
-        APP_MANAGER_CONFIG_FILES, CONCENT_SUPPORTED_APPS
+    APP_MANAGER_CONFIG_FILES, CONCENT_SUPPORTED_APPS
 )
 from golem.core.common import get_golem_path
 from golem.environments.environment import SupportStatus
@@ -21,7 +21,7 @@ class App(object):
 
     @property
     def concent_supported(self):
-        return self.task_type_info().id in CONCENT_SUPPORTED_APPS
+        return self.task_type_info().id in CONCENT_SUPPORTED_APPS  # noqa pylint:disable=not-callable
 
 
 class AppsManager(object):
@@ -53,7 +53,7 @@ class AppsManager(object):
                 setattr(app, opt, getattr(module, name))
 
             self.apps[section] = app
-            self.task_types[app.task_type_info().id] = app
+            self.task_types[app.task_type_info().id] = app  # noqa pylint:disable=not-callable
 
     def get_env_list(self):
         return [app.env() for app in self.apps.values()]
--- a/apps/appsmanager.py
+++ b/apps/appsmanager.py
@@ -77,5 +77,5 @@ class AppsManager(object):
     def _benchmark_enabled(env):
         return env.check_support() == SupportStatus.ok()
 
-    def get_app(self, task_type_id: str):
+    def get_app(self, task_type_id: str) -> App:
         return self.task_types.get(task_type_id)
--- a/apps/appsmanager.py
+++ b/apps/appsmanager.py
@@ -3,7 +3,9 @@ from configparser import ConfigParser
 from collections import OrderedDict
 from importlib import import_module
 
-from golem.config.active import APP_MANAGER_CONFIG_FILES
+from golem.config.active import (
+        APP_MANAGER_CONFIG_FILES, CONCENT_SUPPORTED_APPS
+)
 from golem.core.common import get_golem_path
 from golem.environments.environment import SupportStatus
 
@@ -17,11 +19,16 @@ class App(object):
         self.benchmark = None  # inherit from Benchmark
         self.benchmark_builder = None  # inherit from TaskBuilder
 
+    @property
+    def concent_supported(self):
+        return self.task_type_info().id in CONCENT_SUPPORTED_APPS
+
 
 class AppsManager(object):
     """ Temporary solution for apps detection and management. """
     def __init__(self):
         self.apps = OrderedDict()
+        self.task_types = dict()
 
     def load_all_apps(self):
         for config_file in APP_MANAGER_CONFIG_FILES:
@@ -46,6 +53,7 @@ class AppsManager(object):
                 setattr(app, opt, getattr(module, name))
 
             self.apps[section] = app
+            self.task_types[app.task_type_info().id] = app
 
     def get_env_list(self):
         return [app.env() for app in self.apps.values()]
@@ -68,3 +76,6 @@ class AppsManager(object):
     @staticmethod
     def _benchmark_enabled(env):
         return env.check_support() == SupportStatus.ok()
+
+    def get_app(self, task_type_id: str):
+        return self.task_types.get(task_type_id)
--- a/apps/registered_test.ini
+++ b/apps/registered_test.ini
@@ -1,10 +1,3 @@
-[Blender_NVGPU]
-env=apps.blender.blenderenvironment.BlenderNVGPUEnvironment
-builder=apps.blender.task.blenderrendertask.BlenderNVGPURenderTaskBuilder
-task_type_info=apps.blender.task.blenderrendertask.BlenderNVGPUTaskTypeInfo
-benchmark=apps.blender.benchmark.benchmark.BlenderNVGPUBenchmark
-benchmark_builder=apps.blender.task.blenderrendertask.BlenderNVGPURenderTaskBuilder
-
 [Dummy]
 env=apps.dummy.dummyenvironment.DummyTaskEnvironment
 builder=apps.dummy.task.dummytask.DummyTaskBuilder
@@ -18,6 +11,7 @@ builder=apps.wasm.task.WasmTaskBuilder
 task_type_info=apps.wasm.task.WasmTaskTypeInfo
 benchmark=apps.wasm.benchmark.WasmTaskBenchmark
 benchmark_builder=apps.wasm.task.WasmBenchmarkTaskBuilder
+
 [GLambda]
 env=apps.glambda.glambdaenvironment.GLambdaTaskEnvironment
 builder=apps.glambda.task.glambdatask.GLambdaTaskBuilder
--- a/golem/task/rpc.py
+++ b/golem/task/rpc.py
@@ -59,7 +59,7 @@ class CreateTaskError(Exception):
 
 def _validate_task_dict(client, task_dict) -> None:
     task_type = task_dict.get('type')
-    known_task_types = [t for t in client.apps_manager.task_types.keys()]
+    known_task_types = list(client.apps_manager.task_types.keys())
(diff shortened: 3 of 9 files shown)


apps/appsmanager.py
```diff
@@ -17,11 +19,16 @@ def __init__(self):
         self.benchmark = None  # inherit from Benchmark
         self.benchmark_builder = None  # inherit from TaskBuilder
 
+    @property
+    def concent_supported(self):
+        return self.task_type_info().id in CONCENT_SUPPORTED_APPS  # noqa pylint:disable=not-callable
```
etam, line 24: Does `TaskTypeInfo` has `id` field? I think it has `name`.

apps/appsmanager.py
```diff
@@ -17,11 +19,16 @@ def __init__(self):
...
+
 
 class AppsManager(object):
     """ Temporary solution for apps detection and management. """
     def __init__(self):
         self.apps = OrderedDict()
+        self.task_types = dict()
```
etam, line 31: Please merge b0.20 into develop and add type hints here.

apps/appsmanager.py
```diff
@@ -68,3 +76,6 @@ def get_benchmarks(self):
     @staticmethod
     def _benchmark_enabled(env):
         return env.check_support() == SupportStatus.ok()
+
+    def get_app(self, task_type_id: str):
```
etam: Return type hint please

apps/registered_test.ini
```diff
@@ -1,10 +1,3 @@
-[Blender_NVGPU]
```
etam, line 1: Why removed?

etam: because it's already in `registered.ini`

etam: what's the difference between those two files?

etam: `registered_test.ini` is used only it testnet. Sorry for being so verbose.

golem/task/rpc.py
```diff
@@ -58,6 +58,14 @@ class CreateTaskError(Exception):
 
 
 def _validate_task_dict(client, task_dict) -> None:
+    task_type = task_dict.get('type')
+    known_task_types = [t for t in client.apps_manager.task_types.keys()]
```
etam:
```suggestion
    known_task_types = set(client.apps_manager.task_types.keys())
```

apps/appsmanager.py
shadeofblue: that's a bit more involved problem ;)

apps/appsmanager.py
shadeofblue: well, yes, it has now https://github.com/golemfactory/golem/pull/4647/files#diff-798280b9faa873b5f61d16829c133666R52

golem/task/rpc.py
```diff
@@ -103,6 +111,12 @@ def _validate_task_dict(client, task_dict) -> None:
                     else 'disabled'
                 ),
             )
+        if not client.apps_manager.get_app(
+                task_dict['type'].lower()
```
etam: Above you check task type name verbatim, so here `lower` seems useless.

apps/appsmanager.py
shadeofblue:
```suggestion
    def get_app(self, task_type_id: str) -> App:
```

golem/task/rpc.py
```diff
@@ -115,10 +129,13 @@ def validate_client(client):
 
 
 def prepare_and_validate_task_dict(client, task_dict):
+    task_type_id = task_dict.get('type', '').lower()
```
etam, line 132: Can't we just expect task type to be in correct case? Or is there already another place, where this is case-insensitive?

golem/task/rpc.py
shadeofblue: ah, yeah, this code predates the above one, thx

etam: and below you already make this string lowercase

golem/task/rpc.py
etam: Oh, I see. It's below in taskmanager.

Ok, now I'll just stop adding comments on the fly, read the whole thing, and add comments at once, because I'm asking questions and answering most of them myself :)

apps/appsmanager.py
etam: You'll have conflicts to resolve, because I added a ton of type hints in this file in b0.20 :)

golem/task/taskbase.py
```diff
@@ -48,6 +48,13 @@ def for_purpose(self, purpose: TaskPurpose) -> 'TaskTypeInfo':
     def get_preview(cls, task, single=False):
         pass
 
+    @property
+    def id(self) -> str:
```
etam, line 52: I'm concerned that people will get confused about, whether to use `name` or `id`.
We should consistently use one name across whole application and use case-insensitive comparison where required.

etam: If you want to have everything lowercase, then just edit ini files.

apps/appsmanager.py
shadeofblue: yeah, and there are more conflicts between b0.20 and develop already... I don't think it's a problem pertaining to this particular pull request

tests/golem/task/test_rpc.py
```diff
@@ -117,39 +117,57 @@ def add_new_task(task, *_args, **_kwargs):
     ),
 )
 class TestCreateTask(ProviderBase, TestClientBase):
+    @staticmethod
+    def _get_task_dict(**data):
+        task_dict = dummytaskstate.DummyTaskDefinition().to_dict()
```
etam, line 122: This probably should look like this:
```python
task_dict = dummytask.DummyTaskBuilder.build_dictionary(dummytaskstate.DummyTaskDefinition())
```
because `build_dictionary` adds `subtask_data_size` and `difficulty` to dict.
Maybe not required, if tests are working without this. Dunno.

See also #2424

etam (changes requested):
I'm definitely against adding `TaskTypeInfo.id`. See comments I added earlier.

tests/golem/task/test_rpc.py
shadeofblue: > I'm definitely against adding `TaskTypeInfo.id`. See comments I added earlier.

@etam what is the problem with using a normalized form for comparisons?

golem/task/taskbase.py
shadeofblue: well, I don't _wan't_ everything in lowercase -> rather, I wanted to have a way to compare that doesn't care about the case exactly so that one does not need to remember what the case was - especially that there already was a precedent in Golem's code and I'm not super confident that just updating the labels won't have further consequences

apps/appsmanager.py
etam: If you have `get_app` for accessing this dict, then this should probably be a private field.

etam:
In `taskmanager.py` there is already a lot of `.lower()` on task type name. It's already all messed up :(

I can approve this, because it's about concent and blender, not about task type names. This can be cleaned up in a separate issue.

etam (approved):
I'm not quite happy about it, but whatever. Let's clean this case-sensitivity stuff elsewhere.

etam:
#4650

Issue #3965, opened 2019-03-07
Refactoring idea: split TaskSession into TaskSessionRequestor and TaskSessionProvider


etam:
Our task session is clearly asymmetrical. Provider initiates a connection to requestor, they talk about the task, requestor sends data to provider, provider returns results, requestor validates results. Both ends of this session is handled by a single class `TaskSession`. It has references to both `TaskManager`, which is used by requestor and `TaskComputer`, which is used by provider.

I think it makes sense to split this class into two classes `TaskSessionRequestor` and `TaskSessionProvider`.

This might be also a good opportunity to make it use a [state machine](https://wiki.python.org/moin/FiniteStateMachine).

shadeofblue:
@etam there's a pretty good chance that once we implement message queues, the tasksession's scope will change and especially, we will be able to remove state from `TaskSession` altogether... 

cc: @jiivan

etam:
Those message queues: #2223 ? It seems to me that it adds persistence, which allows to resume a session if a node crashes/restarts. Still you need to have handling of those messages implemented somewhere and I still think it's a good idea to do it using a state machine with separated requestor from provider.

shadeofblue:
@etam let's talk about it when @jiivan and you are at the office next Monday so that we can all be on the same page ... I'm just of an opinion that it doesn't make sense to start working on this issue here without taking the message queues into consideration

mfranciszkiewicz:
@shadeofblue @etam Please consult the RnD team first.

etam:
![new ideas](https://user-images.githubusercontent.com/220752/53953775-8b8c8f80-40d4-11e9-8a7e-8024ee6899de.jpg)

Wiezzel:
Naming idea: it should be `ProviderTaskSession` and `RequestorTaskSession` not `TaskSessionProvider` and `TaskSessionRequestor`.

mfranciszkiewicz:
The idea is not new. This is exactly why you should talk to RnD.

etam:
On hold until message queues are implemented.

etam:
Message queues were implemented some time ago.

This can be taken a step further to also split `TaskServer`.

etam:
I've done some research about state machine implementations for Python and this one I think is the best: https://github.com/pytransitions/transitions

jiivan:
> I've done some research about state machine implementations for Python and this one I think is the best: https://github.com/pytransitions/transitions

[...]

Issue #3173, opened 2018-08-13
SubtaskStatus is kept in two places


etam:
1. `CoreTask.subtasks_given[subtask_id]['status']`, where `CoreTask` objects are in `TaskManager.tasks[task_id]`
2. `SubtaskState.subtask_status`, where `SubtaskState` objects are in `TaskManager.tasks_states[task_id].subtasks_states[subtask_id]`

They are used in parallel. For example `TaskManager.restart_subtask` changes the state in `tasks_states` and calls `CoreTask.restart_subtask`, which uses and updates the state in `subtasks_given`.

Fortunately it seems that those two state machines are not mixed anywhere. They just use the same enum.

What needs to be done:
- ~~Make sure there's only one state machine of a subtask.~~ ~~Update: These two state machines should stay separate. See comments below.~~ Update: Or maybe not. See comments below.
- Try to make it explicit with clearly visible transitions.
- DOCUMENTATION!!! (GraphVis diagram would be nice)

etam:
More states could be added, like `timeout` and `verification_failed`.

@Elfoniok also suggested creating a state `verification_finished` that would be used before going to `finished`.

etam:
@badb said:
> the state in task manager represents Golem knowledge about child task state. Core task object represents apps knowlege about child task state.
> Golem knowledge should come from app 
> App, ie. task (so Golem Blender task)
> Golem should not care how apps classes are implented (apart from API (ie specific functions right now))

So, these two state machines should stay separate, but with their own enums. Question is: what states should each have?

etam:
Idea:

Having two separate state machines is just asking for trouble. As I understand, the point is to have some generic API to know the state of tasks, which are not necessarily rendering tasks, while a task itself might have more complicated and detailed internal state machine.

So the proper solution should be that a task keeps it's state internally in whatever way it want, but has a method which translates it to a common/generic state. And if a task state needs to be changed (for example a task is aborted) then there should be a common/generic API method for that.

etam:
There is a new thing called "Task API", that should also be considered here.

etam:
I've done some research about state machine implementations for Python and this one I think is the best: https://github.com/pytransitions/transitions

etam:
![vader meme: I find your lack of state machines disturbing](https://user-images.githubusercontent.com/220752/62705141-27585980-b9ed-11e9-9a74-7004fd828f7f.jpg)

etam:
[...]

PR #4654, opened 2019-08-28
verify SRR reason

### LINKED ISSUE #4217: Provider fails to verify the reason for `SubtaskResultsRejected`
## Description

**Golem Version**: `develop`
**Branch** (if launched from source): `develop`

**Description of the issue**:

Golem doesn't verify the reason for `SubtaskResultsRejected`

**Actual result**:
Whatever the reason, Golem is always trying to reach out to the Concent Service if Concent is enabled for the subtask

## Expected behavior + solution

Golem should first verify the reason for `SubtaskResultsRejected` and should:
1) trigger Concent UC4 (Additional Verification) only in case of `VerificationNegative`
2) if it receives an SRR with `ForcedResourcesFailure` and an attached `ForceGetTaskResultFailed`, it should verify if the FGTRF message is indeed signed by the Concent and pertains to the same subtask -> in this case, the Provider needs to stop processing the subtask and just mark the subtask as failed locally with an apropriate log message [ depends on https://github.com/golemfactory/golem-messages/issues/346 ]
3) if it receives an SRR with any other reason and the SRR comes from the Requestor - it should trigger UC3 (Force Accept) (as if it hadn't received the SRR at all - iow, normally, after the timeout: https://github.com/golemfactory/golem/blob/ee5ab87dc1f889c2584b6b2163feb7e9f29b29c9/golem/task/tasksession.py#L841-L851) so that the SRR can be verified independently by the Concent
4) We need to verify that in case the SRR issued is not valid, the Concent triggers a UC3 (Force Accept) payment. There are only two valid reasons for the SRR:
a) genuine `VerificationNegative` -> which will trigger Concent's UC4 (Additional Verification) later on
b) a `ForcedResourcesFailure` in which case the message needs to contain a valid `ForceGetTaskResultFailed` signed by the Concent itself [ depends on https://github.com/golemfactory/golem-messages/issues/346 ]

Any other reason here should be treated the same way as an inability for Concent to receive a correct SRR and should result in UC3 (Force Accept) payment.

### Additional tests
* Add (ensure exits) a Concent Acceptance test that verifies that if an SRR issued during UC3 (Force Accept) has any reason besides 4a + 4b above, Concent treats that as a failure on Requestor's part
* Add (ensure exists) a Concent Acceptance test that verifies that if an SRR issued during UC3 (Force Accept) contains an FGTRF message from Concent UC2 (Force Download) then the UC3 (Force Accept) correctly delivers said SRR to the Provider

etam:
resolves #4217

### DIFF
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -808,7 +808,7 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin):
             )
 
         if msg.task_to_compute.concent_enabled:
-            self._handle_srr_with_concent_enabled(msg)
+            self._handle_srr_with_concent_enabled(msg, subtask_rejected)
         else:
             subtask_rejected()
 
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -839,8 +839,12 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin):
             subtask_rejected()
             return
 
-        # wait for timeout to trigger force accept, so that the SRR can be
-        # verified independently by the Concent
+        # in case the reason for SRR is neither
+        # a `VerificationNegative` nor `ForcedResourcesFailure`
+        # the SRR is effectively broken so we're ignoring it
+        # instead
+        # we wait for timeout to trigger force accept,
+        # so that the SRR can be verified independently by the Concent
 
     def _trigger_concent_additional_verification(
             self, msg: message.tasks.SubtaskResultsRejected):
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -796,57 +796,90 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin):
             'ForceSubtaskResults',
         )
 
-        if msg.task_to_compute.concent_enabled:
-            # if the Concent is enabled for this subtask, as a provider,
-            # knowing that we had done a proper job of computing it,
-            # we are delegating the verification to the Concent so that
-            # we can be paid for this subtask despite the rejection
-
-            amount, expected = msg_helpers.provider_deposit_amount(
-                subtask_price=msg.task_to_compute.price,
-            )
-
-            def ask_for_verification(_):
-                srv = message.concents.SubtaskResultsVerify(
-                    subtask_results_rejected=msg
-                )
-                srv.sign_concent_promissory_note(
-                    deposit_contract_address=self.deposit_contract_address,
-                    private_key=self.my_private_key,
-                )
-
-                self.concent_service.submit_task_message(
-                    subtask_id=msg.subtask_id,
-                    msg=srv,
-                )
-
-            self.task_server.client.transaction_system.\
-                validate_concent_deposit_possibility(
-                    required=amount,
-                    tasks_num=1,
-                )
-            self.task_server.client.transaction_system.concent_deposit(
-                required=amount,
-                expected=expected,
-            ).addCallback(ask_for_verification).addErrback(
-                lambda failure: logger.warning(
-                    "Additional verification deposit failed %s", failure.value,
-                ),
-            )
-
-        else:
+        def subtask_rejected():
             dispatcher.send(
                 signal='golem.message',
                 event='received',
                 message=msg
             )
-
             self.task_server.subtask_rejected(
                 sender_node_id=self.key_id,
                 subtask_id=subtask_id,
             )
 
+        if msg.task_to_compute.concent_enabled:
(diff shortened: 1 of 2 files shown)


golem/task/tasksession.py
```diff
@@ -791,62 +791,94 @@ def _react_to_subtask_results_rejected(
                 RequestorCheckResult.OK:
             self.dropped()
             return
+        dispatcher.send(
```
shadeofblue: @etam I'm unsure this is what we want -> this call mainly serves to update the provider stats ... so, in other words, to get to know the final resolution of the subtask ... normally, when concent is disabled for the task, reception of the SRR equals failure and nothing more can be done... with Concent enabled, the verdict of the additional verification by the Concent that still confirms the verification as rejected is the final failure of the subtask - and just receiving an SRR from the Requestor is not...

golem/task/tasksession.py
```diff
@@ -791,62 +791,94 @@ def _react_to_subtask_results_rejected(
...
+                subtask_id=subtask_id,
             )
+            return
 
-        else:
-            dispatcher.send(
-                signal='golem.message',
-                event='received',
-                message=msg
+        # wait for timeout to trigger force accept, so that the SRR can be
```
shadeofblue: there should be some suggestion that this is the "`else`" part ... otherwise, it looks like this comment is "orphaned"...

golem/task/tasksession.py
```diff
@@ -791,62 +791,94 @@ def _react_to_subtask_results_rejected(
...
+                and (msg.report_computed_task.task_to_compute.subtask_id ==
+                     fgtrf_msg.task_to_compute.subtask_id):
+            subtask_id = msg.report_computed_task.subtask_id
+            logger.info("Received ForcedResourcesFailure message. "
+                        "subtask_id=%s", subtask_id)
+            self.task_server.subtask_rejected(
+                sender_node_id=self.key_id,
+                subtask_id=subtask_id,
             )
+            return
```
shadeofblue, line 840: e.g. here, we should dispatch that failure signal

shadeofblue (commented):
I believe that failure reception signal should not be dispatched until a failure is indeed recognized as final...

golem/task/tasksession.py
etam: https://github.com/golemfactory/golem/issues/3173#issuecomment-519507601

etam: \<rant\>
If we had properly implemented state machines, I wouldn't need to think about this. It would be that when entering `subtask_rejected` state, no matter from which state, this signal and task server would be properly called.
\</rant\>

etam: And here's an issue for that: #3965

shadeofblue: still, the issue persists...

etam: Sure. I'm not saying that I won't fix that, just ranting about how non-obvious implicit state machines are.

etam: Done

golem/task/tasksession.py
etam: I can remove the empty newline.

But doing
```python
if ...:
    return ...
else:
    whatever
<end of function>
```
is considered bad style in python

shadeofblue: well, sure, but you can always put this information into the comment to make the comment itself more informative ...

shadeofblue: (which I just did ... )

golem/task/tasksession.py
```diff
@@ -796,58 +796,95 @@ def _react_to_subtask_results_rejected(
...
+
+        self.task_server.client.transaction_system.\
+            validate_concent_deposit_possibility(
+                required=amount,
+                tasks_num=1,
+            )
+        self.task_server.client.transaction_system.concent_deposit(
+            required=amount,
+            expected=expected,
+        ).addCallback(ask_for_verification).addErrback(
```
jiivan, line 882: Refactoring idea: use `@inlineCallbacks` to increase readability.

jiivan (approved):

golem/task/tasksession.py
```diff
@@ -796,58 +796,95 @@ def _react_to_subtask_results_rejected(
...
                 message=msg
             )
-
             self.task_server.subtask_rejected(
                 sender_node_id=self.key_id,
                 subtask_id=subtask_id,
             )
 
+        if msg.task_to_compute.concent_enabled:
+            self._handle_srr_with_concent_enabled(msg)
```
jiivan:
```suggestion
            self._handle_srr_with_concent_enabled(msg, subtask_rejected)
```

shadeofblue (approved):

PR #4655, opened 2019-08-28
a work-around for the benchmark problem on Windows

shadeofblue:


mfranciszkiewicz (changes requested):
This workaround is for benchmarks (first run) only and won't affect computations / task testing / verification on consecutive runs.

It would be better to implement the quickfix in a method that is always executed during startup / VM restarts (f.e. `VirtualBoxHypervisor.start_vm`). The equivalent of `docker-machine ssh golem ...` in class methods would be:
```
self.command('execute', args=[
   self._vm_name,
   'sudo /sbin/mount.vboxsf c/Users /c/Users',
])
```

shadeofblue:
> This workaround is for benchmarks (first run) only and won't affect computations / task testing / verification on consecutive runs.
> 
> It would be better to implement the quickfix in a method that is always executed during startup / VM restarts (f.e. `VirtualBoxHypervisor.start_vm`). The equivalent of `docker-machine ssh golem ...` in class methods would be:
> 
> ```
> self.command('execute', args=[
>    self._vm_name,
>    'sudo /sbin/mount.vboxsf c/Users /c/Users',
> ])
> ```

sigh... thanks...

shadeofblue:
superseded by: https://github.com/golemfactory/golem/pull/4661

Issue #3926, opened 2019-02-22
Enable json serializer in crossbar router


jiivan:
Provide a way for CLIs to connect with json serializer.
https://github.com/golemfactory/golem/blob/c8fcef1ed252ab381302efc5f225ece908f91a1b/golem/rpc/router.py#L26-L27

Use `--xbar-serializer=msgpack|json` or do it inother way.

etam:
merged to b0.20

PR #4656, opened 2019-08-28
Revert "Enable json serializer in crossbar router"

etam:
This reverts commit 51d8a97c25fb9ddf7ffa2a237d5cce85372e57f9.

This was introduced in #3926, but it's not needed anymore.


shadeofblue:
@etam please rebase that against `develop` -> it's not important for this release

etam:
rebased

kmazurek (approved):
keanu_ok

shadeofblue (approved):

etam:
Somehow this change is making integration tests fail with `ReactorAlreadyInstalledError`. I can reproduce this on my PC.

etam:
It fails, because I removed importing `golem.rpc.router` from `golemapp.py`. Which is totally weird and it shouldn't behave like that.

etam:
I pinned down a bit this twisted weird behavior: In `golem/network/hyperdrive/client.py` there is `from twisted.web.client import readBody`, which internally does `from twisted.internet import [...] reactor`. This installs default reactor. Somehow importing `golem.rpc.router` in `golemapp.py` inhibits installing reactor.

Edit: `import golem.rpc.router` can be replaced by more precise `import crossbar.personality`.

mfranciszkiewicz:
@etam Also `from twisted.web.client import readBody` can be safely converted to a local import in `_async_request`

etam:
> @etam Also `from twisted.web.client import readBody` can be safely converted to a local import in `_async_request`

This seems to be not enough. I've made a workaround for now. Let's resolve this in a separate task.

PR #4657, opened 2019-08-28
Manager class for applications using Task API

Wiezzel:
Stores application definitions and state (enabled/disabled). Uses EnvironmentManager to enforce that the appropriate environment is available and enabled before enabling an application. Definitions can be loaded from JSON files.

### DIFF
--- a/golem/app_manager.py
+++ b/golem/app_manager.py
@@ -1,0 +1,98 @@
+import json
+import logging
+from pathlib import Path
+from typing import Any, Dict, List, Iterator
+
+from dataclasses import dataclass, field
+from dataclasses_json import config, dataclass_json
+from marshmallow import fields as mm_fields
+
+from golem.task.envmanager import EnvironmentManager
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass_json
+@dataclass
+class AppDefinition:
+    name: str
+    requestor_env: str
+    requestor_prereq: Dict[str, Any] = field(metadata=config(
+        encoder=json.dumps,
+        decoder=json.loads,
+        mm_field=mm_fields.Dict(keys=mm_fields.Str())
+    ))
+    max_benchmark_score: float
+    version: str = '0.0'
+    description: str = ''
+    author: str = ''
+    license: str = ''
+
+
+def load_app_from_json_file(json_file: Path) -> AppDefinition:
+    """ Parse application definition from the given JSON file. Raise ValueError
+        if the given file doesn't contain a valid definition. """
+    try:
+        app_json = json_file.read_text(encoding='utf-8')
+        # pylint: disable=no-member
+        return AppDefinition.from_json(app_json)  # type: ignore
+        # pylint: enable=no-member
+    except (OSError, ValueError, KeyError):
+        msg = f"Error parsing app definition from file '{json_file}'."
+        logger.exception(msg)
+        raise ValueError(msg)
+
+
+def load_apps_from_dir(app_dir: Path) -> Iterator[AppDefinition]:
+    """ Read every file in the given directory and attempt to parse it. Ignore
+        files which don't contain valid app definitions. """
+    for json_file in app_dir.iterdir():
+        try:
+            yield load_app_from_json_file(json_file)
+        except ValueError:
+            continue
+
+
+class AppManager:
+    """ Manager class for applications using Task API. """
+
+    def __init__(self, env_manager: EnvironmentManager) -> None:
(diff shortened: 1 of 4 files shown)


golem/app_manager.py
```diff
@@ -0,0 +1,99 @@
...
+    """ Manager class for applications using Task API. """
+
+    def __init__(self, env_manager: EnvironmentManager) -> None:
+        self._env_manager = env_manager
+        self._apps: Dict[str, AppDefinition] = {}
+        self._state: Dict[str, bool] = {}
+
+    def register_app(self, app: AppDefinition) -> None:
+        """ Register an application in the manager. """
+        if app.name not in self._apps:
```
Krigpl: I think it should throw when given name is already registered

golem/app_manager.py
```diff
@@ -0,0 +1,99 @@
...
+            self._apps[app.name] = app
+            self._state[app.name] = False
+            logger.info("Application '%s' registered.", app.name)
+
+    def register_app_from_json_file(self, json_file: Path) -> None:
```
Krigpl: I don't think this method as well as `register_apps_from_dir` should exist within this class.
These are convenience methods and don't require access to any internal state (they just use public api of this class) and it makes sense to have them as standalone functions on a module level. Keeps the responsibility of this class clear. Also if we come up with other ways to register the apps (like from db or some other format) it doesn't make sense to keep extending this class with these helpers methods.

golem/app_manager.py
Wiezzel: Well, ok.

golem/app_manager.py
Wiezzel: Well it looks much more natural for me to call it like this:
```python
app_manager.register_apps_from_dir(json_file)
```
Then like this:
```python
register_apps_from_dir(app_manager, json_file)
```
Functions mutating their arguments are just awkward.

If you insist on moving these out of `AppManager` than I would opt for just changing it into pure-parsing functions (i.e. without registering).

Krigpl: Second approach makes it clear that only public interface is used during that operation which makes it easier to maintain, refactor and test.
But I'm fine either way, leaving that to you.

Krigpl (approved):

golem/app_manager.py
```diff
@@ -0,0 +1,98 @@
...
+from golem.task.envmanager import EnvironmentManager
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass_json
+@dataclass
+class AppDefinition:
+    name: str
+    requestor_env: str
```
mfranciszkiewicz, line 19: How does Golem determine the environment for providers? The Task API doesn't have means to provide that kind of information for a `TaskHeader` object (and IMO shouldn't). 

Unless I'm missing something, there is no other source of information besides `AppDefinition`. `provider_env` might need to be specified here.

Wiezzel: It should be returned from `CreateTask`. Take a look at CLAY-432. Specifying it in `AppDefinition` would limit applications to using one provider environment only and I don't think this is what we want.

mfranciszkiewicz: Well, Blender with CPU / GPU is a good example. We may want to extend `Compute` with environment id (and maybe configuration) to prevent apps from guessing.

mfranciszkiewicz (approved):

golem/app_manager.py
```diff
@@ -0,0 +1,98 @@
...
+        return app_name in self._state and self._state[app_name]
+
+    def set_enabled(self, app_name: str, enabled: bool) -> None:
+        """ Enable or disable an application. Raise an error if the application
+            is not registered or the environment associated with the application
+            is not available. """
+        if app_name not in self._apps:
+            raise ValueError(f"Application '{app_name}' not registered.")
+        env_id = self._apps[app_name].requestor_env
+        if not self._env_manager.enabled(env_id):
```
Krigpl, line 84: EnvManager is only required for this single sanity check, I wonder if it's justified to make it a dependency. Especially that you can disable the environment just after enabling the app, thus introducing "invalid" state anyway. Maybe just performing these checks in the RequestedTaskManager when necessary is better.

Krigpl: My reasoning is basically that if you don't allow to enable an app when the env is disabled you should disable all the apps when disabling the env as well - which we're not doing. It'd be circular dependency meaning that these classes should be merged, which doesn't make sense. Thus removing this check makes sense to me.

PR #4648, opened 2019-08-27
Make RequestedTaskManager tests use asyncio over Twisted

maaktweluit:
This cleanup PR is from all leftover comments from #4632

~in progress~

~Change base to develop after the other PR is merged~

- [x] Make `DatabaseFixture` independent of `unittest.TestCase`
- [x] Remove all Twisted logic from `test_requestedtaskmanager.py`
- [x] Remove all TestCase logic from `test_requestedtaskmanager.py`
  - [x] `assert`
  - [x] `addCleanup(`
- [x] Bonus: use `pytest`'s `tmpdir` fixture for the temporary directory
- [x] Bonus: Use mocks like: https://github.com/golemfactory/golem/pull/4632#discussion_r318147420 or https://stackoverflow.com/questions/32480108/mocking-async-call-in-python-3-5
 - [x] Bonus: try to just use async_test decorator like https://github.com/golemfactory/golem/blob/develop/tests/golem/task/task_api/test_task_api_service.py#L39

PR #4664, opened 2019-08-30
Task API integration test script

Krigpl:
Just the bug in the Task API startup is an issue here (those time.sleep calls).

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -272,7 +272,7 @@ class RequestedTaskManager:
             subtask_id=result.subtask_id,
             status=SubtaskStatus.starting,
             payload=result.params,
-            inputs=result.resources,
+            inputs=list(map(str, result.resources)),
             start_time=default_now(),
             price=task.max_price_per_hour,
             computing_node=node,
--- a/golem/task/task_api/__init__.py
+++ b/golem/task/task_api/__init__.py
@@ -75,4 +75,7 @@ class EnvironmentTaskApiService(TaskApiService):
     async def wait_until_shutdown_complete(self) -> None:
         assert self._runtime is not None
         loop = asyncio.get_event_loop()
-        await self._runtime.wait_until_stopped().asFuture(loop)
+        try:
+            await self._runtime.wait_until_stopped().asFuture(loop)
+        finally:
+            await self._runtime.clean_up().asFuture(loop)
--- a/scripts/task_api_tests/basic_integration.py
+++ b/scripts/task_api_tests/basic_integration.py
@@ -1,0 +1,211 @@
+from pathlib import Path
+from typing import List
+from unittest import mock
+import click
+import json
+import logging
+import shutil
+import tempfile
+import time
+
+from golem import app_manager as appmanager, database, model
+from golem.core.common import install_reactor
+from golem.core.deferred import deferred_from_future
+from golem.task import envmanager, requestedtaskmanager, taskcomputer
+from golem.task.task_api import docker
+from golem.envs.auto_setup import auto_setup
+from golem.envs.docker import cpu, whitelist
+
+from twisted.internet.task import react
+from twisted.internet.defer import ensureDeferred
+
+logging.basicConfig(level=logging.INFO)
+
+
+async def test_task(
+        work_dir: Path,
+        environment: str,
+        env_prerequisites_json: str,
+        task_params_path: Path,
+        resources: List[Path],
+        max_subtasks: int,
+) -> None:
+    env_prerequisites = json.loads(env_prerequisites_json)
+
+    app_manager = appmanager.AppManager()
+    app_name = 'test_app'
+    app_manager.register_app(appmanager.AppDefinition(
+        name=app_name,
+        requestor_env=environment,
+        requestor_prereq=env_prerequisites,
+        max_benchmark_score=1.0,
+    ))
+    app_manager.set_enabled(app_name, True)
+
+    env_manager = envmanager.EnvironmentManager()
+    docker_cpu_config = cpu.DockerCPUConfig(work_dirs=[work_dir])
+    docker_cpu_env = auto_setup(cpu.DockerCPUEnvironment(docker_cpu_config))
+    docker_image = cpu.DockerCPUEnvironment.parse_prerequisites(
+        env_prerequisites).image
+    whitelist.Whitelist.add(docker_image)
+    env_manager.register_env(
+        docker_cpu_env,
+        cpu.DOCKER_CPU_METADATA,
+        docker.DockerTaskApiPayloadBuilder,
+    )
+    env_manager.set_enabled(cpu.DOCKER_CPU_METADATA.id, True)
+
+    rtm_work_dir = work_dir / 'rtm'
+    rtm_work_dir.mkdir()
+    rtm = requestedtaskmanager.RequestedTaskManager(
+        env_manager,
+        app_manager,
+        public_key=b'',
+        root_path=rtm_work_dir,
+    )
+
+    task_computer_work_dir = work_dir / 'computer'
+    task_computer_work_dir.mkdir()
+    task_computer = taskcomputer.NewTaskComputer(
+        env_manager,
+        task_computer_work_dir,
+    )
+
+    output_dir = work_dir / 'output'
+    output_dir.mkdir()
+    golem_params = requestedtaskmanager.CreateTaskParams(
+        app_id=app_name,
+        name='testtask',
+        task_timeout=3600,
+        subtask_timeout=3600,
+        output_directory=output_dir,
+        resources=resources,
+        max_subtasks=max_subtasks,
+        max_price_per_hour=1,
+        concent_enabled=False,
+    )
+    with open(task_params_path, 'r') as f:
+        task_params = json.load(f)
+    task_id = rtm.create_task(golem_params, task_params)
+    print('Task created', task_id)
+    await deferred_from_future(rtm.init_task(task_id))
+    rtm.start_task(task_id)
+    print('Task started')
+
+    assert await deferred_from_future(rtm.has_pending_subtasks(task_id))
+    computing_node = \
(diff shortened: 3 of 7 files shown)


golem/task/requestedtaskmanager.py
```diff
@@ -360,6 +368,8 @@ def get_subtasks_outputs_dir(self, task_id: TaskId) -> Path:
             logger.info(
                 'app_client created for app_id=%r, clients=%r',
                 app_id, self._app_clients[app_id])
+            import time
+            time.sleep(2)
```
Wiezzel: 😿

golem/task/task_api/__init__.py
```diff
@@ -70,3 +70,4 @@ def running(self) -> bool:
         assert self._runtime is not None
         loop = asyncio.get_event_loop()
         await self._runtime.wait_until_stopped().asFuture(loop)
+        await self._runtime.clean_up().asFuture(loop)
```
Wiezzel: How about this?
```python3
try:
    await self._runtime.wait_until_stopped().asFuture(loop)
finally:
    await self._runtime.clean_up().asFuture(loop)
```

golem/task/taskcomputer.py
```diff
@@ -320,6 +320,8 @@ def compute(self) -> defer.Deferred:
         )
 
         self._app_client = await ProviderAppClient.create(task_api_service)
+        import time
+        time.sleep(2)
```
Wiezzel: 😿

scripts/task_api_tests/basic_integration.py
```diff
@@ -0,0 +1,191 @@
...
+from golem.task.task_api import docker
+from golem.envs.auto_setup import auto_setup
+from golem.envs.docker import cpu, whitelist
+
+from twisted.internet.task import react
+from twisted.internet.defer import ensureDeferred, Deferred
+
+import logging
+logging.basicConfig(level=logging.INFO)
+# logging.basicConfig(level=logging.DEBUG)
```
Wiezzel: Plz, remove commented line. Also the import order looks weird.

scripts/task_api_tests/basic_integration.py
```diff
@@ -0,0 +1,191 @@
...
+    )
+    env_manager.set_enabled(environment, True)
+
+    env_prerequisites = json.loads(env_prerequisites_json)
+
+    rtm_work_dir = work_dir / 'rtm'
+    rtm_work_dir.mkdir()
+    rtm = requestedtaskmanager.RequestedTaskManager(
+        env_manager,
+        b'',
```
Wiezzel: Better pass parameters by keywords here because this empty string looks obscure.

golem/task/requestedtaskmanager.py
Krigpl: Dunno what to do about it right now, tbh

Wiezzel: Me neither, but this makes me sad.

scripts/task_api_tests/basic_integration.py
```diff
@@ -0,0 +1,198 @@
...
+from golem.envs.docker import cpu, whitelist
+
+from twisted.internet.task import react
+from twisted.internet.defer import ensureDeferred, Deferred
+
+logging.basicConfig(level=logging.INFO)
+
+
+def as_deferred(f):
+    return Deferred.fromFuture(asyncio.ensure_future(f))
```
Wiezzel: Plz, use `golem.core.deferred_from_future`
https://github.com/golemfactory/golem/blob/239d47d5953832d287302b2aad1bb210df29fa69/golem/core/deferred.py#L67-L68

Wiezzel: Actually, you can modify `golem.core.deferred_from_future` to accept any `Awaitable` object by putting `ansycio.ensure_future()` there.

golem/task/requestedtaskmanager.py
```diff
@@ -362,6 +362,8 @@ def get_subtask_outputs_dir(self, task_id: TaskId) -> Path:
...
```
maaktweluit: Also makes me sad, maybe we can at least use asyncio?

```suggestion
            await asyncio.sleep(2)
```

golem/task/taskcomputer.py
maaktweluit: same as the other sleep:
:( and async?

scripts/task_api_tests/basic_integration.py
```diff
@@ -0,0 +1,196 @@
...
+        resources=resources,
+        max_subtasks=max_subtasks,
+        max_price_per_hour=1,
+        concent_enabled=False,
+    )
+    with open(task_params_path, 'r') as f:
+        task_params = json.load(f)
+    task_id = rtm.create_task(golem_params, task_params)
+    print('Task created', task_id)
+    await deferred_from_future(rtm.init_task(task_id))
```
maaktweluit, line 91: why do you need a deferred here? i think `await rtm.init_task(..)` would work to

scripts/task_api_tests/basic_integration.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+async def _task(
+        environment,
+        env_prerequisites,
+        task_params_path,
+        resource,
+        max_subtasks,
+) -> None:
+    work_dir = Path(tempfile.mkdtemp())
```
maaktweluit: maybe we can use `tmp_path` fixture from pytest here too?

scripts/task_api_tests/basic_integration.py
```diff
@@ -0,0 +1,196 @@
...
+    print('work_dir', work_dir)
+    db = database.Database(
```
maaktweluit, line 156: we can use `AsyncDatabaseFixture` for this too https://github.com/golemfactory/golem/pull/4648/files#diff-70d60ac05925af68aa36648b86338e99R181

maaktweluit (changes requested):
LGTM!

Only the deferred q is important, the others are minor.

I know this is not using `pytest` now, but think it would be nice to share common components between tests.

Krigpl:
@maaktweluit The reason I didn't use pytest is that it needs to take command lint arguments to run. Does pytest support that?

golem/task/requestedtaskmanager.py
Krigpl: Maybe let's just move it to the Task API repo since it's an issue there.

maaktweluit: 👍

maaktweluit:
@Krigpl yes 
you need to add them to the options in `conftest.py`, like https://github.com/golemfactory/golem/blob/develop/scripts/node_integration_tests/conftest.py#L39

golem/task/requestedtaskmanager.py
mfranciszkiewicz: Server health check may help here https://github.com/golemfactory/task-api/pull/36

scripts/task_api_tests/basic_integration.py
Wiezzel: Yea, this is a very good point. Why do you even need deferreds when the method is `async`?

Krigpl: This is some magic related to incorporating twisted and asyncio together.
I followed this article https://meejah.ca/blog/python3-twisted-and-asyncio
It doesn't work without converting to deferred.
tl;dr; of the article is that you need to be aware of what context are you really in when inside an async function - twisted vs asyncio.

Krigpl:
@maaktweluit I find it a little awkward to introduce global flag available for 1k+ tests for the sake of a single test. Can still do it if you think it's worth it.

Krigpl:
Also this isn't meant to be run together with the rest of the test suite. It just doesn't make much sense to me to use pytest here.

Wiezzel:
> Also this isn't meant to be run together with the rest of the test suite. It just doesn't make much sense to me to use pytest here.

This is a fair point. It was meant to be a tool for developing apps, not for testing the Golem codebase.

scripts/task_api_tests/basic_integration.py
maaktweluit: ahh, clear, had that fight a few times too the last weeks

maybe we can remove twisted from this test, would be nice to have it native asyncio

maaktweluit:
> @maaktweluit I find it a little awkward to introduce global flag available for 1k+ tests for the sake of a single test. Can still do it if you think it's worth it.

@Krigpl agree with that, but the `conftest.py` file can also be applied locally ( like the linked sample for `node_integration_tests` )

> Also this isn't meant to be run together with the rest of the test suite. It just doesn't make much sense to me to use pytest here.

Agree with this too, but its another test inside the golem-core repo, and most other tests are compatible to run with pytest ( unit, node_e2e, concent_e2e )

I do not mind if we use pytest here or not, the main point of my comment was to be able to re-use test utilities, since `tmpdir` and `DatabaseFixture` are implemented again in this file.

scripts/task_api_tests/basic_integration.py
Krigpl: I doubt we can remove twisted here completely considering the dependent components still use it, i.e. twisted reactor still has to be installed for some golem components to work properly.

Krigpl:
@maaktweluit Right, but another thing is that I am considering making `work_dir` a flag, since it can't be tmpdir in Windows (because of docker mounts) and whether to remove it automatically also configurable by a flag. Thus not reusable from pytest anymore.

scripts/task_api_tests/basic_integration.py
maaktweluit: ok, that is a bummer, lets leave it then for now :)

maaktweluit (approved):
lets leave it without pytest and with twisted for now

if we want to move there later the `tmpdir` is also a flag there

golem/task/requestedtaskmanager.py
maaktweluit: that will solve it :)

PR #4667, opened 2019-09-02
Use multiprocessing in LocalhostEnvironment

Wiezzel:
Running GRPC server in a subprocess instead of a thread provides a great speed up and allows to implement proper stop().

### DIFF
--- a/requirements-lint.txt
+++ b/requirements-lint.txt
@@ -4,6 +4,7 @@ certifi==2018.1.18
 chardet==3.0.4
 codecov==2.0.15
 coverage==4.5.1
+dill==0.3.0
 factory-boy==2.9.2
 Faker==0.8.9
 flake8==3.7.7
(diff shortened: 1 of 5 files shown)


Krigpl (approved):
thumbupglitch

requirements.txt
```diff
@@ -24,6 +24,7 @@ crossbar==19.6.2
 cryptography==2.7
 cytoolz==0.9.0.1
 dataclasses==0.6
+dill==0.3.0
```
Krigpl: That looks like it belongs in `requirements-test`

Wiezzel: Meh, putting it in `requirements-test.txt` breaks buildbot. 😢 
```
E: tests/golem/envs/localhost.py:8:0: Unable to import 'dill' (import-error)\n
```

Wiezzel: Maybe it should be in `requirements-lint.txt` as well?

Krigpl: lint has `-r requirements-test.txt` in it

Krigpl: Not sure if that works tho

Wiezzel: I'll try.

Wiezzel: Adding to `requirements-lint.txt` helped.

PR #4680, opened 2019-09-04
Remove new environment benchmark scores on configuration change

mfranciszkiewicz:


tests/golem/task/test_taskserver.py
```diff
@@ -1520,9 +1522,10 @@ def _check(callback, _):
         task_computer.change_config.assert_called_once_with(config_desc)
         task_computer.lock_config.assert_called_once_with(False)
         run_benchmarks.assert_called_once()
+        assert delete_performance.call_count > 0
```
Wiezzel: Why not `assert_called_once()`?

Wiezzel (approved):

tests/golem/task/test_taskserver.py
```diff
@@ -1487,10 +1487,11 @@ def test(self):
         change_pcs_config.assert_called_once_with(self.ts, config_desc)
 
 
+@patch('golem.task.envmanager.Performance.delete')
```
Wiezzel: I think mocking `remove_cached_performance` here and testing it separately would be a better idea. Now you cannot easily assert that the right entries have been removed from database.

PR #4682, opened 2019-09-04
Allow users to enable / disable discovered docker images

mfranciszkiewicz:
- new: `DockerWhitelistRPC` for exposing Whitelist methods via RPC
- emit events when new images are discovered
- remember recently discovered images (including basic stats)

### DIFF
--- a/golem/envs/docker/whitelist.py
+++ b/golem/envs/docker/whitelist.py
@@ -4,7 +4,7 @@ import golem.model
 
 
 def repository_from_image_name(image_name: str) -> str:
-    return image_name.split('/')[0]
+    return image_name.rsplit('/', maxsplit=1)[0]
 
 
 class Whitelist:
--- a/golem/envs/docker/whitelist.py
+++ b/golem/envs/docker/whitelist.py
@@ -1,7 +1,21 @@
+from typing import List
+
 import golem.model
 
 
+def repository_from_image_name(image_name: str) -> str:
+    return image_name.split('/')[0]
+
+
 class Whitelist:
+
+    @staticmethod
+    def get() -> List[str]:
+        return [
+            whitelisted.repository for whitelisted in
+            golem.model.DockerWhitelist.select().execute()
+        ]
+
     @classmethod
     def add(cls, repository: str) -> bool:
         """
@@ -26,7 +40,7 @@ class Whitelist:
 
     @staticmethod
     def is_whitelisted(image_name: str) -> bool:
-        repository = image_name.split('/')[0]
+        repository = repository_from_image_name(image_name)
         query = \
             golem.model.DockerWhitelist.select().where(
                 golem.model.DockerWhitelist.repository == repository,
--- a/golem/envs/docker/whitelist.py
+++ b/golem/envs/docker/whitelist.py
@@ -10,7 +10,7 @@ def repository_from_image_name(image_name: str) -> str:
 class Whitelist:
 
     @staticmethod
-    def get() -> List[str]:
+    def get_all() -> List[str]:
         return [
             whitelisted.repository for whitelisted in
             golem.model.DockerWhitelist.select().execute()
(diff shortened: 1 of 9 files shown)


golem/node.py
```diff
@@ -38,7 +38,8 @@
 from golem.ethereum.transactionsystem import TransactionSystem
 from golem.model import DB_MODELS, db, DB_FIELDS
 from golem.network.transport.tcpnetwork_helpers import SocketAddress
-from golem.report import StatusPublisher, Component, Stage, report_calls
+from golem.report import StatusPublisher, Component, Stage, report_calls, \
+    EventPublisher
```
Wiezzel: I encourage using this style according to our guide
```python
from golem.report import (
    StatusPublisher,
    ...
)
```

golem/task/server/whitelist.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self) -> None:
+        self._discovered: List[DiscoveredDockerImage] = list()
+
+    def _docker_image_discovered(
+            self,
+            name: str
+    ) -> None:
+        if Whitelist.is_whitelisted(name):
+            return
+        for discovered in self._discovered:
```
Wiezzel: This definitely looks like a job for a `dict`. Why iterate over all discovered images if you can just check it in constant time?

golem/task/server/whitelist.py
```diff
@@ -0,0 +1,82 @@
...
+            for discovered in self._discovered
+        }
+
+    @staticmethod
+    @rpc_utils.expose('env.docker.images.whitelist')
+    def _docker_whitelist_get() -> List[str]:
+        return Whitelist.get()
+
+    @rpc_utils.expose('env.docker.images.whitelist.add')
+    def _docker_whitelist_add(self, image_name: str) -> None:
```
Wiezzel: This is potentially misleading. Someone might think it adds only this particular image to the whitelist while it actually adds all images from a repository.

(*Actually, we're using the term repository wrong here, but let's keep this aside*)

golem/task/server/whitelist.py
```diff
@@ -0,0 +1,82 @@
...
+        repository = repository_from_image_name(image_name)
+        Whitelist.add(repository)
+        self._docker_refresh_discovered_images()
+
+    @staticmethod
+    @rpc_utils.expose('env.docker.images.whitelist.remove')
+    def _docker_whitelist_remove(image_name: str) -> None:
```
Wiezzel: See comment for `_docker_whitelist_add`

golem/envs/docker/whitelist.py
```diff
@@ -1,7 +1,21 @@
...
 
 
+def repository_from_image_name(image_name: str) -> str:
+    return image_name.split('/')[0]
+
+
 class Whitelist:
+
+    @staticmethod
+    def get() -> List[str]:
```
Wiezzel: `get_all()` maybe?

Wiezzel (commented):
Some more and less important remarks and one general idea:

How about changing `DockerWhitelist` model into something like `DockerWhiteAndBlackList` (needs a better name for sure) with `whitelisted` boolean field? Then no separate storing of observed images would be needed and the whole thing would be persistent.

golem/task/server/whitelist.py
mfranciszkiewicz: To keep them in order - might as well be done by `OrderedDict`

Wiezzel: AFAIK dicts are insert-ordered in Python 3.6+

mfranciszkiewicz: That's a broader topic but that statement is true for cpython

mfranciszkiewicz:
@Wiezzel I agree, it sounds like a separate task though. We'd need to update stats for every (including known) image and maintain a separate maximum of N observed images in the same table (or create a new one)

Wiezzel (approved):

PR #4685, opened 2019-09-05
Check for payment tables existence in migrations

kmazurek:
Updated peewee migrations from 31 to 33 to include a check for the existence
of required database tables before performing the actual migration.


jiivan (approved):
buakaw_throw

Issue #4690, opened 2019-09-06
 KeyError: 'work_dirs';  docker_config = DockerGPUConfig.from_dict(config_dict)


ederenn:
## Description

**Golem Version**: 0.20.2+dev501.g1347ae6

**Golem-Messages version** (leave empty if unsure): 3.11.0

**Electron version** (if used):

**OS** [e.g. Windows 10 Pro]: Linux 18.04

**Branch** (if launched from source): b0.21

**Mainnet/Testnet**: mainnet

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

Golem crashes after encountering critical
```
CRITICAL [golem.client                       ] Can't start network. Giving up.
Traceback (most recent call last):
  File "/home/ederenn/projects/golem/golem/client.py", line 340, in start
    self.start_network()
  File "/home/ederenn/projects/golem/golem/client.py", line 391, in start_network
    task_finished_cb=self._task_finished_cb,
  File "/home/ederenn/projects/golem/golem/task/taskserver.py", line 154, in __init__
    NonHypervisedDockerGPUEnvironment.default(docker_config_dict))
  File "/home/ederenn/projects/golem/golem/envs/docker/non_hypervised.py", line 39, in default
    docker_config = DockerGPUConfig.from_dict(config_dict)
  File "/home/ederenn/projects/golem/golem/envs/docker/cpu.py", line 71, in from_dict
    _work_dirs = data.pop('work_dirs')
KeyError: 'work_dirs'
```

## Steps To Reproduce
_Short description of steps to reproduce the behavior:_
e.g.
1. Launch the app from source on branch b0.21
2. See critical
 

## Logs and any additional context

possibly related:
https://github.com/golemfactory/golem/pull/4618

ederenn:
Node without Nvidia starts normally

PR #4692, opened 2019-09-06
Modify a copy of config dict in DockerCPUConfig.from_dict

mfranciszkiewicz:
Resolves https://github.com/golemfactory/golem/issues/4690


maaktweluit (approved):

ederenn:
Checked on two nodes: with and without Nvidia.
Both nodes are starting

PR #4698, opened 2019-09-06
Add CPU usage reporting in benchmarks

kmazurek:


golem/environments/environment.py
```diff
@@ -109,11 +112,13 @@ def get_min_accepted_performance(cls) -> float:
...
         logger.info('Running benchmark for %s', cls.get_id())
         performance = make_perf_test()
         logger.info('%s performance is %.2f', cls.get_id(), performance)
+
         if save:
-            Performance.update_or_create(cls.get_id(), performance)
-        return performance
+            Performance.update_or_create(cls.get_id(), performance, 0)
+
+        return BenchmarkResult(performance, 0)
```
mplebanski, line 124: One thing I don't understand here is: why there is a performance for this benchmark, but no usage?

apps/core/benchmark/benchmarkrunner.py
```diff
@@ -66,11 +69,17 @@ def is_success(self, task_thread: TaskThread) -> bool:
         return self.benchmark.verify_result(res["data"])
 
     def computation_success(self, task_thread: TaskThread) -> None:
-        res, _ = task_thread.result
+        # pylint: disable=no-member
+        provider_stats = dt_stats.ProviderStats(**task_thread.stats)
```
mplebanski, line 73: Are task_thread.stats guaranteed to contain all keys for __slots__ in ProviderStats? If not, then maybe some exception handling should be added.

tests/golem/task/test_taskserver.py
```diff
@@ -174,7 +175,7 @@ def _prepare_env(self, *,
             -> None:
         env = Mock(spec=OldEnv)
         env.get_min_accepted_performance.return_value = min_accepted_perf
-        env.get_performance = Mock(return_value=0.0)
+        env.get_benchmark_result = BenchmarkResult()
```
mplebanski: Suprisingly this is not an assignment of a Callable object and mock's`return_value` is not used, therefore is this even used?

mplebanski:
CPU usage for glambda is 0:

```
INFO     [golem.task.benchmarkmanager        ] Running benchmark for glambda
INFO     [golem.task.benchmarkmanager        ] glambda benchmark finished. performance=653.32, cpu_usage=0
```

do you know why?

apps/core/benchmark/benchmarkrunner.py
kmazurek: Yes, the keys are defined in the [entrypoint wrapper](https://github.com/golemfactory/docker-cgroups-stats). Also, none of the slots for `ProviderStats` is marked as required, so they will be assigned the default value `None`. I don't think we need additional error handling at this point, especially since the slots should stay optional.

tests/golem/task/test_taskserver.py
kmazurek: You're right, this should actually fail. I will change this to a proper mock.

apps/core/benchmark/benchmarkrunner.py
mplebanski: OK.

golem/database/database.py
```diff
@@ -61,7 +61,7 @@ def execute_sql(self, sql, params=None, require_commit=True):
 
 
 class Database:
-    SCHEMA_VERSION = 35
+    SCHEMA_VERSION = 36
```
Krigpl: I think you need to rebase now.

golem/environments/environment.py
Krigpl: This benchmark doesn't use Docker, so I guess cpu stats are not supported.

golem/task/envmanager.py
```diff
@@ -111,17 +111,25 @@ def get_performance(self, env_id) -> Deferred:
         finally:
             self._running_benchmark = False
 
-        Performance.update_or_create(env_id, result)
+        Performance.update_or_create(
+            env_id=env_id,
+            performance=result.performance,
+            cpu_usage=result.cpu_usage
```
Krigpl, line 118: afaict `Environment::run_benchmark` method has not been modified anywhere and the result from line 103 will still return just a float.

golem/envs/__init__.py
```diff
@@ -78,6 +79,16 @@ class EnvSupportStatus(NamedTuple):
     nonsupport_reason: Optional[str] = None
 
 
+@dataclass
+class BenchmarkResult:
+    performance: float = 0.0
+    cpu_usage: int = 0
+
+    @classmethod
```
Krigpl: How come linter doesn't complain about unused `cls` arg? This can be a staticmethod

kmazurek:
@mplebanski there's a problem with collecting usage stats for glambda since it uses gVisor as its runtime. This isn't a blocker since glambda is not going to be included in usage marketplace at this point.

golem/database/database.py
kmazurek: I think you're referring to `develop`, the base here is `b0.21` and there were no recent changes to the DB schema.

golem/environments/environment.py
kmazurek: That's correct.

golem/envs/__init__.py
kmazurek: Good catch, same goes for `Performance::update_or_create`.

golem/task/envmanager.py
kmazurek: Do you mean [this one](https://github.com/golemfactory/golem/blob/6bfd375193ab32ebf6cf066bcd612ff022fce97e/golem/envs/docker/cpu.py#L554)?

Krigpl: rickandmorty_yes

golem/database/database.py
Krigpl: yesquiteindeed

golem/task/envmanager.py
kmazurek: afaik task API is going to take a different approach to collecting stats from containers (doing it from outside of the container itself). I'm relying on `task_thread` for CPU usage, which I cannot use here. Therefore, I'm going to return 0 as CPU usage for now.

Krigpl: OK_thumb

mplebanski (approved):

kmazurek:
Actually, the base here should be updated to `develop`, since these changes are no longer in scope for the 0.21 release.

golem/database/database.py
```diff
@@ -61,7 +61,7 @@ def execute_sql(self, sql, params=None, require_commit=True):
 
 
 class Database:
-    SCHEMA_VERSION = 36
+    SCHEMA_VERSION = 37
```
Wiezzel: Sorry, bro, I was faster:
https://github.com/golemfactory/golem/blob/f01f91c3b1658a37684f62aa6dd1ff7716407b98/golem/database/database.py#L64

kmazurek: no

PR #4701, opened 2019-09-09
ACL Database

mdtanrikulu:
- Moves ACL deny/allow list from text file to database to keep persistence.
- `node_name` parameter added to the list, to provide clear recognition for users.
- type of `node_id` parameter in `net.peer.block`, `net.peer.allow`, `net.peer.block_ip` and `net.peer.allow_ip` changed as union (string, list) so it will accept list of node id's as well.

### DIFF
--- a/golem/node.py
+++ b/golem/node.py
@@ -49,6 +49,7 @@ from golem.rpc.session import (
 from golem import terms
 from golem.tools.uploadcontroller import UploadController
 from golem.tools.remotefs import RemoteFS
+from golem.task.acl import migrate_txt
 
 if TYPE_CHECKING:
     # pylint:disable=unused-import
@@ -135,6 +136,9 @@ class Node(HardwarePresetsMixin):
         # Initialize database
         self._db = Database(
             db, fields=DB_FIELDS, models=DB_MODELS, db_dir=datadir)
+
+        # ACL txt migration
+        migrate_txt(Path(datadir))
 
         self.client: Optional[Client] = None
 
--- a/golem/node.py
+++ b/golem/node.py
@@ -49,7 +49,6 @@ from golem.rpc.session import (
 from golem import terms
 from golem.tools.uploadcontroller import UploadController
 from golem.tools.remotefs import RemoteFS
-from golem.task.acl import migrate_txt
 
 if TYPE_CHECKING:
     # pylint:disable=unused-import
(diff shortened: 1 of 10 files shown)


prekucki (changes requested):
I do not see data migration from deny.txt. what is your plan ? special script during installation ?

mdtanrikulu:
> I do not see data migration from deny.txt. what is your plan ? special script during installation ?

Thanks! It was a good reminder.

Here is the migration script as @mfranciszkiewicz suggests, I moved it from database migration schema to node.py.

https://github.com/golemfactory/golem/pull/4701/files#diff-1bef601c50a4151f88f0ecda16e1aa10R141

golem/node.py
```diff
@@ -136,6 +137,9 @@ def __init__(self,  # noqa pylint: disable=too-many-arguments
         self._db = Database(
             db, fields=DB_FIELDS, models=DB_MODELS, db_dir=datadir)
 
+        # ACL txt migration
+        migrate_txt(Path(datadir))
```
shadeofblue: I strongly disagree with putting migration code inside Golem's code component ... 

migrations are _exactly_ for this purpose and the code performing them should be included there

what if we wanted to have a _reverse_ migration? you can't do that with a solution like this one ...

mfranciszkiewicz:
The basis for my suggestion was the separation of schema and data migrations. Schema changes are orthogonal to locating and parsing configuration files and simply should be done separately. I understand the desire to mix those two together since the current migrator class is not flexible enough to allow that.
What I missed is that data migration won't be performed at the same stage as the newly introduced migration script and would require extra maintenance.

@mdtanrikulu Please move the `migrate_txt` function back to the schema script. Sorry for all the fuss caused.

golem/node.py
mdtanrikulu: > The basis for my suggestion was the separation of schema and data migrations. Schema changes are orthogonal to locating and parsing configuration files and simply should be done separately. I understand the desire to mix those two together since the current migrator class is not flexible enough to allow that.
> What I missed is that data migration won't be performed at the same stage as the newly introduced migration script and would require extra maintenance.
> 
> @mdtanrikulu Please move the `migrate_txt` function back to the schema script. Sorry for all the fuss caused.

👍 done.

prekucki (approved):

shadeofblue (approved):

PR #4705, opened 2019-09-10
provide correct address for the new GNTDeposit smart contract for Con…

shadeofblue:
…cent `staging` and `test`


etam (approved):
seems to be working

PR #4706, opened 2019-09-10
fix the `run_nodes` command when no `datadir`s are provided

shadeofblue:


scripts/node_integration_tests/run_nodes.py
```diff
@@ -41,12 +41,18 @@ def make_node_configs(node_names: typing.Iterable[str],
         for i, node_name in enumerate(node_names)
     }
 
-    for node_name, datadir in override_datadirs.items():
-        node_id = NodeId(node_name)
-        if node_id not in node_configs:
-            raise Exception("can't override datadir for undefined node"
-                            f" '{node_name}'")
-        node_configs[node_id].datadir = datadir
+    if override_datadirs:
```
etam, line 45: Is this `if` required? If no datadir is provided in command line is this `None` or empty dict? If it's `None` I'd rather stick to type hint and use
```python
make_node_configs(args.nodes, args.datadir or {})
```

scripts/node_integration_tests/run_nodes.py
```diff
@@ -41,12 +41,18 @@ def make_node_configs(node_names: typing.Iterable[str],
...
+            node_id = NodeId(node_name)
+            if node_id not in node_configs:
+                raise Exception("can't override datadir for an undefined node"
+                                f" '{node_name}'")
+            node_configs[node_id].datadir = datadir
+
+    for node_id, node_config in node_configs.items():
+        if not node_config.datadir:
+            node_config.datadir = helpers.mkdatadir(node_id.value)
+        print(f"Launching node `{node_id.value}`: {node_config}")
```
etam: This seems a bit out of place, because function is about making configs, not launching anything. (I won't block you on that. I'm just bikeshedding a bit)

shadeofblue: right

scripts/node_integration_tests/run_nodes.py
shadeofblue: well, what's the difference really? updated the type hint to match ...

etam (approved):
ok

PR #4707, opened 2019-09-10
Update to Task API v0.16.1

Wiezzel:
* Import environment IDs from Task API instead of declaring them in Golem code.
* Read the environment ID and prerequisites from CreateTask reply. Save the obtained values in RequestedTask model.

~Requirements will be updated when https://github.com/golemfactory/task-api/pull/37 is merged and Task API v0.16.0 is released.~

### DIFF
--- a/tests/golem/task/test_requestedtaskmanager.py
+++ b/tests/golem/task/test_requestedtaskmanager.py
@@ -2,6 +2,7 @@
 # ^^ Pytest fixtures in the same file require the same name
 
 import asyncio
+import json
 from pathlib import Path
 
 from freezegun import freeze_time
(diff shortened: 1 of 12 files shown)


tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -219,6 +250,10 @@ def test_task_not_exists(self):
     async def test_abort(self, mock_client):
         # given
         self._add_next_subtask_to_client_mock(mock_client)
+        mock_client.create_task.return_value = Mock(
```
Krigpl: That's a noticeable amount of repeated code, maybe create a helper method or just make the `mock_client` fixture mock this by default. It can always be overridden to a custom value when needed and would make the fixture ready to use out of the box.

maaktweluit (approved):
LGTM!

+1 for @Krigpl 's proposal about less repeating mocks

golem/database/schemas/036_schema.py
```diff
@@ -0,0 +1,17 @@
+# pylint: disable=no-member
+# pylint: disable=unused-argument
+# pylint: disable=unused-variable
+
+import peewee as pw
+
+SCHEMA_VERSION = 36
```
Krigpl: Heads up - it's already 36 in develop.

Krigpl (approved):

PR #4710, opened 2019-09-10
RequestedTaskManager - Start and stop the envs

maaktweluit:
Added quit method to RTM to shutdown all running app clients.

Mocked the `TaskServer.quit()` where needed to ensure these tests do not need asyncio

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -598,17 +598,16 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
     @inlineCallbacks
     def pause(self):
         logger.info("Pausing ...")
-        logger.info('services = %r', self._services)
         for service in self._services:
             if service.running:
                 service.stop()
 
         if self.p2pservice:
-            logger.info("Pausing p2pservice")
+            logger.debug("Pausing p2pservice")
             self.p2pservice.pause()
             self.p2pservice.disconnect()
         if self.task_server:
-            logger.info("Pausing task_server")
+            logger.debugs("Pausing task_server")
             yield self.task_server.pause()
         logger.info("Paused")
 
--- a/golem/client.py
+++ b/golem/client.py
@@ -30,6 +30,7 @@ from golem.core.common import (
     string_to_timeout,
     to_unicode,
 )
+from golem.core.deferred import sync_wait
 from golem.core.fileshelper import du
 from golem.hardware.presets import HardwarePresets
 from golem.core.keysauth import KeysAuth
@@ -359,7 +360,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
         if self.concent_filetransfers.running:
             self.concent_filetransfers.stop()
         if self.task_server:
-            self.task_server.quit()
+            sync_wait(self.task_server.quit())
         if self.use_monitor and self.monitor:
             self.diag_service.stop()
             # This effectively removes monitor dispatcher connections (weakrefs)
@@ -595,7 +596,6 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             self.port_mapper.quit()
 
     @rpc_utils.expose('ui.stop')
-    @inlineCallbacks
     def pause(self):
         logger.info("Pausing ...")
         for service in self._services:
@@ -607,8 +607,8 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             self.p2pservice.pause()
             self.p2pservice.disconnect()
         if self.task_server:
-            logger.debugs("Pausing task_server")
-            yield self.task_server.pause()
+            logger.debug("Pausing task_server")
+            sync_wait(self.task_server.pause())
         logger.info("Paused")
 
     @rpc_utils.expose('ui.start')
--- a/golem/client.py
+++ b/golem/client.py
@@ -595,6 +595,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             self.port_mapper.quit()
 
     @rpc_utils.expose('ui.stop')
+    @inlineCallbacks
     def pause(self):
         logger.info("Pausing ...")
         logger.info('services = %r', self._services)
@@ -608,7 +609,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             self.p2pservice.disconnect()
         if self.task_server:
             logger.info("Pausing task_server")
-            self.task_server.pause()
+            yield self.task_server.pause()
         logger.info("Paused")
 
     @rpc_utils.expose('ui.start')
--- a/golem/client.py
+++ b/golem/client.py
@@ -30,7 +30,6 @@ from golem.core.common import (
     string_to_timeout,
     to_unicode,
 )
-from golem.core.deferred import sync_wait
 from golem.core.fileshelper import du
 from golem.hardware.presets import HardwarePresets
 from golem.core.keysauth import KeysAuth
@@ -596,6 +595,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             self.port_mapper.quit()
 
     @rpc_utils.expose('ui.stop')
+    @inlineCallbacks
     def pause(self):
         logger.info("Pausing ...")
         for service in self._services:
@@ -608,7 +608,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             self.p2pservice.disconnect()
         if self.task_server:
             logger.debug("Pausing task_server")
-            sync_wait(self.task_server.pause())
+            yield self.task_server.pause()
         logger.info("Paused")
 
     @rpc_utils.expose('ui.start')
--- a/golem/client.py
+++ b/golem/client.py
@@ -348,7 +348,6 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
         logger.debug('Started client services')
 
     @report_calls(Component.client, 'stop', stage=Stage.post)
-    @inlineCallbacks
     def stop(self):
         logger.debug('Stopping client services ...')
         self.stop_network()
@@ -360,7 +359,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
         if self.concent_filetransfers.running:
             self.concent_filetransfers.stop()
         if self.task_server:
-            yield self.task_server.quit()
+            self.task_server.quit()
         if self.use_monitor and self.monitor:
             self.diag_service.stop()
             # This effectively removes monitor dispatcher connections (weakrefs)
@@ -596,7 +595,6 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             self.port_mapper.quit()
 
     @rpc_utils.expose('ui.stop')
-    @inlineCallbacks
     def pause(self):
         logger.info("Pausing ...")
         logger.info('services = %r', self._services)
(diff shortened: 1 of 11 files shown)


golem/task/requestedtaskmanager.py
```diff
@@ -383,7 +396,7 @@ def _get_task_api_service(
             raise RuntimeError(
                 "Error connecting to app, environment not enabled."
                 f" env={env_id}, app={app_id}")
-        env = self._env_manager.environment(env_id)
+        env = auto_setup(self._env_manager.environment(env_id))
```
Krigpl: This isn't needed, the environment is already registered as `auto_setup`.

golem/task/requestedtaskmanager.py
```diff
@@ -350,6 +351,18 @@ def get_subtask_outputs_dir(self, task_id: TaskId) -> Path:
...
         await self._shutdown_app_client(task.app_id)
 
+    async def quit(self):
+        logger.info('quit()')
+        # Abort active tasks?
+
+        # Shutdown registered app_clients
+        for app_id, app_client in self._app_clients:
+            logger.info('Shutting down app. app_id=%r', app_id)
+            await app_client.Shutdown()
```
Krigpl: Isn't the method called `shutdown` with non capital `s`?

golem/task/taskserver.py
```diff
@@ -298,11 +299,19 @@ def sync_network(self, timeout=None):
...
+        yield self.quit()
 
     def resume(self):
         super().resume()
         CoreTask.VERIFICATION_QUEUE.resume()
 
+    @inlineCallbacks
+    def quit(self):
+        quit_f = asyncio.ensure_future(self.requested_task_manager.quit())
+        yield deferred_from_future(quit_f)
```
Krigpl: `deferred_from_future` already calls `ensure_future` inside

golem/task/taskserver.py
```diff
@@ -298,11 +299,19 @@ def sync_network(self, timeout=None):
...
         super().pause()
         yield CoreTask.VERIFICATION_QUEUE.pause()
+        self.disconnect()
+        yield self.quit()
 
     def resume(self):
         super().resume()
         CoreTask.VERIFICATION_QUEUE.resume()
 
+    @inlineCallbacks
```
Krigpl: Not ideal but I think keeping this method synchronous would eliminate 90% changes from this PR, which I'd consider lesser evil since most of these changes look hacky to make the tests work.

Krigpl (changes requested):

golem/task/taskserver.py
maaktweluit: old habbits die hard, forgot you added it

golem/task/requestedtaskmanager.py
maaktweluit: nice, didnt know that, fixed :)

golem/task/requestedtaskmanager.py
maaktweluit: 🙈

fixed

golem/task/taskserver.py
maaktweluit: Made this call sync, it removed some hacks but added ( less ) new ones :)

tests/golem/task/test_taskserver.py
```diff
@@ -157,7 +157,10 @@ def tearDown(self):
         testutils.DatabaseFixture.tearDown(self)
 
         if hasattr(self, "ts") and self.ts:
-            self.ts.quit()
+            # Hack to not call the asycio call RTM.quit()
+            # Uncomment when this test works with asyncio
```
Krigpl, line 161: I think you can just mock RTM in the `setUp` method just like `TaskComputerAdapter` is being mocked and this won't be necessary.

tests/golem/test_client.py
```diff
@@ -127,14 +128,19 @@ def make_client(*_, **kwargs):
...
-class TestClientBase(testwithreactor.TestDatabaseWithReactor):
+class TestClientBase(DatabaseFixture):
 
     def setUp(self):
         super().setUp()
         self.client = make_client(datadir=self.path)
 
     def tearDown(self):
-        self.client.quit()
+        # Only call taskcomputer.quit, TaskServer.quit now requires asyncio
```
Krigpl, line 138: Similar, I think you could just patch RTM during the `mock_client` call.

tests/golem/test_client.py
```diff
@@ -212,7 +218,7 @@ def test_task_cleaning_enabled(self, task_cleaner, *_):
 
         task_cleaner.assert_called()
 
-    def test_collect_gossip(self, *_):
+    def test_collect_gossip(self, gossip_mock, *_):
```
Krigpl: Where did that come from?

golem/task/requestedtaskmanager.py
```diff
@@ -350,6 +350,18 @@ def get_subtask_outputs_dir(self, task_id: TaskId) -> Path:
...
         await self._shutdown_app_client(task.app_id)
 
+    async def quit(self):
+        logger.info('quit()')
+        # Abort active tasks?
+
+        # Shutdown registered app_clients
+        for app_id, app_client in self._app_clients:
+            logger.info('Shutting down app. app_id=%r', app_id)
+            await app_client.shutdown()
```
Krigpl: Maybe put that into a try/catch block so that failing to shut down a single app doesn't interfere with others.

golem/task/taskserver.py
```diff
@@ -298,11 +299,18 @@ def sync_network(self, timeout=None):
     def pause(self):
         super().pause()
         yield CoreTask.VERIFICATION_QUEUE.pause()
+        self.disconnect()
```
Krigpl, line 301: Is this fine that there's no `connect` in the corresponding `resume` method?

tests/golem/task/test_rpc.py
```diff
@@ -11,14 +11,14 @@
 from golem_messages.factories.datastructures import p2p as dt_p2p_factory
 from mock import call, Mock
 from twisted.internet import defer
+from twisted.trial.unittest import TestCase as TwistedTestCase
```
Krigpl, line 14: How come any changes are needed in this file?

tests/golem/test_client.py
maaktweluit: one of the old hacks, forgot to revert this part

golem/task/requestedtaskmanager.py
maaktweluit: good one, will add this

golem/task/taskserver.py
maaktweluit: i copied this from the `client.py` since it was the only place it was called, there is no `connect` in the resume there so i think its fine.

tests/golem/task/test_rpc.py
maaktweluit: the first change was no longer needed, it was related to the client getting `inlineCallbacks` methods

The second is still needed, using sync_wait here will time out

tests/golem/task/test_taskserver.py
maaktweluit: Tried this but the call still errors, either the sync_wait times out or the new Mocked call does not work with `defer_from_future()`

tests/golem/test_client.py
maaktweluit: tried this one too, same problems as above

tests/golem/task/test_taskserver.py
Krigpl: Notice that this branch is not based on newest develop, and `defer_from_future` doesn't actually call `ensure_future` inside.

Krigpl: Either way, I also can't make it work, so let's leave it.

golem/task/requestedtaskmanager.py
```diff
@@ -350,6 +350,22 @@ def get_subtask_outputs_dir(self, task_id: TaskId) -> Path:
 
         await self._shutdown_app_client(task.app_id)
 
+    async def quit(self):
```
Krigpl: Please add a unit test for that method.

Krigpl (changes requested):

tests/golem/task/test_taskserver.py
maaktweluit: rebased onto develop, but will still leave it for now

golem/task/requestedtaskmanager.py
maaktweluit: That was very needed 😊 

Added it and fixed the errors

Krigpl (approved):

golem/client.py
```diff
@@ -608,9 +608,7 @@ def pause(self):
             self.p2pservice.disconnect()
         if self.task_server:
             logger.debug("Pausing task_server")
-            yield self.task_server.pause()
-            self.task_server.disconnect()
-            self.task_server.task_computer.quit()
+            sync_wait(self.task_server.pause())
```
mfranciszkiewicz: (edge case) `yield CoreTask.VERIFICATION_QUEUE.pause()` will lock the event loop for the remaining time of the verification process and trigger various timeouts (e.g. disconnecting RPC clients).

golem/task/taskserver.py
```diff
@@ -298,11 +298,20 @@ def sync_network(self, timeout=None):
...
+        self.quit()
 
     def resume(self):
         super().resume()
         CoreTask.VERIFICATION_QUEUE.resume()
 
+    def quit(self):
+        defer_rtm_quit = deferred_from_future(
+            self.requested_task_manager.quit()
```
mfranciszkiewicz: We don't re-instantiate app clients on `resume`. Is that intentional?

mfranciszkiewicz (dismissed):

golem/task/taskserver.py
maaktweluit: yes, they are made and cached again on the next call

golem/client.py
maaktweluit: bad verification queue, it should pause like it is asked to..

Is there something i can do about this to make it not time out other calls?

mfranciszkiewicz: Well, you can remove `sync_wait`

maaktweluit: good one, done!

golem/task/requestedtaskmanager.py
```diff
@@ -363,6 +363,22 @@ def get_subtask_outputs_dir(self, task_id: TaskId) -> Path:
...
+        logger.debug('quit()')
+        # Abort active tasks?
+
+        # Shutdown registered app_clients
+        for app_id, app_client in self._app_clients.items():
+            logger.info('Shutting down app. app_id=%r', app_id)
+            try:
+                await app_client.shutdown()
```
Wiezzel, line 372: After calling `shutdown()` you should call `wait_until_shutdown_complete()` to clean up the runtime.

Wiezzel (changes requested):

Wiezzel:
I'm also not sure about naming the newly introduced methods `quit` because it seem to suggest an irreversible shutdown. Maybe `stop` would make a better name?

golem/task/requestedtaskmanager.py
maaktweluit: Isnt this already called inside `app_client.shutdown()`?

Would be a pain to add here, since we do not store the service anywhere, just the client

maaktweluit:
@Wiezzel Made the change to `stop` agree its a better name

Will wait with pushing till we clarify if wait_for_shutdown is needed to not trigger tests twice

golem/task/requestedtaskmanager.py
Wiezzel: > Isnt this already called inside `app_client.shutdown()`?
>
> Would be a pain to add here, since we do not store the service anywhere, just the client

Well, you're right that it would be a pain. Actually it's called in `RequestorAppClient`'s `_soft_shutdown` but I case the soft shutdown gets timed out we should call it as well. So I guess this should be changed in Task API, not here.

maaktweluit: good one! i'll move it to the "finally" part of that try block in task-api

Wiezzel (commented):
You forgot about the tests when renaming.

maaktweluit:
🙈

Good catch, fixed

Wiezzel (approved):

PR #4712, opened 2019-09-11
Restart task in RTM

Krigpl:


maaktweluit (approved):
LGTM!

Nice test cleanup

golem/task/requestedtaskmanager.py
```diff
@@ -358,6 +358,11 @@ def get_subtask_outputs_dir(self, task_id: TaskId) -> Path:
 
         await self._shutdown_app_client(task.app_id)
 
+    async def restart_task(self, task_id: TaskId) -> None:
+        task = RequestedTask.get(RequestedTask.task_id == task_id)
+        task.status = TaskStatus.waiting
+        task.save()
```
maaktweluit, line 364: should we also discard the completed results?

maaktweluit (changes requested):

golem/task/requestedtaskmanager.py
Krigpl: There are multiple ways of task restarts we wanna support
- restart just the failed subtasks - that's the one
- restart the whole task by creating a new one from scratch
- restart particular subtasks

maaktweluit: ok, should we make issues for the other 2? Do we want to support all 3 from the first release?

maaktweluit (approved):

PR #4721, opened 2019-09-12
Add a parameter to return only created tasks in Client.get_tasks

mfranciszkiewicz:
Requested by @mdtanrikulu

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -889,7 +889,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
     def get_tasks(
             self,
             task_id: Optional[str] = None,
-            return_created_tasks: bool = False,
+            return_created_tasks_only: bool = False,
     ) -> Union[Optional[dict], Iterable[dict]]:
         if not self.task_server:
             return []
@@ -901,7 +901,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
         tasks = (self.get_task(task_id) for task_id in task_keys)
         filter_fn = None
 
-        if return_created_tasks:
+        if return_created_tasks_only:
             filter_fn = self._filter_task_created_status
 
         return list(filter(filter_fn, tasks))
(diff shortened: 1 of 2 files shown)


golem/client.py
```diff
@@ -885,18 +886,31 @@ def get_task(self, task_id: str) -> Optional[dict]:
...
 
-        task_ids = list(self.task_server.task_manager.tasks.keys())
-        tasks = (self.get_task(task_id) for task_id in task_ids)
-        # Filter Nones because get_task returns Optional[dict]
-        return list(filter(None, tasks))
+        task_keys = self.task_server.task_manager.tasks.keys()
+        tasks = (self.get_task(task_id) for task_id in task_keys)
+        filter_fn = None
+
+        if return_created_tasks:
```
mdtanrikulu: I think this should be `if not return_created_tasks:`,  cuz `True` means don't filter and include creating and creating error statuses to the list.

golem/client.py
```diff
@@ -885,18 +886,31 @@ def get_task(self, task_id: str) -> Optional[dict]:
...
+            filter_fn = self._filter_task_created_status
+
+        return list(filter(filter_fn, tasks))
+
+    @staticmethod
+    def _filter_task_created_status(task: Dict) -> bool:
+        return bool(task) and task['status'] not in (
+            TaskStatus.creating,
+            TaskStatus.errorCreating)
```
mdtanrikulu: They will probably be (TaskStatus.creating.value, TaskStatus.errorCreating.value), otherwise it creates enum tuple.

golem/client.py
mfranciszkiewicz: `True` means to filter out tasks that are not yet created (created != being created).

Renamed to `return_created_tasks_only`

golem/client.py
mfranciszkiewicz: That's true, I remember implementing it using values initially. I don't remember the exact reason why I changed it - sorry for the confusion.

golem/client.py
mdtanrikulu: so as default nothing will be filtered. that's fine, I will pass True on electron then.

mfranciszkiewicz: Yes, that's to keep it compatible with CLI

mdtanrikulu (approved):

PR #4722, opened 2019-09-12
Introduced app IDs and added app loading

Wiezzel:
* Apps are no longer identified by names. Instead IDs are used which are computed as MD5 digests of app definitions. This prevents possible name clashes.
* During `TaskServer` setup app directory is created if not existent and all app definitions from that directory are loaded and registered in `AppManager`.

### DIFF
--- a/golem/app_manager.py
+++ b/golem/app_manager.py
@@ -1,13 +1,16 @@
+import hashlib
 import json
 import logging
 from pathlib import Path
-from typing import Any, Dict, List, Iterator
+from typing import Any, Dict, List, Iterator, Tuple
 
 from dataclasses import dataclass, field
 from dataclasses_json import config, dataclass_json
 from marshmallow import fields as mm_fields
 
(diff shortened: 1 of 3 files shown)


maaktweluit (approved):
LGTM

mfranciszkiewicz (dismissed):
It's unintuitive that the app definition file is still valid when the properties are out of order and the app_id differs from the ordered version. I'd suggest to make sure that they are in a proper (alphabetical?) order before serializing them _and_ calculating the checksum.

Also, maybe something more modern than MD5? E.g. SHA3, blake

Wiezzel:
@mfranciszkiewicz Improved the ID generation. Take a :eyes: at https://github.com/golemfactory/golem/commit/f29b2e648a0190a2f94c2767807b96053abd28b1

Wiezzel:
@mfranciszkiewicz Any more comments?

mfranciszkiewicz (approved):

PR #4724, opened 2019-09-13
Early reactor import fix

mfranciszkiewicz:


etam:
https://twistedmatrix.com/trac/wiki/FrequentlyAskedQuestions#WhenItrytoinstallmyreactorIgeterrorsaboutareactoralreadybeinginstalled.Whatgives

> When I try to install my reactor, I get errors about a reactor already being installed. What gives?
> 
> Here's the rule - installing a reactor should always be the first thing you do, and I do mean first. Importing other stuff before you install the reactor can break your code.

Maybe we should go this way? Install reactor in golemapp before importing anything else. I even wanted to do this, but found comments about issues with pyinstaller. I have no experience with pyinstaller, so I decided not to touch anything.

etam:
https://twistedmatrix.com/documents/current/core/howto/choosing-reactor.html

> IMPORTANT: installing a reactor should be the first thing done in the app, since any code that does `from twisted.internet import reactor` will automatically install the default reactor if the code hasn’t already installed one.

mfranciszkiewicz:
That IMO would be the best solution. Apart from pyinstaller, some unit tests will be affected by refactoring, making the needed changes a little bit more complicated.

etam (approved):
ok

PR #4725, opened 2019-09-13
Fix VbR BucketVerifier and add tests

mplebanski:
Fixing an issue that verification by redundancy was not giving a verdict when all 3 results differ (in case where referee is called into action).


mbenke (approved):
Well spotted.

PR #4730, opened 2019-09-16
TaskHeaders and RequestedTaskManager

maaktweluit:
Add `RequestedTaskManager.get_task_headers()` and send them to the network

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -3,23 +3,27 @@ import json
 from datetime import timedelta
 import logging
 import os
-from pathlib import Path
 import shutil
+import time
+from pathlib import Path
 from typing import Any, Dict, List
 
 from dataclasses import dataclass
 from golem_messages import idgenerator
+from golem_messages.datastructures.tasks import TaskHeader
 from golem_task_api import constants
 from golem_task_api.client import RequestorAppClient
 from peewee import fn
 
 from golem.app_manager import AppManager
+from golem import constants as gconst
 from golem.model import (
     ComputingNode,
     default_now,
     RequestedTask,
     RequestedSubtask,
 )
+from golem.network.p2p.local_node import LocalNode
 from golem.task.envmanager import EnvironmentManager, EnvId
 from golem.task.taskstate import TaskStatus, SubtaskStatus
 from golem.task.task_api import EnvironmentTaskApiService
@@ -102,15 +106,15 @@ class RequestedTaskManager:
             self,
             env_manager: EnvironmentManager,
             app_manager: AppManager,
-            public_key: bytes,
+            node: LocalNode,
             root_path: Path,
     ):
-        logger.debug('RequestedTaskManager(public_key=%r, root_path=%r)',
-                     public_key, root_path)
+        logger.debug('RequestedTaskManager(node=%r, root_path=%r)',
+                     node, root_path)
         self._dir_manager = DirManager(root_path)
         self._env_manager = env_manager
         self._app_manager = app_manager
-        self._public_key: bytes = public_key
+        self._node: LocalNode = node
         self._app_clients: Dict[EnvId, RequestorAppClient] = {}
 
     def create_task(
@@ -124,7 +128,7 @@ class RequestedTaskManager:
                      golem_params, app_params)
 
         task = RequestedTask.create(
-            task_id=idgenerator.generate_id(self._public_key),
+            task_id=idgenerator.generate_id(self._node.node_id),
             app_id=golem_params.app_id,
             name=golem_params.name,
             status=TaskStatus.creating,
@@ -208,6 +212,55 @@ class RequestedTaskManager:
         # FIXME: add self.notice_task_updated(task_id, op=TaskOp.STARTED)
         logger.info("Task %s started", task_id)
 
+    def get_task_headers(self):
+        pending_tasks = RequestedTask.select().where(
+            # FIXME: duplicate list with TaskStatus.is_active()
+            RequestedTask.status.in_([
+                TaskStatus.sending,
+                TaskStatus.waiting,
+                TaskStatus.starting,
+                TaskStatus.computing,
+            ])
+        ).execute()
+        # filter out Task.needs_computation() ??
+
+        def _make_task_header(db_task: RequestedTask) -> TaskHeader:
+            return TaskHeader(
+                min_version=str(gconst.GOLEM_MIN_VERSION),
+                task_id=db_task.task_id,
+                # FIXME: use value from database after #4707 is mergeds
+                # environment=db_task.env_id,
+                environment='docker_cpu',
+                task_owner=self._node,
+                deadline=int(db_task.deadline.timestamp()),
+                subtask_timeout=db_task.subtask_timeout,
+                subtasks_count=db_task.max_subtasks,
+                # estimated_memory=task_definition.estimated_memory,
+                max_price=db_task.max_price_per_hour,
+                # Concent Disabled for task-api
+                concent_enabled=False,
+                timestamp=int(time.time()),
+            )
+
+        return [_make_task_header(task) for task in pending_tasks]
+
+    def got_wants_to_compute(self, task_id: str):
+        """
+        Updates number of offers to compute task.
+
+        For statistical purposes only, real processing of the offer is done
+        elsewhere. Silently ignores wrong task ids.
+
+        :param str task_id: id of the task in the offer
+        :return: Nothing
+        :rtype: None
+        """
+        if self.task_exists(self.tasks):
+            logger.debug('Got compute offer')
+            # self.notice_task_updated(task_id,
+            #                         op=TaskOp.WORK_OFFER_RECEIVED,
+            #                         persist=False)
+
     @staticmethod
     def task_exists(task_id: TaskId) -> bool:
         """ Return whether task of a given task_id exists. """
@@ -269,7 +322,7 @@ class RequestedTaskManager:
         )
 
         # Check not providing for own task
-        if node.node_id == self._public_key:
+        if node.node_id == self._node.node_id:
             raise RuntimeError(f"No subtasks for self. task_id={task_id}")
 
         # Check should accept provider, raises when waiting on results or banned
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -368,6 +368,13 @@ class RequestedTaskManager:
 
         await self._shutdown_app_client(task.app_id)
 
+    @staticmethod
+    def get_started_tasks():
+        return RequestedTask.select().where(
+            RequestedTask.status.in_(TASK_STATUS_ACTIVE),
+            RequestedTask.start_time is not None
+        ).execute()
+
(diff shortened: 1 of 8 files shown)


golem/task/requestedtaskmanager.py
```diff
@@ -202,6 +206,53 @@ def start_task(task_id: TaskId) -> None:
         # FIXME: add self.notice_task_updated(task_id, op=TaskOp.STARTED)
         logger.info("Task %s started", task_id)
 
+    def get_task_headers(self):
+        pending_tasks = RequestedTask.select().where(
+            # FIXME: duplicate list with TaskStatus.is_active()
```
Krigpl: One way to fix these issues is to create a list of active statuses that is accessible from here and make `is_active` use it as well.

golem/task/requestedtaskmanager.py
```diff
@@ -202,6 +206,53 @@ def start_task(task_id: TaskId) -> None:
...
+                task_id=db_task.task_id,
+                environment=db_task.env_id,
+                task_owner=self._node,
+                deadline=int(db_task.deadline.timestamp()),
+                subtask_timeout=db_task.subtask_timeout,
+                subtasks_count=db_task.max_subtasks,
+                # estimated_memory=task_definition.estimated_memory,
+                max_price=db_task.max_price_per_hour,
+                # Concent Disabled for task-api
+                concent_enabled=False,
```
Krigpl: Use the value from the db - it's set to `False` during create_task anyway.

golem/task/requestedtaskmanager.py
```diff
@@ -202,6 +206,53 @@ def start_task(task_id: TaskId) -> None:
...
+                TaskStatus.computing,
+            ])
+        ).execute()
+        # filter out Task.needs_computation() ??
+
+        def _make_task_header(db_task: RequestedTask) -> TaskHeader:
+            return TaskHeader(
+                min_version=str(gconst.GOLEM_MIN_VERSION),
+                task_id=db_task.task_id,
+                environment=db_task.env_id,
```
Krigpl: What about env_prerequisites?

golem/task/requestedtaskmanager.py
```diff
@@ -202,6 +206,53 @@ def start_task(task_id: TaskId) -> None:
...
+                # estimated_memory=task_definition.estimated_memory,
+                max_price=db_task.max_price_per_hour,
+                # Concent Disabled for task-api
+                concent_enabled=False,
+                timestamp=int(time.time()),
+            )
+
+        return [_make_task_header(task) for task in pending_tasks]
+
+    def got_wants_to_compute(self, task_id: str):
```
Krigpl: I wouldn't introduce this method until it's really needed and does something useful, but up to you.

golem/task/requestedtaskmanager.py
```diff
@@ -202,6 +206,53 @@ def start_task(task_id: TaskId) -> None:
...
+                environment=db_task.env_id,
+                task_owner=self._node,
+                deadline=int(db_task.deadline.timestamp()),
+                subtask_timeout=db_task.subtask_timeout,
+                subtasks_count=db_task.max_subtasks,
+                # estimated_memory=task_definition.estimated_memory,
+                max_price=db_task.max_price_per_hour,
+                # Concent Disabled for task-api
+                concent_enabled=False,
+                timestamp=int(time.time()),
```
Krigpl: Mind the fact that the timestamp here will be updated during each call, meaning providers will override the headers anytime they receive them.

golem/task/taskserver.py
```diff
@@ -639,7 +639,9 @@ def disconnect(self):
                 logger.error("Error closing session: %s", exc)
 
     def get_own_tasks_headers(self):
-        return self.task_manager.get_tasks_headers()
+        old_headers = self.task_manager.get_tasks_headers()
+        new_headers = self.requested_task_manager.get_task_headers()
```
Krigpl: You need to sign the headers somewhere.

Krigpl (commented):
Not a fan of putting the TaskHeaders dependency inside the RTM (vs doing the conversion outside the class) but I understand that's the easiest.

golem/task/requestedtaskmanager.py
maaktweluit: Fixed, made it a constant inside `taskstatus.py`

golem/task/requestedtaskmanager.py
maaktweluit: good one, added them :)

golem/task/requestedtaskmanager.py
maaktweluit: Update the timestamp of the header to be read from the database too

golem/task/taskserver.py
maaktweluit: they are now signed too, good catch!

golem/task/requestedtaskmanager.py
maaktweluit: done

golem/task/requestedtaskmanager.py
maaktweluit: removed, replaced with a fixme for next PR

golem/task/taskserver.py
```diff
@@ -639,7 +641,39 @@ def disconnect(self):
...
 
     def get_own_tasks_headers(self):
-        return self.task_manager.get_tasks_headers()
+        old_headers = self.task_manager.get_tasks_headers()
+        new_headers = self._get_and_sign_headers()
+        return old_headers + new_headers
+
+    def _get_and_sign_headers(self):
+
+        pending_tasks = RequestedTask.select().where(
```
Krigpl: That query and the checks from line 657 should really live inside the RTM, e.g. in the `get_active_tasks` method for proper encapsulation.

tests/golem/task/test_taskserver.py
```diff
@@ -891,6 +892,27 @@ def test_add_task_header_invalid_sig(self):
...
+            max_price_per_hour=1,
+            max_subtasks=1,
+            output_directory=Path(self.tempdir) / 'output',
+            start_time=default_now(),
+        )
+        self.ts.keys_auth._private_key = b'123'
+        # when
+        result = self.ts.get_own_tasks_headers()
+        print(result)
+        # then
```
Krigpl, line 908: Considering that it came up during review you should also put an assertion that the header is signed.

golem/task/taskserver.py
```diff
@@ -639,7 +641,39 @@ def disconnect(self):
...
+        logger.debug('pending_tasks=%r', pending_tasks)
+        signed_headers = []
+        for db_task in pending_tasks:
+            if db_task.start_time is None or db_task.deadline is None:
+                continue
+            task_header = dt_tasks.TaskHeader(
+                min_version=str(gconst.GOLEM_MIN_VERSION),
+                task_id=db_task.task_id,
+                environment=db_task.env_id,
+                environment_prerequisites=json.loads(db_task.prerequisites),
```
Krigpl: I might be wrong but `RequestedTask.prerequisites` is a `JsonField` which automatically converts its string value to a dict, so the `loads` call is not needed here.

mfranciszkiewicz: That's true, `loads` is not needed

maaktweluit: i thought so too, then got this error:
```
golem_messages.exceptions.FieldError: dict is expected not <class 'str'> [environment_prerequisites:'{}']
```

Maybe it needs some `.as_dict()` magic?

mfranciszkiewicz: This looks like the default value for the `JsonField`. Are we sure that it should be a string (not a dict returning _function_)?

golem/task/taskserver.py
maaktweluit: good one, moved the query and checks back

tests/golem/task/test_taskserver.py
maaktweluit: Added a mock for the sign function to check, since keys_auth is also mocked

golem/task/taskserver.py
maaktweluit: we have a winner! that was the case, the json fields in RequestedTask are fixed and the json.loads call is removed :)

Wiezzel (approved):

golem/task/taskserver.py
```diff
@@ -181,10 +182,10 @@ def __init__(self,
             finished_cb=task_finished_cb,
         )
         self.requested_task_manager = RequestedTaskManager(
-            env_manager=new_env_manager,
             app_manager=AppManager(),
-            root_path=TaskServer.__get_task_manager_root(client.datadir),
+            env_manager=new_env_manager,
```
Krigpl, line 193: `env_manager` is still the first parameter, isn't it?

tests/golem/task/test_taskserver.py
```diff
@@ -891,6 +892,28 @@ def test_add_task_header_invalid_sig(self):
         result = self.ts.add_task_header(Mock())
         self.assertFalse(result)
 
+    @patch('golem.task.taskserver.dt_tasks.TaskHeader.sign')
+    def test_get_own_task_headers(self, mock_sign):
+        # given
+        task_id = 'abc'
+        RequestedTask.create(
```
Krigpl: This isn't a unit test not an integration test.
In a unit test `RTM::get_started_tasks` would be mocked to return provided value.
In an integration test a task would be created and started in a production like style (i.e. calling create and start on rtm).
This is either since it creates some task entry in the db artificially and doesn't reflect anything. That approach should be avoided (as creating arbitrary DB entries in general when there's a dedicated class for managing that stuff).

golem/task/taskserver.py
maaktweluit: the order is now alphabetical, it should not matter when using kwargs, right?

can change it back, would be a good one for the style guide, the order of arguments vs alphabetical order

tests/golem/task/test_taskserver.py
maaktweluit: at least its still a test ;)
Was thinking about it when moving the query to the RTM.. 

can split the tests in 2, where the RTM one makes a row in the database ( and stays a half baked unit test ) and the taskserver one with a mocked value, would that improve it?

golem/task/taskserver.py
Krigpl: Right, I forgot for a moment it doesn't matter for kwargs. I'm fine either way then.

tests/golem/task/test_taskserver.py
Krigpl: Since you introduced a new method in the RTM there for sure should be a method in the test_rtm to test this method. And that one should use create and start methods to create the db entry and check that it gets returned from this method.

As for this one, test_taskserver isn't really well designed and there isn't a good solution here. As a bandaid solution I'd say to patch RTM::get_started_tasks within this test and work with that, since a proper integration would have much larger scope to cover.

Krigpl (approved):
Looks good !

tests/golem/task/test_taskserver.py
maaktweluit: done, looks a lot better now, thanks!

mfranciszkiewicz (approved):

PR #4732, opened 2019-09-16
Implement duplicate task method

Krigpl:
Another way of restarting the task is to duplicate the task and start it from scratch.

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -2,6 +2,7 @@ import asyncio
 import json
 from datetime import timedelta
 import logging
+import os
 from pathlib import Path
 import shutil
 from typing import Any, Dict, List
@@ -70,7 +71,7 @@ class DirManager:
     def prepare_task_dir(self, app_id: str, task_id: TaskId) -> Path:
         task_dir = self._get_task_dir(app_id, task_id)
         task_dir.mkdir()
-        task_inputs_dir = task_dir / constants.TASK_INPUTS_DIR
+        task_inputs_dir = self.get_task_inputs_dir(app_id, task_id)
         task_inputs_dir.mkdir()
         subtask_inputs_dir = self.get_subtask_inputs_dir(app_id, task_id)
         subtask_inputs_dir.mkdir()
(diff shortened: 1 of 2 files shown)


golem/task/requestedtaskmanager.py
```diff
@@ -80,6 +81,10 @@ def prepare_task_dir(self, app_id: str, task_id: TaskId) -> Path:
         subtask_outputs_dir.mkdir()
         return task_inputs_dir
 
+    def get_task_inputs_dir(self, app_id: str, task_id: TaskId) -> Path:
```
maaktweluit, line 84: would be nice to use this function on line 74 since its now duplicated there

golem/task/requestedtaskmanager.py
```diff
@@ -366,6 +371,24 @@ def get_subtask_outputs_dir(self, task_id: TaskId) -> Path:
         task.status = TaskStatus.waiting
         task.save()
 
+    async def duplicate_task(self, task_id: TaskId, output_dir: Path) -> TaskId:
+        task = RequestedTask.get(RequestedTask.task_id == task_id)
+        inputs_dir = self._dir_manager.get_task_inputs_dir(task.app_id, task_id)
+        resources = map(lambda f: inputs_dir / f, os.listdir(inputs_dir))
+        golem_params = CreateTaskParams(
+            app_id=task.app_id,
+            name=f'{task.name} copy',
```
maaktweluit, line 380: This is not safe when making a copy of a copy of a copy, think there is a limit on max characters

tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -266,6 +266,30 @@ def test_task_not_exists(self):
         await self.rtm.restart_task(task_id)
         assert not self.rtm.is_task_finished(task_id)
 
+    @pytest.mark.asyncio
+    async def test_duplicate_task(self, mock_client):
+        resources_dir = self.tmp_path / 'resources'
```
maaktweluit, line 271: should we use the task-api constant here?

maaktweluit (commented):
LGTM!

Some small q's will approve on reply

tests/golem/task/test_requestedtaskmanager.py
Krigpl: No, this is an arbitrary input directory.

golem/task/requestedtaskmanager.py
Krigpl: It's as not safe as creating a task with a long name - that should be handled in the `create_task` then imo.

maaktweluit: true dat :)

maaktweluit (approved):

PR #4746, opened 2019-09-18
Upgrade to Task API v0.18.3 and some fixes

Wiezzel:
* `RequestorAppClient.create_task` reply has `prerequisites` dictionary, not `prerequisites_json`.
* `RequestorAppClient.next_subtask` requires `opaque_node_id` parameter.
* `RequestorAppClient.next_subtask` may return `None` when the application refuses to assign a subtask to a particular given node. Such situation has to be handled by `RequestedTaskManager`.
* Node name should not belong to the search query when checking `ComputingNode` because it may change.
* When calling `RequestedTaskManager.get_next_subtask` in `TaskSession`, `computing_node` parameter has to be provided and the result has to be properly awaited.

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -1,11 +1,11 @@
 import asyncio
-import json
+import hashlib
 from datetime import timedelta
 import logging
 import os
 import shutil
 from pathlib import Path
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from dataclasses import dataclass
 from golem_messages import idgenerator
@@ -188,7 +188,7 @@ class RequestedTaskManager:
             task.app_params,
         )
         task.env_id = reply.env_id
-        task.prerequisites = json.loads(reply.prerequisites_json)
+        task.prerequisites = reply.prerequisites
         task.save()
         logger.debug('init_task(task_id=%r) after', task_id)
 
@@ -254,7 +254,7 @@ class RequestedTaskManager:
             self,
             task_id: TaskId,
             computing_node: ComputingNodeDefinition
-    ) -> SubtaskDefinition:
+    ) -> Optional[SubtaskDefinition]:
         """ Return a set of data required for subtask computation. """
         logger.debug(
             'get_next_subtask(task_id=%r, computing_node=%r)',
@@ -265,7 +265,7 @@ class RequestedTaskManager:
         task = RequestedTask.get(RequestedTask.task_id == task_id)
         node, _ = ComputingNode.get_or_create(
             node_id=computing_node.node_id,
-            name=computing_node.name
+            defaults={'name': computing_node.name}
         )
 
         # Check not providing for own task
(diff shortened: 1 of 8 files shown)


golem/task/requestedtaskmanager.py
```diff
@@ -283,7 +283,16 @@ def get_subtask_outputs_dir(self, task_id: TaskId) -> Path:
...
 
         app_client = await self._get_app_client(task.app_id)
-        result = await app_client.next_subtask(task.task_id)
+        result = await app_client.next_subtask(
+            task_id=task.task_id,
+            opaque_node_id=hashlib.sha3_256(node.node_id.encode()).hexdigest()  # noqa pylint: disable=no-member
+        )
+
+        if result is None:
+            raise RuntimeError(
```
Krigpl: That shouldn't be an exception imo - this is expected behavior after all and should be handled normally.

Krigpl (approved):

golem/task/requestedtaskmanager.py
```diff
@@ -283,7 +283,17 @@ def get_subtask_outputs_dir(self, task_id: TaskId) -> Path:
                 f"Task not pending, no next subtask. task_id={task_id}")
 
         app_client = await self._get_app_client(task.app_id)
-        result = await app_client.next_subtask(task.task_id)
+        result = await app_client.next_subtask(
+            task_id=task.task_id,
+            opaque_node_id=hashlib.sha3_256(node.node_id.encode()).hexdigest()  # noqa pylint: disable=no-member
```
maaktweluit, line 288: who is `no-member` here? if its `encode` maybe you can use `str.encode(value_to_enc)`

tests/golem/envs/localhost.py
```diff
@@ -99,10 +100,14 @@ def __init__(self, prereq: LocalhostPrerequisites) -> None:
             task_work_dir: Path,
             max_subtasks_count: int,
             task_params: dict
-    ) -> None:
-        pass
+    ) -> Task:
+        return await self._prereq.create_task()  # type: ignore
```
maaktweluit, line 104: why do we have to ignore the type here? would be nice to keep new files clean of linter hacks

golem/task/tasksession.py
```diff
@@ -465,8 +466,17 @@ def _get_next_ctd(
         if self.requested_task_manager.task_exists(task_id):
             if not self.requested_task_manager.has_pending_subtasks(task_id):
                 return None
-            subtask_definition = \
-                self.requested_task_manager.get_next_subtask(task_id)
+            subtask_definition = yield deferred.deferred_from_future(
+                self.requested_task_manager.get_next_subtask(
+                    task_id=task_id,
+                    computing_node=ComputingNodeDefinition(
+                        name='',  # FIXME: Provide proper node name
```
maaktweluit, line 473: is there an issue for this? 

maybe we do not need the name stored here at all

maaktweluit (approved):
LGTM!

Some nitty comments, still approving

golem/task/requestedtaskmanager.py
Wiezzel: No, it's `sha3_256`. I'm afraid there's nothing we can do.

tests/golem/envs/localhost.py
Wiezzel: That's what mypy says:
```error: Attribute function "create_task" with type "Callable[[], Awaitable[Optional[Subtask]]]" does not accept self argument```

golem/task/tasksession.py
Wiezzel: I'll create a task. It's going to be useful when we would like to display who's computing the subtask in the GUI.

golem/task/requestedtaskmanager.py
maaktweluit: 😭

tests/golem/envs/localhost.py
maaktweluit: bummer

https://github.com/python/mypy/issues/708 ( issue from 2015 still open )

golem/task/tasksession.py
maaktweluit: yes, but it is already stored and updated at `KnownHosts` ( i think )

would be nice to have a single truth about what name belongs to what id

Wiezzel: OK, that will go to the newly created task. :wink:

PR #4751, opened 2019-09-19
Readjust concent force payment integration test

jiivan:
Requires: #4750


shadeofblue (approved):
darth-vader-thumbs-up

PR #4757, opened 2019-09-20
fixed 1 small bugs in RTM

maaktweluit:
- subtask_id was not used in the loop of discard
- ~subtasks status was not updated on timeout~ ( bigger then expected, made an issue on CLAY )

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -483,16 +483,13 @@ class RequestedTaskManager:
             shared_dir=shared_dir
         )
 
-    def _check_task_timeout(self, task_id: TaskId) -> None:
+    @staticmethod
+    def _check_task_timeout(task_id: TaskId) -> None:
         task = RequestedTask.get(RequestedTask.task_id == task_id)
         if task.status.is_active():
             logger.info("Task timed out. task_id=%r", task_id)
             task.status = TaskStatus.timeout
             task.save()
-            pending_subtasks = self._get_pending_subtasks(task_id)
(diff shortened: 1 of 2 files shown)


golem/task/requestedtaskmanager.py
```diff
@@ -483,13 +483,16 @@ def _get_task_api_service(
...
-    def _check_task_timeout(task_id: TaskId) -> None:
+    def _check_task_timeout(self, task_id: TaskId) -> None:
         task = RequestedTask.get(RequestedTask.task_id == task_id)
         if task.status.is_active():
             logger.info("Task timed out. task_id=%r", task_id)
             task.status = TaskStatus.timeout
             task.save()
+            pending_subtasks = self._get_pending_subtasks(task_id)
+            for subtask in pending_subtasks:
+                subtask.status = SubtaskStatus.failed
```
Krigpl: Please note that the we may receive the results from the provider anyway and they will handled thus overriding the status to success (I think).

Krigpl (approved):
A good practice when fixing bugs is to write a unit test to prevent them from happening ever again :)

golem/task/requestedtaskmanager.py
maaktweluit: meh, lets not open this can of worms now ;)

Made an issue to look into this properly

maaktweluit:
@Krigpl good one, extended the unit test to use multiple subtasks to catch this

Krigpl (approved):
Awesome, thanks for fixing this.

PR #4770, opened 2019-09-25
 Upgrade to Task API v0.20.0 

Wiezzel:
* Return VerifyStatus instead of bool from verify method.
* Use dirtuils for managing requestor and provider directories.

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -9,11 +9,12 @@ from typing import Any, Dict, List, Optional
 
 from dataclasses import dataclass
 from golem_messages import idgenerator
-from golem_task_api import constants
+from golem_task_api.dirutils import RequestorDir, RequestorTaskDir
+from golem_task_api.enums import VerifyResult
 from golem_task_api.client import RequestorAppClient
 from peewee import fn
 
-from golem.app_manager import AppManager
+from golem.app_manager import AppManager, AppId
 from golem.model import (
     ComputingNode,
     default_now,
@@ -24,7 +25,6 @@ from golem.task.envmanager import EnvironmentManager, EnvId
 from golem.task.taskstate import TaskStatus, SubtaskStatus, TASK_STATUS_ACTIVE
 from golem.task.task_api import EnvironmentTaskApiService
 from golem.task.timer import ProviderComputeTimers
-
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


golem/task/requestedtaskmanager.py
```diff
@@ -152,12 +144,11 @@ def create_task(
             task.task_id,
             task.app_id,
         )
-        task_inputs_dir = self._dir_manager.prepare_task_dir(
-            task.app_id,
-            task.task_id)
+        task_dir = self._dir_manager.task_dir(task.app_id, task.task_id)
```
maaktweluit: do we still need self._dir_manager? there is only one function left that is more then 1 line

maaktweluit (changes requested):
LGTM!

One small q, approve on reply or solve :)

golem/task/requestedtaskmanager.py
Wiezzel: Heh, not really. Will refactor.

maaktweluit (approved):
LGTM!

Not sure if we need all those dirs as public functions, some might be only used by the RTM and then `self._task_dir(...).some_dir` would suffice

golem/task/requestedtaskmanager.py
```diff
@@ -59,60 +59,55 @@ class SubtaskDefinition:
...
         self._public_key: bytes = public_key
         self._app_clients: Dict[EnvId, RequestorAppClient] = {}
 
+    def _app_dir(self, app_id: AppId) -> RequestorDir:
+        app_dir = RequestorDir(self._root_path / app_id)
+        app_dir.mkdir(exist_ok=True)
+        return app_dir
+
+    def _task_dir(self, task_id: TaskId) -> RequestorTaskDir:
+        task = RequestedTask.get(RequestedTask.task_id == task_id)
```
maaktweluit, line 85: Seems like this query is always duplicate now, maybe we can make the dir part take a RequestedTask as argument?

Wiezzel: `get_subtask_inputs_dir` and `get_subtask_outputs_dir` are called from outside RTM. Other methods might probably be called as well at some moment. I don't see how we get it to work with `RequestedTask` as an argument instead of `task_id`.

maaktweluit: yeah, that will only work nice when called from inside RTM.. lets leave it for now

PR #4774, opened 2019-09-30
fix enable_{talkback,monitor} typing

### LINKED ISSUE #4478: `ValueError: invalid literal for int() with base 10: 'True'` while loading app_cfg.ini
## Description

**Golem Version**: `0.20.0+dev9.g13fafea`
**OS** [e.g. Windows 10 Pro]: `Ubuntu 18.04`
**Branch** (if launched from source): `b0.20`
**Mainnet/Testnet**: mainnet
**Description of the issue**:

```
'Reading config from file /home/blue/hyperg-test/mainnet/app_cfg.ini' ... failed with an exception
  File "/home/blue/PycharmProjects/golem/golem/core/simpleconfig.py", line 125, in __init__
    self.__read_options(cfg)
  File "/home/blue/PycharmProjects/golem/golem/core/simpleconfig.py", line 194, in __read_options
    prop.set_value_from_str(self.__read_option(cfg, prop))
  File "/home/blue/PycharmProjects/golem/golem/core/simpleconfig.py", line 48, in set_value_from_str
    value_type = self._value_type(val)
ValueError: invalid literal for int() with base 10: 'True'
```

## Steps To Reproduce
1. Launch a brand-new instance of Golem on mainnet directly from shell (using `golemapp.py`)
2. Have it act as a provider?
3. Stop Golem, Start Golem

## Expected behavior
No error

etam:
fixes #4478

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -1095,7 +1095,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
             self.task_server.change_config(self.config_desc,
                                            run_benchmarks=run_benchmarks)
 
-        self.enable_talkback(self.config_desc.enable_talkback)
+        self.enable_talkback(bool(self.config_desc.enable_talkback))
         self.app_config.change_config(self.config_desc)
 
         dispatcher.send(
@@ -1372,7 +1372,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
         return None
 
     @staticmethod
-    def enable_talkback(value):
+    def enable_talkback(value: bool):
         enable_sentry_logger(value)
 
     @rpc_utils.expose('net.peer.block')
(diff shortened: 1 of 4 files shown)


etam:
Explanation: in `golem.appconfig.NodeConfig` and `golem.clientconfigdescriptor.ClientConfigDescriptor` values are stored as `int`, but in `golem.node.Node` fields `_use_monitor` and `_use_talkback` are *supposed* to be `bool`. Because types were not checked nor enforced, bools were saved in `ClientConfigDescriptor` and serialized to `app_cfg.ini`.

golem/client.py
```diff
@@ -1372,7 +1372,7 @@ def activate_hw_preset(self, name, run_benchmarks=False):
         return None
 
     @staticmethod
-    def enable_talkback(value):
+    def enable_talkback(value: int):
```
mfranciszkiewicz: IMO this (and `value` in `enable_sentry_logger`) should be a `bool`ean value. This way we won't spread `config_desc` issues to the rest of the APIs. Any configuration setting can be cast to `bool` when needed.

etam: done

mfranciszkiewicz (approved):

PR #4779, opened 2019-10-01
explicitly pass the stats dict as `ProviderStats`

shadeofblue:
explicitly pass the stats dict as `ProviderStats` instead of a plain dict to make the serialization and deserialization consistent

fixes the issue with Concent's Force Report node integration test


kmazurek (approved):
sorry

PR #4783, opened 2019-10-04
Task API verification queue

mfranciszkiewicz:
A new queue for verification purposes.

- persistent: database backed
- properly handles the `AWAITING_DATA` verification result preventing "forever loops" caused by re-queueing (see class docs)
- used exclusively by `RequestedTaskManager`

### DIFF
--- a/golem/task/verification/queue/queue.py
+++ b/golem/task/verification/queue/queue.py
@@ -50,7 +50,6 @@ class VerificationQueue:
         self._queue = backend or DatabaseQueueBackend()
         # In-memory store for pending calls
         self._pending: Dict[Tuple[TaskId, SubtaskId], asyncio.Future] = dict()
-        self._pending: Dict[Tuple[TaskId, SubtaskId], asyncio.Future] = dict()
         # Tells whether the queue processing is running
         self._processing = False
         # Tells whether queue processing was paused by the user
--- a/golem/task/verification/queue/queue.py
+++ b/golem/task/verification/queue/queue.py
@@ -1,0 +1,136 @@
+import asyncio
+import logging
+from concurrent import futures
+from typing import Dict, Callable, Optional, Coroutine, Any, Tuple
+
+from golem_task_api.enums import VerifyResult
+
+from golem.core.common import get_timestamp_utc
+from golem.task import SubtaskId, TaskId
+from golem.task.verification.queue.backend import QueueBackend, \
+    DatabaseQueueBackend
+
+logger = logging.getLogger(__name__)
+
+VerifyFn = Callable[[TaskId, SubtaskId], Coroutine[Any, Any, VerifyResult]]
+
+
+def _next_priority() -> int:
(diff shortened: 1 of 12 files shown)


mfranciszkiewicz:
TODO in a follow-up PR: trigger pending verifications on startup

tests/golem/task/verification/queue/test_queue.py
```diff
@@ -0,0 +1,194 @@
...
+    await asyncio.sleep(0.5)
+    return VerifyResult.SUCCESS
+
+
+@pytest.mark.usefixtures('pytest_database_fixture')
+class TestPut:
+
+    @pytest.mark.asyncio
+    async def test_put_duplicate(self):
+        queue = VerificationQueue(verify_fn)
```
jiivan: This line is repeated in every test here. It oculd be avoided by using *unittest* and `.setUp()`. Also IMO assert gives less readable output in case of failures than *unittest* assert methods do.

tests/golem/task/verification/queue/test_backend.py
```diff
@@ -0,0 +1,71 @@
...
+        assert self.backend.get() == (f"{TASK_ID}1", f"{SUBTASK_ID}1")
+        assert self.backend.get() is None
+
+    def test_update_priorities(self):
+        counter = 0
+
+        def priority_fn() -> int:
+            nonlocal counter
+            counter += 1
+            return counter
```
jiivan: Instead of `priority_fn` maybe using `itertools.count().next()` would be easier?

golem/task/verification/queue/queue.py
```diff
@@ -0,0 +1,135 @@
...
+            result = VerifyResult.FAILURE
+            logger.error("Verification timeout: subtask_id=%s", subtask_id)
+
+        if result is VerifyResult.AWAITING_DATA:
+            self._queue.put(
+                task_id,
+                subtask_id,
+                priority=None)
+        else:
+            # The queue is persistent, there may be no Future object in memory
```
jiivan, line 132: If I understand correctly, shouldn't we skip verification if pending entry is missing?

mfranciszkiewicz: Pending entries are only missing after an application restart, when we no longer wait for a specific subtask verification to finish. Verification can still be performed and the subtask state will in turn be updated by the `RequestedTaskManager._verify` method. 

I didn't see a need to provide means to `await` each subtask verification when Golem is restarted. It's still possible to `await` verification of all items in the queue. If you think otherwise, I'm all ears.

tests/golem/task/verification/queue/test_queue.py
mfranciszkiewicz: I've created a `setUp`-like fixture for the tests. I wouldn't want to convert these to `unittest` because of the `pytest-asyncio` plugin / integration and the overall minimalistic approach.

jiivan (approved):

PR #4785, opened 2019-10-08
Include market strategy in app definition

mfranciszkiewicz:
Currently a `DEFAULT_REQUESTOR_MARKET_STRATEGY` is enforced


Wiezzel (commented):
Looks OK but what's the deal with thousands of changed lines in taskserver and tasksession?

mfranciszkiewicz:
That was a small hiccup with my IDE. Fixed

golem/marketplace/__init__.py
```diff
@@ -15,5 +17,36 @@
...
+            return cls[name]
+        except KeyError:
+            return cls.default()
+
+    @classmethod
+    def default(cls):
+        raise NotImplemented
+
+
+class RequestorMarketStrategies(NameSerializableEnum):
```
Wiezzel: Why plural name? We never used plural names for enums or any other classes.

golem/marketplace/__init__.py
```diff
@@ -15,5 +17,36 @@
...
+    Brass = RequestorBrassMarketStrategy
+    Wasm = RequestorWasmMarketStrategy
```
Wiezzel: Our styleguide doesn't mention enums but i think the options should be either in `ALL_CAPS` or `snake_case`.

golem/marketplace/__init__.py
```diff
@@ -15,5 +17,36 @@
...
+
+    @classmethod
+    def default(cls) -> 'RequestorMarketStrategies':
+        return cls.Brass
+
+
+class ProviderMarketStrategies(NameSerializableEnum):
```
Wiezzel: See comments about `RequestorMarketStrategies`.

Wiezzel (commented):
Some code style suggestions.

golem/marketplace/__init__.py
mfranciszkiewicz: `RequestorMarketStrategy` already exists. Other names I came up with weren't good enough. `RequestorMarketStrategyEnum`?

Wiezzel: Oh, let's leave it then. I cannot think of any better name now.

Wiezzel (approved):

mfranciszkiewicz:
@Wiezzel Please re-review. One of the tests was failing but due to the impact of the bug on the rest of the codebase I've changed the solution

Wiezzel (approved):
OK

PR #4788, opened 2019-10-10
New subtasks count limit

### LINKED ISSUE #2428: Cleanup usage of Defaults classes in Apps
We have TaskDefaults class and all the classes that inherits from it, ie. BlenderDefaults, LuxRenderDefaults, etc. They were used in old GUI, but are probably ignored in the new one. They are partially used in building task definition from dictionary, but not all fields are included. 
We should clean usage of these classes and make its function more explicit.

### LINKED ISSUE #4003: Cannot create tasks with subtasks counts > 50

etam:
- resolves #4003
- resolves #2428
- removes `TaskDefinition.optimize_total`, which seems to be unused and quite vague.
- removes `comp.task.subtasks.count` rpc call for good this time (see https://github.com/golemfactory/golem-electron/pull/464)


etam:
@shadeofblue Do we want this in b0.21 (#4003 is labeled for this release), or develop (because I think it's quite significant change)?

shadeofblue:
@etam develop it should be, then

apps/core/task/coretask.py
```diff
@@ -280,9 +280,8 @@ def get_tasks_left(self):
         return (self.get_total_tasks() - self.last_task) \
             + self.num_failed_subtasks
 
-    # pylint:disable=unused-argument
-    @classmethod
-    def get_subtasks(cls, part):
+    # pylint:disable=unused-argument,no-self-use
+    def get_subtasks(self, part) -> Dict[str, dict]:
```
jiivan, line 284: ?

apps/rendering/task/renderingtask.py
```diff
@@ -244,24 +245,8 @@ class RenderingTaskBuilderError(Exception):
...
-        optimize_total: bool,
-        defaults: RendererDefaults) -> int:
-    if optimize_total:
-        return defaults.default_subtasks
-
-    if defaults.min_subtasks <= subtasks_count <= defaults.max_subtasks:
-        return subtasks_count
-
-    logger.warning("Cannot set total subtasks to %s. Changing to %s",
-                   subtasks_count, defaults.default_subtasks)
```
jiivan, line 258: Maybe move this log into `framerenderingtask._calculate..`

jiivan (approved):
heavyrain_nod

apps/core/task/coretask.py
etam: This method is overridden in deriving classes. You can't override a classmethod with a normal method. At least mypy is complaining.

apps/rendering/task/renderingtask.py
etam: done

mdtanrikulu (approved):

magdasta (approved):

PR #4802, opened 2019-10-17
Usage factor persistence

shadeofblue:
closes GR-178


golem/database/schemas/040_usage_factor.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @migrator.create_model
+    class UsageFactor(pw.Model):
+        created_date = pw.UTCDateTimeField(default=dt.datetime.now)
+        modified_date = pw.UTCDateTimeField(default=dt.datetime.now)
+        provider_node = pw.ForeignKeyField(db_column='provider_node_id', rel_model=migrator.orm['computingnode'], related_name='usage_factor', to_field='node_id', unique=True)
+        usage_factor = pw.FloatField(default=1.0)
+
+        class Meta:
+            db_table = "usagefactor"
```
mbenke: Does it need a separate model/table? Isn't it better to just add a field  in `LocalRank`?

mbenke:
It might be worthwhile to add some tests along the lines of  `TestLocalRank`

golem/model.py
```diff
@@ -803,6 +803,12 @@ class Meta:
         primary_key = CompositeKey('task_id', 'subtask_id')
 
 
+class UsageFactor(BaseModel):
+    provider_node = ForeignKeyField(
+        ComputingNode, null=False, unique=True, related_name='usage_factor')
+    usage_factor = FloatField(default=1.0)
+
```
mbenke, line 820: All other models have

```
    class Meta:
        database = db
```

don't we need it here?

golem/database/schemas/040_usage_factor.py
```diff
@@ -0,0 +1,27 @@
+# pylint: disable=no-member
+import datetime as dt
+import peewee as pw
+
+SCHEMA_VERSION = 40
+
+try:
+    import playhouse.postgres_ext as pw_pext
```
kmazurek: Is this import required? How is it used here?

mbenke:
Also, `TestRequestorMarketStrategy` seems to need  to inherit `DatabaseFixture` now

golem/database/schemas/040_usage_factor.py
shadeofblue: it was added automatically by peewee_migrate...

golem/database/schemas/040_usage_factor.py
```diff
@@ -0,0 +1,27 @@
...
+except ImportError:
+    pass
+
+
+def migrate(migrator, database, fake=False, **kwargs):
+
+    @migrator.create_model
+    class UsageFactor(pw.Model):
+        created_date = pw.UTCDateTimeField(default=dt.datetime.now)
+        modified_date = pw.UTCDateTimeField(default=dt.datetime.now)
```
kubkon: I assume that we somehow use these for tracking usage factors in time? If not, are they not somewhat redundant?

golem/database/schemas/040_usage_factor.py
shadeofblue: I'm unsure but maybe I know too little about the marketplace in general -> how does a usage factor relate to LocalRank? is a usage factor always accompanied by a local rank record? does one implicitly assume the other or can they exist independently? are the properties in the LocalRank used  for the same purpose as the usage factor?

golem/marketplace/wasm_marketplace.py
```diff
@@ -44,15 +44,34 @@ def set_my_usage_benchmark(cls, benchmark: float) -> None:
...
+                defaults={'usage_factor': uf})
+        return usage_factor.usage_factor
+
+    @classmethod
+    def update_usage_factor(cls, provider_id: str, delta: float):
+        usage_factor = model.UsageFactor.select().where(
+            model.UsageFactor.provider_node_id == provider_id).first()
+
+        r = delta * usage_factor.usage_factor
+        logger.info("RWMS: adjust R for provider %s: %.3f -> %.3f",
```
kubkon, line 68: Silly question but what's "RWMS"? Weighted RMS perhaps?

kubkon (commented):
Definitely not an expert on this, but here's a couple of nitpicks.

golem/model.py
shadeofblue: hmm... peewee's docs say it's not needed:
http://docs.peewee-orm.com/en/latest/peewee/models.html#models-and-fields

, plus:

```
In [1]: from golem.model import BaseModel, GenericKeyValue, UsageFactor, db                                                                                                                                        

In [2]: BaseModel._meta.database is db                                                                                                                                                                             
Out[2]: True

In [3]: GenericKeyValue._meta.database is db                                                                                                                                                                       
Out[3]: True

In [4]: UsageFactor._meta.database is db                                                                                                                                                                           
Out[4]: True
```

...

golem/database/schemas/040_usage_factor.py
shadeofblue: these are inherited from BaseModel so, as standard feature of all golem models, so to say...

golem/marketplace/wasm_marketplace.py
shadeofblue: @kubkon no idea, as you may see, this logging line was just copied from the other place

shadeofblue: I believe it's more of `RequestorWasmMarketStrategy` than whatever else though, @kubkon

golem/database/schemas/041_usage_factor.py
```diff
@@ -0,0 +1,28 @@
+# pylint: disable=no-member
+import datetime as dt
+import peewee as pw
+
+SCHEMA_VERSION = 40
```
kmazurek: Note: the schema versions were shifted in #4807. I see you merged with develop, there's now a unit test which detects schema version collisions. Also, I'm adding a migration in #4811 - the race is on.
race

shadeofblue: @zakaprov yeah, I noticed ... I find our migration system horribly broken in that regard... migration numbering (as opposed to their contents) should NOT be something you have to take into account when resolving merge conflicts... :/

golem/database/schemas/040_usage_factor.py
shadeofblue: @mbenke ^ ?

golem/database/schemas/041_usage_factor.py
mfranciszkiewicz: `SCHEMA_VERSION` can be dropped altogether. Filenames decide the order that migrations are performed in.

If you believe that numbering is horribly broken please propose an alternative solution.

shadeofblue:
> It might be worthwhile to add some tests along the lines of `TestLocalRank`

@mbenke some tests added ...

golem/database/schemas/041_usage_factor.py
shadeofblue: @mfranciszkiewicz numbering itself is not the issue ... conflict resolution of those numbered migrations is the issue... it's okay for the numbering to determine the order but the existence of two migrations that share the same number should not be an issue. and as long as those migrations don't modify the same model, they should be able to co-exist... 

good example could be Django's `south` (later just renamed to "django migrations") which handles such cases gracefully... though, of course, it being a django-specific module, it cannot/should not be used in a project like ours ...

golem/database/schemas/040_usage_factor.py
mbenke: `LocalRank` is used basically for the same purpose as usage factors, i.e. provider choice.

golem/marketplace/wasm_marketplace.py
mbenke: Indeed, `RWMS` stands for `RequestorWasmMarketStrategy`. We use this abbreviation in log messages to make them shorter, but perhaps this abbreviation renders them less readable.

mbenke (approved):

golem/database/schemas/041_usage_factor.py
shadeofblue: @zakaprov lost it ;)

PR #4803, opened 2019-10-18
limit benchmark-based usage factors

mbenke:
Sometimes usage benchmark results can get misreported. This PR introduces sanity checks for usage factors based on these benchmark results, limiting them to the [0.25,1.5] range. 

Additionally, we prevent dividing by 0 (or other very small value of the usage benchmark).

### DIFF
--- a/golem/marketplace/wasm_marketplace.py
+++ b/golem/marketplace/wasm_marketplace.py
@@ -25,8 +25,11 @@ UsageReport = Tuple[ProviderId, SubtaskId, Usage]
 logger = logging.getLogger(__name__)
 
 
+USAGE_SECOND = 1e9  # usage is measured in nanoseconds
+
+
 class RequestorWasmMarketStrategy(RequestorPoolingMarketStrategy):
-    DEFAULT_USAGE_BENCHMARK: float = 1.0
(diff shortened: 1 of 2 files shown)


kmazurek (approved):

golem/marketplace/wasm_marketplace.py
```diff
@@ -49,8 +54,13 @@ def get_usage_factor(cls, provider_id, usage_benchmark):
             # Division goes this way since higher benchmark val means
             # faster processor
             usage_factor = cls.get_my_usage_benchmark() / usage_benchmark
-            if not usage_factor:
-                usage_factor = 1.0
+
+            # Sanity check against misreported benchmarks
+            usage_factor = min(max(usage_factor, 0.25), 1.5)
```
shadeofblue, line 59: why are we assuming that the usage cannot be farther away than that from our own? can't a really slow or really fast CPU have such - out of range - results?

mbenke: It can, but if it is, it will be caught by the usage factor adjustment. It may make sense to widen the limits though.

mbenke:
Abandoned in favour of #4827, which is a version of this PR rebased to develop after mewrging #4802

Issue #4380, opened 2019-06-24
sqlite3.ProgrammingError: Cannot operate on a closed database


ederenn:
## Description

**Golem Version**:    ] GOLEM Version: 0.19.2+dev610.g7dab45b

**Golem-Messages version** (leave empty if unsure):  golem_messages Version: 3.7.0

**Electron version** (if used): 0.2.0

**OS** [e.g. Windows 10 Pro]: Linux 18.04

**Branch** (if launched from source): b0.20

**Mainnet/Testnet**: mainnet

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

after restart of an app error appears in logs

## Steps To Reproduce
_Short description of steps to reproduce the behavior:_
e.g.

prerequisites: some tasks finished and in timeout

1. Launch  Golem 
2. Restart an app.
3. Check logs for error

## Logs and any additional context

```ERROR    [golem.task.taskserver              ] TaskServer.sync_network job <bound method TaskMessagesQueueMixin.connect_to_nodes of <golem.task.taskserver.TaskServer object at 0x7f0fc51d56d8>> failed
Traceback (most recent call last):
  File "/home/ederenn/projects/golem/golem/task/taskserver.py", line 208, in sync_network
    job()
  File "/home/ederenn/projects/golem/golem/task/server/queue_.py", line 97, in connect_to_nodes
    for node_id in msg_queue.waiting():
  File "/home/ederenn/projects/golem/golem/network/transport/msg_queue.py", line 69, in waiting
    ).group_by(model.QueuedMessage.node):
  File "/home/ederenn/projects/golem-env/lib/python3.6/site-packages/peewee.py", line 2289, in next
    obj = self.qrw.iterate()
  File "/home/ederenn/projects/golem-env/lib/python3.6/site-packages/peewee.py", line 2340, in iterate
    row = self.cursor.fetchone()
sqlite3.ProgrammingError: Cannot operate on a closed database.
```

## Proposed Solution?
_(Optional: What could be a solution for that issue)_

ederenn:
encountered again on Linux 18.04
rc 0.20.0
https://buildbot.golem.network/buildbot/#/builders/2/builds/4749

```
[...]

PR #4804, opened 2019-10-18
Handle errors during msg_queue iteration

jiivan:
fixes: #4380

### DIFF
--- a/golem/network/transport/msg_queue.py
+++ b/golem/network/transport/msg_queue.py
@@ -1,11 +1,13 @@
 import datetime
 import logging
+import sqlite3
 import threading
 import typing
 
 import golem_messages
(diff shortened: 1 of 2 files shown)


etam:
Is the source of "Cannot operate on a closed database" error well understood? Because that sounds for me very serious. And the proposed solution just looks like hiding the error message.

jiivan:
> Is the source of "Cannot operate on a closed database" error well understood? Because that sounds for me very serious. And the proposed solution just looks like hiding the error message.

From what I know it happens sometimes during startup sequence. It has no effect on subsequent calls. One could implement a retry based solution similar to the one implemented in `GolemSqliteDatabse` but IMO it would be an overkill in this case.

etam (approved):
ok

PR #4807, opened 2019-10-18
Fixed version collision in migration scripts

kmazurek:
There were two different migration scripts with schema version 36. This resulted in Golem failing to start when upgrading from b0.21 to develop.


etam (approved):

jiivan (commented):
Maybe it would be possible to write a test (`tests/golem/database/test_migration.py`) that looks for duplicates?

We should also rename migrations to match their numbers.

kmazurek:
@jiivan Migration files were renamed, I think it's not clearly indicated in the GitHub UI though. I'll look into creating a test for this case.

PR #4812, opened 2019-10-19
Build fixes

dmikushin:
This patch adds fixes to get golem `whl` built and `golemapp` to run from the `develop` branch. Some missing `__init__.py` files were added. A call to `move_wheel` has been commented out, as it makes the generated package uninstallable. Clearly, this dirty hack was on purpose, but it no longer works as expected.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -110,8 +110,8 @@ if not (in_appveyor() or in_travis() or
         print('Exception occurred:', exc)
         DockerManager().build_images()
 
-if building_wheel:
-    move_wheel()
+#if building_wheel:
+#    move_wheel()
 
 if not building_migration:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -110,8 +110,8 @@
         print('Exception occurred:', exc)
         DockerManager().build_images()
 
-if building_wheel:
-    move_wheel()
+#if building_wheel:
+#    move_wheel()
```
Wiezzel, line 114: Please remove these lines instead of commenting out.
Also, please remove the `setup_util.setup_commons.move_wheel` import.

Wiezzel (changes requested):

PR #4813, opened 2019-10-21
New logic for setting up and cleaning up environments

Wiezzel:
EnvironmentManager now ensures that exactly one environment is active at a time. Once an environment is needed it is automatically prepared. It would be cleaned up when another environment is going to be used.

This brings a disadvantage of keeping up idle environments but allows to avoid lengthy setups and cleanups if one environment is used a few times consecutively.


golem/task/envmanager.py
```diff
@@ -25,6 +27,36 @@ def __init__(self):
         self._envs: Dict[EnvId, EnvironmentManager.EnvEntry] = {}
         self._state = EnvStates()
         self._running_benchmark: bool = False
+        self._lock = DeferredLock()
+        self._active_env: Optional[Environment] = None
+
+    @inlineCallbacks
+    def _start_usage(self, env: Environment) -> Deferred:
+        yield self._lock.acquire()
```
maaktweluit, line 35: should it acquire the lock each time if the `_active_env == env`?

maaktweluit (approved):
LGTM!

Small comment, still approving!

golem/task/envmanager.py
Wiezzel: Yes.
Having the lock acquired means that the environment is currently running computation. After the computation ends, it will release the lock. On the other hand, `_active_env` means environment that is in `ENABLED` status.

mfranciszkiewicz (approved):

PR #4817, opened 2019-10-22
Use App Definition in task_api_tests/basic_integration.py

maaktweluit:


scripts/task_api_tests/basic_integration.py
```diff
@@ -1,60 +1,48 @@
...
-        environment: str,
-        env_prerequisites_json: str,
         task_params_path: Path,
+        app_definition_path: Path,
         resources: List[Path],
         max_subtasks: int,
 ) -> None:
-    env_prerequisites = json.loads(env_prerequisites_json)
 
+    app_definition = load_app_from_json_file(Path(app_definition_path))
```
mfranciszkiewicz, line 35:
```suggestion
    app_definition = load_app_from_json_file(app_definition_path)
```

scripts/task_api_tests/basic_integration.py
```diff
@@ -126,9 +114,13 @@
...
             subtask_def.subtask_id).mkdir()
         result_path = await task_computer.compute()
+        print(f'result_path={result_path}')
+        subtask_output_dir = rtm.get_subtask_outputs_dir(
+            task_id, subtask_def.subtask_id)
+        subtask_output_dir.mkdir()
         shutil.copy2(
             result_path,
-            rtm.get_subtask_outputs_dir(task_id, subtask_def.subtask_id),
+            subtask_output_dir / result_path.name,
```
mfranciszkiewicz:
```suggestion
            subtask_output_dir,
```
In general `copy` functions allow the second argument to be a directory

scripts/task_api_tests/basic_integration.py
```diff
@@ -1,60 +1,48 @@
...
-        docker.DockerTaskApiPayloadBuilder,
-    )
-    env_manager.set_enabled(cpu.DOCKER_CPU_METADATA.id, True)
+    register_built_in_repositories()
+    register_environments(
+        work_dir=work_dir,
+        env_manager=env_manager)
 
     rtm_work_dir = work_dir / 'rtm'
     rtm_work_dir.mkdir()
```
mfranciszkiewicz, line 48: Maybe `rtm_work_dir.mkdir(parents=True, exist_ok=True)`? (a GH suggestion here modifies whitespace)

`exist_ok=True` may also apply to other `mkdir` calls. What do you think?

mfranciszkiewicz (commented):
Nothing really important, maybe besides the first `mkdir` call

scripts/task_api_tests/basic_integration.py
```diff
@@ -1,60 +1,48 @@
...
-    docker_image = cpu.DockerCPUEnvironment.parse_prerequisites(
-        env_prerequisites).image
-    whitelist.Whitelist.add(docker_image)
-    env_manager.register_env(
-        docker_cpu_env,
-        cpu.DOCKER_CPU_METADATA,
-        docker.DockerTaskApiPayloadBuilder,
-    )
-    env_manager.set_enabled(cpu.DOCKER_CPU_METADATA.id, True)
+    register_built_in_repositories()
```
Wiezzel, line 42: Actually, if someone is testing their own app, registering built-in repositories won't help much. We need a mechanism for whitelisting the very image that the tested app uses. Which doesn't seem that obvious because app interface is abstract and whitelist is a docker-specific thing. I'm not sure if this issue belongs to this task but it surely has to be addressed.

scripts/task_api_tests/basic_integration.py
maaktweluit: tried this first, it gives this error :(

```
  File "/Users/mwu-gol/src/golem-project/golem/golem/apps/__init__.py", line 76, in load_app_from_json_file
    app_json = json_file.read_text(encoding='utf-8')
builtins.AttributeError: 'str' object has no attribute 'read_text'
```

seems something is wrong in Click with paths, will do some digging

Wiezzel (changes requested):
The code doesn't quite match the acceptance criteria:
> * Providing a JSON file still works
> * There is an additional option to run the test on app_id

There is no option to provide app_id instead of a JSON file.

scripts/task_api_tests/basic_integration.py
maaktweluit: works as intended: https://github.com/pallets/click/issues/405

Will update the types to be correct for the arguments :)

scripts/task_api_tests/basic_integration.py
maaktweluit: By default the testing folders are deleted at the end of each run, so i think it works as intended to require a clean dir each run.

If we want to drop this "feature" i like your proposal :)

scripts/task_api_tests/basic_integration.py
Wiezzel: Why not use https://pypi.org/project/click-pathlib/ ?

scripts/task_api_tests/basic_integration.py
maaktweluit: You are right, bit strange i remember trying without result_path.name first and it gave a not found error..

scripts/task_api_tests/basic_integration.py
maaktweluit: i think its a bit overkill to add a package, import and component that can break stuff to replace a single `Path(...)` cast

scripts/task_api_tests/basic_integration.py
maaktweluit: good one!

In the next PR ( app_id ) this call should be moved to the `task-by-app-id` call.

Would prefer to leave it as is in this PR since it works ( and its last weeks issue )

maaktweluit:
@Wiezzel this is last weeks issue, using a JSON only. 

app_id is a draft #4856  and will be open for review when this PR is merged

scripts/task_api_tests/basic_integration.py
Wiezzel: OK, fine.

Wiezzel (approved):

scripts/task_api_tests/basic_integration.py
maaktweluit: Updated the argument types

scripts/task_api_tests/basic_integration.py
mfranciszkiewicz: Let’s keep it as it is

mfranciszkiewicz (approved):

scripts/task_api_tests/basic_integration.py
maaktweluit: @Wiezzel updated the app_id PR accordingly, please check and comment there

Resolving this conversation

PR #4822, opened 2019-10-23
Return expected amount in unconfirmed incomes

jiivan:
Return more usable value in old `pay.incomes` rpc. This issue was brought up by a user on *#tech* channel.

awaitingincomes


shadeofblue (approved):
oh, okay

Issue #4801, opened 2019-10-16
preview FileNotFoundError


ederenn:
## Description

**Golem Version**: 0.21.0

**Golem-Messages version** (leave empty if unsure): 3.11.0

**Electron version** (if used): 0.2.3

**OS** [e.g. Windows 10 Pro]:Any

**Branch** (if launched from source): -

**Mainnet/Testnet**: mainnet

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

During requesting a task DiscBoyBlender  error appeared in `golem.log`
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/ederenn/.local/share/golem/default/mainnet/ComputerRes/519c058a-f018-11e9-856e-5ad86f3605c2/tmp/current_task_preview1.PNG_tmp.png' -> '/home/ederenn/.local/share/golem/default/mainnet/ComputerRes/519c058a-f018-11e9-856e-5ad86f3605c2/tmp/current_task_preview1.PNG'
```

Might be related, RPC was failing at that time. 

**Actual result**:

Golem might have problem with finishing subtasks/task.

**Screenshots**:

_If applicable, add screenshots to help explain your problem._

## Steps To Reproduce
_Short description of steps to reproduce the behavior:_
e.g.
1. Launch the Golem
2. Start a task at least with 40 subtasks. In This case, DiscBoyBlender, resolution and samples from file, frames 40-69, 60 subtasks , 2h for a task, 20 min for subtask, bid 0.1
3. Open console for electron - toggle developer tools in Golem view menu, watch for errors.
4.Check logs for  error.

## Logs and any additional context

```
ERROR    [golem.task.taskserver              ] TaskServer.sync_network job <bound method TaskServer.__remove_old_tasks of <golem.task.taskserver.TaskServer object at 0x7f7dc9672128>> failed
Traceback (most recent call last):
  File "/home/buildbot-worker/worker/buildpackage_linux/build/golem/task/taskserver.py", line 285, in sync_network
  File "/home/buildbot-worker/worker/buildpackage_linux/build/golem/task/taskserver.py", line 1108, in __remove_old_tasks
[...]

PR #4823, opened 2019-10-23
catch openvc errors better

etam:
related to #4801


jiivan (approved):
i_have_no_idea_what_im_doin

PR #4827, opened 2019-10-24
Fix division and limit benchmark-based usage factors

mbenke:
Sometimes usage benchmark results can get misreported. This PR introduces sanity checks for usage factors based on these benchmark results, limiting them to the [0.1,2] range.

This PR also modifies initial usage factor calculation to take into account the fact that with usage benchmarks, (in contrast to efficiency benchmarks), faster machines correspond to smaller values.

Additionally, we prevent dividing by 0 (or other very small value of the usage benchmark).

Note: this PR is basically #4803 rebased to develop after merging #4802


shadeofblue (approved):
darth-vader-thumbs-up

PR #4837, opened 2019-10-28
Provider/requestor switch

Wiezzel:
* Stop accepting tasks when a Task API task is requested.
* Resume accepting tasks when no more Task API tasks are requested.

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -32,6 +32,7 @@ from golem.task.taskstate import (
     TaskOp,
     TaskStatus,
     TASK_STATUS_ACTIVE,
+    TASK_STATUS_COMPLETED,
 )
 from golem.task.task_api import EnvironmentTaskApiService
 from golem.task.timer import ProviderComputeTimers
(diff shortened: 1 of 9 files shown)


golem/task/requestedtaskmanager.py
```diff
@@ -235,6 +236,12 @@ def is_task_finished(task_id: TaskId) -> bool:
         task = RequestedTask.get(RequestedTask.task_id == task_id)
         return task.status.is_completed()
 
+    @staticmethod
+    def has_unfinished_tasks() -> bool:
+        """ Return True iff there are any tasks that need computation. """
```
maaktweluit, line 241:
```suggestion
        """ Return True if there are any tasks that need computation. """
```

maaktweluit (approved):
LGTM!

One small comment, still approving!

golem/task/requestedtaskmanager.py
Wiezzel: 'iff' is a shorthand for 'if and only if'

maaktweluit: ok, TIL ^^

mfranciszkiewicz:
There is another side of the issue, unresolved - we're still able to request tasks while computing a subtask. But I’m fine with doing so separately

Wiezzel:
> There is another side of the issue, unresolved - we're still able to request tasks while computing a subtask. But I’m fine with doing so separately

Take a look at https://github.com/golemfactory/golem/pull/4837/commits/4d97349c05a84b79454819d91bc070ceda5d9e17

mfranciszkiewicz (approved):

PR #4838, opened 2019-10-29
Bump golemfactory/wasm to 0.5.3

kubkon:
This PR bumps our Wasm sandbox to latest release `0.5.3` in Docker Hub. The corresponding changes can be found at golemfactory/sp-wasm#37. For completeness though, here's the summary of major changes to the sandbox:
* polyfill time functions with a stub that always returns the same time of day
* modify the default Emscripten `_usleep` function with a custom function which underneath uses `std::thread::sleep` to perform the actual sleeping of the running thread


kmazurek (approved):
LGTM

PR #4840, opened 2019-10-29
+ calculate max usage helper

shadeofblue:
mv `compute_subtask_value` to a separate `helpers` module and rename to `calculate_subtask_amount`

+ `calculate_max_usage`
+ tests

### DIFF
--- a/golem/marketplace/brass_marketplace.py
+++ b/golem/marketplace/brass_marketplace.py
@@ -7,7 +7,7 @@ from golem.marketplace import (Offer, ProviderMarketStrategy, ProviderPricing)
 from golem.marketplace.pooling_marketplace import\
     RequestorPoolingMarketStrategy
 from golem.task import timer
-from golem.task.taskkeeper import compute_subtask_value
+from golem.task.helpers import calculate_subtask_payment
 import golem.ranking.manager.database_manager as dbm
 
 from .rust import order_providers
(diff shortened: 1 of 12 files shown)


golem/marketplace/brass_marketplace.py
```diff
@@ -7,7 +7,7 @@
 from golem.marketplace.pooling_marketplace import\
     RequestorPoolingMarketStrategy
 from golem.task import timer
-from golem.task.taskkeeper import compute_subtask_value
+from golem.task.helpers import calculate_subtask_amount
```
mbenke: Maybe it's just me but "amount" here does not feel right. What was wrong with "value"?

mbenke (approved):

golem/marketplace/brass_marketplace.py
shadeofblue: in programming context, `value` has a different connotation and obviously, may refer to something that's not necessarily a number and specifically, even if number, it may not a "monetary" number ... `amount` does not have such a connotation and the meaning itself may indeed be used to describe monetary quantities

if you're saying "subtask value", it could be understood as e.g. the result of the subtask or some other "value" (as in, a parameter) related to a subtask... in this respect `subtask amount` feels more unambiguous

shadeofblue: @mbenke after a tiny bit more thought, maybe just `calculate_subtask_payment` would be better altogether ... ?

PR #4841, opened 2019-10-29
Fix MaskUpdateService

mfranciszkiewicz:
- `_run` must not be decorated with `inlineCallbacks`
- do not call `rtm.has_pending_subtasks` as it is potentially extremely expensive; trust RTM's internal state instead (via `get_started_tasks`)
- fix invalid `task_id` references in `_run` (RTM)
~- catch `JSONDecodeError` in monitor as it frequently shows up during shutdown~ Reverted

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -1602,8 +1602,7 @@ class MaskUpdateService(LoopingCallService):
         self._interval = interval_seconds
         super().__init__(interval_seconds)
 
-    @inlineCallbacks
-    def _run(self) -> Deferred:
+    def _run(self):
         logger.debug('Updating masks')
         old_task_manager = self._old_task_manager
         requested_task_manager = self._requested_task_manager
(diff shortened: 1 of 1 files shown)


golem/client.py
```diff
@@ -1621,23 +1620,16 @@ def _run(self) -> Deferred:
             logger.info('Updating mask for task %r Mask size: %r',
                         task_id, task.header.mask.num_bits)
 
-        started_tasks = requested_task_manager.get_started_tasks()
-        # Using list() because tasks could be changed by another thread
-        for db_task in list(started_tasks):
```
maaktweluit, line 1626: are you sure the `list()` is no longer needed here?

maaktweluit (approved):
LGTM!

One small question, still approving

golem/client.py
mfranciszkiewicz: Yes, these are database query results converted to peewee models. Few lines above, with `old_task_manager.tasks` we reference a collection that may change any time.

Wiezzel (approved):

PR #4845, opened 2019-10-31
Populate output_path with data from Wasm task def

kubkon:
~~This commit populates the so-far unused `TaskBuilder` options field `output_path` with data from the gWasm task definition. This field is now a required field in the gWasm task's manifest. This commit partially addresses golemfactory/g-flite#28.~~

This commit overloads the `TaskBuilder`'s `get_output_path` for gWasm tasks. If the task def consists of `output_path` field, then its value is used as the output path of the task. Otherwise, `get_output_path` returns the value of `output_dir` that is supplied within the task def.

~~cc @lukasz-glen FYI, this is a potentially breaking change for the raw gWasm apps in our `wasm-store`.~~

cc @lukasz-glen I've made this field optional, so this shouldn't be a breaking change anymore.

cc @mdtanrikulu FYI, this commit together with golemfactory/g-flite#33 and golemfactory/gwasm-rust-api#10 should address golemfactory/g-flite#28.

### DIFF
--- a/apps/wasm/task.py
+++ b/apps/wasm/task.py
@@ -1,3 +1,4 @@
+import os
 from dataclasses import dataclass
 from copy import deepcopy
 from pathlib import Path, PurePath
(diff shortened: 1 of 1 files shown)


mdtanrikulu (approved):
Here is your flower 🌹

mbenke (approved):
LGTM

PR #4848, opened 2019-11-04
Update to Task API v0.23.0

Wiezzel:
Task API v0.23.0 provider AbortSubtask call which is now invoked when a subtask times out. Subtasks in verification are not timed out. Updated the Blender app to v0.6.0 which is compatible with Task API v0.23.0

### DIFF
--- a/golem/apps/default.py
+++ b/golem/apps/default.py
@@ -7,11 +7,11 @@ from pathvalidate import sanitize_filename
 from golem.apps import AppId, AppDefinition, save_app_to_json_file
 from golem.marketplace import RequestorBrassMarketStrategy
 
-BlenderAppDefinition_v0_5_0 = AppDefinition(
+BlenderAppDefinition_v0_6_0 = AppDefinition(
     name='golemfactory/blenderapp',
     author='Golem Factory GmbH',
     license='GPLv3',
-    version='0.5.0',
+    version='0.6.0',
     description=(
         'Rendering with Blender, the free and open source '
         '3D creation suite'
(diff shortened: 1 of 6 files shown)


golem/apps/default.py
```diff
@@ -7,26 +7,26 @@
 from golem.apps import AppId, AppDefinition, save_app_to_json_file
 from golem.marketplace import RequestorBrassMarketStrategy
 
-BlenderAppDefinition_v0_5_0 = AppDefinition(
+BlenderAppDefinition = AppDefinition(
```
maaktweluit, line 10: Why did you remove the version from the app definition? it was meant to be able to run multiple versions of the same app side by side for compatibility reasons

maaktweluit (changes requested):
LGTM!

One small comment, asked this q before and got this reply from @mfranciszkiewicz

golem/apps/default.py
mfranciszkiewicz: If you believe that including version in the name is an overkill, we may approach the problem differently:
- add version only to alternative versions of apps and treat the version-less one as "latest"
- inline app definitions in the `APPS` list

golem/task/requestedtaskmanager.py
```diff
@@ -729,17 +729,22 @@ def _time_out_subtask(
             RequestedSubtask.task == task_id,
             RequestedSubtask.subtask_id == subtask_id
         )
-        if subtask.status.is_active():
+        # Do *not* time out subtasks during verification
+        if subtask.status in (
```
mfranciszkiewicz: This could be a `not in` condition with a return to decrease code indentation

mfranciszkiewicz (approved):

mfranciszkiewicz:
Please change the target branch to `develop`

golem/apps/default.py
Wiezzel: Inlining doesn't seem an option because this particular definition is imported.
I think we can put the version back in the name if we *actually include two versions of blenderapp* in the built-in definitions. For now, it's not necessary.

Wiezzel:
@maaktweluit Do you accept?

golem/apps/default.py
maaktweluit: I like that, this way we keep the latest without version number and give legacy versions a longer variable name.

This will keep the imports cleaner.

maaktweluit (approved):

mfranciszkiewicz:
Merging since only the Task API computation flow is affected

PR #4851, opened 2019-11-04
Reporting exceeded CPU limit

kmazurek:
Part of GR-179

Implements informing the requestor about CPU budget overruns during task computation. This fact is reported in the `TaskFailure` message using the `reason` field (with value `BudgetExceeded`).


mbenke (approved):

shadeofblue (approved):
incredible_contribution

PR #4854, opened 2019-11-05
adjust the expected amount after we know the actual payment value

shadeofblue:
(which, in case of usage marketplace, is usually lower than the budget)

#closes gr260


mbenke (changes requested):
It might be better to adjust when submitting ReportComputedTask

shadeofblue:
> It might be better to adjust when submitting ReportComputedTask

indeed, added that adjustment to `SRR` handling because it's more explicit there and more in-line with the existing budget/payment checks

mbenke (approved):
LGTM

PR #4856, opened 2019-11-05
Add option to load app_definition by app_id

maaktweluit:


golem/apps/default.py
```diff
@@ -45,3 +45,10 @@ def save_built_in_app_definitions(path: Path) -> List[AppId]:
             app_ids.append(app_id)
 
     return app_ids
+
+
+def get_built_in_app_by_id(app_id: AppId) -> Optional[AppDefinition]:
```
Wiezzel: Better make `APPS` a dictionary so we can use simple `get()`.

scripts/task_api_tests/basic_integration.py
```diff
@@ -27,19 +30,21 @@
...
-    app_definition = load_app_from_json_file(Path(app_definition_path))
     env_prerequisites = app_definition.requestor_prereq
     app_manager = appmanager.AppManager()
     app_manager.register_app(app_definition)
     app_manager.set_enabled(app_definition.id, True)
 
     env_manager = envmanager.EnvironmentManager()
-    register_built_in_repositories()
+    docker_repo = repository_from_image_name(
+        app_definition.requestor_prereq['image'])
```
Wiezzel: That's fine for now but it's a hack and would require some action once we add other environments.  Please, put a comment here to remember about it.

golem/apps/default.py
maaktweluit: good one, updated!

scripts/task_api_tests/basic_integration.py
maaktweluit: Added

Wiezzel (approved):

mfranciszkiewicz (approved):

shadeofblue (approved):
lgtm

PR #4858, opened 2019-11-05
Reporting graceful shutdown via status publisher

kmazurek:
Adds calls to status publisher whenever Golem enters and exits graceful shutdown mode.


mdtanrikulu (approved):
LGTM

jiivan (approved):

PR #4862, opened 2019-11-06
Docker stats POC

Wiezzel:
Run e.g.
```
./scripts/docker_stats_poc.py busybox "sh -c 'while true; do echo dupa; sleep 1; done'"
```
Should print something like this:
```
Kernel: 0.000 s User: 0.200 s Total: 0.208 s Memory: 0.637 MiB
Kernel: 0.000 s User: 0.200 s Total: 0.209 s Memory: 0.641 MiB
Kernel: 0.000 s User: 0.200 s Total: 0.210 s Memory: 0.637 MiB
Kernel: 0.000 s User: 0.200 s Total: 0.210 s Memory: 0.637 MiB
Kernel: 0.000 s User: 0.200 s Total: 0.211 s Memory: 0.516 MiB
...
```
Hit <kbd>CTRL</kbd>+<kbd>C</kbd> to stop.

### DIFF
--- a/scripts/docker_stats_poc.py
+++ b/scripts/docker_stats_poc.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+
+import signal
+import sys
+from pathlib import Path
+
+from mock import patch
+from twisted.internet import defer, reactor
+
+from golem.docker.client import local_client
+from golem.envs import RuntimeStatus
+from golem.envs.docker import DockerPrerequisites, DockerRuntimePayload
+from golem.envs.docker.cpu import DockerCPUEnvironment, DockerCPUConfig
+
+
+@defer.inlineCallbacks
+def main(image, tag, command):
+    patch('golem.envs.docker.cpu.Whitelist.is_whitelisted', return_value=True)\
+        .start()
+    env = DockerCPUEnvironment(DockerCPUConfig(work_dirs=[Path('.')]))
+    yield env.prepare()
+    yield env.install_prerequisites(DockerPrerequisites(
+        image=image,
+        tag=tag
+    ))
+    runtime = env.runtime(DockerRuntimePayload(
+        image=image,
+        tag=tag,
+        env={},
+        command=command
+    ))
+
+    @defer.inlineCallbacks
+    def _stop(*_, **__):
+        print('Stopping...')
+        yield runtime.stop()
+
+    signal.signal(signal.SIGTERM, _stop)
+    signal.signal(signal.SIGINT, _stop)
+
+    yield runtime.prepare()
+    yield runtime.start()
+
+    client = local_client()
+    stream = client.stats(runtime._container_id, decode=True, stream=True)
+    for stats in stream:
+        # Container cannot be removed when the stream is being read and the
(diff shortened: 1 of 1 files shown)


scripts/docker_stats_poc.py
```diff
@@ -0,0 +1,82 @@
...
+
+from golem.docker.client import local_client
+from golem.envs import RuntimeStatus
+from golem.envs.docker import DockerPrerequisites, DockerRuntimePayload
+from golem.envs.docker.cpu import DockerCPUEnvironment, DockerCPUConfig
+
+
+@defer.inlineCallbacks
+def main(image, tag, command):
+    patch('golem.envs.docker.cpu.Whitelist.is_whitelisted', return_value=True)\
```
maaktweluit, line 18: why patch this over just whitelisting the image in this line?

scripts/docker_stats_poc.py
```diff
@@ -0,0 +1,82 @@
...
+            user = stats['cpu_stats']['cpu_usage']['usage_in_usermode']
+            total = stats['cpu_stats']['cpu_usage']['total_usage']
+            mem = stats['memory_stats']['usage']
+            print(f'Kernel: {(kernel / 1000000000):.3f} s '
+                  f'User: {(user / 1000000000):.3f} s '
+                  f'Total: {(total / 1000000000):.3f} s '
+                  f'Memory: {(mem / 1024 / 1024):.3f} MiB')
+        except KeyError:
+            # The last message in the stream usually raises KeyError
+            print('Wrong stats format')
```
maaktweluit, line 63: maybe add if status == running here to, so it can be an "error" while running and a info ( or no message ) when its not

scripts/docker_stats_poc.py
```diff
@@ -0,0 +1,82 @@
...
+
+    print('Cleaning up...')
+    yield runtime.wait_until_stopped()
+    yield runtime.clean_up()
+    yield env.clean_up()
+    reactor.stop()
+
+
+if __name__ == '__main__':
+    if len(sys.argv) != 3:
```
maaktweluit, line 73: would recommend to use `Click` for args parsing if this poc file is going to stay

maaktweluit (changes requested):
LGTM!

Some small comments.

Would move this file to `scripts/task_api` together with the basic integration tests if it will stay.

scripts/docker_stats_poc.py
Wiezzel: Because whitelisting requires a properly initialized database. :stuck_out_tongue:

scripts/docker_stats_poc.py
Wiezzel: Nice idea.
However, I'm not sure if the last message is sent after the container is stopped or just before. Need to check it.

scripts/docker_stats_poc.py
Wiezzel: It's not going to stay.

maaktweluit (approved):

mfranciszkiewicz:
Merged in #4901

PR #4864, opened 2019-11-06
Common RPC call for creating tasks

mfranciszkiewicz:
- use the existing `comp.task.create` RPC call for creating Task API tasks
- new Task API task dict (JSON) format: `{ "golem": { ... }, "app": { ... } }` for `golem_params` and `app_params` accordingly


golem/task/requestedtaskmanager.py
```diff
@@ -63,40 +63,9 @@ class CreateTaskParams:
     resources: List[Path]
     max_subtasks: int
     max_price_per_hour: int
-    min_memory: int
+    min_memory: int  # FIXME: remove with new Task API ver
```
maaktweluit: maybe add task-api version or issue # so if its forgotten we can track it back more easily

golem/task/rpc.py
```diff
@@ -483,15 +484,36 @@ def requested_task_manager(
...
-    def create_task(self, task_dict, force=False) \
-            -> typing.Tuple[typing.Optional[str],
-                            typing.Optional[typing.Union[str, typing.Dict]]]:
+    def create_task(
+            self,
+            task_dict: dict,
+            force: bool = False,
+    ) -> typing.Tuple[
+        typing.Optional[TaskId],
+        typing.Optional[typing.Union[str, typing.Dict]]
```
maaktweluit, line 494: as far as i can see this return value is never used, `None` on all paths.
Seems errors are set by raising instead of using this return var

if the RPC still requires it maybe we can set it to `None` here so its only in a single place

scripts/node_integration_tests/playbooks/golem/task_api/playbook.py
```diff
@@ -2,4 +2,4 @@
 
 
 class Playbook(NodeTestPlaybook):
-    RPC_TASK_CREATE = 'comp.task_api.create'
+    pass
```
maaktweluit: when the playbook is empty the test can be simplified, you could move the config one level up like `exr.py`

golem/task/rpc.py
maaktweluit: its in `@safe_run`, nvm.. 

would be nicer now to have a single try block in this function

maaktweluit (approved):
LGTM!

issue # is not needed if its picked up soon, but good practice to always add

golem/task/rpc.py
```diff
@@ -522,46 +544,33 @@ def create_task(self, task_dict, force=False) \
...
+            output_directory=Path(golem_params['output_directory']),
+            max_price_per_hour=int(golem_params['max_price_per_hour']),
+            max_subtasks=int(golem_params['max_subtasks']),
+            min_memory=int(golem_params['min_memory']),  # FIXME: new API ver
+            task_timeout=int(golem_params['task_timeout']),
+            subtask_timeout=int(golem_params['subtask_timeout']),
+            concent_enabled=bool(golem_params.get('concent_enabled', False)),
+            resources=list(map(Path, golem_params['resources'])),
+        )
+        app_params['resources'] = [r.name for r in create_task_params.resources]
```
Wiezzel: I know this line is just moved an this PR doesn't change it, but nevertheless:

Shouldn't `app_params` be completely opaque to Golem?

golem/task/rpc.py
mfranciszkiewicz: I've simplified the `legacy` and `task_api` methods but decorated the main calling method.   IMO it's advisable for the method to behave as it used to.

Wiezzel (approved):

scripts/node_integration_tests/playbooks/golem/task_api/playbook.py
mfranciszkiewicz: Maybe we should keep it? We need additional tests anyway.

maaktweluit: why would you like to keep it? new tests will make new playbook files or folders

we can migrate it back to have a playbook script if required, for now it seems cleaner to just keep the config

mfranciszkiewicz: It's just `task_api_blender` (docker cpu + blender app) doesn't seem to be generic enough to be called a Task API configuration. But it's the way the integration tests are structured at the moment, so I don't really care. Changing.

PR #4868, opened 2019-11-07
Dump docker logs

Wiezzel:
Docker container output (and runtime output in general) is needed for debugging purposes. Implemented an environment wrapper that will automatically dump these outputs to text files and print the filenames in Golem logs.

Slight changes in the `Environment` and `Runtime` interfaces were needed to make the solution generic. Also done some refactoring in the `auto_setup` wrapper to avoid code repetition.


golem/core/common.py
```diff
@@ -238,6 +238,12 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+def get_log_dir(data_dir: Optional[str] = None) -> str:
+    if data_dir is None:
+        data_dir = simpleenv.get_local_datadir("default")
+    return os.path.join(data_dir, 'logs')
```
maaktweluit: would be nice to use `Path` over `os.path`

maaktweluit (approved):
LGTM!

One minor comment, still approving!

mfranciszkiewicz (approved):

PR #4871, opened 2019-11-07
Only include task_api apps on testnet

maaktweluit:
So they can be safely tested on testnet first


Wiezzel (commented):
What about provider part? Will it go in another PR?

maaktweluit:
@Wiezzel good one, will look at this tomorrow, in this PR :) Was thinking if a task is not able to be requested the provider "switch" is not needed, but it is a good practice to include it on both sides.

golem/task/taskserver.py
```diff
@@ -139,8 +140,17 @@ def __init__(self,
...
 
         self.app_manager = app_mgr = app_manager.AppManager()
+        from golem.config.active import TASK_API_IMG_NAMES
+        logger.debug('Enabled app_names from config: %r', TASK_API_IMG_NAMES)
         for app_def in golem.apps.load_apps_from_dir(app_dir):
-            app_mgr.register_app(app_def)
+            logger.debug(
+                'App found in dir. app_id=%r, app_name=%r',
+                app_def.id, app_def.name)
+            if app_def.requestor_prereq['image'] in TASK_API_IMG_NAMES:
```
Wiezzel: I liked the app-based approach better than this. It relies on Docker-specific `'image'` key thus breaking the abstraction.

golem/task/taskserver.py
```diff
@@ -729,6 +739,28 @@ def add_task_header(self, task_header: dt_tasks.TaskHeader):
         if task_header.environment_prerequisites:
             image_name = task_header.environment_prerequisites['image']
             self._docker_image_discovered(image_name)
+            if not Whitelist.is_whitelisted(
+                    repository_from_image_name(image_name)
```
Wiezzel: `repository_from_image_name` is already being called inside `Whitelist.is_whitelisted`. No need to call it here. Maybe prepend the function name with underscore to avoid this kind of mistakes.

golem/task/taskserver.py
```diff
@@ -729,6 +739,28 @@ def add_task_header(self, task_header: dt_tasks.TaskHeader):
...
+                logger.info(
+                    "Task not added, image not whitelisted."
+                    " task_id=%r, image=%r",
+                    task_header.task_id,
+                    image_name
+                )
+                logger.debug("task_header=%r", task_header)
+                return False
+            from golem.config.active import TASK_API_IMG_NAMES
+            if image_name not in TASK_API_IMG_NAMES:
```
Wiezzel: This effectively blocks all apps other then blender. We don't want it.

Wiezzel (changes requested):
I don't approve this approach. I suggest to do what the task description says:
> don't register envs

This should effectively block all Task API-related activity both on requestor and provider side.

golem/task/taskserver.py
maaktweluit: only task_api related apps, so i think this is as expected

golem/task/taskserver.py
maaktweluit: with an app based approach the data is not available in the task_header, so if we only make it requestor side this is possible, for both sides we need to do it image based

golem/task/taskserver.py
maaktweluit: fixed the duplicate call and made the function private

golem/task/taskserver.py
Wiezzel: How are developers supposed to do anything with the Task API then?

Wiezzel: Summon @mfranciszkiewicz

maaktweluit:
@Wiezzel saw your main comment after fixing all the others 🤦‍♂ 

Moved the logic to use envs instead, indeed much neater.

golem/config/environments/mainnet.py
```diff
@@ -79,3 +79,5 @@ def __init__(self):
     'blender',
     'blender_nvgpu'
 )
+
+TASK_API_IMG_NAMES = []
```
Wiezzel:
```suggestion
TASK_API_ENVS = []
```

maaktweluit: Whoops, global find / replace fail, amended the previous commit

golem/task/taskserver.py
maaktweluit: in the above example they would have to add their image name to the list.

Now its all moved to envs the comment no longer applies and developers can use all apps on the enabled envs

Wiezzel (approved):

mfranciszkiewicz (approved):

PR #4874, opened 2019-11-12
Cgi/transcoding/dir structure

nieznanysprawiciel:


tests/apps/ffmpeg/task/test_ffmpegtranscoding.py
```diff
@@ -38,19 +38,22 @@ def setUp(self):
     def test_extract_and_split_video(self):
         for parts in [1, 2]:
             with self.subTest('Testing splitting', parts=parts):
+                task_id = str(uuid.uuid4())
```
bidzyyys, line 42: Can it be genereted in `setUp()` ?

tests/apps/ffmpeg/task/test_ffmpegtranscoding.py
```diff
@@ -38,19 +38,22 @@ def setUp(self):
...
-                        self.RESOURCE_STREAM, parts, self.dir_manager,
-                        str(uuid.uuid4()))
+                        self.RESOURCE_STREAM, parts,
+                        self.dir_manager.get_task_temporary_dir(task_id),
+                        task_id)
                 self.assertEqual(len(chunks), parts)
 
     def test_extract_and_split_invalid_video(self):
         with self.assertRaises(ffmpegException):
+            task_id = str(uuid.uuid4())
```
bidzyyys, line 52: Same as above

apps/transcoding/ffmpeg/utils.py
```diff
@@ -92,7 +90,7 @@ def extract_video_streams_and_split(self, # noqa pylint: disable=too-many-locals
             except ffmpegException as exception:
                 raise ffmpegExtractSplitError(str(exception)) from exception
 
-        split_result_file = os.path.join(host_dirs['output'],
+        split_result_file = os.path.join(directory_mapping.output,
```
bidzyyys: Can it throw an exception?

bidzyyys (approved):

apps/transcoding/ffmpeg/utils.py
```diff
@@ -92,7 +90,7 @@ def extract_video_streams_and_split(self, # noqa pylint: disable=too-many-locals
...
                                          Commands.EXTRACT_AND_SPLIT.value[1])
```
bidzyyys: Can it throw an exception?

tests/apps/ffmpeg/task/test_ffmpegtranscoding.py
nieznanysprawiciel: It's not good idea.
1) I prefer explicit tests, that show what they do, even if it makes test longer. And here it doesn't make it longer so much.
Setup is reserved for environment initialization.
2) task_id is used to generate directory for task and I don't want tasks directories to overlap. Golem doesn't behave this way and tests shouldn't either.

apps/transcoding/ffmpeg/utils.py
nieznanysprawiciel: Hmm. It seems to me it can't. directory_mapping output field will exist for sure. output is valid as long as task_dir was valid. And task_dir is generated by paths manager, so it should be valid.
Of course it is generated by outside function so extract_video_streams_and_split could be misused, but probably we don't want to check everything.

Do you see any other cases?

PR #4878, opened 2019-11-13
there's no strong rationale behind breaking connection with another n…

shadeofblue:
…ode just because we couldn't add a single task header that it helped propagate...

### DIFF
--- a/golem/network/p2p/peersession.py
+++ b/golem/network/p2p/peersession.py
@@ -365,10 +365,7 @@ class PeerSession(BasicSafeSession):
         logger.debug("Running handler for `Tasks`. msg=%r", msg)
         for t in msg.tasks:
             logger.debug("Task information received. task header: %r", t)
-            if not self.p2p_service.add_task_header(t):
-                self.disconnect(
-                    message.base.Disconnect.REASON.BadProtocol
-                )
+            self.p2p_service.add_task_header(t)
 
     def _react_to_remove_task(self, msg):
         if not self._verify_remove_task(msg):


etam:
I'm not entirely sure. Nodes should not send invalid task headers. If one does, then IHMO it's reasonable to disconnect from it and connect to others that are sending valid task headers.

etam:
\<rant\>
> there's no strong rationale [...]

AFAIK, there's no rationale for any part of the protocol written down. All we can do is to infer it from the code.
\</rant\>

jiivan (approved):

shadeofblue:
> I'm not entirely sure. Nodes should not send invalid task headers. If one does, then IHMO it's reasonable to disconnect from it and connect to others that are sending valid task headers.

* why is it reasonable to disconnect from it?

etam:
Because it's violating the protocol (if the protocol is defined as such).

shadeofblue:
> Because it's violating the protocol (if the protocol is defined as such).

is it breaking the p2p protocol? or the task protocol?

if it's the latter (which is the case here), a more reasonable action is just to drop the task header (which, again, is the case here) thus, disallowing that one piece of data from influencing the network but allowing the connections to continue unobstructed otherwise, instead of disconnecting the node itself, which causes unnecessary obstacles in the p2p network as a whole (a possible attacker could e.g. construct such a header that's valid when first distributed but which happens to time-out shortly after being sent and thus cause a lot of unnecessary disconnections)

etam (approved):

PR #4885, opened 2019-11-14
cast the wei budget to int

shadeofblue:


apps/wasm/task.py
```diff
@@ -583,7 +583,7 @@ def build_full_definition(
             logger.warning("Assigning task default budget: %d",
                            task_def.budget / denoms.ether)
         else:
-            task_def.budget = dictionary.get('budget') * denoms.ether
+            task_def.budget = int(dictionary.get('budget') * denoms.ether)
```
kubkon: Could we do some rounding here though? It kinda hurts my eyes when I see simple truncation...

tests/apps/wasm/test_task.py
```diff
@@ -108,6 +111,7 @@ def test_build_full_definition(self):
             WasmTaskTypeInfo(), TEST_TASK_DEFINITION_DICT,
         )
         self.assertEqual(task_def.subtasks_count, 2)
+        self.assertEqual(task_def.budget, int(0.5 * denoms.ether))
```
kubkon: Ditto!

apps/wasm/task.py
shadeofblue: sure... though, do we imagine specifying values to 18th decimal place in the task definition JSON? ...

mbenke (approved):
Seems to work with https://github.com/golemfactory/g-flite/pull/37

kubkon (approved):
LGTM!

PR #4636, opened 2019-08-22
Task API computation flow fixes + integration test

maaktweluit:
Original description (obsolete):
> I tried to use this test to build the requestedTaskManager but it turned out harder than i thought.
> 
> This branch / pr can be used to build the e2e integration tests, the error currently showing on mac and linux is:
> 
> ```
> 2019-08-22 14:17:56 ERROR    golem.task.rpc                      init_task() - exception=Exception('Could not connect to socket (172.17.0.2, 50005)',)
> ```
> 
> I build and tagged the current `blenderapp` image from the master branch to make this test work

Edit:
The integration test itself was flawed and couldn't possibly work.

Since the integration is a minor part of the commitment, this PR contains a large amount of fixes to actually perform the work.

PR #4886, opened 2019-11-14
B0.22 task-api: Also add app_ids when they are already saved to disk

maaktweluit:
Apps are currently only enabled the first run of golem. This patch enables them at all times.


Wiezzel (commented):
This is a feature, not a bug. We don't want to re-enable apps every time Golem is started. 
See https://github.com/golemfactory/golem/pull/4636#discussion_r335916329

mfranciszkiewicz:
Fixed in https://github.com/golemfactory/golem/pull/4907. Can we close this PR?

maaktweluit:
Yes we can :D

PR #4891, opened 2019-11-15
[P3] Transcoding: Validations for subtitle and data stream whitelists

cameel:
Changes:
1) `strip_unsupported_data_streams`/`strip_unsupported_subtitle_streams` task parameters added in P4 (#4639) are now being passed to `validate_transcoding_params()` so that the new validations can be bypassed if the user requests all unknown streams to be stripped.
2) The test for `UnsupportedStream` has been restored. This case now causes a validation failure instead of crashing the merge step.

### Notes
1) This change of course does not completely eliminate the possibility of an ffmpeg error at the merge step but such errors were possible even before P3/P4 and fixing it is outside of the scope of this pull request. This needs to be fixed separately by adding a new state just like the old comment from the test suggested.

### Dependencies
This pull request has no dependencies on the Golem side. It's directly on `CGI/transcoding/master`.

As for ffmpeg-tools - https://github.com/golemfactory/ffmpeg-tools/pull/20 needs to get merged and released for this code to work and pass CI tests.

### DIFF
--- a/apps/transcoding/task.py
+++ b/apps/transcoding/task.py
@@ -135,7 +135,9 @@ class TranscodingTask(CoreTask):  # pylint: disable=too-many-instance-attributes
 
             validation.validate_transcoding_params(
                 dst_params,
-                video_metadata
+                video_metadata,
+                self.task_definition.options.strip_unsupported_data_streams,
+                self.task_definition.options.strip_unsupported_subtitle_streams,
             )
 
         except validation.InvalidVideo as e:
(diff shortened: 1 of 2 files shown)


cameel:
I see that there was a big refactor merged into `CGI/transcoding/master`. I have rebased this pull request on top of it to make sure my branch can be merged without conflicts.

nieznanysprawiciel (approved):

nieznanysprawiciel:
Will be merged together with:
https://github.com/golemfactory/golem/pull/4925

PR #4894, opened 2019-11-15
Restore task-api task timeouts after restarting node

maaktweluit:
Schedule or trigger the timeouts of task-api tasks when restarting the golem node

### DIFF
--- a/tests/golem/task/test_requestedtaskmanager.py
+++ b/tests/golem/task/test_requestedtaskmanager.py
@@ -3,12 +3,13 @@
 
 import asyncio
 from pathlib import Path
+import datetime
 
 from freezegun import freeze_time
 from golem_task_api.client import RequestorAppClient
 from golem_task_api.enums import VerifyResult
 from golem_task_api.structs import Subtask, Infrastructure
-from mock import ANY, Mock
+from mock import ANY, call, MagicMock, Mock, patch
 import pytest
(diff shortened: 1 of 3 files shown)


Wiezzel (commented):
Looks good. I have two comments though:
1. What about subtask timeouts?
2. Some unit tests for this feature would be nice.

tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -73,6 +74,64 @@ def setup_method(self, tmpdir, monkeypatch):
...
+            mock_loop.time.return_value = datetime.datetime.now().timestamp()
+            # when
+            with patch(
+                    'golem.task.requestedtaskmanager.asyncio.get_event_loop',
+                    return_value=mock_loop
+            ):
+                self.rtm.restore_tasks()
+            # then
+            assert mock_loop.call_at.call_count == 2
+            mock_loop.call_at.calls == [
```
Wiezzel: I think it should be `assert_has_calls`. Currently, this line does nothing.

maaktweluit: thank you, that was the one i was looking for

Also this function was indented wrong after adding the other, so it never ran.

Both are fixed :)

Wiezzel (approved):

Issue #4904, opened 2019-11-18
Failed to update balances: TypeError("unsupported operand type(s) for -=: 'NoneType' and 'int'",)


ederenn:
## Description

**Golem Version**:     GOLEM Version: 0.21.0+dev424.gea67a5c

**Golem-Messages version** (leave empty if unsure): golem_messages Version: 3.14.0

**Electron version** (if used): 0.21.0

**OS** [e.g. Windows 10 Pro]: Linux 18.04

**Branch** (if launched from source): b0.22

**Mainnet/Testnet**: mainnet 

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

I'm recieving a lot of warnings in logs

```
2019-11-18 15:10:39 WARNING  golem.ethereum.transactionsystem    Failed to update balances: TypeError("unsupported operand type(s) for -=: 'NoneType' and 'int'",)
2019-11-18 15:10:52 WARNING  golem.ethereum.transactionsystem    Failed to update balances: TypeError("unsupported operand type(s) for -=: 'NoneType' and 'int'",)
```

after not connecting with GETH

## Logs and any additional context

``````2019-11-18 14:25:08 WARNING  golem.ethereum.web3.providers       GETH: request failure, retrying: 502 Server Error: Bad Gateway for url: https://0.geth.golem.network:2137/
2019-11-18 14:25:08 WARNING  golem.ethereum.web3.providers       GETH: request failure, retrying: 502 Server Error: Bad Gateway for url: https://0.geth.golem.network:2137/
2019-11-18 14:25:08 WARNING  golem.ethereum.web3.providers       GETH: request failure, retrying: 502 Server Error: Bad Gateway for url: https://0.geth.golem.network:2137/
2019-11-18 14:25:08 WARNING  golem.ethereum.web3.providers       GETH: reconnecting to another provider
2019-11-18 14:25:08 INFO     golem.ethereum.web3.providers       GETH: connecting to remote RPC interface at https://0.geth.golem.network:55555
2019-11-18 14:25:08 WARNING  golem.ethereum.web3.providers       GETH: request failure, retrying: 502 Server Error: Bad Gateway for url: https://0.geth.golem.network:55555/
[...]

PR #4906, opened 2019-11-18
Safely update balances

jiivan:
fixes: #4904 

1. Decrease verbosity of known geth connection issue to INFO
1. Try to update other balances even if one fails

### DIFF
--- a/golem/ethereum/transactionsystem.py
+++ b/golem/ethereum/transactionsystem.py
@@ -1,3 +1,5 @@
+# pylint: disable=too-many-lines
+import contextlib
 import datetime
 import functools
 import json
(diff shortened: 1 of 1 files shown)


golem/ethereum/transactionsystem.py
```diff
@@ -956,15 +958,33 @@ def _refresh_balances(self) -> None:
 
         # Sometimes web3 may throw but it's fine here, we'll just update the
         # balances next time
-        try:
+        @contextlib.contextmanager
+        def safe_update(currency):
+            print("safe update", currency)
```
maaktweluit: why use `print` here? would be cleaner to use logger.debug i think

maaktweluit (changes requested):
LGTM!

One small q, will approve on fix or reply

golem/ethereum/transactionsystem.py
jiivan: My bad. Those are my debug debris

jiivan: Fixed

maaktweluit (approved):

PR #4910, opened 2019-11-19
Fix: use system time instead of UTC in ResourceManager.share

mfranciszkiewicz:


golem/resource/resourcemanager.py
```diff
@@ -48,7 +48,8 @@ def share(
             # Missing / invalid timeouts are validated by the client.
             # Don't pre-check it here and set the timeout to max int.
             timeout = client_options.timeout or sys.maxsize
-            now = get_timestamp_utc()
+            # Use system time like simple-transfer does
+            now = datetime.now().timestamp()
```
maaktweluit, line 52: shouldnt this be `default_now` like in the database?

maaktweluit (approved):
LGTM!

One small comment and a q:
Shouldnt this be fixed in simple-transfer instead?

golem/resource/resourcemanager.py
mfranciszkiewicz: `default_now` returns a UTC `datetime` object

mfranciszkiewicz:
@maaktweluit We'd have to release a new version of simple-transfer and afaict that wasn't planned

PR #4912, opened 2019-11-20
Use full precision in payment log messages

mbenke:
Using a full .18f precision in payment log messages should make it easier to track pending and recveived payments


shadeofblue (approved):
wolfofwallstreet

maaktweluit (approved):

PR #4914, opened 2019-11-20
Fix: disable only new envs in EnvironmentManager.register_env

mfranciszkiewicz:


maaktweluit (approved):

Wiezzel (approved):
Looks fine.. but shouldn't the check in line 75 disallow for re-registering?

mfranciszkiewicz:
@Wiezzel `self._state` is persistent, consecutive Golem runs will overwrite the stored value

PR #4918, opened 2019-11-21
fix Client init when `in_shutdown` config flag is set

shadeofblue:


etam:
What is the fix for? What is broken and how this resolves it?

mdtanrikulu (approved):

Issue #4890, opened 2019-11-14
Newly created Gwasm tasks are not appearing on top of a task list


ederenn:
## Description

**Golem Version**: 0.21.0+dev413.g4be8f69

**Golem-Messages version** (leave empty if unsure): 3.14.0

**Electron version** (if used): 0.21.0

**OS** [e.g. Windows 10 Pro]: Linux 18.04

**Branch** (if launched from source): b0.22

**Mainnet/Testnet**: mainnet

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

When Gwasm task is created it appears in some random place at the task list in GUI, not at the top of the task list. Easy visible when there are numerous tasks on the task list.

**Screenshots**:

![Screenshot from 2019-11-14 17-25-07](https://user-images.githubusercontent.com/32061891/68878433-fd0eea00-0707-11ea-8e48-953397d0b46a.png)

![Screenshot from 2019-10-23 16-46-14](https://user-images.githubusercontent.com/32061891/68878560-3ba4a480-0708-11ea-9cb3-7ba8ad096a98.png)

## Steps To Reproduce
_Short description of steps to reproduce the behavior:_
e.g.
1. Launch Golem at b0.22 brach
2. Launch electron
3. On task list there should be already some tasks created previously
4. Run gflite task with any settings
5. Check position of a new task on a task list  

## Expected behavior
New task should appear at the top of the task list

## Proposed Solution?
_(Optional: What could be a solution for that issue)_

kmazurek:
Closed by https://github.com/golemfactory/golem/pull/4920

PR #4920, opened 2019-11-21
Sorting tasks returned by comp.tasks RPC endpoint

kmazurek:
Fixes: #4890

Changes the order of tasks returned by comp.tasks RPC endpoint.
Tasks are now sorted by their `time_started` field.


golem/client.py
```diff
@@ -935,7 +935,8 @@ def get_tasks(
         if return_created_tasks_only:
             filter_fn = self._filter_task_created_status
 
-        return list(filter(filter_fn, tasks))
+        filtered_tasks = list(filter(filter_fn, tasks))
+        return sorted(filtered_tasks, key=lambda task: task['time_started'])
```
maaktweluit, line 943: `time_started` is not put on the new task-api style task dicts, can you add it on #847 ? should be task.start_time

maaktweluit (changes requested):

maaktweluit:
Please also note start_time is empty for newly created tasks, think these should count as the "highest" times

golem/client.py
kmazurek: Good catch, thanks! I'm not sure about the issue you wanted to link here though, since 847 is from more than 2 years back and not relevant to these changes.

maaktweluit: whoops, it was 874

https://github.com/golemfactory/golem/pull/4920/files#diff-9f351727974c5655127c6a4dc0671995R874

kmazurek: @maaktweluit added, can you take a second look?

maaktweluit (approved):

maaktweluit:
Looks great! Thanks for the fix :)

PR #4911, opened 2019-11-20
[P5a] Transcoding: Move audio transcoding to the replace step

cameel:
A simple change to make Golem pass audio parameters to the "replace" step. So far it was passing them to the "transcode" step where they had no effect on the final video.

### Dependencies
https://github.com/golemfactory/ffmpeg-tools/pull/22 needs to get merged and released for this code to work and pass CI tests.

This pull request is based on #4892 (`transcoding-task-o3-validate-target-fps-even-if-not-specified-in-the-task` branch) and should be merged after it. Please remember to change the base branch back to `CGI/transcoding/master` before you merge.

PR #4925, opened 2019-11-21
[P5b] Transcoding: Default audio codec detection and validation

cameel:
### Changes
1) `extract-and-split` command can now fetch muxer info (including the name of the default audio codec) for the target container if it gets the name of that container (it's an optional parameter).
    - As requested, there's also a separate command for getting that information (`query-muxer-info`). Having both ways to get it gives us the flexibility to call it separately if needed while not having to do multiple calls if we're already fetching the metadata and doing extract+split.
2) Muxer info is now being passed to `validate_transcoding_params()` to let it validate the target audio codec even if it's not given explicitly.
3) `container` parameter has been renamed to `intermediate_container` (the one used temporarily before transcoding) or `target_container` (the one we get after transcoding) in most commands. This is to prevent confusing them with each other now that `extract-and-split` needs both.

### Dependencies
https://github.com/golemfactory/ffmpeg-tools/pull/23 needs to get merged and released for this code to work and pass CI tests.

This pull request is based on #4911 (`transcoding-task-p5a-move-audio-transcoding-to-the-replace-step` branch) and should be merged after it. Please remember to change the base branch back to `CGI/transcoding/master` before you merge.


nieznanysprawiciel (approved):

nieznanysprawiciel:
PR will be merged soon. I want to do additional tests with transcoding webservice.

nieznanysprawiciel:
@cameel 
There's bug in ffmpeg tools. If I understand, what is happening, there're cyclic imports. formats.py imports validation and validation tries to import formats.
That's why python can't find symbol FrameRate.

To reproduce this bug you can just start Golem, which will fail with message:
```ERROR    [golem.node                         ] Stopping because of 'Cannot start the client' error: module 'ffmpeg_tools.formats' has no attribute 'FrameRate'```

Check results of buildbot/test_e2e_source_smoke, they report the same error.

nieznanysprawiciel (changes requested):
Waiting for bug fix

cameel:
OK. I'll look into that in a moment.

cameel:
There's indeed an import cycle in ffmpeg-tools. `codecs` -> `formats` -> `validations` -> `codecs`. I cannot reproduce this particular failure by running unit tests of either Golem or ffmpeg-tools though. I can only trigger it on purpose if I specifically use something from `codecs` at module level. I did not dig deeper into how `test_e2e_source_smoke` imports stuff but this must be related to the order in which it does it.

Anyway, since the underlying issue seems to be a cycle, I refactored ffmpeg-tools to remove it. The reason for circular imports is mostly the fact that other modules need exceptions from `validations`. I moved all exceptions to a separate module which should fix it.

I'm running tests on my machine right now and I'll push the commits in a moment it the tests pass. Let me know if there are still some problems after this fix - buildbot needs a published version of ffmpeg-tools so I won't be able to see on my own if it passes or not. If it works, I'll rebase other branches on my changes to save you the trouble of resolving conflicts (there's going to be a lot of tiny ones in ffmpeg-tools).

cameel:
Branch updated. Cycle should be gone now.

nieznanysprawiciel (approved):

nieznanysprawiciel (changes requested):
Something is still wrong with smoke tests and I can't run Golem on my local computer either.

cameel:
Apparently there's still a cycle. `formats` imports `codecs` which then uses `formats.FrameRate` at module level.

I can extract `FrameRate` into a separate module. It isn't a format anyway. What would you call the new module? Would `frame_rate` be fine of do you prefer something generic like `types` or `utils`?

nieznanysprawiciel:
You can name it frame_rate.py.

cameel:
Done. Hopefully this time this fixes things.

nieznanysprawiciel:
Did you try to run golem ?

cameel:
I did not but I ran Golem tests for `tests/apps/ffmpeg`, including `--runslow`. I did not run the whole set though - I'm going to do it as soon as I finish resolving conflicts I caused in ffmpeg-tools and can switch back to the right branch. I wanted to push it as soon as I could to hopefully let you accept this today.

cameel:
I ran the full suite of tests and they all pass. I have also manually reviewed all imports in `ffmpeg_tools` for potential cycles. After previous changes there are only two such instances (`codecs` -> `formats` -> `codecs` and `meta` -> `commands` -> `meta`) and in all of them the imported module is used only inside a function so it should not cause this kind of problem.

So I think it should all be fine now.

BTW, I have rebased all branches on my changes from P5b and solved all conflicts. There were quite a few of them. You'll probably get lots of notifications about newly pushed commits but nothing  has really changed inside. The only changes I managed to do today were the ones for P5b.

nieznanysprawiciel (approved):

nieznanysprawiciel (approved):

magdasta (approved):

nieznanysprawiciel:
I merge it and ignore mac tests problems, because it is known error and hopefully will be fixed.

PR #4926, opened 2019-11-22
Add subtask_budget to task-api task_headers

maaktweluit:
New changes not applied to task-api

Calculate subtask_budget and set it in task_header

### DIFF
--- a/golem/task/taskserver.py
+++ b/golem/task/taskserver.py
@@ -76,10 +76,11 @@ from golem.task import helpers as task_helpers
 from golem.task import timer
 from golem.task.acl import get_acl, setup_acl, AclRule, _DenyAcl as DenyAcl
 from golem.task.exceptions import ComputationInProgress
-from golem.task.server.whitelist import DockerWhitelistRPC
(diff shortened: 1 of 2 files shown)


golem/task/taskserver.py
```diff
@@ -693,6 +694,10 @@ def _get_and_sign_headers(self):
         started_tasks = self.requested_task_manager.get_started_tasks()
         signed_headers = []
         for db_task in started_tasks:
+            subtask_budget = calculate_subtask_payment(
```
mfranciszkiewicz, line 700: Could you make `subtask_budget` a part of `RequestedTask` and calculate the value during task creation? We could support calculating budgets for different marketplaces in RTM later on. A FIXME comment would be in place

mfranciszkiewicz: As we've talked offline, this will be implemented in a follow up PR

mfranciszkiewicz:
@maaktweluit Please add a FIXME comment and update the test so we can merge this PR.

mfranciszkiewicz (approved):

PR #4929, opened 2019-11-22
DO NOT MERGE Common crypto package - RSA keys, DH params, X509 certs

mfranciszkiewicz:
**DO NOT MERGE** This PR will be rebased onto develop when b0.22 is merged.

A prerequisite branch for introducing gRPC SSL support.

New module for:
- RSA key generation and (de)serialization
- DH params generation and (de)serialization
- X509 cert generation and (de)serialization

### DIFF
--- a/golem/crypto.py
+++ b/golem/crypto.py
@@ -1,0 +1,141 @@
+import ipaddress
+from datetime import datetime, timedelta
+from pathlib import Path
+from typing import Iterable, Optional, Union
+
+from cryptography import x509
+from cryptography.x509.oid import NameOID
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import dh, rsa
+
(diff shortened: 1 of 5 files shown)


maaktweluit (changes requested):
LGTM! Nice cleanup :)

Maybe add some tests for the new file `golem/crypto.py`?

Wiezzel (approved):
Nice cleanup.
Agreed with @maaktweluit that some unit tests would be nice.

mfranciszkiewicz:
These are basically wrappers for the provided crypto functions. New unit tests would only check whether those functions were called with the provided parameters and IMO that's not worth the effort. It makes little sense to (re)test `cryptography`.

The `CertificateManager` test uses these functions under the hood and provides a better test case. The E2E test validates whether the generated key, DH params and the certificate are generated properly and conform to general requirements (e.g. running with `Crossbar`).

maaktweluit:
Sounds good to me, maybe add them to the coverage exclusion list with a small comment like above

maaktweluit (approved):

PR #4932, opened 2019-11-24
Task API gRPC SSL

mfranciszkiewicz:
Requires #4929

### DIFF
--- a/golem/apps/ssl.py
+++ b/golem/apps/ssl.py
@@ -1,0 +1,92 @@
+import logging
+import shutil
+import ssl
+from ipaddress import IPv4Address
+from pathlib import Path
+from typing import Optional
+
+import golem_task_api as api
+
+from golem.crypto import (
+    generate_rsa_private_key,
+    generate_x509_certificate,
+    save_rsa_private_key,
+    save_x509_certificate,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class SSLContextConfig:
+    key_and_cert_directory: Optional[Path] = None
+
+
+def create_golem_ssl_context(
+        shared_directory: Path
+) -> Optional[ssl.SSLContext]:
+    if not SSLContextConfig.key_and_cert_directory:
+        raise RuntimeError("Golem Task API SSL context was not set up")
+    return api.ssl.create_client_ssl_context(
+        SSLContextConfig.key_and_cert_directory,
+        shared_directory)
+
+
+def create_golem_ssl_context_files(directory: Path):
+    SSLContextConfig.key_and_cert_directory = directory
+
+    create_ssl_context_files(
+        directory,
+        key_file_name=api.ssl.CLIENT_KEY_FILE_NAME,
+        cert_file_name=api.ssl.CLIENT_CERT_FILE_NAME,
+        label="Golem Task API",
+    )
+
+
+def create_app_ssl_context_files(
+        shared_directory: Path
+) -> None:
+    if not SSLContextConfig.key_and_cert_directory:
+        raise RuntimeError("Golem Task API SSL context was not set up")
+    create_ssl_context_files(
+        shared_directory,
+        key_file_name=api.ssl.SERVER_KEY_FILE_NAME,
+        cert_file_name=api.ssl.SERVER_CERT_FILE_NAME,
+        label="App",
+    )
+    shutil.copy(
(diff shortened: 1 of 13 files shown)


golem/apps/ssl.py
```diff
@@ -0,0 +1,92 @@
...
+        cert_file_name=api.ssl.SERVER_CERT_FILE_NAME,
+        label="App",
+    )
+    shutil.copy(
+        SSLContextConfig.key_and_cert_directory / api.ssl.CLIENT_CERT_FILE_NAME,
+        shared_directory
+    )
+
+
+def create_ssl_context_files(
```
maaktweluit, line 62: this function is not used outside this file, could be private

golem/apps/ssl.py
```diff
@@ -0,0 +1,92 @@
...
+        cert = generate_x509_certificate(
+            key,
+            names=['localhost'],
+            ip_addresses=[IPv4Address('127.0.0.1')])
+
+        directory.mkdir(parents=True, exist_ok=True)
+        save_rsa_private_key(key_file, key)
+        save_x509_certificate(cert_file, cert)
+    except Exception:  # pylint: disable=broad-except
+        logger.exception('Unable to create Task API SSL context filex')
```
maaktweluit, line 88: should we log some extra info here like path and of stacktrace?

tests/golem/apps/test_ssl.py
```diff
@@ -0,0 +1,78 @@
...
+    create_ssl_context_files,
+    create_golem_ssl_context_files,
+    generate_rsa_private_key,
+    SSLContextConfig,
+)
+from golem.testutils import TempDirFixture
+
+
+@mock.patch.object(SSLContextConfig, 'key_and_cert_directory', None)
+class TestSslContext(TempDirFixture):
```
maaktweluit, line 18: i think its better to use the pytest fixture tmpdir ( and tmp_path after we updated pytest )

example: https://github.com/golemfactory/golem/blob/develop/tests/golem/task/test_requestedtaskmanager.py#L51

tests/golem/task/test_requestedtaskmanager.py
```diff
@@ -74,8 +76,8 @@ def setup_method(self, tmpdir, monkeypatch):
             lambda *_: TaskState())
 
     @pytest.mark.asyncio
-    @pytest.mark.freeze_time("1000")
-    async def test_restore_tasks_timedout(self, freezer, mock_client):
+    @pytest.mark.freeze_time("2020")
```
maaktweluit, line 79: just curious, why did the dates changed? i liked testing the middle ages ;)

maaktweluit (approved):
LGTM!

Some small comments, still approving

golem/apps/ssl.py
Wiezzel: `logger.exception()` automatically logs stacktrace

tests/golem/task/test_requestedtaskmanager.py
mfranciszkiewicz: Certificate validation failed due to invalid dates

tests/golem/apps/test_ssl.py
mfranciszkiewicz: Simply, these tests are unittest-based, not pytest-based.

golem/apps/ssl.py
```diff
@@ -0,0 +1,92 @@
...
+    generate_x509_certificate,
+    save_rsa_private_key,
+    save_x509_certificate,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class SSLContextConfig:
+    key_and_cert_directory: Optional[Path] = None
```
Wiezzel, line 21: I'm not really convinced of this global variable approach. Could the path be passed as argument for `TaskComputer`, `AppBenchmarkManager` and `RequestedTaskManager`? This argument could be optional.

Wiezzel (commented):
Sorry for late review. The PR looks OK, but I have some doubts about the singleton design.

golem/apps/ssl.py
mfranciszkiewicz: I also had doubts about this approach and personally do not like it. But when I started refactoring, the solution seemed messy and forced. We really need a better way to store runtime configuration.

PR #4940, opened 2019-11-26
Task API task control RPC calls

mfranciszkiewicz:
- implement missing task / subtask control RPC calls
- mark selected RPC calls as unsupported by Task API


golem/task/requestedtaskmanager.py
```diff
@@ -564,9 +591,34 @@ def get_node_id_for_subtask(
 
     async def restart_task(self, task_id: TaskId) -> None:
         task = RequestedTask.get(RequestedTask.task_id == task_id)
+        subtask_ids = self.get_requested_task_subtask_ids(task_id)
+
+        app_client = await self._get_app_client(task.app_id)
+        await app_client.discard_subtasks(task_id, subtask_ids)
```
maaktweluit: i think this should be a separate restart command. AFAIR there are 3 ways of restarting:
- Restart task ( no discard )
- Restart task, discard all work
- Restart subtask / frame ( create new task only marking the single subtask failed )

Not sure if we want to keep it this way, up to you :)

maaktweluit (approved):
LGTM!

Small comment on the restart change, still approving!

golem/client.py
```diff
@@ -719,26 +720,41 @@ def check_test_status(self) -> Optional[Dict[str, Any]]:
...
     @rpc_utils.expose('comp.task.subtask.restart')
+    @inlineCallbacks
     def restart_subtask(self, subtask_id):
-        logger.debug("restarting subtask %s", subtask_id)
-        task_manager = self.task_server.task_manager
+        logger.debug("Restarting subtask %s", subtask_id)
+
+        rtm = self.task_server.requested_task_manager
+        if rtm.subtask_exists(subtask_id):
+            yield deferred_from_future(rtm.abort_subtask(subtask_id))
```
maaktweluit: this only aborts the subtask, i think it should call `restart_subtask(`

maaktweluit (dismissed):

golem/client.py
mfranciszkiewicz: We achieve restarts by cancelling the subtask. To restart a subtask means to create a new subtask with the same input data and this is managed by the application.

We have no "restart" Task API calls.

golem/task/requestedtaskmanager.py
mfranciszkiewicz: This can be either discard or abort. @Wiezzel ?

golem/client.py
maaktweluit: ahh, of course :) nvm, i was confused

golem/task/requestedtaskmanager.py
mfranciszkiewicz: Changing to `abort`

golem/client.py
Wiezzel: `abort_subtask` is not the right method to call here because it calls `abort_subtask` RPC and sets subtask status to `cancelled`. We should call `discard_subtasks` RPC and set subtask status to `restarted`.

golem/task/requestedtaskmanager.py
Wiezzel: Discard is the right RPC for restarting but there are other things which bother me here:
* Non-Task API tasks are restarted in such a way that a new task with a new ID is created. Here, we re-use the old task ID. Wouldn't this lack of consistency cause any troubles?
* What about task timeouts? There is no handling of timeouts here. And without it the restarted task will have the same deadline as the old one.

Wiezzel (dismissed):
I haven't reviewed it completely, but I already do have some important remarks.

golem/client.py
```diff
@@ -719,34 +720,50 @@ def check_test_status(self) -> Optional[Dict[str, Any]]:
...
+    @inlineCallbacks
     def restart_subtask(self, subtask_id):
-        logger.debug("restarting subtask %s", subtask_id)
-        task_manager = self.task_server.task_manager
+        logger.debug("Restarting subtask %s", subtask_id)
+
+        rtm = self.task_server.requested_task_manager
+        if rtm.subtask_exists(subtask_id):
+            yield deferred_from_future(rtm.restart_subtask(subtask_id))
+            return
```
Wiezzel, line 742: What about `self.funds_locker.add_subtask(task_id)`? Don't we need this for Task API tasks as well?

golem/task/requestedtaskmanager.py
```diff
@@ -81,6 +83,33 @@ class SubtaskDefinition:
     deadline: int
 
 
+class CallScheduler:
```
Wiezzel: Not a big deal, but this class looks like it belongs to some common/utils package. There's nothing specific to RequestedTaskManager here.

golem/task/requestedtaskmanager.py
```diff
@@ -563,9 +622,39 @@ def get_node_id_for_subtask(
             return None
 
     async def restart_task(self, task_id: TaskId) -> None:
+        subtask_ids = self.get_requested_task_subtask_ids(task_id)
+        await self.restart_subtasks(task_id, subtask_ids)
```
Wiezzel: I believe we don't need this call. Why restart subtasks in the *old* task when a new task is created by `duplicate_task`? Please note that `duplicate_task` *does not copy state* from the old task but merely creates a new one with the same parameters.

golem/task/requestedtaskmanager.py
```diff
@@ -563,9 +622,39 @@ def get_node_id_for_subtask(
...
+
         task = RequestedTask.get(RequestedTask.task_id == task_id)
-        task.status = TaskStatus.waiting
+        task.status = TaskStatus.restarted
         task.save()
+        self._schedule_task_timeout(task, task.task_timeout)
```
Wiezzel: Same as above. We don't need to reschedule timeout for the old task.

golem/task/rpc.py
```diff
@@ -759,6 +809,54 @@ def restart_subtasks(
...
+            ignore_gas_price,
+            disable_concent)
+
+    @defer.inlineCallbacks
+    def restart_task_api_task_subtasks(
+            self,
+            task_id: str,
+            subtask_ids: typing.List[str],
+            ignore_gas_price: bool = False,
+    ):
```
Wiezzel, line 832: This method has different semantics than `restart_legacy_task_subtasks`. I'm not sure if this is intended.
This one:
  * Assumes task timeout has *not* passed.
  * Simply calls restart on each of the provided subtask IDs.

The legacy one:
  * Assumes task timeout has passed.
  * Creates a new task, copying results of all finished subtasks except for the ones provided in the list.

Wiezzel (dismissed):

golem/task/requestedtaskmanager.py
mfranciszkiewicz: If we skip this call the application will never be notified that the state has changed

golem/task/requestedtaskmanager.py
mfranciszkiewicz: True, we actually need to cancel it or wait for the timeout and handler to ignore it. This code hasn't been updated, thanks for spotting that out

golem/task/rpc.py
mfranciszkiewicz: Yes, this is intended. It would be easy to implement on RTM side but it's currently not possible to duplicate subtask state in the app (or force the input data for non-finished subtasks)

golem/task/requestedtaskmanager.py
Wiezzel: And what is wrong about that? Why does the application need to be notified that the state of the old task has changed? The old task is already over, I believe we don't care about its state.

golem/task/rpc.py
Wiezzel: OK, please put a comment then. Someone might get confused reading this code an take it to be a bug.

golem/task/requestedtaskmanager.py
mfranciszkiewicz: > Why does the application need to be notified that the state of the old task has changed?

Let me ask the other way around - why don't we want to notify the application? Not caring about the state after restarts is a heavy assumption on how the application is implemented.

Wiezzel: It is already assumed in the Task API design that:
  1. The application does not initiate any activity by itself but only reacts to RPC calls from Golem.
  2. Tasks are isolated. When computing one task the application should be oblivious of other tasks.

After restarting task X there will be no more calls from Golem regarding task X. And calls regarding other tasks should not need any information about task X to be answered. So the state of task X does not need to be updated.

mfranciszkiewicz: Calling `abort_task` instead of `restart_subtasks`

golem/client.py
```diff
@@ -719,34 +720,52 @@ def check_test_status(self) -> Optional[Dict[str, Any]]:
...
     def restart_subtask(self, subtask_id):
-        logger.debug("restarting subtask %s", subtask_id)
-        task_manager = self.task_server.task_manager
+        logger.debug("Restarting subtask %s", subtask_id)
+
+        rtm = self.task_server.requested_task_manager
+        subtask = rtm.get_requested_subtask(subtask_id)
+        if subtask:
+            yield deferred_from_future(rtm.restart_subtask(subtask_id))
+            self.funds_locker.add_subtask(subtask.task.task_id)
```
Wiezzel, line 740: Shouldn't it lock funds *first* so that subtask does not get restarted if the user doesn't have enough funds?

golem/task/requestedtaskmanager.py
```diff
@@ -562,10 +594,41 @@ def get_node_id_for_subtask(
         except DoesNotExist:
             return None
 
-    async def restart_task(self, task_id: TaskId) -> None:
+    async def restart_task(self, task_id: TaskId) -> Optional[TaskId]:
         task = RequestedTask.get(RequestedTask.task_id == task_id)
-        task.status = TaskStatus.waiting
-        task.save()
+        if not task.status.is_active():
```
Wiezzel: I believe we should allow restarting finished tasks. This is the most common case after all. Just put the abort call in a conditional:
```python
if task.status.is_active():
    await self.abort_task(task_id)
```

golem/task/requestedtaskmanager.py
```diff
@@ -562,10 +594,41 @@ def get_node_id_for_subtask(
...
+            return None
+
+        new_task_id = await self.duplicate_task(task_id, task.output_directory)
+        await self.init_task(new_task_id)
```
Wiezzel, line 602: We should probably call `self.start_task` as well after `init_task`.

golem/client.py
mfranciszkiewicz: Yes, we should. We should actually check for the funds first, restart the subtask and lock the funds in a transactional way since restarting may fail for some reason.

Wiezzel (approved):
:+1:

Issue #4945, opened 2019-11-27
Cannot set total subtasks to 4. Changing to 4


ederenn:
## Description

**Golem Version**: GOLEM Version: 0.21.0+dev469.ge527148

**Golem-Messages version** (leave empty if unsure): 3.14.0

**Electron version** (if used): 0.21.0

**OS** [e.g. Windows 10 Pro]: Linux 18.04; Windows 10 Pro

**Branch** (if launched from source): b0.22 based

**Mainnet/Testnet**: mainnet

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**: 
After requesting a task with two frames and four subtasks appeared error:
` Cannot set total subtasks to 4. Changing to 4`

**Screenshots**:

_If applicable, add screenshots to help explain your problem._

## Steps To Reproduce
_Short description of steps to reproduce the behavior:_
e.g.
1. Launch Golem on b0.22 branch
2. Start a task Small_shark.blend: resolution from file, 2 frames, 4 subtasks, 10 min for a task 5 min for a subtask, bid 0.1 
3. Check logs

## Logs and any additional context

```
2019-11-26 17:05:12 INFO     golem.client                        Loading presets for Blender
2019-11-26 17:05:12 INFO     golem.task.rpc                      Running test task "{'name': 'scene-Helicopter-27-cycl', 'resources': ['/home/ederenn/Desktop/Host_Share_Folder/test files/pliki testowe/scene-Helicopter-27-cycles.blend'], 'compute_on': 'cpu', 'type': 'Blender', 'subtasks_count': 1}" ...
2019-11-26 17:05:22 INFO     golem.task.rpc                      Test task succes "{'name': 'scene-Helicopter-27-cycl', 'resources': ['/home/ederenn/Desktop/Host_Share_Folder/test files/pliki testowe/scene-Helicopter-27-cycles.blend'], 'compute_on': 'cpu', 'type': 'blender', 'subtasks_count': 1, 'concent_enabled': False}"
[...]

PR #4950, opened 2019-11-28
Omit printing erroneous subtask count warning

kmazurek:
Fixes: https://github.com/golemfactory/golem/issues/4945

### DIFF
--- a/apps/rendering/task/framerenderingtask.py
+++ b/apps/rendering/task/framerenderingtask.py
@@ -82,7 +82,7 @@ def _calculate_subtasks_count(
         # whole number of subtasks.
         new_subtasks_count = _round_int(subtasks_count, num_frames)
 
-    if new_subtasks_count > max_subtasks_count:
+    if new_subtasks_count >= max_subtasks_count:
         return max_subtasks_count
 
     logger.warning("Cannot set total subtasks to %s. Changing to %s",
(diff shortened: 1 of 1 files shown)


apps/rendering/task/framerenderingtask.py
```diff
@@ -82,7 +82,7 @@ def _calculate_subtasks_count(
         # whole number of subtasks.
         new_subtasks_count = _round_int(subtasks_count, num_frames)
 
-    if new_subtasks_count > max_subtasks_count:
+    if new_subtasks_count >= max_subtasks_count:
```
shadeofblue: @zakaprov hmm... generally, it would seem that the warning should be generated any time `new_subtasks_count` is different from `subtasks_count` ... and just that new value returned should not be larger than the maximum ... 

so... 

```
    new_subtasks_count = min(new_subtasks_count, max_subtasks_count)
    if new_subtasks_count != subtasks_count:
        logger.warning("Cannot set total subtasks to %s. Changing to %s",
                       subtasks_count, new_subtasks_count)
    return new_subtasks_count
```

maybe?

kmazurek: Agreed, that makes sense.

etam (approved):

PR #4952, opened 2019-12-02
[P6b] Transcoding: Audio sample rate validation

cameel:
### Changes
Changes are pretty similar to P5b (#4925).
1) `extract-and-split` command can now fetch encoder info for specified codec. The feature is optional and the data is fetched if either the codec name is specified or if we're fetching muxer info anyway (which means that the default audio codec might be used as the target).
    - As requested, there's also a separate command for getting that information (`query-codec-info`).
2) Encoder info is now being passed to `validate_transcoding_params()`.
3) There are two small fixes for P5b: the command name in `QUERY_MUXER_INFO` was wrong and so was one of the type annotations.

### Dependencies
https://github.com/golemfactory/ffmpeg-tools/pull/27 needs to get merged and released for this code to work and pass CI tests.

This pull request is based on #4925 (`transcoding-task-p5b-audio-codec-detection-and-validation` branch) and should be merged after it. Please remember to change the base branch back to `CGI/transcoding/master` before you merge.

### DIFF
--- a/apps/transcoding/ffmpeg/ffmpeg_docker_api.py
+++ b/apps/transcoding/ffmpeg/ffmpeg_docker_api.py
@@ -6,6 +6,7 @@ import glob
 from pathlib import Path
 from typing import List, Optional
 from threading import Lock
+from ffmpeg_tools.codecs import AudioCodec
 from ffmpeg_tools.formats import Container
 
 from apps.transcoding.common import ffmpegException, ffmpegExtractSplitError, \
@@ -54,7 +55,9 @@ class FfmpegDockerAPI:
     def extract_video_streams_and_split(self,
                                         input_file_on_host: str,
                                         parts: int,
-                                        target_container: Optional[Container]):
+                                        target_container: Optional[Container],
+                                        target_audio_codec: Optional[AudioCodec
+                                                                    ],):
 
         input_file_basename = os.path.basename(input_file_on_host)
 
@@ -79,12 +82,18 @@ class FfmpegDockerAPI:
         else:
             target_container_str = target_container.value
 
+        if target_audio_codec is None:
+            target_audio_codec_str = None
+        else:
+            target_audio_codec_str = target_audio_codec.value
+
         extra_data = {
             'entrypoint': FFMPEG_ENTRYPOINT,
             'command': Commands.EXTRACT_AND_SPLIT.value[0],
             'input_file': input_file_in_container,
             'parts': parts,
             'target_container': target_container_str,
+            'target_audio_codec': target_audio_codec_str,
         }
 
         logger.debug(
--- a/apps/transcoding/ffmpeg/ffmpeg_docker_api.py
+++ b/apps/transcoding/ffmpeg/ffmpeg_docker_api.py
@@ -40,7 +40,7 @@ class Commands(enum.Enum):
     TRANSCODE = ('transcode', '')
     MERGE_AND_REPLACE = ('merge-and-replace', '')
     COMPUTE_METRICS = ('compute-metrics', '')
-    QUERY_MUXER_INFO = ('compute-metrics', 'query-muxer-info-results.json')
+    QUERY_MUXER_INFO = ('query-muxer-info', 'query-muxer-info-results.json')
     QUERY_ENCODER_INFO = (
         'query-encoder-info',
         'query-encoder-info-results.json')
--- a/apps/transcoding/ffmpeg/ffmpeg_docker_api.py
+++ b/apps/transcoding/ffmpeg/ffmpeg_docker_api.py
@@ -41,6 +41,9 @@ class Commands(enum.Enum):
     MERGE_AND_REPLACE = ('merge-and-replace', '')
     COMPUTE_METRICS = ('compute-metrics', '')
     QUERY_MUXER_INFO = ('compute-metrics', 'query-muxer-info-results.json')
+    QUERY_ENCODER_INFO = (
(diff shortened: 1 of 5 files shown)


apps/transcoding/ffmpeg/ffmpeg_docker_api.py
```diff
@@ -51,7 +55,9 @@ def __init__(self, directory_mapping: DockerDirMapping):
     def extract_video_streams_and_split(self,
                                         input_file_on_host: str,
                                         parts: int,
-                                        target_container: Container):
+                                        target_container: Optional[Container],
+                                        target_audio_codec: Optional[AudioCodec
+                                                                    ],):
```
nieznanysprawiciel, line 60: I thing there's to many parameters here, that aren't directly necessary for this command and are used only to query ffmpeg parameters.
We could avoid this by returning from split command muxer info and audio codec info for all supported by us codecs and formats. Probably this won't be performance problem, but if it would be, we could cache these parameters in Transcoding integration after first use.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
```diff
@@ -101,6 +151,13 @@ def do_extract_and_split(input_file,
     if target_container is not None:
         results["muxer_info"] = commands.query_muxer_info(target_container)
 
+    encoder_info = _fetch_encoder_info_if_requested(
+        target_audio_codec,
+        results["muxer_info"] if "muxer_info" in results else None,
+    )
+    if encoder_info is not None:
+        results["encoder_info"] = encoder_info
+
```
nieznanysprawiciel, line 160: I expected, that extract_and_split will use the same code for quering encoder info. Why do we need 2 functions ?

apps/transcoding/ffmpeg/ffmpeg_docker_api.py
cameel: I think this is just a different design choice. I did it this way because I assumed that minimizing the number of ffmpeg calls would be best. Given what you said about passing `muxer_info` and `encoder_info` to validations I can see why you want it done differently and it does make sense. It just looks like another task rather than something to be fixed in a review.

Maybe let's first get the pull requests merged. The change is not hard but it it's also not small because it will touch a few different places both in Golem and in ffmpeg-tools - especially in validations and the places that test or use them. I would prefer not to do it as a part of review because it's not a bug that prevents things from working so changing it here is not critical and would just cause conflicts and add more work for me. If it's just this small task I can do it after it all gets in. If there's more (from other reviews), we'll have to decide how to go about it.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
cameel: You mean that you expected `do_extract_and_split` command and `query_encoder_info` command to have the same internal code for obtaining encoder info internally? They do, kinda - they both eventually call `commands.query_encoder_info()` function from ffmpeg-tools. It's just that `do_extract_and_split` has more additional logic needed to decide what to do in various corner cases (what if the audio codec has no encoder? or is not supported? or the codec name is not specified and `muxer_info` is not available? what if `muxer_info` is available?). `query_muxer_info` command is much simpler because it just expects the encoder name.

Actually I could make `query_muxer_info` command do the same thing as `do_extract_and_split` but I did not think it was worth the effort because in my opinion this command is completely superfluous right now and it's not certain what you'll really need in the future. Originally you told me to add it and just call it from split because the code will later be refactored so that it's all inside Docker container and additional command calls will stop being a performance issue. I added the command but I don't think we'll really want to use it separately even if it's all in a single container. What we'll probably need is a combined command for getting `muxer_info` and `encoder_info`. But that depends on future requirements that I don't know yet. I implemented it in the simplest possible way because I expected you may want to get rid of it for now after you see it. I can change it if you already have a good idea of what you'll need.

Maybe it would be best done together with the design change you requested above: https://github.com/golemfactory/golem/pull/4952/files#r354919877

apps/transcoding/ffmpeg/ffmpeg_docker_api.py
nieznanysprawiciel: Since we don't have time for further work, we will accept this as is.

apps/transcoding/ffmpeg/resources/ffmpeg-scripts/ffmpeg_task.py
nieznanysprawiciel: Since we don't have time for further work, we will accept this as is.

nieznanysprawiciel:
@cameel Check this PR https://github.com/golemfactory/golem/pull/4984
One test is not passing.
I checked CGI/transcoding/master to be sure and it works on master

nieznanysprawiciel:
Ok new build passed :/
I don't understand this, but these tests worked locally, so maybe it wasn't bug.

cameel:
Weird. But these tests do not seem fully deterministic. Though it's pretty much expected in such a big application with so many moving parts.

BTW, sorry for the exception imports I missed and you had to fix in 73617afe6dd4b355e8a059fb54adc59e9db724e5. I just checked how this could happen and it seems that it still the spotty coverage of the Golem's part of the code running in the container. These `raise`/`except` blocks don't even seem to be reached by the interpreter on my machine when I run the tests so I missed them.

PR #4953, opened 2019-12-02
Transcoding: Fix duplicate container in extra_data

cameel:
The dict contains two fields: `format` and `container`. `format` was there from the beginning but was never actually passed down to the command running in the Docker container. I have later added `container` in many places and looks like I overlooked the fact that this place already had a similar field.

Well, `container` is now used basically everywhere so I'm removing `format`.

### Dependencies
This change does not really depend on branches below and could be easily rebased on master.
https://github.com/golemfactory/ffmpeg-tools/pull/28 contains a similar-looking change in `ffmpeg_tools` but they affect different dicts and don't really depend on each other.

This pull request is based on #4952 (`transcoding-task-p6b-audio-sample-rate-validation` branch) and should be merged after it. Please remember to change the base branch back to `CGI/transcoding/master` before you merge.

### DIFF
--- a/apps/transcoding/ffmpeg/task.py
+++ b/apps/transcoding/ffmpeg/task.py
@@ -46,18 +46,21 @@ class ffmpegTask(TranscodingTask):
         resolution = video_params.resolution
         resolution = [resolution[0], resolution[1]] if resolution else None
         vc = video_params.codec.value if video_params.codec else None
-        container = transcoding_options.output_container
+        if transcoding_options.output_container is not None:
+            target_container_str = transcoding_options.output_container.value
+        else:
+            target_container_str = None
+
         extra_data = {
             'track': chunk,
             'targs': {
-                'container': container.value if container is not None else None,
+                'container': target_container_str,
                 'video': {
                     'codec': vc,
                     'bitrate': video_params.bitrate
                 },
                 'resolution': resolution,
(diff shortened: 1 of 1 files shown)


apps/transcoding/ffmpeg/task.py
```diff
@@ -46,18 +46,16 @@ def _get_extra_data(self, subtask_num: int):
         resolution = video_params.resolution
         resolution = [resolution[0], resolution[1]] if resolution else None
         vc = video_params.codec.value if video_params.codec else None
-        container = transcoding_options.output_container
         extra_data = {
             'track': chunk,
             'targs': {
-                'container': container.value if container is not None else None,
```
nieznanysprawiciel, line 53: Are we sure here, that container won't be None?

cameel: I'm not sure. It depends on the code that creates the task. If it allows the user not to specify the container and sets it to `None` in that case then it's possible.

In tests it never happens because the `'format'` line below has been here for a long time and it wasn't failing. We always set the target container there. If we didn't validations would fail. On top of that, I have actually added the check again in #4954 to make handling of container and codecs consistent in all places.

So it's a non-issue but since it's not certain that #4954 will be accepted and because each pull request should be self-contained I backported this check so that it's performed here too.

I just updated the commit.

PR #4954, opened 2019-12-02
[I6] Transcoding: Automatic detection of target video codec and resolution

cameel:
### Changes
1) Until now specifying the video codec and/or resolution was mandatory even if user did not intend to change them. Now the corresponding task parameters can be set to `None` and they'll be filled in just before validations.
    - Note: this is done only in Golem. Validations (and ffmpeg-tools in general) expect these parameters to already have values when they're passed in.
2) There are a few corner cases here:
    - **No video stream in the input file**. Such file won't pass validations but the autofill happens before then. In that case the fields are just left empty and we expect validations to deal with them when they run.
    - **Multiple video streams**: We take the value from the first stream and log a warning. This is not very likely but possible. If user prefers different values, he should specify them explicitly.
    - **Unsupported codec**: Since this happens before validations, we can get a target video codec that's not in `VideoCodec`. Until now the task class required it to be an enum so I enhanced it to support both enums and strings. I made it work this way for containers and audio codecs too for consistency.
        - `if`s were getting pretty inconvenient so I added a simple helper to `golem.core.common`: `value_if_enum()` either returns what it gets (if not an enum instance) or calls `.value` on what it gets (if enum instance).

### Dependencies
This change is based on #4953 but does not really depend on branches below and could be easily rebased on master. It uses the new helpers added to `ffmpeg_tools.meta` in https://github.com/golemfactory/ffmpeg-tools/pull/29 though so that pull request must be merged for this one to work.

**NOTE**: Base branch of this pull request is not set to `CGI/transcoding/master`. Please remember to change it before merge.

### DIFF
--- a/apps/transcoding/task.py
+++ b/apps/transcoding/task.py
@@ -12,6 +12,7 @@ import ffmpeg_tools.validation as validation
 import ffmpeg_tools.exceptions as exceptions
 from ffmpeg_tools.codecs import VideoCodec, AudioCodec
 from ffmpeg_tools.formats import Container
+from ffmpeg_tools.meta import get_codecs, get_resolutions
 
 import apps.transcoding.common
 from apps.transcoding.ffmpeg.ffmpeg_docker_api import FfmpegDockerAPI
@@ -91,6 +92,32 @@ class TranscodingTask(CoreTask):  # pylint: disable=too-many-instance-attributes
         super(TranscodingTask, self).__setstate__(state)
         self.lock = Lock()
 
+    def _autofill_video_codec_from_metadata(self, metadata: dict) -> None:
+        target_video_codec = self.task_definition.options.video_params.codec
+        if target_video_codec is None:
+            src_codecs = get_codecs(metadata, 'video')
+            if len(src_codecs) > 0:
+                self.task_definition.options.video_params.codec = src_codecs[0]
+
+                if len(src_codecs) > 1:
+                    logger.warn(
+                        "Found multiple video streams while trying to determine"
+                        " source video codec. Ignoring all but the first one.")
+
+    def _autofill_resolution_from_metadata(self, metadata: dict) -> None:
+        target_resolution = self.task_definition.options.video_params.resolution
+        if target_resolution is None:
+            src_resolutions = get_resolutions(metadata)
+            if len(src_resolutions) > 0:
+                self.task_definition.options.video_params.resolution = \
+                    src_resolutions[0]
+
+                if len(src_resolutions) > 1:
+                    logger.warn(
+                        "Found multiple video streams while trying to determine"
+                        " source video resolution. Ignoring all but the first"
+                        " one.")
+
     def initialize(self, dir_manager: DirManager):
         super(TranscodingTask, self).initialize(dir_manager)
 
@@ -132,6 +159,9 @@ class TranscodingTask(CoreTask):  # pylint: disable=too-many-instance-attributes
         self.task_resources = streams
         self.chunks = streams
         self.task_definition.subtasks_count = len(chunks)
+
+        self._autofill_video_codec_from_metadata(video_metadata)
+        self._autofill_resolution_from_metadata(video_metadata)
 
         try:
             validation.validate_video(video_metadata)
--- a/apps/transcoding/task.py
+++ b/apps/transcoding/task.py
@@ -392,13 +392,13 @@ class TranscodingTaskBuilder(CoreTaskBuilder):
                 not output_container.is_supported_audio_codec(audio_codec):
             raise AudioCodecNotSupportedByContainer(
                 'Container {} does not support {}'.format(
-                    output_container.value, audio_codec.value))
+                    output_container, audio_codec))
 
         if video_codec and \
                 not output_container.is_supported_video_codec(video_codec):
             raise VideoCodecNotSupportedByContainer(
                 'Container {} does not support {}'.format(
-                    output_container.value, video_codec.value))
+                    output_container, video_codec))
 
     @classmethod
     def build_minimal_definition(cls, task_type: CoreTaskTypeInfo,
--- a/apps/transcoding/task.py
+++ b/apps/transcoding/task.py
@@ -99,7 +99,7 @@ class TranscodingTask(CoreTask):  # pylint: disable=too-many-instance-attributes
                 raise exceptions.InvalidVideo(
                     "Files with more than 1 video stream are not supported.")
 
-            src_codec = get_video_codec(metadata)
+            src_codec = VideoCodec(get_video_codec(metadata))
             self.task_definition.options.video_params.codec = src_codec
 
     def _autofill_resolution_from_metadata(self, metadata: dict) -> None:
--- a/apps/transcoding/task.py
+++ b/apps/transcoding/task.py
@@ -12,7 +12,7 @@ import ffmpeg_tools.validation as validation
 import ffmpeg_tools.exceptions as exceptions
 from ffmpeg_tools.codecs import VideoCodec, AudioCodec
 from ffmpeg_tools.formats import Container
-from ffmpeg_tools.meta import get_codecs, get_resolutions
+from ffmpeg_tools.meta import count_streams, get_video_codec, get_resolution
 
 import apps.transcoding.common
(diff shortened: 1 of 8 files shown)


nieznanysprawiciel (changes requested):
I expected, that everything will be done on the lowest level possible - probably somewhere in ffmpeg near commands generation. Validation would just assume, that if we don't specify parameters, they don't change. Command generation probably even don't need to specify these parameters if they remain the same.
But doing this on transcoding integration level introduces many problems and requires many additional checks, that probably could be avoided.

Am I right or are there any problems, which I can't see?

cameel:
Doing it my way actually requires less logic. The checks are limited to the very top level of the call hierarchy - the point in Golem where you fill in these values. After that, all the way down, you can assume that these parameters are mandatory. You can see that I didn't even need to change anything in ffmpeg-tools for this to work properly (there's no I6 pull request in ffmpeg-tools). After the initial validation that required parameters are present, you can just assume everywhere that video codec and resolution are known. Otherwise you'd have to handle both variants in every place you needed to access those values.

Second problem: you would also have to make sure you have the same logic for determining the value in every place you use it. That's perfectly doable but adds some new opportunities for things to go out of sync which is not the case with the current implementation.

Another downside of doing it at a lower level is that it breaks the separation of concerns a bit. Currently you just needs access to the target parameters. After doing what you propose, you'd have to have access to input file metadata everywhere you wanted to access video codec or resolution. For example you'd have to either pass input file metadata to provider who runs the `transcode` command or have the provider analyze the file himself which carries a slight risk that the results will be inconsistent. In theory ffmpeg should give you the same values but we know that there are many weird corner cases and I am not sure that values for a segment will always match the input it was taken from. Recall for example the one weird file that we ignored in tests that after extract+split magically gains a (broken) audio stream and fails transcoding.

I can see some good reasons for doing it your way though - doing it in Golem does not provide any benefit to the transcoding service or any other code that uses ffmpeg-tools. That's a perfectly valid reason to want it done differently. I admit that when writing it I assumed that Golem will be the main user of the library and it wouldn't matter in practice.

Changing this requires basically implementing it again though. And this functionality improves usability but is not really critical in my opinion so would suggest to either just accept the current design or drop this feature entirely. Another implementation of this would really be a separate task.

Sorry it turned out this way, I really thought this was what you wanted and I was wrong.

apps/transcoding/task.py
```diff
@@ -86,6 +87,32 @@ def __setstate__(self, state):
...
+            src_resolutions = get_resolutions(metadata)
+            if len(src_resolutions) > 0:
+                self.task_definition.options.video_params.resolution = \
+                    src_resolutions[0]
+
+                if len(src_resolutions) > 1:
+                    logger.warn(
+                        "Found multiple video streams while trying to determine"
+                        " source video resolution. Ignoring all but the first"
+                        " one.")
```
nieznanysprawiciel: Please remove this warning.

apps/transcoding/task.py
```diff
@@ -86,6 +87,32 @@ def __setstate__(self, state):
...
+        target_video_codec = self.task_definition.options.video_params.codec
+        if target_video_codec is None:
+            src_codecs = get_codecs(metadata, 'video')
+            if len(src_codecs) > 0:
+                self.task_definition.options.video_params.codec = src_codecs[0]
+
+                if len(src_codecs) > 1:
+                    logger.warn(
+                        "Found multiple video streams while trying to determine"
+                        " source video codec. Ignoring all but the first one.")
```
nieznanysprawiciel: Please remove this warning.

apps/transcoding/task.py
```diff
@@ -86,6 +87,32 @@ def __setstate__(self, state):
         super(TranscodingTask, self).__setstate__(state)
         self.lock = Lock()
 
+    def _autofill_video_codec_from_metadata(self, metadata: dict) -> None:
+        target_video_codec = self.task_definition.options.video_params.codec
+        if target_video_codec is None:
+            src_codecs = get_codecs(metadata, 'video')
```
nieznanysprawiciel: Use function for getting single codec that I requested in other review

apps/transcoding/task.py
```diff
@@ -86,6 +87,32 @@ def __setstate__(self, state):
...
+
+                if len(src_codecs) > 1:
+                    logger.warn(
+                        "Found multiple video streams while trying to determine"
+                        " source video codec. Ignoring all but the first one.")
+
+    def _autofill_resolution_from_metadata(self, metadata: dict) -> None:
+        target_resolution = self.task_definition.options.video_params.resolution
+        if target_resolution is None:
+            src_resolutions = get_resolutions(metadata)
```
nieznanysprawiciel: Use function for getting single codec that I requested in other review

golem/core/common.py
```diff
@@ -166,6 +167,21 @@ def short_node_id(node_id):
...
+    Returns the thing it gets unless the value is an enum member - in that case
+    returns its `.value`.
+    """
+
+    if not isinstance(x, Enum):
+        return x
+
+    return x.value
+
+
```
nieznanysprawiciel: I don't want to support both enums and strings. If we get invalid string, Enum will just throw exception and thats ok.
Please revert all changes that are related to this.

nieznanysprawiciel (changes requested):
We will accept this version without design changes that I wanted previously. But please check my ne comments.

apps/transcoding/task.py
cameel: Done.

apps/transcoding/task.py
cameel: Done.

apps/transcoding/task.py
cameel: Done.

Note that these functions have sanity checks that will fail if you get a video with more than one stream (as mentioned in https://github.com/golemfactory/ffmpeg-tools/pull/29#discussion_r354960606). I'm not sure what you want to happen in that situation (as I said in https://github.com/golemfactory/ffmpeg-tools/pull/29#issuecomment-564666048, ffmpeg-tools currently can handle files with multiple audio/video streams just fine). For now I added a check that raises a validation error if the file has multiple video streams.

apps/transcoding/task.py
cameel: Done. See my remark on a similar comment above (https://github.com/golemfactory/golem/pull/4954#discussion_r356780676).

golem/core/common.py
cameel: I originally made it support both because in that function I'm dealing with a codec name that has not been validated yet. If I put that name in the video codec field in the task and let it run, it fails before it reaches validations. I decided that it would be better for the string to reach validations and fail there than to duplicate the validation logic.

But now I reverted my solution and changed it according to your request. The video codec is now validated in the autofill function. To be honest, it turned out to be easier than I initially expected (I just run `VideoCodec` and let it fail if the codec is not in the enum) so maybe that's for the better.

apps/transcoding/task.py
nieznanysprawiciel: Why do you think it should fail? I think it could get to validation phase. If validation rejects it - ok. If not, it will be transcoded.

cameel: Because I think that by default it's better to fail with a validation error than to do something other than the user intended. And this situation is ambiguous.

In this case the user might not even be aware that his file contains multiple video streams. If I were the user I'd prefer to be asked if anything about my parameters was ambiguous instead of having the target codec/resolution silently chosen for me based on a random stream. Adding special code to ask the user in this very rare corner case is probably not worth it though (I suspect files with multiple video streams are not very common) so the next best thing in my opinion is to at least raise validation error and force the user to select the codec/resolution explicitly.

That's my default. Now I know that you don't want it to work that way so I removed the check.

cameel:
@nieznanysprawiciel `test_dont_retry_failed_subtask_more_than_1_time` is behaving erratically for me. It sometimes passes and sometimes fails. Is it maybe the test that was failing for you before? I didn't dig deeper into it but there must be something non-deterministic about it. Maybe a race condition between `_fail_next_subtask_and_verify()` and `start_task()`?

nieznanysprawiciel:
Good to know. We will investigate this test.

Issue #4957, opened 2019-12-02
Invocation surge detected golem.rpc.session 'evt.net.connection', {'listening': True, 'port_statuses':


ederenn:
## Description

**Golem Version**: GOLEM Version: 0.21.0+dev470.g7817ff3

**Golem-Messages version** (leave empty if unsure): 3.14.1

**Electron version** (if used): 0.21.0

**OS** [e.g. Windows 10 Pro]: Windows 10 Pro

**Branch** (if launched from source): b0.22

**Mainnet/Testnet**: mainnet

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

Appears in stacks in logs
```
WARNING  [golem.decorators                   ] Invocation surge detected. func=golem.rpc.session.Publisher.publish, hottest_args=("(<golem.rpc.session.Publisher object at 0x00000252B2297710>, 'evt.net.connection', {'listening': True, 'port_statuses': {40102: 'open', 40103: 'open', 3282: 'open'}, 'connected': True, 'msg': 'Port(s) 40102: open, 40103: open, 3282: open. Connected'}){}", 11)
```

## Logs and any additional context

after problems connecting to geth
```
WARNING  [golem.ethereum.transactionsystem   ] Failed to update ETH balance: ValueError({'code': -32000, 'message': 'missing trie node 74051d0118c348500323a1b502e466399e0c6449b593d0cb1b44b89f876d6fc1 (path )'},)
WARNING  [golem.ethereum.transactionsystem   ] Failed to update GNT balance: ValueError({'code': -32000, 'message': 'missing trie node 74051d0118c348500323a1b502e466399e0c6449b593d0cb1b44b89f876d6fc1 (path )'},)
WARNING  [golem_sci.implementation           ] Exception while processing eth subscription: ValueError({'code': -32000, 'message': 'missing trie node 74051d0118c348500323a1b502e466399e0c6449b593d0cb1b44b89f876d6fc1 (path )'},)
[...]

PR #4958, opened 2019-12-02
Send evt.net.connection only when status changes

jiivan:
fixes: #4957

### DIFF
--- a/tests/golem/test_client.py
+++ b/tests/golem/test_client.py
@@ -641,7 +641,9 @@ class TestNetworkConnectionPublisherService(testwithreactor.TestWithReactor):
         )
 
     @patch('golem.client.logger')
-    def test_run(self, logger):
+    @patch('golem.client.NetworkConnectionPublisherService.poll')
+    def test_run(self, poll_mock, logger):
+        poll_mock.return_value = {'random_key': random.random()}
         self.service._run()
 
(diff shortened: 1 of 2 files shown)


etam (approved):
ok

tests/golem/test_client.py
```diff
@@ -641,7 +641,9 @@ def setUp(self):
         )
 
     @patch('golem.client.logger')
-    def test_run(self, logger):
+    @patch('golem.client.NetworkConnectionPublisherService.poll')
+    def test_run(self, poll_mock, logger):
+        poll_mock.return_value = {'random_key': random.random()}
```
etam, line 646: This might very rarely fail, because it might produce the same value twice. You can assign an array to `side_effect` and it will return next value for each call.

etam (changes requested):

tests/golem/test_client.py
jiivan: It won't fail, because service is initialized in `.setUp()` which uses original `.poll()` implementation, that returns totally different kind of dict. Also this value is returned only once.

https://github.com/golemfactory/golem/blob/3b92e82da512a7d3af9c8e31fbdf50638630dc1a/golem/client.py#L1329-L1342

etam (approved):

PR #4961, opened 2019-12-03
0.22 concent mainnet

shadeofblue:


maaktweluit (approved):
LGTM!

Did not check the terms ;)

golem/CONCENT_TERMS.html
```diff
@@ -1,7 +1,371 @@
-<h5><u>Disclaimer</u></h5>
+<h4><u>Terms of Use of Concent Service</u></h4>
+<p><em>Last Update: June 2019.</em></p>
+<p>
+    These Terms of Use of Concent Service (hereinafter the “Terms of Use”)
```
etam: I'd make sure this is a simple ascii text file, not utf8. So all `“` should be changed to `"`.

etam: and `’` to `'`

shadeofblue: yup, detected by the test suite and fixed already :)

PR #4969, opened 2019-12-04
task-api task timeout bugfix

maaktweluit:
- Schedule task timeout from the moment it started, not created
- ~Convert milliseconds to seconds before passing to scheduler~
- Use seconds instead of milliseconds as input parameter, updated tests and comments
- restore_tasks() marks tasks as errorCreating if they are stuck in creating

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -277,8 +277,7 @@ class RequestedTaskManager:
         task.status = TaskStatus.waiting
         task.start_time = default_now()
         task.save()
-        task_timeout = timedelta(milliseconds=task.task_timeout).total_seconds()
-        self._schedule_task_timeout(task, task_timeout)
+        self._schedule_task_timeout(task, task.task_timeout)
         self._notice_task_updated(task, op=TaskOp.STARTED)
         logger.info("Task %s started", task_id)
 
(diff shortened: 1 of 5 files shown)


golem/task/requestedtaskmanager.py
```diff
@@ -276,6 +277,8 @@ def start_task(self, task_id: TaskId) -> None:
         task.status = TaskStatus.waiting
         task.start_time = default_now()
         task.save()
+        task_timeout = timedelta(milliseconds=task.task_timeout).total_seconds()
```
mfranciszkiewicz: Could we store `task_timeout` and `subtask_timeout` in seconds instead?

Wiezzel: To avoid confusion I suggest appending `_ms` or `_s` to every variable representing time duration.

maaktweluit: both would work for me, i think storing a float for seconds is the best of both worlds, will update

maaktweluit: we rather not do a migration from Int to Float now on a release hotfix, will keep seconds as int

mfranciszkiewicz (approved):

maaktweluit:
Bump, hope this will trigger buildbot again

Wiezzel (approved):

PR #4971, opened 2019-12-05
refactor the web3 ProviderProxy retry logic

shadeofblue:
to ease the load off the geth nodes

+ additional logging to ethereum.web3.providers

rationale:

the allegedly "simple" logic of ProviderProxy causes erroneous queries (e.g. ones that are bound to time-out on all geth nodes) to spam a node indefinitely since - if other queries succeed in the meantime, the retry counter is reset and ProviderProxy doesn't even know how many times such query has been run ... 

this pull request makes the retries of a single query separate from general "health" of the geth node Golem is connected to and the retry logic clearer and more explicit... 

moreover, it doesn't make reaching the last node on the list critical on its own, as the node list is explicitly cycled

### DIFF
--- a/golem/ethereum/web3/providers.py
+++ b/golem/ethereum/web3/providers.py
@@ -63,7 +63,7 @@ class ProviderProxy(HTTPProvider):
                 self._register_retry()
 
                 retry = retries < SINGLE_QUERY_RETRY_LIMIT \
-                    or self._retries < RETRY_COUNT_LIMIT
+                    and self._retries < RETRY_COUNT_LIMIT
 
                 logger.debug(
                     "GETH: request failure%s"
--- a/golem/ethereum/web3/providers.py
+++ b/golem/ethereum/web3/providers.py
@@ -72,6 +72,7 @@ class ProviderProxy(HTTPProvider):
 
                 if not retry:
                     nodes_tried += 1
+                    retries = 0
                     self._handle_remote_rpc_provider_failure(
                         method,
                         nodes_tried >= self._single_query_node_limit
--- a/golem/ethereum/web3/providers.py
+++ b/golem/ethereum/web3/providers.py
@@ -12,7 +12,7 @@ logger = logging.getLogger(__name__)
 RETRY_COUNT_LIMIT = 10
 RETRY_COUNT_INTERVAL = 1800  # seconds
 SINGLE_QUERY_RETRY_LIMIT = 3
-SINGLE_QUERY_NODE_LIMIT = 3
+SINGLE_QUERY_NODE_LIMIT = 2
 
 
 class ProviderProxy(HTTPProvider):
--- a/golem/ethereum/web3/providers.py
+++ b/golem/ethereum/web3/providers.py
@@ -1,6 +1,5 @@
 import itertools
 import logging
-import random
 import socket
 import time
 import typing
@@ -20,9 +19,6 @@ class ProviderProxy(HTTPProvider):
 
     def __init__(self, initial_addr_list) -> None:
         super().__init__()
-        initial_addr_list = list(initial_addr_list)
-        random.shuffle(initial_addr_list)
-
         self._node_addresses = itertools.cycle(initial_addr_list)
         self.provider = self._create_remote_rpc_provider()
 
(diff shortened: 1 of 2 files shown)


golem/ethereum/web3/providers.py
```diff
@@ -1,65 +1,102 @@
...
+SINGLE_QUERY_RETRY_LIMIT = 3
+SINGLE_QUERY_NODE_LIMIT = 3
+
 
 class ProviderProxy(HTTPProvider):
 
     def __init__(self, initial_addr_list) -> None:
         super().__init__()
-        self.initial_addr_list = initial_addr_list
-        self.addr_list = initial_addr_list
```
mfranciszkiewicz, line 18: Wouldn't this work just fine if `addr_list` was always a copy of `initial_addr_list` (e.g. shuffled)?

golem/ethereum/web3/providers.py
```diff
@@ -1,65 +1,102 @@
...
+        self._initial_addr_list = initial_addr_list
```
jiivan: Looks like there is no need to store this initial list.
```suggestion
```

jiivan (approved):
LGTM

golem/ethereum/web3/providers.py
jiivan: Maybe shuffle this list before passing it to `cycle()`?

golem/ethereum/web3/providers.py
shadeofblue: @mfranciszkiewicz well, it wouldn't cycle then...

mfranciszkiewicz: What I mean is that `self.addr_list` and `self.initial_addr_list` are basically references to `initial_addr_list` and this code does not work as intended.

How wouldn't it cycle? It tries to `self.addr_list = self.initial_addr_list` but since they're the same, there is no cycle

golem/ethereum/web3/providers.py
shadeofblue: gotcha, done

golem/ethereum/web3/providers.py
shadeofblue: @mfranciszkiewicz well, yes, we could make a copy and always reset to a copy ... _or_ we can utilize the `cycle` iterator... (in which case we don't need to do any resetting after the initialization of the iterator)

mfranciszkiewicz: I agree, cycling through the list is better. But overall, the code got more complicated and I'm not a fan of that

shadeofblue: @mfranciszkiewicz well, I would say it got _clearer_, especially when you take into consideration the fact that it didn't even work the way it seemed because of the singleton-like nature of `ProviderProxy` and the fact that it's called from multiple threads anyway... 

so ... its alleged simplicity is only an illussion ... and one that has potential to disrupt our geth nodes too by spamming them indefinitely ...

shadeofblue: @mfranciszkiewicz I have put some more explanation in the rationale above

maaktweluit (approved):

golem/ethereum/web3/providers.py
mfranciszkiewicz: In general, I have a feeling that you describe a different mechanism than implemented

- retries weren't implemented as retries per specific request, we cared only about successive failures for requests made,
- list shuffling was broken in the last commit,
- cycling through the node list was broken in the last commit,
- reaching the last node was a trigger to count how many times we've cycled in order to fail (it used to hang indefinitely during startup)

shadeofblue: @mfranciszkiewicz yes, the mechanism here is modified ... 

but anyway, the fact that we only cared about successive failures meant that an unsuccessful query would hit geth nodes _repeatedly_ and _indefinitely_, potentially causing strain on them

I don't know which "last commit" you're talking about ... I confirmed the issue was even more pronounced in b0.20.2 (when we were trying to filter _all_ blocks from genesis onward) so I'm pretty sure that's no "last commit broke it" case ...

mfranciszkiewicz: > so I'm pretty sure that's no "last commit broke it" case

Above, I've pointed out _what_ was broken in the last commit made to the edited file.

PR #4973, opened 2019-12-05
b0.22 - task-api - disable providing while requesting without config

maaktweluit:
replace `accept_tasks` config toggle with `requested_task_manager.has_unfinished_tasks()`

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -1431,6 +1431,12 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
                 'environment': environment
             }
 
+        # computing a task
+        if self.task_server.requested_task_manager.has_unfinished_tasks():
+            return {
+                'status': 'Requesting a task',
+            }
+
         # not accepting tasks
         if not self.config_desc.accept_tasks:
             return {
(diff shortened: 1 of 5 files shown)


mfranciszkiewicz (changes requested):
There is a race condition within this solution - the node may request a subtask while a task is being created. Subtasks are requested frequently enough for this issue to be apparent. The config change was a faster state trigger (side effects aside).

What I'd suggest is to return early from `Client.change_config` for chosen config properties (e.g. whitelisted by the config approver) or create a separate method that only does `self.config = self.config_approver.change_config(new_config_desc)` for those properties.

maaktweluit:
@mfranciszkiewicz i do not see a race created here, now tasks are no longer picked up from `requested_task_manager.create_task` that happens a few lines before the config was turned off.

mfranciszkiewicz:
@maaktweluit It simply takes longer from calling `comp.task.create` to creating a model in the database (so that `has_unfinished_tasks()` returns True) than changing an option in config (and relying on that value in code).
 In the meantime, the Golem node may request a subtask from the network and if that succeeds, the computation flow will continue in parallel with requesting a task.

mfranciszkiewicz:
Alternative solution merged #4976

PR #4974, opened 2019-12-05
Strip newline from docker-machine ip output

Wiezzel:
Docker-machine outputs a newline character at the end of command output. In case of `ip` command this results in invalid address string if the trailing newline is not stripped.


mfranciszkiewicz (approved):

maaktweluit (approved):
LGTM!

Would be nice to strip inside the `.command(` and/or the `DockerCommandHandler(`

Wiezzel:
@maaktweluit Nice idea but we don't have much time now so I'm merging it as is.

PR #4978, opened 2019-12-06
Use utc timestamp when checking task deadline

maaktweluit:
Found by QA team while testing task-api.
`time.time()` was compared to `default_now().timestamp()`

- Made `get_timestamp_utc()` also use `default_now()`
- Moved `default_now` to golem.core.common to avoid circular imports
- Replaced `time.time()` with `get_timestamp_utc()`

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -897,7 +897,7 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
                 return None
             subtask_ids = rtm.get_requested_task_subtask_ids(task_id)
             if task.start_time is None:
-                time_started = model.default_now().timestamp()
+                time_started = get_timestamp_utc()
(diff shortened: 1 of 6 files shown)


Wiezzel (approved):

mfranciszkiewicz (approved):

etam:
FYI: There is no such thing as "UTC timestamp". It's just timestamp. It's by definition relative time to Epoch, which is a 00:00:00 **UTC** on 1 January 1970. Anything else is just not a timestamp.

Wiezzel:
@etam I don't know where did you get your "definition" but it is clearly some arbitrary usurpation of the term "timestamp". Timestamp is just a mark specifying when something occurred. So writing "UTC timestamp" is **not** a pleonasm.

PR #4980, opened 2019-12-09
Mainnet integration tests

jiivan:


maaktweluit (commented):
Please note when the granary is used for main net the same keys will be used for both.

I think its best to disable the granary when mainnet until the granary has a switch for mainnet / testnet to keep the keys separated

shadeofblue (approved):
girl a okay stockphoto

PR #4982, opened 2019-12-09
Container not shut down on task creation failure

mfranciszkiewicz:
- shuts down the container if no more tasks are pending in order to free resources (i.e. bound ports)
- changes `RequestedTask`'s status to `errorCreating` on failure

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -1,13 +1,14 @@
 import asyncio
 import hashlib
 import logging
-from functools import partial
-
 import os
 import shutil
 from datetime import timedelta
+from functools import partial
 from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional
+
+import async_generator
 
 from dataclasses import dataclass
 from golem_messages import idgenerator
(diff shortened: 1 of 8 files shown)


golem/task/rpc.py
```diff
@@ -529,13 +531,14 @@ def _create_legacy_task(
...
-        return task_id
+        return task_id, None
 
+    @defer.inlineCallbacks
     def _create_task_api_task(
             self,
             golem_params: dict,
             app_params: dict,
-    ) -> TaskId:
+    ):
```
maaktweluit, line 541: this still returns a `TaskId`

maaktweluit (changes requested):
LGTM!

One small return type comment

golem/task/rpc.py
mfranciszkiewicz: The type returned is not `TaskId` but a `Generator` generic type due to `inlineCallbacks`

mfranciszkiewicz: I have issues with various linters regarding the returned type when I try to hint it as `Generator[...]` - I haven't seen anyone getting it right

maaktweluit: ahh, damn linters :(

ok, lets leave it like this then :)

maaktweluit (approved):

golem/task/requestedtaskmanager.py
```diff
@@ -265,6 +282,32 @@ def create_task(
...
+                task = RequestedTask.get(task_id=task_id)
+                task.status = TaskStatus.errorCreating
+                task.save()
+                if not app_id:
+                    app_id = task.app_id
+            except RequestedTask.DoesNotExist:
+                pass
+            try:
+                if app_id:
+                    await self._shutdown_app_client(app_id)
```
Wiezzel, line 305: Shouldn't we check here if there are no more tasks running on this client?

mfranciszkiewicz: The check is inside `_shutdown_app_client`

Wiezzel: Oh, of course.

Wiezzel (approved):

PR #4989, opened 2019-12-10
fix force payment concent integration test

shadeofblue:
additionally
* don't send `SystemExit` and `Exit` to sentry in any case
* don't try to freeze raven's time calls

### DIFF
--- a/golem/ethereum/incomeskeeper.py
+++ b/golem/ethereum/incomeskeeper.py
@@ -208,6 +208,7 @@ class IncomesKeeper:
             return
 
         for income in incomes:
+            logger.debug("Marking payment as overdue: sender: %s, amount: %s ")
             income.wallet_operation.status = \
                 model.WalletOperation.STATUS.overdue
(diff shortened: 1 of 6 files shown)


golem/ethereum/incomeskeeper.py
```diff
@@ -208,6 +208,7 @@ def update_overdue_incomes() -> None:
             return
 
         for income in incomes:
+            logger.debug("Marking payment as overdue: sender: %s, amount: %s ")
```
maaktweluit:
```suggestion
            logger.debug("Marking payment as overdue. sender=%s, amount=%s ")
```
would format data as kwargs, according to logging guide

also no values are passed for the formatters.

maaktweluit (changes requested):
LGTM!

small nit and bug in the log message

jiivan (approved):

maaktweluit (approved):

PR #4990, opened 2019-12-12
update requirements

etam:
I just used `pip freeze` on a fresh venv and manually adjusted multidict version. It has no further dependencies, so it should be fine.


maaktweluit (changes requested):
`python-dateutil==2.8.1`

is now duplicate and should be removed from `requirements-lint.txt`
https://buildbot.golem.network/buildbot/#/builders/8/builds/16523

maaktweluit (approved):

shadeofblue:
5 thumbs up

etam:
by the way: multidict issue with pyinstaller should be fixed by now: https://github.com/pyinstaller/pyinstaller/issues/4572

PR #4992, opened 2019-12-12
Task-api handle `cannot_compute_task` message

maaktweluit:
Missing check for RTM, added in this PR

### DIFF
--- a/golem/task/requestedtaskmanager.py
+++ b/golem/task/requestedtaskmanager.py
@@ -996,7 +996,13 @@ class RequestedTaskManager:
         )
         update_provider_efficacy(node_id, op)
         if subtask_timeout is not None:
-            update_provider_efficiency(node_id, subtask_timeout, comp_time)
+            if comp_time:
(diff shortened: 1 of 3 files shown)


golem/task/requestedtaskmanager.py
```diff
@@ -996,7 +996,13 @@ def _finish_subtask(self, subtask: RequestedSubtask, op: SubtaskOp):
         )
         update_provider_efficacy(node_id, op)
         if subtask_timeout is not None:
-            update_provider_efficiency(node_id, subtask_timeout, comp_time)
+            if comp_time:
```
mfranciszkiewicz, line 999: At this point we have obtained computation time. Is this check here because it might be equal to 0 for some reason?

mfranciszkiewicz (approved):

Wiezzel (approved):
LGTM

golem/task/requestedtaskmanager.py
maaktweluit: yep, it showed n error when it was 0.

This check was copied from the old task manager

PR #4999, opened 2019-12-16
Fixed recovery logic for DockerMachineHypervisor

Wiezzel:
* Set timeout of 120 s for regenerating certificates and restarting the VM (on my machine it takes approx. 60 s so I guess 120 s should be enough for slower PCs). Lack of timeout could cause the command to hang forever.
* After restarting the VM try to regenerate certificates once more.


maaktweluit (approved):
LGTM!

mfranciszkiewicz (approved):

PR #5003, opened 2019-12-17
Resolved misc task-api related warnings

maaktweluit:
- Fix trust mod warning for task-api tasks
- Fix "can not iterate NoneType object" when there are no offers
- Fix race where task funds where unlocked before subtask funds

Each fix in its own commit

### DIFF
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -380,13 +380,11 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin):
         )
 
         def resolution(market_strategy, task_id):
-            offers = market_strategy.resolve_task_offers(task_id)
-            if offers:
-                for offer in offers:
-                    try:
-                        offer.callback()
-                    except Exception as e:
-                        logger.error(e)
+            for offer in market_strategy.resolve_task_offers(task_id):
+                try:
+                    offer.callback()
+                except Exception as e:
+                    logger.error(e)
 
         market_strategy.add(msg.task_id, offer)
         logger.debug("Offer accepted & added to pool. offer=%s", offer)
--- a/golem/task/tasksession.py
+++ b/golem/task/tasksession.py
@@ -380,11 +380,13 @@ class TaskSession(BasicSafeSession, ResourceHandshakeSessionMixin):
         )
 
         def resolution(market_strategy, task_id):
-            for offer in market_strategy.resolve_task_offers(task_id):
-                try:
-                    offer.callback()
-                except Exception as e:
-                    logger.error(e)
+            offers = market_strategy.resolve_task_offers(task_id)
+            if offers:
+                for offer in offers:
+                    try:
+                        offer.callback()
+                    except Exception as e:
+                        logger.error(e)
 
         market_strategy.add(msg.task_id, offer)
         logger.debug("Offer accepted & added to pool. offer=%s", offer)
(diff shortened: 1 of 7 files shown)


golem/task/tasksession.py
```diff
@@ -380,11 +380,13 @@ def __init__(
         )
 
         def resolution(market_strategy, task_id):
-            for offer in market_strategy.resolve_task_offers(task_id):
-                try:
-                    offer.callback()
-                except Exception as e:
-                    logger.error(e)
+            offers = market_strategy.resolve_task_offers(task_id)
```
jiivan: If there is a case of `market_strategy.resolve_task_offers()` returning **None**, maybe it wuld be better for it to always return an iterable?

maaktweluit: Good one! was thinking about it too but did not fix it yet.

Updated the code now :)

jiivan (approved):

golem/task/requestedtaskmanager.py
```diff
@@ -524,22 +524,24 @@ def has_unfinished_tasks() -> bool:
...
-                    self._move_task_results(
-                        task_id,
-                        Path(task.output_directory))
-                    logger.info("Task finished. task_id=%r", task.task_id)
-                    await self._shutdown_app_client(task.app_id)
-                    self._notice_task_updated(task, op=TaskOp.FINISHED)
-
         return result
 
+    async def verify_success(self, task_id: TaskId):
```
mfranciszkiewicz: Please do not introduce a new indirection (RTM -> TS -> RTM) and keep the flow RTM-only.

I cannot see the race condition in question since all operations are done in sequence, including the `verification_finished` call. Could you provide more information?

maaktweluit: was trying that first, to keep the flow RTM only, but with the queue in between the callback was not so easy to pass. This is how its done in the old task manager.

You are right, there is no race condition ( i thought it since the verify part was fire and forget ) the order of callbacks is wrong. The subtask success handler, unlocking funds and sending results, should run before the task success handler `verify_success(` in the new code.

For the release i think its best to leave this as is, on develop i can try to make the callbacks cleaner and in the proper order.

maaktweluit: was trying that first, to keep the flow RTM only, but with the queue in between the callback was not so easy to pass. This is how its done in the old task manager.

You are right, there is no race condition ( i thought it since the verify part was fire and forget ) the order of callbacks is wrong. The subtask success handler, unlocking funds and sending results, should run before the task success handler verify_success( in the new code.

For the release i think its best to leave this as is, on develop i can try to make the callbacks cleaner and in the proper order.

mfranciszkiewicz: > The subtask success handler, unlocking funds and sending results, should run before the task success handler verify_success( in the new code

I disagree - we should persist the new task and subtask states before sending the verification result.

AFAICT things break because of one the "finished" event handlers (emitted in `_notice_task_updated` via `dispatcher`) which mutates the state of funds locker.

If the funds are properly unlocked, wouldn't a quickfix be to (conditionally) suppress the warning message?

mfranciszkiewicz (approved):

golem/task/requestedtaskmanager.py
maaktweluit: you are right, the funds are properly unlocked so i suppressed the warning

PR #5004, opened 2019-12-18
Properly handle timeouts/cancellations of Task API subtasks

Wiezzel:
If task gets timed out or cancelled a failure message should be sent to requestor. In case of timeout requestor trust should be decreased but not in case of cancellation (it's not requestor's fault that provider has cancelled computation).


maaktweluit (approved):

etam (approved):
ok

mfranciszkiewicz (approved):

PR #5010, opened 2019-12-19
Generic cache in TransactionSystem

jiivan:


golem/ethereum/transactionsystem.py
```diff
@@ -125,17 +127,33 @@ def __init__(self, datadir: Path, config) -> None:
...
-        self._last_eth_update: Optional[float] = None
-        self._last_gnt_update: Optional[float] = None
         self._payments_locked: int = 0
         self._gntb_locked: int = 0
         self._gntb_withdrawn: int = 0
         # Amortized gas cost per payment used when dealing with locks
         self._eth_per_payment: int = 0
 
+    @property
+    def _eth_balance(self) -> int:
```
shadeofblue, line 145: if the try/except was part of `cache_get`, I think you could implement these properties as `functools.partial`s ...

golem/ethereum/transactionsystem.py
```diff
@@ -754,10 +772,19 @@ def on_receipt(receipt) -> None:
...
+            self,
+            account_address: Optional[str] = None,
+            cached: bool = True,
+    ) -> int:
         self._sci: SmartContractsInterface
         if account_address is None:
             account_address = self._sci.get_eth_address()
+        if cached and (account_address == self._sci.get_eth_address()):
+            try:
+                self.cache_get('concent_balance')
```
shadeofblue: could be `GNTDeposit` if the others are `GNT`, `ETH` and `GNTB` ... plus, maybe `GNT`, `ETH`, `GNTB` and `GNTDeposit` should be a StringEnum?

golem/ethereum/transactionsystem.py
jiivan: `self` is not available at the moment of class definition. Added `default` keyword to cache though.

golem/core/cache.py
```diff
@@ -0,0 +1,36 @@
+import time
+import typing
+
+from dataclasses import dataclass
+
+
+@dataclass()
```
kmazurek, line 7:
```suggestion
@dataclass
```

golem/core/cache.py
```diff
@@ -0,0 +1,36 @@
...
+            self._cache_store: typing.Dict[str, CacheEntry] = {}
+        return self._cache_store
+
+    def cache_get(self, key, **kwargs) -> object:
+        try:
+            return self._cache[key].value
+        except KeyError:
+            if 'default' in kwargs:
+                return kwargs['default']
+            raise
```
kmazurek, line 31: I think this would be safer and cleaner if instead of raising the function would return `None` on a cache miss. My rationale is the following:
1. A cache miss is an expected behaviour rather than an error.
2. `Optional[object]` as return type would be enough to document the behaviour.
3. Passing in an explicit default would be optional, simplifying the calls to `cache_get`.

jiivan: It would mean that you can't cache `None`. It won't be a problem as long as you only store ints, but in general I think it's a problem.

golem/core/cache.py
jiivan: Why?

kmazurek: For consistency. Although this is not explicitly stated in the code style, the code base seems to stick to omitting parentheses for decorators with no arguments.

maaktweluit: lets make it a code style rule!

https://github.com/golemfactory/styleguide

maaktweluit (approved):

golem/core/cache.py
jiivan: Actually none of current golem decorators support this double invocation style.
SEE: https://github.com/golemfactory/golem/blob/develop/golem/decorators.py

And also:
https://github.com/golemfactory/golem/blob/f909422af772bab2798639a6a51139b47f9abe9f/golem/ethereum/transactionsystem.py#L58-L79

golem/core/cache.py
kmazurek: I see, makes sense. One option would be to return some special value in case of a miss and have the return type be a `Union` with the special value type.
In other case, I think it would make sense to mention the raising behaviour in docs.

golem/core/cache.py
kmazurek: Hm, okay, I was not aware that the `@dataclass` decorator was an exception. Still, in the Golem code base it is always invoked with no parentheses.

kmazurek (approved):

PR #5018, opened 2019-12-20
Added legacy fields to task-api comp.task rpc command

maaktweluit:
This way electron does not crash after adding a task-api task.

### DIFF
--- a/golem/client.py
+++ b/golem/client.py
@@ -23,6 +23,7 @@ from typing import (
 
 from ethereum.utils import denoms
 from golem_messages import datastructures as msg_datastructures
+from golem_task_api.envs import DOCKER_GPU_ENV_ID
 from pydispatch import dispatcher
 from twisted.internet.defer import (
     inlineCallbacks,
@@ -923,6 +924,10 @@ class Client:  # noqa pylint: disable=too-many-instance-attributes,too-many-publ
                 last_updated = get_timestamp_utc()
             else:
                 last_updated = task.end_time.timestamp()
+            # compute_on
+            compute_on = 'cpu'
+            if task.env_id == DOCKER_GPU_ENV_ID:
+                compute_on = 'gpu'
             # estimated_cost and estimated_fee
             subtask_price = calculate_subtask_payment(
                 task.max_price_per_hour,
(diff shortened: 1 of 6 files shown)


kmazurek:
I think you should change the base branch here to `b0.22`.

maaktweluit:
@zakaprov you are correct, thanks!

Updated the base

golem/client.py
```diff
@@ -897,14 +899,59 @@ def get_task(self, task_id: str) -> Optional[dict]:
...
                 'id': task.task_id,
+                'time_remaining': time_remaining,
+                'subtasks_count': task.max_subtasks,
                 'status': task.status.value,
+                'progress': progress,
                 'time_started': time_started,
+                'last_updated': last_updated,
+                'name': task.name,
+                'bid': float(task.max_price_per_hour) / denoms.ether,
+                'compute_on': task.env_id,
```
mfranciszkiewicz: Afair this should be either cpu or gpu

golem/client.py
```diff
@@ -897,14 +899,59 @@ def get_task(self, task_id: str) -> Optional[dict]:
             if not task:
                 return None
             subtask_ids = rtm.get_requested_task_subtask_ids(task_id)
+            # time_started
             if task.start_time is None:
                 time_started = get_timestamp_utc()
+                if task.status == taskstate.TaskStatus.errorCreating:
+                    time_started -= 1
```
mfranciszkiewicz, line 907: I believe you meant `time_started = -1`. Maybe use `None` if it’s handled properly?

golem/client.py
maaktweluit: do we want to match the values for old tasks? i think `docker_cpu` and `docker_gpu` describe better where the task is computed on.

In my tests this was not breaking the GUI

@mdtanrikulu what do you think?

golem/client.py
maaktweluit: the idea was to show tasks in the order:
- Tasks that started
- errorCreating
- creating

When both creating and errorCreating get assigned to `now()` there is no order

`None` would be best but i'm not sure if its handled properly, that would be a fix for develop

golem/client.py
mfranciszkiewicz: If the goal is to add legacy fields, then I believe we do

maaktweluit: looking at electron this value is only printed: https://github.com/golemfactory/golem-electron/search?q=compute_on&unscoped_q=compute_on

my proposal is to extend the possible values of this field with the new envs:
- `cpu` -> non task-api cpu
- `docker_cpu` -> task-api cpu
- etc...

this means we do not have to store a mapping from `env_id` to `compute_on`

why do you think its better to only stick with the old values? happy to change it :)

mfranciszkiewicz (approved):

PR #5030, opened 2020-01-02
fix the `update_requestor_efficiency` call

shadeofblue:
+ logs

### DIFF
--- a/golem/task/timer.py
+++ b/golem/task/timer.py
@@ -13,6 +13,12 @@ class ActionTimer:
         self._started: Optional[float] = None
         self._finished: Optional[float] = time.time()
 
+    def __repr__(self):
+        return f"{self.__class__.__name__}: " \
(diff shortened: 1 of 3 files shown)


jiivan (approved):

golem/task/timer.py
```diff
@@ -13,6 +13,12 @@ def __init__(self) -> None:
         self._started: Optional[float] = None
         self._finished: Optional[float] = time.time()
 
+    def __repr__(self):
```
jiivan, line 16: Have you considered using https://github.com/golemfactory/golem/blob/develop/golem/core/printable_object.py ?

shadeofblue: @jiivan well, I forgot about it :) ... still, I think in this case, this custom implementation will give a better digest of the state of the timer

Issue #5042, opened 2020-01-14
Task-api: generate app resources list from golem resources


maaktweluit:
Currently when making a task-api `task.json` file the resources list needs to be repeated.

Maybe we can generate the list in the `app` block from the list in the `golem` block.

When fixed, please update this guide:
https://github.com/golemfactory/golem/blob/develop/scripts/task_api_tests/task-api-readme.md#appresources

PR #5043, opened 2020-01-14
Task-api-docs: removed todo's

maaktweluit:
- Removed all TODO's
- made #5042
- specify GUI is not supported, since it will show now since #5018


jamuszyn (commented):
LGTM

jamuszyn (approved):

PR #5046, opened 2020-01-16
Add task-api-dev mode setting

maaktweluit:
While writing the [task-api docs](
https://taskapi.docs.golem.network/#running-the-application-locally) i found there are some code hacks needed to run the `tutorialapp` locally. 

Introduced `task-api-dev` mode to make this easier, this new mode disables:
- image whitelist checks
- pulling of images

### DIFF
--- a/golem/core/variables.py
+++ b/golem/core/variables.py
@@ -18,6 +18,7 @@ APP_NAME = "Brass Golem"
 PRIVATE_KEY = "keystore.json"
 DEFAULT_PROC_FILE = "node_processes.ctl"
 MAX_PROC_FILE_SIZE = 1024 * 1024
+ENV_TASK_API_DEV = 'ENV_TASK_API_DEV'
 
 #################
 # NETWORK CONST #
--- a/golem/core/variables.py
+++ b/golem/core/variables.py
@@ -18,7 +18,7 @@ APP_NAME = "Brass Golem"
 PRIVATE_KEY = "keystore.json"
 DEFAULT_PROC_FILE = "node_processes.ctl"
 MAX_PROC_FILE_SIZE = 1024 * 1024
-ENV_TASK_API_DEV = 'ENV_TASK_API_DEV'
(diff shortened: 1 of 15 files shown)


Wiezzel (approved):
Looks nice.
I suggest preparing a documentation update for this feature (to be published when it's released ofc).

maaktweluit:
@Wiezzel Great tip! will add it to my list

etam:
@maaktweluit Will it be written to config file? Is it good?

maaktweluit:
@etam good catch, currently it is toggle-able from config and from arguments.

First i made it only config based, but added the argument later. Now i think the argument will update the config by accident..

What do you think? should we :
- fix both options 
- only make it an argument
- only make it config

think the code will be most clean with only config, but the most dangerous option if someone leaves it on by accident

etam:
> think the code will be most clean with only config, but the most dangerous option if someone leaves it on by accident

That's exactly what I thought. I think it would be better to make it configurable only through command line.

mfranciszkiewicz (approved):

maaktweluit:
@etam fixed!

Removed the setting from config, added the option to use an ENV var.

@wiezzel @mfranciszkiewicz please also check again

golem/core/variables.py
```diff
@@ -18,6 +18,7 @@
 PRIVATE_KEY = "keystore.json"
 DEFAULT_PROC_FILE = "node_processes.ctl"
 MAX_PROC_FILE_SIZE = 1024 * 1024
+ENV_TASK_API_DEV = 'ENV_TASK_API_DEV'
```
etam:
```suggestion
ENV_TASK_API_DEV = 'GOLEM_TASK_API_DEV'
```

I think it makes more sense to not use `ENV` in env, but make it more specific to golem.

maaktweluit: good one! updated :)

etam (approved):
ok

PR #5052, opened 2020-01-17
address a security vulnerability in pyinstaller < 3.6

shadeofblue:


jiivan (approved):
tworecapproves

maaktweluit (approved):
Would tripple check the produced binaries, they could have issues building or running when hooks changed.

shadeofblue:
> Would tripple check the produced binaries, they could have issues building or running when hooks changed.

well, sure, first they need to build at all :)

Issue #4985, opened 2019-12-10
When restarting task fails task stays in "preparing for computations' state


ederenn:
## Description

**Golem Version**: GOLEM Version: 0.21.0+dev504.g5e5363a 

**Golem-Messages version** (leave empty if unsure): 3.14.1

**Electron version** (if used): 0.21.0

**OS** [e.g. Windows 10 Pro]: Windows 10 Pro

**Branch** (if launched from source): 

**Mainnet/Testnet**: mainnet

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

Task was restarted from interface. Task was created in logs and on a task list in GUI but it failed to be restarted. 
`2019-12-09 17:22:15 ERROR    golem.task.rpc                      Cannot restart task 'b6cb73ba-1a87-11ea-82f9-350634f06032': [WinError 206] The filename or extension is too long: 'C:\\Users\\ederenn\\Documents\\scene-Helicopter-27-cyc2_2019-12-05_16-02-57\\scene-Helicopter-27-cyc2_2019-12-05_16-05-59\\scene-Helicopter-27-cyc2_2019-12-06_15-50-14\\scene-Helicopter-27-cyc2_2019-12-09_14-27-57\\scene-Helicopter-27-cyc2_2019-12-09_17-22-15' `
Instead of changing state to failure task hangs in "preparing for computations' state over timeout time.

**Screenshots**:

![Screenshot from 2019-12-09 17-26-05](https://user-images.githubusercontent.com/32061891/70528519-a4880c80-1b4e-11ea-9d34-64ec4bfd973a.png)

## Steps To Reproduce
_Short description of steps to reproduce the behavior:_
e.g.
1. Launch Golem
2. Start a short task, e.g. helicopter.blend, res: 400x400, sample 30, 1 frame, 4 subtasks, 10 min for a task, 5 minutes for subtask, bid 0.1 Don't rename task, leave name from file.
3. Wait for task to finish.
4. Restart same task at least 6 times, restarting task will fail eventually
5. Check logs for error.

## Logs and any additional context

```
2019-12-09 17:22:15 INFO     golem.task.taskmanager              Task b6cb73ba-1a87-11ea-82f9-350634f06032 put into restarted state 
[...]

PR #5060, opened 2020-01-22
Handle task failure during creation

jiivan:
fixes: #4985

### DIFF
--- a/golem/task/rpc.py
+++ b/golem/task/rpc.py
@@ -433,9 +433,20 @@ def _create_task_error(e, _self, task_dict, *args, **_kwargs) \
     return None, str(e)
 
 
-def _restart_task_error(e, _self, task_id, *args, **_kwargs) \
-        -> typing.Tuple[None, str]:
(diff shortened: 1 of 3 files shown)


golem/task/rpc.py
```diff
@@ -433,9 +433,20 @@ def _create_task_error(e, _self, task_dict, *args, **_kwargs) \
     return None, str(e)
 
 
-def _restart_task_error(e, _self, task_id, *args, **_kwargs) \
-        -> typing.Tuple[None, str]:
+def _restart_task_error(
+        e: Exception,
+        self: 'ClientProvider',
```
maaktweluit, line 438: is self a reserved word?

maaktweluit (approved):
LGTM!

Small nit about a maybe reserved keyword

golem/task/rpc.py
jiivan: It's not. It's a convention.

kmazurek (approved):

PR #5078, opened 2020-01-29
Follow symlinks for sys.executable

maaktweluit:
After the pyinstaller update to 3.6 the sys.executable no longer follows symlinks. Will make an issue on their side since the docs state it should.

from [docs](https://pyinstaller.readthedocs.io/en/stable/runtime-information.html):
> If the user launches the app by way of a symbolic link, sys.argv[0] uses that symbolic name, while sys.executable is the actual path to the executable. Sometimes the same app is linked under different names and is expected to behave differently depending on the name that is used to launch it. For this case, you would test os.path.basename(sys.argv[0])

### DIFF
--- a/apps/blender/benchmark/benchmark.py
+++ b/apps/blender/benchmark/benchmark.py
@@ -3,7 +3,7 @@ import sys
 import tempfile
 from typing import Type
 
-from os.path import dirname, join
+from os.path import dirname, join, realpath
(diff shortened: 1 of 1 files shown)


shadeofblue (approved):
wolfofwallstreet

kmazurek (approved):

Issue #4363, opened 2019-06-19
RPC for ACL management


prekucki:
## Rationale

Currently, there are only operations for adding block rule to ACL: `net.peer.block`, 
`net.peer.block_ip`. But there is no way to:
 - list current rules
 - remove node from blacklist.

### Functional specification

New operations:

| uri | name | description |
| -- | -- | -- |
| net.peer.allow | allow_node | Whitelists node |
| net.peer.allow_ip | allow_ip | Removes ip from blacklist |
| net.peer.acl | acl_status | Returns ACL configuration for nodes |
| net.peer.acl_ip | acl_ip_status | Returns ACL for ip addresses | 
| net.peer.acl.new | acl_setup | Resets nodes ACL to new configuration |

Existing operations

| uri | name | description |
| -- | -- | -- |
| net.peer.block | disallow_node | Blacklists node |
| net.peer.block_ip | disallow_ip | Adds ip to blacklist |

#### `allow_node`

**URI**: `net.peer.allow`

| In/Out | name | type | is optional | 
| -- | -- | -- | -- |
| IN | node_id | str | No |
| IN | persist | bool | Yes, default `true` |
| OUT | - | None | |

#### `allow_ip`

**URI**: `net.peer.allow_ip`

| In/Out | name | type | is optional | 
| -- | -- | -- | -- |
| IN | ip | str | No |
| IN | persist | bool | Yes, default `true` |
| OUT | - | None | |

#### acl_status

**URI**: `net.peer.acl`

| In/Out | name | type | is optional | 
| -- | -- | -- | -- |
| OUT | AclStatus |  | |

#### acl_ip_status

**URI**: `net.peer.acl_ip`

| In/Out | name | type | is optional | 
| -- | -- | -- | -- |
| OUT | AclStatus |  | |

#### acl_setup

**URI**: `net.peer.acl.new`

| In/Out | name | type | description | 
| -- | -- | -- | -- |
| IN | default_rule | str | required, valid values: `allow`, `deny` |
| IN | exceptions | `List[str]` | required, list of exceptions to default rule |

#### AclStatus

| Field | type | description | 
| default_rule | str | enum: `allow`, `deny` |
| rules | List[Tuple[str, str, Optional[int]]] | list of triplets in form: `(item, rule, duration)`. |

**Examples:**

```json
{"default_rule": "allow",
    "rules": [
      [
[...]

PR #5095, opened 2020-02-13
ACL improvement

### LINKED ISSUE #630: ACL list not the same as from golemcli
## Description

**Golem Version**: 0.22.0 release

**Golem-Messages version** (leave empty if unsure):

**Electron version**: packed with release b0.22

**OS** [e.g. Windows 10 Pro]: ubuntu

**Branch** (if launched from source):

**Mainnet/Testnet**: mainnet

???

**Description of the issue**:

Banned nodes do not show in GUI

**Actual result**:

GUI does not show any banned nodes. while `golemcli acl show` does

**Screenshots**:

Clipboard - January 16, 2020 2_50 PM

## Steps To Reproduce
_Short description of steps to reproduce the behavior:_
e.g.
1. Launch the app
2. Wait for handshake to fail
3. Go to ACL tab in GUI
4. See empty list

## Expected behavior
Node that are banned also show on the GUI

_(Any additional information that could help to resolve the issue, which systems were checked to reproduce the issue)_
_Please upload your logs if possible_

## Proposed Solution?
_(Optional: What could be a solution for that issue)_

mdtanrikulu:
Resolves golemfactory/golem-electron#630.

Requires those pr's to be merged;
- golemfactory/golem-electron#635
- golemfactory/golem-client#99

#### Changes
- Drop `net.peer.disallow` RPC due to not being used.
- Move Block node definition from `client.py` to `acl.py`
- Drop `persist` argument from `net.peer.block`, while there is `timeout (default = -1)` argument, it's unnecessary to have another argument.
- Check if nodes already being added or removed from/to the list
- Update tests


etam:
FYI: https://github.com/golemfactory/golem/issues/4363

golem/task/acl.py
```diff
@@ -169,14 +168,16 @@ def disallow(self, node_id: str,
             assert isinstance(node_deadlines, SortedList)
             node_deadlines.add(self._deadline(timeout_seconds))
 
-        if persist and timeout_seconds == -1:
+        if timeout_seconds < 0:
```
etam: I think it would be more readable, to declare `persist = timeout_seconds < 0` at the beginning of function.

golem/task/acl.py
```diff
@@ -169,14 +168,16 @@ def disallow(self, node_id: str,
...
             try:
-                existNode = ACLDeniedNodes.get(node_id=node_id)
+                ACLDeniedNodes.get(node_id=node_id)
+                raise Exception(node_id)
```
etam: I'd add some explanation like "ACL already defined".

golem/task/acl.py
```diff
@@ -192,11 +193,10 @@ def allow(self, node_id: str, persist: bool = False) -> None:
 
         if persist:
             try:
-                existNode = ACLDeniedNodes.get(node_id=node_id)
+                ACLDeniedNodes.get(node_id=node_id)
             except ACLDeniedNodes.DoesNotExist:
-                existNode = None
-
-            if existNode:
+                raise Exception(node_id)
```
etam: I'd add some explanation like "ACL doesn't exist".

golem/task/acl.py
```diff
@@ -273,24 +273,21 @@ def is_allowed(self, node_id: str) -> Tuple[bool, Optional[DenyReason]]:
         return False, DenyReason.not_whitelisted
 
     def disallow(self, node_id: str,
-                 timeout_seconds: int = 0,
-                 persist: bool = False) -> None:
+                 timeout_seconds: int = -1) -> None:
 
-        if persist:
+        if timeout_seconds < 0:
```
etam: I think it would be more readable, to declare `persist = timeout_seconds < 0` at the beginning of function.

golem/task/acl.py
```diff
@@ -273,24 +273,21 @@ def is_allowed(self, node_id: str) -> Tuple[bool, Optional[DenyReason]]:
...
             self._allow_list = [node for node in self._allow_list if not (
                 node_id == node.node_id)]
             try:
-                existNode = ACLAllowedNodes.get(node_id=node_id)
+                ACLAllowedNodes.get(node_id=node_id)
             except ACLAllowedNodes.DoesNotExist:
-                existNode = None
-
-            if existNode:
+                raise Exception(node_id)
```
etam: I'd add some explanation like "ACL doesn't exist".

golem/task/acl.py
```diff
@@ -312,11 +309,9 @@ def allow(self, node_id: str, persist: bool = False) -> None:
 
         if persist:
             try:
-                existNode = ACLAllowedNodes.get(node_id=node_id)
+                ACLAllowedNodes.get(node_id=node_id)
+                raise Exception(node_id)
```
etam: I'd add some explanation like "ACL already defined".

jiivan:
@mdtanrikulu Should this be rebased to develop?

golem/task/acl.py
mdtanrikulu: make sense, added.

golem/task/acl.py
mdtanrikulu: the exception is a return for failing cases in rpc, not for logging. see here L1256;
https://github.com/golemfactory/golem/blob/c345bba89b44258a2c5c26c9ab93826e45c49ea4/golem/task/taskserver.py#L1243-L1256

golem/task/acl.py
mdtanrikulu: the exception is a return for failing cases in rpc, not for logging. see here L1276;
https://github.com/golemfactory/golem/blob/c345bba89b44258a2c5c26c9ab93826e45c49ea4/golem/task/taskserver.py#L1266-L1276

golem/task/acl.py
mdtanrikulu: > the exception is a return for failing cases in rpc, not for logging. see here L1256;
> 
> https://github.com/golemfactory/golem/blob/c345bba89b44258a2c5c26c9ab93826e45c49ea4/golem/task/taskserver.py#L1243-L1256

golem/task/acl.py
mdtanrikulu: > the exception is a return for failing cases in rpc, not for logging. see here L1276;
> 
> https://github.com/golemfactory/golem/blob/c345bba89b44258a2c5c26c9ab93826e45c49ea4/golem/task/taskserver.py#L1266-L1276

mdtanrikulu:
@jiivan I can, but I see that there is still couple of commits ahead of dev in b0.22. Will we merge them?

golem/task/acl.py
etam: So I'd do something like this:
```python
class ACLAlreadyDefined(Exception):
    def __init__(self, node_id: str) -> None:
        self.node_id = node_id
```
and in handler:
```python
try:
    [...]
except ACLAlreadyDefined as e:
    return False, e.node_id
except Exception as e:  #  pylint: disable=broad-except
    return False, str(e)
```
This way you make it clear what is the purpose of exception and where it's handled.

etam: Also: `node_id` may be a list. If one of entries causes an exception, the rest will be omitted. Maybe instead of throwing exceptions, `disallow` method may return a bool (or enum if more options are needed) indicating the result, and in `disallow_node` you would first gather all results and then return collected status.

jiivan:
@mdtanrikulu yes, I've created a PR https://github.com/golemfactory/golem/pull/5147 (but those commits are unrelevant to ACL as far as I can see)

golem/task/acl.py
```diff
@@ -147,16 +147,15 @@ def is_allowed(self, node_id: str) -> Tuple[bool, Optional[DenyReason]]:
         return True, None
 
     def disallow(self, node_id: str,
-                 timeout_seconds: int = -1,
-                 persist: bool = False) -> None:
+                 timeout_seconds: int = -1) -> Union[bool, str]:
```
etam: This function now returns `True` if node was disallowed or `node_id` that it gets on input if it already was. I think it should be just a `bool`.

Also, there is a `return` without value in this funcion (I can't comment there, because it wasn't changed).

golem/task/taskserver.py
```diff
@@ -1240,17 +1240,23 @@ def should_accept_requestor(self, node_id):
...
-            self.acl.disallow(item, timeout_seconds, persist)
+    ) -> Tuple[bool, List[str], Optional[str]]:
+        results: List[str] = []
+        try:
+            if isinstance(node_id, str):
+                node_id = [node_id]
+            for item in node_id:
+                result = self.acl.disallow(item, timeout_seconds)
+                if isinstance(result, str):
+                    results.append(result)
```
etam:
```python
if not self.acl.disallow(item, timeout_seconds):
    results.append(item)
```

golem/task/taskserver.py
```diff
@@ -1240,17 +1240,23 @@ def should_accept_requestor(self, node_id):
...
             node_id: Union[str, list],
             timeout_seconds: int = -1,
-            persist: bool = False
-    ) -> None:
-        if isinstance(node_id, str):
-            node_id = [node_id]
-        for item in node_id:
-            self.acl.disallow(item, timeout_seconds, persist)
+    ) -> Tuple[bool, List[str], Optional[str]]:
+        results: List[str] = []
```
etam: I'd call it `not_changed`

golem/task/taskserver.py
```diff
@@ -1240,17 +1240,23 @@ def should_accept_requestor(self, node_id):
...
+        try:
+            if isinstance(node_id, str):
+                node_id = [node_id]
+            for item in node_id:
+                result = self.acl.disallow(item, timeout_seconds)
+                if isinstance(result, str):
+                    results.append(result)
+            return True, results, None
+        except Exception as e:  # pylint: disable=broad-except
+            return False, results, str(e)
```
etam: I think that a dict, that gives names to those values would be more readable.

etam (changes requested):

golem/task/acl.py
```diff
@@ -147,41 +147,43 @@ def is_allowed(self, node_id: str) -> Tuple[bool, Optional[DenyReason]]:
...
+        if persist:
             self._deny_deadlines[node_id] = self._always
         else:
             if node_id not in self._deny_deadlines:
                 self._deny_deadlines[node_id] = SortedList(key=operator.neg)
             node_deadlines = self._deny_deadlines[node_id]
 
             if node_deadlines is self._always:
-                return
+                return True
```
etam: I think it's `False`, because node is already disallowed and nothing is changed.

golem/task/taskserver.py
```diff
@@ -1240,17 +1240,30 @@ def should_accept_requestor(self, node_id):
...
-        for item in node_id:
-            self.acl.disallow(item, timeout_seconds, persist)
+    ) -> Tuple[bool, List[str], Optional[str]]:
+        '''
+        return Tuple
+        [is_disallow_succeed, list_of_already_disallowed_nodes, err_message]
+
+        Success: [True, [], None]
+        Success with exist node: [True, ['node_id'], None]
+        Error: [False, [], 'message']
```
etam:
```suggestion
        (is_disallow_success, list_of_already_disallowed_nodes, err_message)

        Success: (True, [], None)
        Success with existing node: (True, ['node_id'], None)
        Error: (False, [], 'message')
```

golem/task/taskserver.py
```diff
@@ -1261,12 +1274,29 @@ def disallow_ip(self, ip: Union[str, list],
...
+            node_id: Union[str, list],
+            persist: bool = True
+    ) -> Tuple[bool, List[str], Optional[str]]:
+        '''
+        return Tuple
+        [is_allow_succeed, list_of_already_allowed_nodes, err_message]
+
+        Success: [True, [], None]
+        Success with exist node: [True, ['node_id'], None]
+        Error: [False, [], 'message']
```
etam:
```suggestion
        (is_allow_success, list_of_already_allowed_nodes, err_message)

        Success: (True, [], None)
        Success with existing node: (True, ['node_id'], None)
        Error: (False, [], 'message')
```

golem/task/taskserver.py
etam: It's either "is success" or "did succeed"

golem/task/acl.py
mdtanrikulu: I rather in favor of keep this `True`.
1. There is no use of temporary ban by user, unless user wants to do it in purpose.
2. No point to show warning when golem itself makes temporary bans
3. If there is no warning needed when Golem makes temp bans, it means it actually doesn't matter what returns
4. The check is not comes from db, comes from memory.

etam: ~I'm trying to understand the meaning of return value and I'm a bit puzzled.~

~Here's what I came up with (I put in _italic_ the result we're commenting on):~

Let's keep this table as description of return values;

| | not persist | persist |
| --- | --- | --- | 
| **not banned** | True (a temporary ban is added to memory) | True (a persistent ban is added to database) |
| **is persistently banned** | _False_ (nothing changes, because existing ban is stronger) | False (nothing changes) |
| **is temporarily banned** | True (a temporary ban is extended) | True (temporary ban is removed and a persistent ban is added to database) |

~So my first suggestion: Add documentation comments in `Acl` base class in~ ~`disallow` and `allow` methods what is the meaning of return value. Preferably in the~ ~form "returns True when ..." or "returns False when ...".~

etam: Returns False if node_id is already banned in database

etam (approved):
ok

PR #5103, opened 2020-02-21
task-api - Added rpc calls for managing apps

### LINKED ISSUE #5097: Add RPC call to read, update and delete task-api apps
As a task-api requestor i want to enable new apps so i can request tasks for them

Add RPC calls to:
- Read apps
- Update apps ( set status to enabled / disabled )
- Delete apps ( remove out of date apps )

maaktweluit:
Resolves #5097 

Added new commands for:
```
golemcli debug rpc apps.list
golemcli debug rpc apps.update <app_id> <enabled_flag>
golemcli debug rpc apps.delete <app_id>
```

Other improvements are:
- Moved app definition json loading to `AppManager` ( @mfranciszkiewicz please double check )
- Added `delete()` to `AppManager`
- Added missing `tearDown()`s in tests that use `TmpDirFixture`

Creates new tasks:
- Add full argument to `apps.list` rpc call (https://github.com/golemfactory/golem/issues/5104)
- Add rpc commands to golemcli (https://github.com/golemfactory/golem-client/issues/100)

### DIFF
--- a/golem/apps/manager.py
+++ b/golem/apps/manager.py
@@ -2,9 +2,7 @@ import logging
 from typing import Dict, List, Tuple
 from pathlib import Path
 
-from golem.apps import (
-    AppId, AppDefinition, load_apps_from_dir, delete_app_from_dir,
-)
+from golem.apps import AppId, AppDefinition, load_apps_from_dir
 from golem.apps.default import save_built_in_app_definitions
 from golem.model import AppConfiguration
 
@@ -17,14 +15,15 @@ class AppManager:
     def __init__(self, app_dir: Path, save_apps=True) -> None:
         self._apps: Dict[AppId, AppDefinition] = {}
         self._state = AppStates()
-        self._app_dir = app_dir
+        self._app_file_names: Dict[AppId, Path] = dict()
 
         # Save build in apps, then load apps from path
         built_in_apps: List[AppId] = []
         if save_apps:
             built_in_apps = save_built_in_app_definitions(app_dir)
-        for app_def in load_apps_from_dir(app_dir):
+        for (app_def_path, app_def) in load_apps_from_dir(app_dir):
             self.register_app(app_def)
+            self._app_file_names[app_def.id] = app_def_path
         for app_id in built_in_apps:
             self.set_enabled(app_id, True)
 
@@ -78,8 +77,8 @@ class AppManager:
     def delete(self, app_id: AppId) -> bool:
         # Delete self._state from the database first
         del self._state[app_id]
-        delete_app_from_dir(self._app_dir, self._apps[app_id])
         del self._apps[app_id]
+        self._app_file_names[app_id].unlink()
         return True
 
 
--- a/golem/apps/manager.py
+++ b/golem/apps/manager.py
@@ -77,12 +77,7 @@ class AppManager:
 
     def delete(self, app_id: AppId) -> bool:
         # Delete self._state from the database first
-        try:
-            AppConfiguration.delete() \
-                .where(AppConfiguration.app_id == app_id).execute()
-        except AppConfiguration.DoesNotExist:
-            logger.warning('Can not delete app, not found. id=%e', app_id)
-            return False
+        del self._state[app_id]
         delete_app_from_dir(self._app_dir, self._apps[app_id])
         del self._apps[app_id]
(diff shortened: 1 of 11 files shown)


tests/golem/network/concent/test_received_handler.py
```diff
@@ -187,6 +187,8 @@ def tearDown(self):
         # Remove registered handlers
         del self.task_server
         gc.collect()
+        testutils.DatabaseFixture.tearDown(self)
+        testutils.TestWithClient.tearDown(self)
```
etam: https://stackoverflow.com/a/30187306

etam: see also `setUp` above

golem/apps/manager.py
```diff
@@ -58,6 +75,18 @@ def app(self, app_id: AppId) -> AppDefinition:
         """ Get an app with given ID (assuming it is registered). """
         return self._apps[app_id]
 
+    def delete(self, app_id: AppId) -> bool:
+        # Delete self._state from the database first
```
mfranciszkiewicz, line 78: Maybe implement `__delitem__` in `AppStates` to preserve cohesion? Then it would be enough to `del self._states[app_id]`

golem/apps/manager.py
```diff
@@ -10,9 +14,22 @@
 class AppManager:
     """ Manager class for applications using Task API. """
 
-    def __init__(self) -> None:
+    def __init__(self, app_dir: Path, save_apps=True) -> None:
```
mfranciszkiewicz, line 15: Is there a case we don't want to save app definitions? I cannot see any other usage of this flag.

mfranciszkiewicz (commented):
App definition loading seems to be on par with the previous version

golem/apps/manager.py
maaktweluit: great idea!
looked into this a bit but did not find `__delitem__`

Now its fixed!

golem/apps/manager.py
maaktweluit: currently only during tests

tests/golem/network/concent/test_received_handler.py
maaktweluit: this also confuses me so much, updated the `tearDown()` to be like `setUp`.

is this what you meant?

etam: Yes. I just had a look and it seems, that to be completely sure it works, in `TempDirFixture` in `setUp` and `tearDown` methods it also needs to call `super().thisMethod()`.

golem/apps/rpc.py
```diff
@@ -0,0 +1,55 @@
...
+
+if typing.TYPE_CHECKING:
+    # pylint:disable=unused-import, ungrouped-imports
+    from golem.client import Client
+
+logger = logging.getLogger(__name__)
+
+
+class ClientAppProvider:
+    def __init__(self, client: 'Client'):
```
mfranciszkiewicz: `client` is not needed, only an `app_manager` object. Not like in case of `ClientProvider`, which does indeed use methods of `client`

maaktweluit: nice catch! updated :)

mfranciszkiewicz (approved):

golem/apps/manager.py
```diff
@@ -10,9 +12,23 @@
...
         self._apps: Dict[AppId, AppDefinition] = {}
         self._state = AppStates()
+        self._app_file_names: Dict[AppId, Path] = dict()
+
+        # Save build in apps, then load apps from path
+        built_in_apps: List[AppId] = []
+        if save_apps:
+            built_in_apps = save_built_in_app_definitions(app_dir)
+        for (app_def_path, app_def) in load_apps_from_dir(app_dir):
```
etam:
```suggestion
        for app_def_path, app_def in load_apps_from_dir(app_dir):
```
It's not an error, though.

etam (approved):
ok

golem/apps/manager.py
maaktweluit: Fixed it, thanks!

Wanted to rename `build_in_apps` to `new_apps` since its more accurate, but did not want to rebuild for only this ;)

PR #5106, opened 2020-02-26
Add deadlines to message queue entries

### LINKED ISSUE #5105: Resource Handshake Error due to outdated Message Queue messages
## Description

Due to the unexpected termination of Golem, old handshake messages in Message Queue hangs and this causes handshake issues for the further handshakes.

**Golem Version**: latest observation is in 0.22, but seen in previous versions as well.

**Golem-Messages version** 3.14.1

**Electron version** (if used): n/a

**OS** [e.g. Windows 10 Pro]: latest observation is on MacOS 10.14.6, seen in different OS'es as well.

**Branch** (if launched from source): b.0.22 and others

**Mainnet/Testnet**: tested on Testnet

**Description of the issue**:

While with clear installation everything is fine, after some time later node starts to struggle with handshake issues. Currently whenever I am going to create a task, unable to get a computation due to resource handshake timeout errors.

**Actual result**:

_What is the observed behavior and/or result in this issue_

## Steps To Reproduce

1. Launch Golem
2. Create a task and terminate the Golem in starting handshake step
3. Re-Launch the Golem
4. Create a task
5. From that moment you will have handshake timeout error in your node whenever you create a task.

## Expected behavior
Handshake should be successful

```
INFO     [golem.task.taskkeeper              ] Adding task 'decb15e8-549e-11ea-af5d-7c67a4b1777f' support=<SupportStatus ok ({})> 
INFO     [golem.task.server.resources        ] Starting resource handshake with '7c67a4b1..25b230ca' 
INFO     [golem.task.server.resources        ] Resource handshake timeout. node=7c67a4b1..25b230ca 
INFO     [golem.task.acl                     ] Banned node. node_id=7c67a4b1..25b230ca, timeout=7200s 
```

## Proposed Solution?
Golem should remove old handshake messages from the message queue before starting new handshake and handshake messages should have an expiration mechanism to be cleared out.

kmazurek:
Adds a deadline field to the the database model QueuedMessage. Extends
message queue logic to discard messages which are past their deadline.
When putting new messages to the queue an optional timeout parameter
can be specified. A message's deadline is derived from this timeout.

This change also adds a timeout for ResourceHandshakeStart and
WantToComputeTask messages.

fixes: #5105

### DIFF
--- a/golem/model.py
+++ b/golem/model.py
@@ -653,6 +653,7 @@ class NetworkMessage(BaseModel):
 def default_msg_deadline() -> datetime.datetime:
     return default_now() + MESSAGE_QUEUE_MAX_AGE
 
+
 class QueuedMessage(BaseModel):
     node = CharField(null=False, index=True)
     msg_version = VersionField(null=False)
--- a/golem/model.py
+++ b/golem/model.py
@@ -32,6 +32,7 @@ import semantic_version
 from golem.core import common
 from golem.core.common import datetime_to_timestamp, default_now
 from golem.core.simpleserializer import DictSerializable
+from golem.core.variables import MESSAGE_QUEUE_MAX_AGE
 from golem.database import GolemSqliteDatabase
 from golem.ranking.helper.trust_const import NEUTRAL_TRUST
 from golem.ranking import ProviderEfficacy
@@ -649,12 +650,18 @@ class NetworkMessage(BaseModel):
         return msg
 
 
+def default_msg_deadline() -> datetime.datetime:
+    return default_now() + MESSAGE_QUEUE_MAX_AGE
+
 class QueuedMessage(BaseModel):
     node = CharField(null=False, index=True)
     msg_version = VersionField(null=False)
     msg_cls = CharField(null=False)
     msg_data = BlobField(null=False)
-    deadline = UTCDateTimeField(null=True)
+    deadline = UTCDateTimeField(
+        null=False,
+        default=default_msg_deadline()
+    )
 
     class Meta:
         database = db
@@ -674,7 +681,7 @@ class QueuedMessage(BaseModel):
             golem_messages.__version__,
         )
         instance.msg_data = golem_messages.dump(msg, None, None)
-        instance.deadline = deadline    # type: ignore
+        instance.deadline = deadline or default_msg_deadline()  # type: ignore
         return instance
 
     def as_message(self) -> message.base.Message:
--- a/golem/model.py
+++ b/golem/model.py
@@ -654,12 +654,17 @@ class QueuedMessage(BaseModel):
     msg_version = VersionField(null=False)
     msg_cls = CharField(null=False)
     msg_data = BlobField(null=False)
+    deadline = UTCDateTimeField(null=True)
 
     class Meta:
         database = db
 
     @classmethod
-    def from_message(cls, node_id: str, msg: message.base.Message):
+    def from_message(
+            cls,
+            node_id: str,
+            msg: message.base.Message,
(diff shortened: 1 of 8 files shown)


golem/model.py
```diff
@@ -654,12 +654,17 @@ class QueuedMessage(BaseModel):
     msg_version = VersionField(null=False)
     msg_cls = CharField(null=False)
     msg_data = BlobField(null=False)
+    deadline = UTCDateTimeField(null=True)
```
jiivan: I wouldn't allow `NULL` values. I'd just set `now() + variables.MESSAGE_QUEUE_MAX_AGE` as default

golem/model.py
```diff
@@ -669,6 +681,7 @@ def from_message(cls, node_id: str, msg: message.base.Message):
             golem_messages.__version__,
         )
         instance.msg_data = golem_messages.dump(msg, None, None)
+        instance.deadline = deadline or default_msg_deadline()  # type: ignore
```
etam, line 685: why `type: ignore`?

golem/network/transport/msg_queue.py
```diff
@@ -90,12 +107,18 @@ def waiting() -> typing.Iterator[str]:
...
-    """Sweep ancient messages"""
+    """Sweep messages"""
     with READ_LOCK:
-        oldest_allowed = datetime.datetime.now() \
+        count = 0
+        count += model.QueuedMessage.delete().where(
+            model.QueuedMessage.deadline <= default_now()
+        ).execute()
+
+        oldest_allowed = default_now() \
```
etam: Avoid possible inconsistency, by calling `default_now` only once.
```suggestion
        now = default_now()
        count = 0
        count += model.QueuedMessage.delete().where(
            model.QueuedMessage.deadline <= now
        ).execute()

        oldest_allowed = now \
```

golem/model.py
kmazurek: mypy complains about incompatible types used in this assignment. The right hand side is either `datetime` or `Union[Any, datetime]` (depending on whether a type hint is used on the `deadline` parameter), while the left hand side is expected to be `UTCDateTimeField`. However, this is what peewee expects (assigning python types to DB mapped fields). I can either drop the type hints on both the `deadline` parameter of `from_message` and on `default_msg_deadline` or I can use `type: ignore` on this line. In my opinion the latter is a better option.

golem/network/transport/msg_queue.py
kmazurek: Good catch, will do.

etam (approved):
ok

etam:
BTW, why is this needed?

golem/network/transport/msg_queue.py
```diff
@@ -90,12 +107,20 @@ def waiting() -> typing.Iterator[str]:
...
+        count = 0
+
+        count += model.QueuedMessage.delete().where(
+            model.QueuedMessage.deadline <= now
+        ).execute()
+
+        oldest_allowed = now \
             - variables.MESSAGE_QUEUE_MAX_AGE
-        count = model.QueuedMessage.delete().where(
+        count += model.QueuedMessage.delete().where(
```
jiivan: Please use only deadline here

kmazurek:
@etam Together with setting the timeout for `ResourceHandshakeStart` it resolves https://github.com/golemfactory/golem/issues/5105. Apart from that it can be used for other messages which have a clearly defined lifetime (e.g. `WantToComputeTask`, I set it to use the deadline from its task header).

jiivan (approved):

PR #5109, opened 2020-02-28
BroadcastProtocol introduction

jiivan:


golem/model.py
```diff
@@ -877,6 +880,105 @@ def __repr__(self):
...
+                b,
+            )
+            self.broadcast_type = self.TYPE(broadcast_type)  # type: ignore
+        except (ValueError, struct.error):
+            from golem.network import broadcast
+            raise broadcast.BroadcastError('Invalid header')
+
+    @classmethod
+    def from_bytes(cls, b: bytes) -> 'Broadcast':
+        # Remember to verify signature
```
maaktweluit: should this be a new issue or was it a note to self?

there does not seem to be any verification here

golem/model.py
```diff
@@ -877,6 +880,105 @@ def __repr__(self):
...
+
+    def sign(self, private_key: bytes) -> None:
+        assert self.signature is None
+        self.signature = cryptography.ecdsa_sign(
+            privkey=private_key,
+            msghash=self.get_hash(),
+        )
+
+    def verify_signature(self, public_key: bytes) -> None:
+        # XXX MultiSig solution? Like two of three?
```
maaktweluit: new issue and paste the number here?

golem/network/broadcast.py
```diff
@@ -0,0 +1,76 @@
...
+        return db.read_all()
+
+
+def prepare_handshake() -> BroadcastList:
+    query = model.Broadcast.select().where(
+        model.Broadcast.broadcast_type == model.Broadcast.TYPE.Version,
+    )
+    bl = BroadcastList()
+    if query.exists():
+        bl.append(query.order_by('-timestamp')[0])
```
maaktweluit, line 59: would be nice to order in the request to the database, or does peewee magic do this for us?

golem/network/broadcast.py
```diff
@@ -0,0 +1,76 @@
...
+    logger.debug('Prepared handshake: %s', bl)
+    return bl
+
+
+@decorators.run_with_db()
+def sweep() -> None:
+    logger.info('Sweeping broadcasts')
```
maaktweluit: maybe one of those can be a debug? think the first one has least value

tests/golem/network/test_broadcast.py
```diff
@@ -0,0 +1,36 @@
...
+    def test_single(self):
+        model.Broadcast.create_and_sign(self.privkey, 1, b'1.3.3.7')
+        broadcast.sweep()
+        self.assertEqual(
+            model.Broadcast.select(peewee.fn.Count()).scalar(),
+            1,
+        )
+
+    @freeze_time("2018-01-01 00:00:00", as_arg=True)
+    def test_two(frozen_time, self):  # noqa pylint: disable=no-self-argument, no-self-use
```
maaktweluit: freezer argument mess, maybe a pytest fixture would make it better

https://pypi.org/project/pytest-freezegun/

maaktweluit (approved):
LGTM!

Small comments, please fix or reply before merge

Approving to get those full tests running

golem/network/broadcast.py
```diff
@@ -0,0 +1,76 @@
...
+logger = logging.getLogger(__name__)
+
+
+class BroadcastError(Exception):
+    pass
+
+
+class BroadcastList(list):
+    @classmethod
+    def from_bytes(cls, b: bytes) -> typing.List[model.Broadcast]:
```
etam: Shouldn't this return a `BroadcastList`?

etam: Or you could just remove this class and have free functions instead.

golem/network/broadcast.py
etam: I think it makes it a part of database request, to order by timestamp.

golem/network/transport/network.py
```diff
@@ -1,9 +1,76 @@
...
 
 
+logger = logging.getLogger(__name__)
+
+
+class GolemStateMachine(transitions.Machine):
+    def add_transition_callback(  # pylint: disable=too-many-arguments
+            self,
+            trigger: str,
+            source,
```
etam:
```suggestion
            source: str,
```

golem/network/transport/network.py
```diff
@@ -1,9 +1,76 @@
...
+import transitions
 from twisted.internet.protocol import Factory, Protocol, connectionDone
 
 from .tcpnetwork_helpers import TCPConnectInfo, TCPListenInfo, TCPListeningInfo
 
 
+logger = logging.getLogger(__name__)
+
+
+class GolemStateMachine(transitions.Machine):
```
etam: This name suggests it's a machine that describes Golem. If it's just our extension to `transitions.Machine`, I'd call it `Machine` or `StateMachine`.

golem/network/transport/network.py
```diff
@@ -19,81 +86,84 @@ def stop_listening(self, listening_info: TCPListeningInfo):
...
-    # Protocol function
     def connectionMade(self):
+        super().connectionMade()
+        # map twisted Protocol event into transition
+        self.connectionMadeTransition()  # pylint: disable=no-member
+
+    def connectionLost(self, reason=connectionDone):
+        super().connectionLost(reason=reason)
+        # map twisted Protocol event into transition
+        self.connectionLostTransition(reason=reason)  # noqa pylint: disable=no-member
```
etam, line 172: Shouldn't this be `self.machine.connectionLostTransition`? (If so, the `Transition` suffix is superfluous)

golem/model.py
jiivan: It's a note to a future developer using this function. I'll rephrase it.

golem/network/broadcast.py
jiivan: _This is the way_ of using sql `ORDER BY` in peewee.

golem/network/broadcast.py
jiivan: I made it the same way as in msg_queue
https://github.com/golemfactory/golem/blob/51fb79c9ae61acb4f6920cdedc4b5b83a1ccd685/golem/network/transport/msg_queue.py#L107-L116

IMHO it's ok, because it's triggered once a day.

jiivan: Removed it

tests/golem/network/test_broadcast.py
etam: or use it as context manager https://github.com/spulec/freezegun#context-manager

golem/network/transport/network.py
```diff
@@ -19,81 +86,84 @@ def stop_listening(self, listening_info: TCPListeningInfo):
...
+                'disconnected',
+            ],
+            initial='initial',
+            auto_transitions=False,
+        )
+        self.machine.add_transition(
+            'connectionMadeTransition',
+            'initial',
+            'connected',
+            after=self.create_session or True,  # always True
```
etam:
```suggestion
            after=self.create_session
```

golem/network/transport/network.py
```diff
@@ -19,81 +86,84 @@ def stop_listening(self, listening_info: TCPListeningInfo):
...
+        )
+        self.machine.add_transition(
+            'connectionLostTransition',
+            '*',
+            'disconnected',
+            after=lambda reason: (
+                delattr(self, 'session') or True  # always True
+            ),
```
etam: Why is `or True` necessary here? And `reason` argument?
```suggestion
            after=lambda: delattr(self, 'session'),
```

golem/network/transport/tcpnetwork.py
```diff
@@ -269,11 +273,18 @@ class BasicProtocol(SessionProtocol):
...
+        super().__init__(session_factory)
         self.db = DataBuffer()
         self.spam_protector = SpamProtector()
+        self.machine.add_transition_callback(
+            'connectionLostTransition', 'connected', 'disconnected',
+            'before',
+            lambda reason: (
+                (self.session.dropped() if self.session else True)
+                or True  # always True
+            ),
```
etam: State machine should guarantee you that in `connected` state `self.session` exists.
```suggestion
            lambda: self.session.dropped(),
```

golem/network/transport/network.py
jiivan: or True is unnecessary - removed.
Reason is passed from original [connectionLost](https://twistedmatrix.com/documents/current/api/twisted.internet.tcp._BaseBaseClient.html#connectionLost)

golem/network/transport/network.py
```diff
@@ -19,81 +86,84 @@ def stop_listening(self, listening_info: TCPListeningInfo):
...
+        super().connectionMade()
+        # map twisted Protocol event into transition
+        self.connectionMadeTransition()  # pylint: disable=no-member
+
+    def connectionLost(self, reason=connectionDone):
+        super().connectionLost(reason=reason)
+        # map twisted Protocol event into transition
+        self.connectionLostTransition(reason=reason)  # noqa pylint: disable=no-member
+
+    def create_session(self) -> bool:
```
etam:
```suggestion
    def create_session(self) -> None:
```

golem/network/transport/tcpnetwork.py
```diff
@@ -269,11 +273,15 @@ class BasicProtocol(SessionProtocol):
...
-        super().__init__()
-        self.opened = False
+    def __init__(self, session_factory, **_kwargs):
+        super().__init__(session_factory)
         self.db = DataBuffer()
         self.spam_protector = SpamProtector()
+        self.machine.add_transition_callback(
+            'connectionLostTransition', 'connected', 'disconnected',
+            'before',
+            lambda reason: self.session.dropped(),
```
etam:
```suggestion
            self.session.dropped,
```

golem/network/transport/network.py
```diff
@@ -19,81 +86,84 @@ def stop_listening(self, listening_info: TCPListeningInfo):
...
+        )
+        self.machine.add_transition(
+            'connectionMadeTransition',
+            'initial',
+            'connected',
+            after=self.create_session,
+        )
+        self.machine.add_transition(
+            'connectionLostTransition',
+            '*',
```
etam, line 149: this should go only from `connected`, because `initial` has no session.

golem/network/broadcast.py
maaktweluit: _This is the way_

Since it called `exists()` first i was thinking it would have already run the SELECT too.. TIL

golem/network/transport/tcpnetwork.py
jiivan: Who will eat reason argument then? :)

etam:
FYI this are the current state machines:
SessionProtocol:
0_SessionProtocol
BasicProtocol:
1_BasicProtocol
ServerProtocol:
2_ServerProtocol
BroadcastProtocol:
3_BroadcastProtocol

golem/network/transport/tcpnetwork.py
```diff
@@ -269,11 +284,15 @@ class BasicProtocol(SessionProtocol):
...
```
etam: Why is this here? I think it should be in SessionProtocol

etam:
I've cleaned up them a bit. What you think?

SessionProtocol:
0_SessionProtocol
ServerProtocol:
2_ServerProtocol
BroadcastProtocol:
3_BroadcastProtocol

EDIT: jiivan is right, that from `initial` and `handshaking` states must be a transition to `disconnected`, that has no callbacks.

etam (approved):
ok

In the next step we could add an integration test.

kmazurek:
@jiivan Looking through the logs of the last test run I noticed that `transitions` logs seem to be going to `INFO` level by default, e.g.:
```
INFO [transitions.core                   ] Exited state initial
INFO [transitions.core                   ] Entered state handshaking
INFO [transitions.core                   ] Exited state handshaking
INFO [transitions.core                   ] Entered state connected
```
They seem to be unnecessary for end users and will also attribute to log spam as multiple lines will be printed for each connection.

PR #5113, opened 2020-03-03
fix OSX default temporary directories when running the integration tests

shadeofblue:
* task api integration test
* node integration tests


scripts/tempdir.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+from golem.core.common import is_osx
+
+
+def fix_osx_tmpdir() -> bool:
+    if not is_osx():
+        return False
+    tmpdir = os.environ.get('TMPDIR') or ''
+    if not (tmpdir.startswith('/private') or tmpdir.startswith('/tmp')):
```
maaktweluit: i think its better to only set `tmpdir` when not specified by the user, i usually run it in `/Users/me/tmp` to have the logs in a central location

```suggestion
    if not tmpdir:
```

alternatively we can rename the function to `force_tmp_dir()` then we should also allow the other 2 default docker directories, `/Users` and `/Volumes` ( not sure if that last one is required )
In this case its nice to show a warning to the user that its TMPDIR env variable is wrong and has been overridden

Sorry for the stream of thoughts, pick your best, or ignore ;)

maaktweluit (changes requested):
LGTM!

One q, will approve on reply or fix

scripts/tempdir.py
shadeofblue: @maaktweluit updated

maaktweluit (approved):

scripts/tempdir.py
maaktweluit: nice!

PR #5114, opened 2020-03-03
Load app definition files from Golem releases CDN

### LINKED ISSUE #5099: Load remote app_definition files
As a golem factory we want to add app_definition.json files to the releases CDN so we do not need a golem release to add a new app

- Every 24h ( time is changeable ) check the release CDN and download new app definitions.
- Send out event when new app definition is found, encouraging the user to enable it.

kmazurek:
Resolves: https://github.com/golemfactory/golem/issues/5099

Adds logic for downloading app definition JSON files from CDN in Task API `AppManager`. The definition files are distinguished based on their file name (which, among other things, should contain the hash of the file's contents). Updating apps is done upon `AppManager` init and then periodically (once a day using `DailyJobsService`). For every new definition downloaded an RPC event is fired (`evt.apps.new_definition`).

### DIFF
--- a/golem/core/variables.py
+++ b/golem/core/variables.py
@@ -75,7 +75,7 @@ MESSAGE_QUEUE_MAX_AGE = dateutil.relativedelta.relativedelta(months=6)
 # How long should peer be banned when failing on resources (seconds)
 ACL_BLOCK_TIMEOUT_RESOURCE = 2 * 3600  # s
 
-APP_DEFINITIONS_CDN_URL = 'https://golem-app-definitions.s3.eu-central-1.amazonaws.com/'    # noqa pylint: disable=line-too-long
+APP_DEFINITIONS_CDN_URL = 'https://golem-app-definitions.cdn.golem.network'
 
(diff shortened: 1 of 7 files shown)


maaktweluit (approved):
LGTM!

golem/core/variables.py
```diff
@@ -75,6 +75,8 @@
 # How long should peer be banned when failing on resources (seconds)
 ACL_BLOCK_TIMEOUT_RESOURCE = 2 * 3600  # s
 
+APP_DEFINITIONS_CDN_URL = 'https://golem-app-definitions.s3.eu-central-1.amazonaws.com/'    # noqa pylint: disable=line-too-long
```
shadeofblue: shouldn't that rather be some subdomain in `golem.network` that would redirect to s3? ...

kmazurek: @etam ^

shadeofblue: or maybe @ktomala ?

PR #5115, opened 2020-03-03
Fix sorting order in surge detector hottest_args

kmazurek:
The surge_detector decorator includes reporting of most used arguments
for a given surged call. These arguments were being sorted in ascending
order based on the number of calls, resulting in the least used argument
being incorrectly reported. This changes the sorting order to
descending.


jiivan (approved):
LOL, nice catch

PR #5116, opened 2020-03-04
Use environment Docker tags in tests

kmazurek:
Replaces hardcoded Docker tags in job tests with tags taken from their respective `Environment` classes.


jiivan (approved):
who_would_believe_phasma

PR #5117, opened 2020-03-05
Decrease transitions loglevel

jiivan:


etam:
Do you have an example of before and after?

kmazurek (approved):
go_go_go

jiivan:
> Do you have an example of before and after?

Before:
```
INFO [transitions.core                   ] Exited state initial
INFO [transitions.core                   ] Entered state handshaking
INFO [transitions.core                   ] Exited state handshaking
INFO [transitions.core                   ] Entered state connected
```
After:
```
```

etam:
> Do you have an example of before and after?

I have. Here it is: https://github.com/golemfactory/golem/pull/5109#issuecomment-595179599

etam (approved):
ok

Issue #5122, opened 2020-03-06
in after_disconnection      builtins.AttributeError: 'NoneType' object has no attribute 'dropped'


ederenn:
## Description

**Golem Version**:  GOLEM Version: 0.22.1+dev102.g85a1283 

**Golem-Messages version** (leave empty if unsure):3.15.0

**Electron version** (if used):

**OS** [e.g. Windows 10 Pro]: Ubuntu 18.04, Windows

**Branch** (if launched from source):

**Mainnet/Testnet**: mainnet

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

After starting the app error appears in logs

```
2020-03-06 14:12:51 CRITICAL twisted                             Unhandled Error
Traceback (most recent call last):
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/internet/asyncioreactor.py", line 267, in run
    
  File "/usr/lib/python3.6/asyncio/base_events.py", line 427, in run_forever
    raise RuntimeError(
  File "/usr/lib/python3.6/asyncio/base_events.py", line 1440, in _run_once
    try:
  File "/usr/lib/python3.6/asyncio/events.py", line 145, in _run
    self._callback(*self._args)
--- <exception caught here> ---
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/python/log.py", line 103, in callWithLogger
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/python/log.py", line 86, in callWithContext
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/python/context.py", line 122, in callWithContext
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/python/context.py", line 85, in callWithContext
    
  File "/home/buildbot-worker/worker/buildpackage_linux/build/.venv/lib/python3.6/site-packages/twisted/internet/asyncioreactor.py", line 141, in _readOrWrite
    
[...]

PR #5123, opened 2020-03-06
Handle connection lost during handshake

jiivan:
fixes: #5122


etam:
Does it mean, that `copy_transitions` doesn't do, what it's supposed to do?

jiivan:
> Does it mean, that `copy_transitions` doesn't do, what it's supposed to do?

To the contrary. It copied disconnect handler from connected state (which has session) to handshake state which doesn't.

etam (approved):
ok

PR #5127, opened 2020-03-11
add an integration test that runs two simultanous tasks

shadeofblue:
+ fix execution of simultaneous tasks in Golem


scripts/node_integration_tests/playbooks/base.py
```diff
@@ -304,9 +304,17 @@ def on_success(result):
...
 
-    def step_create_task(self, node_id: NodeId = NodeId.requestor):
-        print("Output path: {}".format(self.output_path))
-        print("Task dict: {}".format(self.config.task_dict))
+    def step_create_task(
+            self, node_id: NodeId = NodeId.requestor,
+            output_path: str = None, task_dict: dict = None
+    ):
+        if not output_path:
+            output_path = self.output_path
```
etam: it seems unused, apart from logging

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -304,9 +304,17 @@ def on_success(result):
 
         return self.call(node_id, 'comp.tasks', on_success=on_success)
 
-    def step_create_task(self, node_id: NodeId = NodeId.requestor):
-        print("Output path: {}".format(self.output_path))
-        print("Task dict: {}".format(self.config.task_dict))
+    def step_create_task(
+            self, node_id: NodeId = NodeId.requestor,
```
maaktweluit: 4 spaces too much?

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -328,21 +336,31 @@ def on_success(result):
...
                 node_id,
                 self.RPC_TASK_CREATE,
-                self.config.task_dict,
+                task_dict,
                 on_success=on_success
             )
 
+    @staticmethod
+    def _identify_new_task_id(
+            tasks: set, known_tasks: set) -> typing.Optional[str]:
```
maaktweluit: according to the style guide when breaking the signature each parameter should have its own line, like above

scripts/node_integration_tests/playbooks/golem/simultaneous_tasks/playbook.py
```diff
@@ -0,0 +1,58 @@
...
+from ...test_config_base import NodeId
+
+if typing.TYPE_CHECKING:
+    from ...test_config_base import TestConfigBase
+
+
+class Playbook(NodeTestPlaybook):
+    def __init__(self, config: 'TestConfigBase') -> None:
+        super().__init__(config)
+        self.output_path2 = tempfile.mkdtemp(
```
maaktweluit: would be nicer to make these lists instead

maaktweluit (commented):
LGTM!

Some small nitty comments, will approve on reply :)

etam:
Is it needed to create a second output path? Isn't golem creating a per-task subdirectory inside it?

scripts/node_integration_tests/playbooks/golem/simultaneous_tasks/playbook.py
etam: and instead of copying playbook methods, add an index argument, which could be 0 by default.

shadeofblue: yeah, but that would require us to rewrite all the other functionality that already uses those which I wanted to avoid

scripts/node_integration_tests/playbooks/base.py
shadeofblue: gotcha

scripts/node_integration_tests/playbooks/base.py
shadeofblue: how? normal indent is 4... this is a parameter indent which is normally farther than the default

shadeofblue: > Is it needed to create a second output path? Isn't golem creating a per-task subdirectory inside it?

maybe you're right, I'll have a look

maaktweluit: you are right, it is correct like this! sorry, brainfart ^^

scripts/node_integration_tests/playbooks/golem/simultaneous_tasks/playbook.py
maaktweluit: maybe we can make a task pointing to this comment, nice to have but no priority now

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -348,21 +358,34 @@ def on_success(result):
...
+    def _identify_new_task_id(
+            tasks: set,
+            known_tasks: set
+    ) -> typing.Optional[str]:
+        new_tasks = tasks - known_tasks
+        if len(new_tasks) != 1:
+            print("Cannot find the new task ({})".format(new_tasks))
+            time.sleep(3)
+            return None
+        else:
```
etam, line 372: minor: this else is unnecessary

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -324,9 +324,19 @@ def on_success(result):
...
         )
 
-    def step_create_task(self, node_id: NodeId = NodeId.requestor):
-        print("Output path: {}".format(self.output_path))
-        print("Task dict: {}".format(self.config.task_dict))
+    def step_create_task(
+            self,
+            node_id: NodeId = NodeId.requestor,
+            output_path: typing.Optional[str] = None,
+            task_dict: typing.Optional[dict] = None,
```
etam: are those used now?

scripts/node_integration_tests/playbooks/golem/simultaneous_tasks/playbook.py
```diff
@@ -0,0 +1,41 @@
...
+                set(map(lambda r: r['id'], result)),
+                self.known_tasks.union({self.task_id})
+            )
+            if task_id:
+                self.task_id2 = task_id
+                self.next()
+
+        return self.call(node_id, 'comp.tasks', on_success=on_success)
+
+    def step_wait_task_finished2(self, node_id: NodeId = NodeId.requestor):
```
etam: optional: this could be replaced by wrapping `step_wait_task_finished` in `partial`

scripts/node_integration_tests/playbooks/base.py
shadeofblue: yeah, though, more explicit

scripts/node_integration_tests/playbooks/base.py
shadeofblue: no, I'll remove `output_path` since indeed it doesn't make too much sense but I'll leave `task_dict` as it seems reasonable to make it possible to run different tasks during a single test run

golem/task/taskmanager.py
```diff
@@ -1023,7 +1023,16 @@ def restart_subtask(
             new_status: SubtaskStatus = SubtaskStatus.restarted,
```
etam, line 1023: You add parameter `new_state` to other `restart_subtask` functions and here it's called `new_status`. I'd suggest keeping consistency.

shadeofblue: I don't want to change the existing calls to _this_ function ... it is a preexisting code

etam: I'm not saying you have to change this. You can change variables you're adding. I just had to place my comment somewhere.

etam (approved):
ok

golem/task/taskcomputer.py
```diff
@@ -747,6 +747,8 @@ def task_given(
                 cpu_limit=cpu_time_limit))
 
     def has_assigned_task(self) -> bool:
+        logger.debug(
+            f"Has assigned task? assigned_subtasks={self.assigned_subtasks}")
```
maaktweluit: better for performance to use native logger formatting
```suggestion
            "Has assigned task? assigned_subtasks=%r", self. assigned_subtasks)

```

scripts/node_integration_tests/playbooks/base.py
```diff
@@ -324,9 +324,16 @@ def on_success(result):
...
+            self,
+            node_id: NodeId = NodeId.requestor,
+            task_dict: typing.Optional[dict] = None,
+    ):
+        if not task_dict:
+            task_dict = self.config.task_dict
+
         print("Output path: {}".format(self.output_path))
-        print("Task dict: {}".format(self.config.task_dict))
+        print("Task dict: {}".format(task_dict))
```
maaktweluit, line 336: format strings are nicer, but can leave it for now :)

maaktweluit (approved):
LGTM!

2 nitty comments

golem/task/taskcomputer.py
shadeofblue: ah, it's not even about performance, afaik, but rather e.g. analysis of logs with tools like sentry ... will include that commit once the tests pass so as not to intefere with it now

etam: It's also about performance. If you format your string manually, it's always done, regardless of log level. With formatting in logger, it can avoid doing unnecessary work.

PR #5131, opened 2020-03-12
add the default `cube` scene to alleviate issues with the JPEG integration tests

shadeofblue:
(the scene's simpler lighting and less abrupt transitions should make the compression artifacts less prominent and the failures thus rarer)


maaktweluit (approved):
LGTM!

Maybe we can replace `cubes` with `cube` as default if it gives more consistent results

jiivan (approved):

scripts/node_integration_tests/tests/test_golem.py
```diff
@@ -56,7 +56,10 @@ def test_jpeg(self):
         verifies if Golem - when supplied with `JPEG` as the format - will
         render the output as JPEG with the proper extension.
         """
-        self._run_test('golem.jpeg')
+        self._run_test(
+            'golem.regular_run_stop_on_reject',
```
etam: you changed the test here, so you're no longer testing jpeg

shadeofblue:
> LGTM!
> 
> Maybe we can replace `cubes` with `cube` as default if it gives more consistent results

I wouldn't do that - especially that other formats don't have this problem... 

(we don't want to test everything on a default scene)

scripts/node_integration_tests/tests/test_golem.py
shadeofblue: [ facepalm ]

scripts/node_integration_tests/tests/test_golem.py
```diff
@@ -56,7 +56,10 @@ def test_jpeg(self):
         verifies if Golem - when supplied with `JPEG` as the format - will
         render the output as JPEG with the proper extension.
         """
-        self._run_test('golem.jpeg')
+        self._run_test(
+            'golem.jpg',
```
etam: No, it's `jpeg`. The `e` in the middle makes a difference.

etam: `jpg` is a different test

etam (approved):
ok

scripts/node_integration_tests/tests/test_golem.py
maaktweluit: `jpeg` is the correct output type, rendering into `.jpg` extension
https://github.com/golemfactory/golem/blob/develop/apps/blender/resources/images/entrypoints/scripts/verifier_tools/file_extension/types.py#L53

while `jpg` is an unsupported output type thus defaulting to PNG
https://github.com/golemfactory/golem/blob/develop/apps/blender/task/blenderrendertask.py#L247

maybe `jpg` should be changed to `abc` to be more clear.

PR #5133, opened 2020-03-13
Detect protocol

jiivan:
* Add protocol marker to detect old clients early
* Assure all received data is processed after transitions

### DIFF
--- a/golem/network/transport/network.py
+++ b/golem/network/transport/network.py
@@ -167,6 +167,15 @@ class SessionProtocol(Protocol):
             after_disconnection,
         )
 
+    def __str__(self):
+        peer = self.transport.getPeer() if self.transport else '<no_transport>'
+        # pylint: disable=no-member
+        return (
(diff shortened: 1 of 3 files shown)


maaktweluit (approved):
LGTM!

etam:
I have an idea on how to avoid adding `dataReceived(b'')`.

This is more or less a pseudocode:
```python
def dataReceived(self, data: bytes) -> None:
    self.db.append_bytes(data)
    while self.processData():
        pass

def processData(self) -> Bool:
    """ process a single chunk of data
    :return: True if a chunk was successfully consumed
    """"
    if self.isInSomeState():
        if not haveEnoughDataInBufffer(self.db):
            logger.debug("waiting for more data")
            return False
        data = self.db.read_some_bytes()
        self.doSomething(data)  # may change state
        return True
    if self.isInSomeOtherState():
        self.youGetTheIdea()
        return True
```

jiivan:
@etam it might work, but I had to rebuild tests five times, so I'm not that enthusiastic about it :)

PR #5136, opened 2020-03-17
Use subtask timeout status in taskmanager

kmazurek:
Part of: https://github.com/golemfactory/golem/pull/5136
Updates the legacy taskmanager to use `SubtaskStatus.timeout` instead of `failure` for subtasks which have timed out.


jiivan (approved):

maaktweluit (approved):
LGTM!

Issue #5012, opened 2019-12-19
When task creation fails due to insufficient funds, task stays in 'Creating' state


kmazurek:
## Description

**Golem Version**: 0.21.0+dev408.g4a05d87

**Golem-Messages version** (leave empty if unsure): 3.14.1

**Electron version** (if used):

**OS** [e.g. Windows 10 Pro]: Ubuntu 18.04.3 LTS

**Branch** (if launched from source): b0.22

**Mainnet/Testnet**: both mainnet and testnet

**Priority label is set to the lowest by default. To setup higher priority please change the label**
_P0 label is set for Severity-Critical/Effort-easy
P1 label is set  for Severity-Critical/Effort-hard
P2 label is set for Severity-Low/ Effort-easy
P3 label is set for Severity-Low/Effort-hard_

**Description of the issue**:

When attempting to create a new task with not enough funds available, the task stays in `Creating` state.

## Steps To Reproduce
1. Launch the Golem app.
2. Try creating a new task for which there is not enough funds available (e.g. set the `bid` field high enough). This was observed when creating a task from the CLI (command `tasks create <path_to_your_task_definition>).
3. Wait for the task creation to fail (i.e. for insufficient funds error to be reported).
4. Check the status of the requested tasks (e.g. call `tasks show`). The status for the attempted task will be reported as `Creating`.

## Expected behavior
An error status should be reported, e.g. `Error creating`.

kmazurek:
One additional side effect is that this effectively blocks graceful shutdown, since the task stuck in `Creating` is considered to be in progress.

etam:
fix merged to b0.23

PR #5137, opened 2020-03-17
Set correct state when creating task with no funds

kmazurek:
Resolves: https://github.com/golemfactory/golem/issues/5012

When task creation fails due to insufficient funds, the task's state is
now set to errorCreating. This commit also changes that state to be
considered final.

### DIFF
--- a/golem/task/taskstate.py
+++ b/golem/task/taskstate.py
@@ -197,7 +197,7 @@ TASK_STATUS_COMPLETED = [
     TaskStatus.aborted,
     TaskStatus.timeout,
     TaskStatus.restarted,
-    TaskStatus.errorCreating
+    TaskStatus.errorCreating,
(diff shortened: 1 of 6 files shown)


golem/task/taskstate.py
```diff
@@ -196,7 +196,8 @@ def is_active(self) -> bool:
     TaskStatus.finished,
     TaskStatus.aborted,
     TaskStatus.timeout,
-    TaskStatus.restarted
+    TaskStatus.restarted,
+    TaskStatus.errorCreating
```
jiivan:
```suggestion
    TaskStatus.errorCreating,

```

jiivan (approved):
lgtm

maaktweluit (approved):
LGTM!

PR #5138, opened 2020-03-19
Enable task-api environments on mainnet

maaktweluit:
Was discussed to enable it but the config was never updated.


etam (approved):
The change in code is OK, but it's more of a [layer 8](https://en.wikipedia.org/wiki/Layer_8) issue.
@shadeofblue ?

badb (approved):
YES.

PR #5140, opened 2020-03-20
Moved slogger fix for ethereum package to SCI

maaktweluit:
this hack was called too late, disabling unfortunate loggers in our project.

Moved it before the first import of `ethereum` inside golem_sci


kmazurek (approved):
LGTM

PR #5150, opened 2020-03-27
add node name from nodeskeeper to subtask

maaktweluit:
fixes the issue:

```
task-api subtasks had no node_name in some CLI commands
```


etam (approved):
ok

PR #5163, opened 2020-04-15
show Task API tasks in provider stats

etam:
Progress is always 0, but at least it's something.

I need it for an upcoming integration test, where I need to know when provider is computing a Task API task.


maaktweluit (approved):
LGTM!

Wiezzel (approved):

PR #5166, opened 2020-04-22
fix wasm tasks `restart_subtask` signature

shadeofblue:


apps/wasm/task.py
```diff
@@ -539,7 +539,11 @@ def _load_requestor_perf(self):
         self.REQUESTOR_MARKET_STRATEGY.set_my_usage_benchmark(
             cpu_usage * NANOSECOND)
 
-    def restart_subtask(self, subtask_id: str):
+    def restart_subtask(
+            self,
+            subtask_id,
+            new_state: Optional[SubtaskStatus] = None,
+    ):
```
jiivan, line 546: Where is `new_state` used?

shadeofblue: it's not - but it's required by the method's signature

jiivan (approved):

PR #5171, opened 2020-04-24
Fix config cannot be updated

### LINKED ISSUE #5151: ValueError: Config can be updated only when the environment is disabled
## Description

**Golem Version**: 0.22.1+dev124.gf5a985e

**Golem-Messages version** (leave empty if unsure): 3.15.0

**Electron version** (if used):

**OS** [e.g. Windows 10 Pro]: Ubuntu 18.04

**Branch** (if launched from source): b0.23

**Mainnet/Testnet**: mainnet

**Description of the issue**:
After pausing golem thera are problems with switching on/off requesting only. In logs appears an error
```ERROR    [twisted                            ] Session.onUserError(): "ValueError: Config can be updated only when the environment is disabled"
  File "/home/ederenn/projects/golem-env/lib/python3.6/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/home/ederenn/projects/golem/golem/task/taskserver.py", line 866, in _change_task_computer_config
    config_changed = yield self.task_computer.change_config(config_desc)
  File "/home/ederenn/projects/golem-env/lib/python3.6/site-packages/twisted/internet/defer.py", line 1613, in unwindGenerator
    return _cancellableInlineCallbacks(gen)
  File "/home/ederenn/projects/golem-env/lib/python3.6/site-packages/twisted/internet/defer.py", line 1529, in _cancellableInlineCallbacks
    _inlineCallbacks(None, g, status)
--- <exception caught here> ---
  File "/home/ederenn/projects/golem/golem/client.py", line 1525, in activate_hw_preset
    result = yield deferred
  File "/home/ederenn/projects/golem/golem/task/taskserver.py", line 858, in change_config
    yield self._change_task_computer_config(config_desc, run_benchmarks)
  File "/home/ederenn/projects/golem/golem/task/taskserver.py", line 866, in _change_task_computer_config
    config_changed = yield self.task_computer.change_config(config_desc)
  File "/home/ederenn/projects/golem-env/lib/python3.6/site-packages/twisted/internet/defer.py", line 1418, in _inlineCallbacks
    result = g.send(result)
  File "/home/ederenn/projects/golem/golem/task/taskcomputer.py", line 263, in change_config
    work_dir=work_dir)
  File "/home/ederenn/projects/golem/golem/task/taskcomputer.py", line 478, in change_config
    docker_cpu.update_config(DockerCPUConfig(**config_dict))
  File "/home/ederenn/projects/golem/golem/envs/wrappers/__init__.py", line 120, in update_config
    self._env.update_config(config)
  File "/home/ederenn/projects/golem/golem/envs/wrappers/__init__.py", line 120, in update_config
    self._env.update_config(config)
  File "/home/ederenn/projects/golem/golem/envs/docker/cpu.py", line 690, in update_config
    "Config can be updated only when the environment is disabled")
builtins.ValueError: Config can be updated only when the environment is disabled
 ```

**Actual result**:

In interface it seems like it is not possible to resume Golem

## Steps To Reproduce
_Short description of steps to reproduce the behavior:_
e.g.
1. Launch the app 
2. Click on Stop Golem button
3. Switch on "I want to act only as a requestor "
4. Click button Start Golem
5. See error in logs

## Proposed Solution?
_(Optional: What could be a solution for that issue)_

### PR DESCRIPTION
**Pull request by etam**
Fixes #5151

### DIFF
--- a/golem/envs/__init__.py
+++ b/golem/envs/__init__.py
@@ -3,7 +3,7 @@ from abc import ABC, abstractmethod
 from copy import deepcopy
 from enum import Enum
 from logging import Logger, getLogger
-from threading import RLock
+from threading import Lock, RLock
 
 from typing import Any, Callable, Dict, List, Optional, NamedTuple, Union, \
     Sequence, Iterable, ContextManager, Set, Tuple, Type, TYPE_CHECKING
@@ -605,6 +605,15 @@ def delayed_config(cls: Type[Environment]) -> Type[Environment]:
     """
     This class decorator allows to save config update and apply it, when env is
     disabled.
+
+    Mutex prevents the following scenario
+    Thread 1                Thread 2
+    call apply_next_config
+                            call update_config
+                            status is disabled, so calls
+                                super().update_config
+    cls.update_config
+        with _next_config
     """
     # FIXME workaround https://github.com/python/mypy/issues/5865
     cls2: Any = cls
@@ -613,24 +622,27 @@ def delayed_config(cls: Type[Environment]) -> Type[Environment]:
         def __init__(self, *args, **kwargs) -> None:
             super().__init__(*args, **kwargs)
             self._next_config: Optional[EnvConfig] = None
+            self._config_lock = Lock()
 
             def apply_next_config(_):
-                if self._next_config is None:
-                    return
-                self._logger.debug("Applying saved config")
-                config = self._next_config
-                self._next_config = None
-                cls.update_config(self, config)
+                with self._config_lock:
+                    if self._next_config is None:
+                        return
+                    self._logger.debug("Applying saved config")
+                    config = self._next_config
+                    self._next_config = None
+                    cls.update_config(self, config)
 
             self.listen(EnvEventType.DISABLED, apply_next_config)
 
         def update_config(self, new_config: EnvConfig) -> None:
-            if self._status == EnvStatus.DISABLED:
-                self._logger.debug("Config applied immediately")
-                super().update_config(new_config)
-                return
+            with self._config_lock:
+                if self._status == EnvStatus.DISABLED:
+                    self._logger.debug("Config applied immediately")
+                    super().update_config(new_config)
+                    return
 
-            self._logger.debug("Config saved for later")
-            self._next_config = new_config
+                self._logger.debug("Config saved for later")
+                self._next_config = new_config
 
     return DelayedConfigWrapper
--- a/golem/envs/__init__.py
+++ b/golem/envs/__init__.py
@@ -6,7 +6,7 @@ from logging import Logger, getLogger
 from threading import RLock
 
 from typing import Any, Callable, Dict, List, Optional, NamedTuple, Union, \
-    Sequence, Iterable, ContextManager, Set, Tuple, TYPE_CHECKING
+    Sequence, Iterable, ContextManager, Set, Tuple, Type, TYPE_CHECKING
 
 from dataclasses import dataclass, field
 from twisted.internet.threads import deferToThread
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Review by Wiezzel — COMMENTED** [review]
I have worries about potential race conditions occurring when the config is updated at the very moment when environment gets enabled/disabled. Environment state could change between status check in `update_config` and calling `super().update_config` or assigning `_next_config`.

**Comment by Wiezzel** [comment]
And you're gonna have a hard time silencing mypy when dealing with class decorators...

**Comment by etam** [comment]
> I have worries about potential race conditions occurring when the config is updated at the very moment when environment gets enabled/disabled. Environment state could change between status check in `update_config` and calling `super().update_config` or assigning `_next_config`.

How about now?

> And you're gonna have a hard time silencing mypy when dealing with class decorators...

It wasn't that bad :)

File: golem/envs/__init__.py [hunk 1]
```diff
@@ -599,3 +599,50 @@ def listen(
             listener: EnvEventListener
     ) -> None:
         self._event_listeners.setdefault(event_type, set()).add(listener)
+
+
+def delayed_config(cls: Type[EnvironmentBase]) -> Type[EnvironmentBase]:
```
Wiezzel: Is `EvironmentBase` necessary here? Wouldn't `Environment` suffice?

Wiezzel (commented):
The lock seems to do the job.
One more small comment.

scripts/node_integration_tests/playbooks/golem/reconfigure_provider_while_working/playbook.py
```diff
@@ -0,0 +1,75 @@
...
+    steps = BasePlaybook.initial_steps + (
+        BasePlaybook.step_enable_app,
+        BasePlaybook.step_create_task,
+        BasePlaybook.step_get_task_id,
+        BasePlaybook.step_get_task_status,
+        wait_for_computing_task,
+        partial(ui_stop, node_id=NodeId.provider),
+        partial(change_config, node_id=NodeId.provider),
+        partial(check_if_test_failed, node_id=NodeId.provider),
+        # No need to wait for subtask to finish, because it's a waste of time
```
shadeofblue: I would remove the comment here ... there are a few other playbooks that don't wait for the task to finish ...

scripts/node_integration_tests/playbooks/golem/reconfigure_provider_while_working/playbook.py
```diff
@@ -0,0 +1,75 @@
...
+
+    steps = BasePlaybook.initial_steps + (
+        BasePlaybook.step_enable_app,
+        BasePlaybook.step_create_task,
+        BasePlaybook.step_get_task_id,
+        BasePlaybook.step_get_task_status,
+        wait_for_computing_task,
+        partial(ui_stop, node_id=NodeId.provider),
+        partial(change_config, node_id=NodeId.provider),
+        partial(check_if_test_failed, node_id=NodeId.provider),
```
shadeofblue, line 72: shouldn't we somehow detect that the configuration has been completed? (instead of waiting the 10 seconds)?

shadeofblue (approved):
looks good as far as I could tell, added a couple of comments

golem/task/taskcomputer.py
```diff
@@ -495,7 +489,7 @@ def change_config(
             )
         )
 
-        # FIXME: Decide how to properly configure environments
+        # FIXME: Decide how to properly configure environments - is it still relevant?
```
etam: @Wiezzel ?

Wiezzel: There was this idea that "new" environments should be configured separately, not depending on the global config descriptor at all. But I believe this will never happen if we're not developing Clay much further.

golem/envs/__init__.py
etam: No. Yes.

scripts/node_integration_tests/playbooks/golem/reconfigure_provider_while_working/playbook.py
etam: I don't know if we should, but in the current "fire and forget" implementation model, it's not easy to achieve. I tried.

scripts/node_integration_tests/playbooks/golem/reconfigure_provider_while_working/playbook.py
etam: done

Wiezzel (approved):

PR #5175, opened 2020-04-29
fix 'pip install' with latest setuptools

etam:


shadeofblue (approved):
as long as the tests pass ...

PR #5186, opened 2020-05-18
fix a desync between taskkeeper's `supported_tasks` and `task_headers`

shadeofblue:


golem/task/taskkeeper.py
```diff
@@ -677,6 +677,7 @@ def get_task(
             if supported_tasks is not None else self.supported_tasks
         if exclude:
```
mdtanrikulu, line 678: Maybe remove this check then? Because anyway `tasks` will be overwritten on #L680.

mdtanrikulu (approved):

PR #5200, opened 2020-12-24
docs: fix simple typo, wheter -> whether

timgates42:
There is a small typo in golem/network/concent/handlers_library.py.

Should read `whether` rather than `wheter`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


shadeofblue:
the CLA is not signed and btw, it's not the only issue in that piece of comment ...

